PR #9, opened 2020-08-23
APU page shows up automaticaly when apu master switch is put to on

noga-aviator:
i did that all thing in A320_Neo_EICAS.js to be able to use changePage() from baseEICAS

by noga#8589, i'm on your discord !


noga-aviator:
thanks for the merge !

PR #27, opened 2020-08-24
New automatic ECAM page switch

noga-aviator:
Now the ECAM switch to the APU page automaticaly when apu master is on. If no external power is connected, only stay on the APU page for 10 seconds then switch to the DOOR page.
ECAM now start on the DOOR page, was previously the FUEL page.
Everything tested in sim, seems to work.


noga-aviator:
also fix a bug where the apu gen info was showing only when the flap opened, should show up when master is on

noga-aviator:
also now the DOOR page switch !

noga-aviator:
ECAM doesn't start with door because it created a bug, i'll look into it another day

PR #35, opened 2020-08-24
Dev 0.1.1

JackZ974:
Changed the size of message APU FLAP OPEN
created new style for APU GEN which should turns amber when APU GEN is OFF and MASTER is ON
Modified style anchor point for units to be aligned to the right


noga-aviator:
i claim that, should be pretty quick. 
What is the amber color hex ?

noga-aviator:
nvm found it

noga-aviator:
👍

PR #63, opened 2020-08-25
Added takeoff memo

SeaRoll:
Added takeoff memo on the EICAS.
the memo shows up on two conditions:

1. Engine starter one & engine starter two is on. (I couldn't find the simvars for n2 for engine one and two).
2. The plane is on the ground and hasn't landed.

Issues:
No underline on T.O Text.
No blue memo. (can someone help with this?)


noga-aviator:
have you tryed everything in the sim ?

PR #64, opened 2020-08-25
APU now shows XX + bug fixes

noga-aviator:
APU gauge now show XX when it's is off. it diseaper a few seconds after the master is pressed
fixed a bug where apu page would not automaticaly show up
this hasn't been tested in sim


noga-aviator:
DONT MERGE

noga-aviator:
still shows xx on N% even tho it is active

PR #70, opened 2020-08-25
EICAS starts on door page

aboodilatif:
probably not the best way to do this, first time tinkering with js. but it works and does not glitch with the fuel page

kindly do not commit unless code is reviewed


noga-aviator:
merging because it works :+1:

PR #78, opened 2020-08-25
Displays are no longer on if there is no GPU/APU/ENG power

ninjomcs:


Username23-23:
That’s how it’s supposed to be

ninjomcs:
Yes it is, this is a change I implemented.

Username23-23:
My bad, I'm new to GitHub. I thought someone was reporting this as a bug.

PR #80, opened 2020-08-25
Fixed 'THR CLB' and 'SPEED' displaying at incorrect times

Curtis-VL:
Fixed 'THR CLB' and 'SPEED' displaying at incorrect times on the FMA.
Please double check this works as intended before merging. 👍

Previously 'THR CLB' would appear any time the MCDU was in the 'Climb' phase, now it is based on logic from some Airbus documentation. It should now only display 'THR CLB' when actually climbing and return to 'SPEED' or 'MACH' when level/hitting ALT* and levelling off.


noga-aviator:
tested in sim ?

Curtis-VL:
Fixed a small issue, now works correctly in sim.

PR #82, opened 2020-08-25
APU BLEED page first version

noga-aviator:
So for now it only shows if the eng1/2 bleed are connected and if apu bleed is connected.

i know the text is offset, i'm fixing that rn.


noga-aviator (commented):
this is really working and tested in sim

PR #87, opened 2020-08-25
Fix 70ft & 60ft callouts

### LINKED ISSUE #86: Remove "60" callout from landing annunciations
The real A320 calls out 100...50...40...30...20...5. Remove 60.

Alexr03:
Fixes #86 and also removes the 70ft callout as this is not what happens in the real aircraft.


noga-aviator:
have you tested it in sim?

Alexr03:
Yes - as far as I hear, the callouts are gone and only hear 50... 40...

PR #116, opened 2020-08-26
APU gen arrow only show when no EXTERN POWER connected

### LINKED ISSUE #20: APU GEN info
My APU GEN info doesn't show up when there's no FLAP OPEN message. 
Microsoft Flight Simulator Screenshot 2020 08 23 - 23 43 26 88

noga-aviator:
Fixes #20 
Couldn't get it to work with engine GEN because there is not enough simvar for it, will wait until someone does a more realistic ELEC system


noga-aviator:
btw tested in sim and working

PR #124, opened 2020-08-26
Fixed 'NAV', 'CLB', and 'DES' appearing with FD disabled.

Curtis-VL:
Fixed 'NAV', 'CLB', and 'DES' appearing with FD disabled.

(It looks like Airbus_FMA.js was removed at some point after my last pull request? This re-adds that file also)


michaeldiguiseppi:
the Airbus_FMA file was moved to within theA320 code instead of shared code. It's still in the project.

Curtis-VL:
Corrected file location, wasn't aware of that, sorry! :)

PR #154, opened 2020-08-27
Corrected various wrongly displayed modes on the FMA with FD disabled.

Curtis-VL:
Corrected various wrongly displayed modes on the FMA, none of these modes should have been appearing with FD disabled.


Curtis-VL:
Oops, forgot to update the layout.json, corrected that with a later commit.

PR #197, opened 2020-08-28
Modify the lower ENGINE ECAM screen to the correct Neo variant

lousybyte:


lousybyte:
Comparison: https://i.postimg.cc/Qtt1WHfn/Comparison.png

PR #205, opened 2020-08-28
Added Checklist Files 

arkar16:
Added Ingame Checklists, currently most are non-interactive, but it is all modular so we can expand on this in the future.


Benjozork:
Some unaligned tags in the XML files, might want to format it in your code editor !

arkar16:
Agreed, @Benjozork , I'll need to format that in the future.

TheoBearman:
@arkar16 Thanks for this update. Are there any plans to make the checklist interactive in the (near) future?

arkar16:
@TheoBearman Yes, we plan to make them interactive! @DoToAdventures can probably also chime in but we have no set release date or anything for now

TheoBearman:
@arkar16 thanks for the clarification! I think it's important as there is a lot of new functionality in the latest release and I'm sure lots more to come. While this is great in terms of realism, newer pilots to the aircraft and/or the game may well struggle. Speaking personally, I have benefited a lot from the interactive checklist when I need to double check something or just when I want to get flying as quickly as possible and just set the game to go through everything automatically ☺️

Issue #53, opened 2020-08-25
APU start button should work even if flap is not open yet


wpine215:
In the airplane, when you hit the APU Master switch, you don't have to wait for the FLAP OPEN to hit the start button. If you hit one after the other, the start waits for the FLAP OPEN to start the cycle.

noga-aviator:
ok, will try to implement

Blumski1:
The starter PB light still illuminates even when the flap is in transition and the starter hasn't engaged yet.

becas22:
However, you should wait 3sec between master on and start press

JackZ974:
I'm on it. reduced the time to 3 seconds.

becas22:
Nooo, the flap shouldn't open in 3 seconds! It's the time you wait to press START P/B

Blumski1:
And that is why I didn't mention the 3 seconds, it has nothing to do with the switch logic. The starter obeys the FLAP OPEN.

Waiting 3 seconds ensures that the "LOW OIL LEVEL" message is not displayed prior to starting the APU. If the message is displayed, refer to the QRH ECAM Advisory section and notify Maintenance.

JackZ974:
Yes. I know. But for now the logic that has been implemented in the js code is that the START button is linked to the FLAP OPEN. The delay for Flap open was set initially to 20 sec (like in the old CEO), so the START button was inop until then. @Dalmont confirmed me that in the NEO the FLAP OPEN Appears after 3 to 5 sec only. SO i reduced the FLAP apparition time from 20 to 3. 

Temporary fix.

JackZ974:
> 
> 
> Nooo, the flap shouldn't open in 3 seconds! It's the time you wait to press START P/B

This is Company procedure, but there is nothing in the FCOM about the fact that you should wait 3 seconds before switching the MASTER ON

JackZ974:
> 
> 
> And that is why I didn't mention the 3 seconds, it has nothing to do with the switch logic. The starter obeys the FLAP OPEN.
> 
> Waiting 3 seconds ensures that the "LOW OIL LEVEL" message is not displayed prior to starting the APU. If the message is displayed, refer to the QRH ECAM Advisory section and notify Maintenance.

In fact, You are allowed to press the start button right after the master. The APU simply won't start until the falp is fully open (around 10 to 20 sec on CEOs, 3 to 5 sec on NEO apparently)

Blumski1:
[...]

PR #223, opened 2020-08-29
Update A320_Neo_LowerECAM_APU.js --APU Startup Time 

JackZ974:
temporary fix modified the APU startup time from 20 sec to 3 according to @Dalmont (Flap opens much quicker on the NEo)


NathanInnes:
#53

JackZ974:
> 
> 
> #53

As said, it's a temporary fix in the meantime to have not pple wait 20 seconds before starting the APU. Needs more elaborate coding

PR #225, opened 2020-08-29
Activate both AP1 and AP2 at the same time

2hwk:
EXCLUSIVE_AP case removed. AP1 and AP2 will now display as engaged at the same time and used already baked-in logic to simulate simultaneous AP1+AP2.


2hwk:
https://github.com/wpine215/msfs-a320neo/issues/119

JackZ974:
Hmmm. Not that simple, alas. 
AP1+2 is only authorized for CAT3DUAL mode approaches, and Precision approaches, namely ILS ones. Ie LOC/GS or Roll out or GO around mode is engaged. Source FCOM: System/flight Guidance/Autopilot/AP Engagement

2hwk:
Alright, makes sense it was deliberately disabled then because the logic isn't actually complete. Closed.

PR #230, opened 2020-08-29
Updates to APU, power delivery to displays, fuel ecam page

ThePilot916:
This PR adds/fixes the following:
1. APU start sequence has been updated with improved logic and EGT now reaches max of 780 instead of prior 930.
2. Fixed Power delivery issue to systems when starting directly from APU.
3. Ovhd Batt display is now represented correctly
4. APU Bleed pressure momentarily drops when bleed value is opened and stabilizes again
5. Added jz fuel pages.


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_APU.js
```diff
@@ -216,28 +248,27 @@ var A320_Neo_LowerECAM_APU;
...
-                    return (135*n)-1320;
-                } else if (n < 20) {
-                    return -1262 + (224*n) - (5.8 * (n*n));
-                } else if (n < 36) {
-                    return ((-5/4)*n) + 925;
+                    return (135*n)-1600;
+                } else if(n < 20){
+                    return (20*n)+105;
+                } else if (n < 37) {
+                    return (17.125*n)+148.5;
```
Benjozork, line 264: Seems like some white space is missing here, might wanna check formatting settings in your editor :)

ThePilot916:
Resolved conflicts.

wpine215:
Seems this PR caused the MCP display to freeze/stop working somehow, so I had to revert it quickly. Need to figure out what may have caused it.

JackZ974:
The APU EGT reaching around 900 at startup is normal and expected, because this is a single stage turbine with a small compressor and not much air left for the combustion chamber isolation at startup. Pretty much like a PT6.The EGT goes down to around 500 after a few seconds. The operational limits for the APU are as follow (source FCOM): 
- Max EGT at startup (below 25000ft): 900C
-Max EGT at startup (above 25000ft): 930C
- Max EGT (continuous): 682C
-MAx EGT that requires emergency shutdown by the APU "FADEC": 700 to 742C

ThePilot916:
> Seems this PR caused the MCP display to freeze/stop working somehow, so I had to revert it quickly. Need to figure out what may have caused it.

Found the issue under FCU, will send in a new PR soon.

ukflyer:
Amazing it was doing all sorts of weird things! :)

Issue #57, opened 2020-08-25
Removal of Taxi Map on ND 


davidonthereg:
The taxi map on ND is currently too illegible, and inaccurate.

AdenFlorian:
I'm working on this, almost got it, maybe.

AdenFlorian:
Got a fix, will clean up and put up a PR tomorrow for it.

AdenFlorian:
Only have half a fix, but better than nothing! #232

Curtis-VL:
Corrected as best as we can in #232, waiting for Asobo fixes to allow it to be off when the weather radar is off! Thanks for the report! 👍

AdenFlorian:
reopening as its still an issue, we can track progress on asobo's side on this ticket

I created a zendesk ticket about it a while ago and got this email today:

![image](https://user-images.githubusercontent.com/3533988/93129429-985d4c80-f696-11ea-939b-1d27f3de8b56.png)

hotshotp:
issue is fixed

AdenFlorian:
only half fixed

hotshotp:
we decided to close the issue as it is fixed with the terrain/wxr not showing at lower altitudes

PR #232, opened 2020-08-29
Hide BingMap on ND when no weather or terrain

AdenFlorian:
Partially solves #57 , it hides the airports on the ND when no terrain or weather is showing, but airports will still show when terrain is on and above 500 AGL.

We are hiding the airports by hiding the BingMap, which draws the airports, terrain, and weather.

When weather is on and terrain is off, the weather radar is displayed, and it clears the background with black, so airports are already not displayed when weather is displayed.

When terrain is on and you are at or above 500 AGL, the airport will still be rendered.
We cannot fix this without a patch from Asobo.
There are no function exposed through Coherent that can control whether the airports are rendered or not (that we found).

The user should not see any differences when weather is being displayed (weather on and terrain off), or when terrain is on and the plane is at or above 500 AGL.

## Before (weather off, terrain off, 0 AGL) (dev branch):
FlightSimulator_1598740347

## After (weather off, terrain off, 0 AGL):
FlightSimulator_1598738390

## Before (weather off, terrain on, 0 AGL) (dev branch):
FlightSimulator_1598740355

## After (weather off, terrain on, 0 AGL):
FlightSimulator_1598738404

## After (weather off, terrain on, 500 AGL):
FlightSimulator_1598738521

The airport is still expected to be seen in the above picture for the reasons mentioned above.


Benjozork:
guys, multiple hours of discord streams and group investigation went into this, hope you like it 😂

AdenFlorian:
I can run the build.py once it's already been run on the branch itself, since the layout.json is out of date.

Curtis-VL (approved):

PR #284, opened 2020-08-31
Update flight_model.cfg

frankgrivel:
Various tweaks for stability.


frankgrivel:
I'm still figuring out how to work with Git properly; if this is not the way to address this please let me know

Benjozork:
Do you have references for the values you tweaked ?

frankgrivel:
> Do you have references for the values you tweaked and why you tweaked them ?

In general through the FCOMs, would you like me to specifically reference each one?

The effectiveness etc. is more to reduce the twitchiness of the control response. It was way too sensitive for myself and others I've spoken. I can't specifically back those tweaks up with numbers.

Benjozork:
> In general through the FCOMs, would you like me to specifically reference each one?

It is generally preferable to reference changes like that so we can backtrace in case of future changes or questions.

frankgrivel:
> > In general through the FCOMs, would you like me to specifically reference each one?
> 
> It is generally preferable to reference changes like that so we can backtrace in case of future changes or questions.
> 
> > The effectiveness etc. is more to reduce the twitchiness of the control response. It was way too sensitive for myself and others I've spoken. I can't specifically back those tweaks up with numbers.
> 
> Fair. I would personally wait for people to test the changes, but that's up to the core team.

Alright, so the references listed:
[W&B]
Airbus Airport & Maintenance Planning - https://www.airbus.com/aircraft/support-services/airport-operations-and-technical-data/aircraft-characteristics.html
MSFS2020 SDK documentation - (mirrored on my website - https://frankgrivel.nl/sdk/03-Content_Configuration/SimObjects/00-Aircraft/Aircraft_Configuration/Flight_Model.html)

[Geometry]
Airbus Airport & Maintenance Planning - https://www.airbus.com/aircraft/support-services/airport-operations-and-technical-data/aircraft-characteristics.html
FCOM & FCTM Part 1 and measurements

[Tuning]
No specific reference

[Speeds]
FCOM A320 Part 3

UncleClapton:
We tested this last night. There's worry that the autopilot might have some hard coded behavior that was modeled after the existing flight model, and it has lead to weird behavior on landings.

AdenFlorian:
@frankgrivel what are your thoughts on how it performs in autopilot?

AdenFlorian:
updated the pr to be based on latest master

AdenFlorian:
asking QA to try this out.

flightsimsimon:
So operated some circuits, firstly extraordinary long time for the SIM to load up which is unusual. 

The flaps seem to be extending on there own, no ref speeds on speed tape, speed select in red, hit APFLOOR on final approach as no VAPP speed though VAPP was in the approach page, ILS wasnt aligned, quite far to the left, whether thats an ASOBO issue or what I dont know, thanks Simon

AdenFlorian:
Closing this as we don't have a good method for testing these kinds of changes yet.

Issue #283, opened 2020-08-31
FOB/fuel on board in single digits is wrong


PCGamesF1Tech:
Fuel on board indication (FOB) is showing an accuracy in single digit kilograms, e.g. 8947kg.

This is wrong, it should be in increments of 20kg, so e.g. 8940kg.

Th3Falc0n:
Working on this.

PCGamesF1Tech:
Phew, that was quick, thanks for your effort! 👍🏻
Is it 10kg? I don't have my Laptop/OM with me on vacation

JackZ974:
Noticed also a discrepancy between Fuel USED between Engines page and FUEL page.  The values shown are different. 
Working on it

Th3Falc0n:
@Curtis-VL this is solved in #287 awaiting the PR to merge

NathanInnes:
Do you have a reference to confirm this behaviour at all?

AdenFlorian:
reference: https://www.youtube.com/watch?v=g32B8_5TbdQ

Th3Falc0n:
I get following things from the video:

All "fuel amounts available" in 20kg increments.
All "fuel amounts used" in 10kg increments.

I will implement that and create a new PR.

PR #287, opened 2020-08-31
FOB on upper ECAM and lower ECAM Fuel Page now shows in increments of 10

Th3Falc0n:
According to issue #283 the FOB should be shown in increments of 10.

IIRC from my plane in FSX this is correct.


AdenFlorian:
please start with a fresh branch made off of master and reapply your changes on that, thanks!

PR #290, opened 2020-08-31
Fix disabling APU starter switch

UncleClapton:
Fixes a typo that allowed the APU starter to be pressed even if master was off.

demonstrating bug


UncleClapton:
Closed since Pr got scuffed

Issue #212, opened 2020-08-29
Linting and code formatting


disposedtrolley:
I've witnessed a tremendous amount of progress being made on this project in the few days I've known about it. It's incredible to see how passionate the flight sim community is in giving up their time to work on freeware!

In an effort to keep this project scalable and maintainable, what do we think about introducing linting and code formatting for the HTML, CSS, and JS files in particular? This would allow for a consistent style across the codebase, reduce the amount of complexity required when dealing with any merge conflicts, and reduce overhead for contributors when thinking about how code should be formatted. The linter can also catch basic issues via static analysis.

I'm happy to investigate some tools and spike it out in a PR if there's interest. [ESLint](https://eslint.org/docs/rules/) is quite comprehensive and does double duty of linting and code formatting.

The main challenge will be to roll this out with minimal disruption to inflight features and PRs. Perhaps we can introduce another step in CI to run the formatter and commit changes?

Benjozork:
I've been thinking about this too.

[EditorConfig](https://editorconfig.org/) would also be useful to guarantee end-of-line and indentation consistency.

michaeldiguiseppi:
We could also implement prettier for code formatting.

Benjozork:
@michaeldiguiseppi, in my opinion Prettier alone is too opiniated and enforces some annoying rules sometimes. I also think it's not very suitable for us since a lot of Asobo files are already weirdly formatted and Its lack of configurability would not let us tune those warnings down.

ESLint can inherit Prettier's rule set anyway and has the advantage of being able to let the user disable certain rules.

Curtis-VL:
Ironically, a lot of the Asobo code follows different formatting rules too, which doesn't make things any easier. :)

Benjozork:
> Ironically, a lot of the Asobo code follows different formatting rules too, which doesn't make things any easier. :)

Yeah, that's the thing.

I think we can start by adding linter config files (ESLint, EditorConfig, not sure what to use for XML) and encouraging people to use them. Let them print out warnings in checks. Encourage people to fix formatting whenever they have the chance.

Then, when a lot of code has better formatting, we can start enforcing it.

AdenFlorian:
i just added eslint, will figure out prettier or something for all other files

AdenFlorian:
closing in favor of #761

PR #325, opened 2020-09-02
build: add .editorconfig

Benjozork:
This PR adds an [EditorConfig](https://editorconfig.org/) file.  This does not force anything, rather it provides a baseline encoding / formatting config for different text editors.

It can be used if desired by contributors, and optionally enforced later at the decision of the team.

Relates to / part of #212.

### DIFF
--- a/A32NX/.editorconfig
+++ b/A32NX/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 180
+tab_width = 4


AdenFlorian:
ah, so this is why my editor started changing line endings and inserting new lines

AdenFlorian:
a lot of the files are already CLRF, and if someones editor auto changes it to LF, and they check it in, then git thinks the entire file changed.

AdenFlorian:
All the files I've looked at are CRLF, so I think we should either set the editorconfig to CRLF or remove the end_of_line config.

Benjozork:
> All the files I've looked at are CRLF, so I think we should either set the editorconfig to CRLF or remove the end_of_line config.

Do you read my mind ? Literally just had a "oh shit" moment and realized that might be an issue 😂

I think we can change it to CRLF. You wanna make the change ?

AdenFlorian:
you can go ahead

PR #340, opened 2020-09-03
build: change newline to CRLF in .editorconfig

Benjozork:
See discussion in #325.

PR #342, opened 2020-09-03
build: set trailing newline to false in .editorconfig

Benjozork:
See discussion in #340.

I swear this is the last one 😆


Benjozork:
Actually don't merge this I think we should move it to the root.

AdenFlorian:
talked to Ben, will merge now.

AdenFlorian (approved):

PR #361, opened 2020-09-03
Removed PTU Sound

benjwri:
In the 320neo the PTU has been significantly dampened and is no longer audible in either the cockpit or cabin, and therefore I have removed it

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/sound/sound.xml
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/sound/sound.xml
@@ -227,18 +227,6 @@
 	
     <!-- Miscellaneous ==========================================================================================  --> 
 	
-	<Sound WwiseData="true" WwiseEvent="ptu_on" LocalVar="XMLVAR_PTU_ON" Continuous="false">
-		<Requires SimVar="TURB ENG N1" Units="PERCENT" Index="1">
-			<Range UpperBound="1" />
-		</Requires>
-	</Sound>
-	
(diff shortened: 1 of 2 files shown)


Benjozork:
for reference : 

https://youtu.be/UuSjXMTQabo

I personally hear it just as much, but one video is not enough to judge. I hope others can chime in on this.

benjwri:
When the first neos were delivered the PTU was louder than in previous models, and therefore Airbus has to come up with a damper system. This has silenced the PTU, however since this was produced after the first neos were delivered, this may not have been installed on the particular aircraft in the video, as it is quite an old video. 

I was travelling on a BA neo yesterday and although you could feel a slight vibration from the PTU, the noise itself was not audible.

Benjozork:
> When the first neos were delivered the PTU was louder than in previous models, and therefore Airbus has to come up with a damper system. This has silenced the PTU, however since this was produced after the first neos were delivered, this may not have been installed on the particular aircraft in the video, as it is quite an old video.
> 
> I was travelling on a BA neo yesterday and although you could feel a slight vibration from the PTU, the noise itself was not audible.

Thanks. So it's one of those things that vary from aircraft to aircraft.

Someone on discord mentioned having it be an option : https://discord.com/channels/738864299392630914/738864300818432023/751212341869281300

benjwri:
> Thanks. So it's one of those things that vary from aircraft to aircraft.

Don't think it varies too much, it was corrected very quickly, and therefore the vast majority of them have the dampers fitted. Regardless the PTU is not really audible in the cockpit in either version.

benjwri:
Closing this as I messed up the file structure, will fix and create now PR

PR #366, opened 2020-09-04
Build.py changes

arkar16:
Created by BiologicalNanobot#5285, approved by kiwi2021#0001

### DIFF
--- a/A32NX/build.py
+++ b/A32NX/build.py
@@ -1,47 +1,34 @@
 import os
 import json
-import sys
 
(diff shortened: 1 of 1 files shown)


AdenFlorian:
Can you describe the changes?

arkar16:
"it updates both layout.json and total_package_size in manifest.json, it also updates dates in layout.json instead of using a pre-set one"

AdenFlorian (approved):
ran the new build.py on my local and looks like it works!

PR #375, opened 2020-09-04
Removed PTU Sound

benjwri:
A320neo has the PTU dampened on almost all aircraft, bar a few of the original ones, and  therefore I have removed the sound, which is not audible in the cockpit either way. Past discussion on the topic in #361


AdenFlorian:
From multiple people, including Ares from the discord, I'm hearing that the PTU sound is more quiet in the neo, but it's still audible, and that the sound in MSFS just needs to be turned down a bit, so I'll make an issue for it to be turned down, and you or someone else can open a PR to turn it down, but not remove it. Thanks!

PR #380, opened 2020-09-04
Remove no team name from FAQ

pepperoni505:
Since there is now a team name, it doesn't make sense to keep the "Q: What is the name of this team?" on the FAQ


AdenFlorian:
Sorry for the trouble, but could you recreate this PR on master? we had to force push to master so it messed up this PR. Thanks!

wpine215:
Already fixed

PR #389, opened 2020-09-04
Runway Turnoff Lights fixed

nobodyn0se:
Fixed circuit mismatch for RWY_TURNOFF (left and right) in the NEO_INTERIOR.xml file to match corresponding circuit IDs as per the system.cfg file. Now, Runway Turnoff Lights do not illuminate the cockpit interior after landing gear is retracted.


AdenFlorian:
Looks good! Lights are only on when gear is mostly deployed.

image

image

AdenFlorian (approved):

PR #410, opened 2020-09-05
Fix the default GW on Lower EICAS not update in realtime

ductrinh1810:
**Summary of Changes**
- The SimVar TOTAL WEIGHT is not updated during the flight so the gross weight is not updated on the Lower EICAS. I changed the calculation of gross weight by GW = Total Payload + Airplane Empty weight + Total Fuel Weight.

- Not update the gross weight unit every time the refeshGrossWeight function is called. It is only updated when there is a change in gross weight value.


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -79,16 +79,22 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
-            if (this.gwUnit)
+                unit = "kg";
+            }
+        } else {
+            if (this.gwUnit) {
                 this.gwUnit.textContent = "LBS";
+                unit = "lbs";
+            }
+        }
+        for (var i = 1; i <= 6; i++) {
```
UncleClapton, line 94: Idk about leaving this hard coded. I assume these are all the different payload categories? While they shouldn't change, there could be a mod in the future that changes it which would break this. Is there a way to find this out from the sim?

UncleClapton (changes requested):

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -79,16 +79,22 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
+            gw += Math.round(SimVar.GetSimVarValue(`PAYLOAD STATION WEIGHT:${i}`, unit));
         }
+        gw += Math.round(SimVar.GetSimVarValue("EMPTY WEIGHT", unit)) + Math.round(SimVar.GetSimVarValue("FUEL TOTAL QUANTITY WEIGHT", unit));
```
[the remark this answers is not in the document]
ductrinh1810: I'm extracting it to variables for cleaning up. Thanks for your advise.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
ductrinh1810: I'm looking on the sim variable table. Let's me try with the payload station number variable.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -79,16 +79,22 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
     refreshGrossWeight(_force = false) {
         let gw = 0;
         let isInMetric = BaseAirliners.unitIsMetric(Aircraft.A320_NEO);
+        let unit;
         if (isInMetric) {
-            gw = Math.round(SimVar.GetSimVarValue("TOTAL WEIGHT", "kg"));
-            if (this.gwUnit)
+            if (this.gwUnit) {
```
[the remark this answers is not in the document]
ductrinh1810: Ok, let me check and refactor them. Thanks for you help.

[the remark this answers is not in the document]
ductrinh1810: @UncleClapton can we move the display unit checking process to init() ? As we just need it to set it only one time ?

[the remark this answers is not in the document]
UncleClapton: This code probably exists because unit settings may be introduced in the future. Probably best to leave it in the update loop for now until that's clarified down the road.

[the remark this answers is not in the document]
ductrinh1810: @UncleClapton  I fixed and refactored the code base on your suggestion.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -77,23 +77,23 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
+        let unit = isInMetric ? "kg" : "lbs";
+        let gw = 0;
+        let payloadWeight = 0;
+        let payloadCount = SimVar.GetSimVarValue("PAYLOAD STATION COUNT", "number");
+        let fuelWeight = Math.round(SimVar.GetSimVarValue("FUEL TOTAL QUANTITY WEIGHT", unit));
+        let emptyWeight = Math.round(SimVar.GetSimVarValue("EMPTY WEIGHT", unit));
+        for (var i = 1; i <= payloadCount; i++) {
+            payloadWeight += Math.round(SimVar.GetSimVarValue(`PAYLOAD STATION WEIGHT:${i}`, unit));
         }
+        gw = emptyWeight + fuelWeight + payloadWeight;
```
UncleClapton, line 90: I mentioned this before but the weight should only be rounded down once so we retain the decimals. If we don't then there's a chance for display errors.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -77,23 +77,23 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
-        if (isInMetric) {
-            gw = Math.round(SimVar.GetSimVarValue("TOTAL WEIGHT", "kg"));
-            if (this.gwUnit)
-                this.gwUnit.textContent = "KG";
-        }
-        else {
-            gw = Math.round(SimVar.GetSimVarValue("TOTAL WEIGHT", "lbs"));
-            if (this.gwUnit)
-                this.gwUnit.textContent = "LBS";
+        let unit = isInMetric ? "kg" : "lbs";
```
UncleClapton, line 81: Any variable that isn't changed after it's declaration should be a `const` instead of a `let` so we know it isn't changed somewhere down the line.

UncleClapton (changes requested):
just a few more things :)

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
ductrinh1810: > I mentioned this before but the weight should only be rounded down once so we retain the decimals. If we don't then there's a chance for display errors.

ok. Let me round the gw before set the value to the UI. This is my first PR for the flybywiresim project. Sorry for that.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
ductrinh1810: ok, I'm working on it and push new commit soon.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/EICAS_Common.js
UncleClapton: It's no biggie!

ductrinh1810: I pushed new code base on your review @UncleClapton

PR #227, opened 2020-08-29
Wind Indicator Fix

NathanInnes:
Fix #216 

If you are below 100kts ground speed the wind indicator on the MFD will not show.

Issue #216, opened 2020-08-29
Wind data on ND shows up below 100kts IAS 


mrgpnaccyt:
The wind data on the ND (Top left corner) should only show up above 100kts on the takeoff roll

PCGamesF1Tech:
I was also wondering why I had a wind indication at a parking position somewhere... 🤔

AdenFlorian:
@NathanInnes fix from #227 is now in master, I tested it and it looks good!

PR #425, opened 2020-09-05
Fix Wind Indicator

NathanInnes:
@AdenFlorian just an extra patch to solve the GS not showing 0.

#227
#216


topiek-fs (commented):
Is removing the entire MDInfo.js file intentional?

NathanInnes:
Yes it was intentional,  I originally imported shared NDInfo.js to override the .padStart(3, "0"); to .padStart(3); to remove the 0s on GS.

But done another work around to force MFD.js to do it on there instead. Once tested it worked the same way.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A320_Neo_MFD.js
```diff
@@ -561,22 +561,27 @@ class A320_Neo_MFD_NDInfo extends NavSystemElement {
...
             ws.textContent = "---";
+            wa.setAttribute("visibility", "hidden");
+        }else {
+            wa.setAttribute("visibility", "visible");
+        }
+        if (ADIRSState != 2){
+            gs.textContent = "---";
+        }else {
+            gs.textContent = groundSpeed.toString().padStart(3);
         }
```
NathanInnes, line 584: As seen here

NathanInnes, line 584: Doing it here instead of doing is on Shared/NDInfo.js is better

NathanInnes (commented):
As seen here

Issue #449, opened 2020-09-06
LS (ILS) button doesn't light up when activated [BUG]


hummelmose:
**Mod Version**

0.2.0

**Describe the bug**

when pressing the LS button for ILS approach it doesn't light up

**To Reproduce**

1. ILS approach
2. press LS 
3. No light

**Expected behavior**

Light in button

**Actual behavior**

No light in button

**References**

**Additional context**

UncleClapton:
LS button is tied to panel brightness for some reason. This bug is also present in the base aircraft. 

Duplicate of #67. In the future, please use the search box at the top of the issues list to try to find your problem before reporting it.

hummelmose:
Hi Uncle Clapton - I did search for LS but didn't find anything. Sorry.

UncleClapton:
No problem! I'll probably be looking into this one soon since nobody has really picked it up

Curtis-VL:
Fixed in #451.

Issue #271, opened 2020-08-31
LS button does not illuminate anymore


flightsimsimon:
Though all ILS functions are working, LS button no longer illuminates when pressed

Simon

ukflyer:
Hi, 
Also raised in https://github.com/wpine215/msfs-a320neo/issues/67
If you turn up the panel brightness its fine :)

flightsimsimon:
I see that thanks - that is still a bug though and shouldnt be affected by floodlighting

ukflyer:
absolutely Simon i believe its on the TO DO list :)

flightsimsimon:
Ah brill

2hwk:
![image](https://user-images.githubusercontent.com/15316958/91783565-341ca080-ec33-11ea-8d4b-03f2e5deb657.png)

LS brightness is linked with panel lighting brightness.

See #67

Curtis-VL:
Fixed in #451.

PR #451, opened 2020-09-06
Fix ILS button brightness

### LINKED ISSUE #67: ILS info display button work but doesn't light up
as you can see ils info is displayed but the button doesn't light up

UncleClapton:
Fixes #67, #271, #449 (those are the ones that are open)

**Summary of Changes**

* Fixes ILS button by using `<POTENTIOMETER_SEQ2>` to tie the panel brightness to only the text instead of the whole button.
* also remove a ton of whitespace.

before, the button was always off if the panel backlight was off.

panel backlight off:

panel backlight on:

**Additional context**

I'm not sure if much can be done about the brightness differences between buttons. the FD button is on a completely different set of templates from the LS button, and the LS button shares templates with the ND display buttons (CSTR, VOR D, WPT, etc...). That seems to be the root of the problem, but trying to find whats so different about each one is a rabbit hole.


NathanInnes (approved):
Tested and working.

Curtis-VL (approved):

PR #453, opened 2020-09-06
Hide Accu/Brake gauge if power is not available

### LINKED ISSUE #424: [BUG] Accu/brake press gauge illuminated when cold & dark
**Branch**
master

**Describe the bug**
Accu/brake press gauge is illuminated when cold & dark.

**To Reproduce**
1. Go to gate/park at night, cold & dark

**Expected behavior**
Cockpit should be completely dark when no power is on.

**Screenshots or Videos**
kuva
kuva

**Additional context**
-

veikkos:
Fixes #424 

**Summary of Changes**
Tie Accu/Break "gauge" visiblility to DC power state, so that it does not illuminate unrealistically when aircraft has no power.

Read `DCPowerAvailable` only once in a hope of a tiny performance improvement.

**Screenshots**
Images dark since Cold & Dark, please bump brightness if needed. :)
kuva
kuva

**Additional context**


UncleClapton:
This is supposed to simulate a physical gauge, so the solution of blanking out the display if it's off isn't ideal. I've been toying with an approach that lets it remain visible without giving off light in the cockpit, but the XML is a pain to understand.

veikkos:
> 
> 
> This is supposed to simulate a physical gauge, so the solution of blanking out the display if it's off isn't ideal. I've been toying with an approach that lets it remain visible without giving off light in the cockpit, but the XML is a pain to understand.

Great if you have better plan! Wouldn't this be better solution until then though? It's not hard to modify later on. I find the illumination more immersion breaking for the time being.

AdenFlorian:
I think this is fine for now, since it does the same thing in vanilla. taking a look at the code now...

JackZ974:
Yes. That was a known default I introduced in the new brake gauge code, because this is a physical gauge, and its not removed when power is off. But i haven't found a way to make it black when at night and cockpit completely unpowered.

PR #476, opened 2020-09-07
Checklists: revise what is INOP and what is not

Venorcis:
**Summary of Changes**
Some things are now supported and shouldn't thus be marked INOP in the checklists, whilst others were inadvertedly not marked as INOP. Also fixed some casing, two typo's (parkig > parking, slide > slides), and changed three double spaces to one space.

Edit: this has been integrated (mostly) in #479, but can be merged separately first too.

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/Checklist/Airbus_A320neo_Checklist.xml
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/Checklist/Airbus_A320neo_Checklist.xml
@@ -446,7 +446,7 @@
 				<Checkpoint ReferenceId="A320_PARKING_PARKING_BRAKE_ACCU_PRESS"/>
 				<Checkpoint ReferenceId="A320_PARKING_PARKING_BRAKE"/>
 				<Checkpoint ReferenceId="A320_PARKING_ANTI_ICE"/>
-				<Checkpoint ReferenceId="APU_Bleed_Off"/>
+				<Checkpoint ReferenceId="APU_Bleed_On"/>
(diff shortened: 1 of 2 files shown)


Venorcis:
Added an additional structural change: APU BLEED should be ON when parking (and OFF again when securing the aircraft, which was correct already). Source: https://youtu.be/TbppXMGrsvw?t=720

DoToAdventures:
Hello, I took the liberty to implement your changes in #479 and gave you access to my fork if you wish to implement further changes.

Venorcis:
> Hello, I took the liberty to implement your changes in #479 and gave you access to my fork if you wish to implement further changes.

Nice! Maybe we should decide on the casing that is used in the action labels, I went for fully uppercase right now (seemed most consistent with the current code) but the built-in checklists don't do that (e.g. OFF vs Off). Also I wasn't sure if the "barking brake" was a mistake or not, if it is the corresponding identifier should probably also be corrected.

DoToAdventures:
I would go with fully uppercase as this is most often found in real SOPs and FCOMs and potentially change the parts of the checklist that currently use functions from the Asobo Checklist Library. I only used those for conveniance and because they are fully interactive already.

Venorcis:
> I would go with fully uppercase as this is most often found in real SOPs and FCOMs and potentially change the parts of the checklist that currently use functions from the Asobo Checklist Library. I only used those for conveniance and because they are fully interactive already.

I agree, so I did that in my final commit; probably good if you integrate that in your pull then too. I also put some comments on it :)

AdenFlorian:
@Venorcis Want to close this one in favor of #479 ?

AdenFlorian:
Thanks for working with them!

PR #480, opened 2020-09-07
[CDU] The AIRWAYS page should only show the waypoints after the curre…

### LINKED ISSUE #301: MCDU "AIRWAYS" page
The current AIRWAYS page in F-PLN currently can only display/edit up to 5airways, cannot cycle through pages, the real A320 CDU should only edit/display the airways from the WAYPOINT you have entered the AIRWAYS> page on, for example below, this page should only edit airways from the LVZ waypoint instead of all.

image

lousybyte:
…nt selected one

Fixes #301
Also fixes an issue with waypoints not being visible sometimes when inserted in the temp plan before commiting.

**Summary of Changes**

The AIRWAYS page should properly show just the enroute waypoints after the current selected one and not all of them. Players are now able to manually enter as many airway connected waypoints as needed.

N/A

**Additional context**

N/A


AdenFlorian (approved):
tested, and LGTM!

PR #488, opened 2020-09-07
Rework chronometer so it runs off in-game time

### LINKED ISSUE #464: [BUG] Chronometer doesn't scale time properly
**Mod Version**

master

**Describe the bug**

The chronometer that has been added always runs at 1x speed (real time), even when the simulation speed is set to something higher like 2x or 4x

**To Reproduce**

1. Start the chronometer
2. Change the simulation rate
3. Observe chronometer timing relative to clock timing

**Expected behavior**

The chronometer should speed up when the simulation speed is increased.

**Actual behavior**

The chronometer always records elapsed time at 1x speed (real time).

**References**

**Additional context**

UncleClapton:
Fixes #464 

**Summary of Changes**

Makes chronometer work off of in-game time instead of a an update delta time. This lets the chronometer be affected by sim rate (time acceleration), with a couple caveats:
1. Time warping from the weather menu will break the chronometer (rather hilariously).
2. The chronometer cannot be stopped by pausing the sim, since time still elapses with the game "paused".
    * this was also a problem with the old approach, but worth stating again since it's still a limitation.

This PR also adds a default build task for VSCode.

**Additional context**


AdenFlorian (approved):
tested and LGTM!

PR #515, opened 2020-09-08
Cleans up alot of code

Meqolo:
This PR goes through all code to:
- Remove all cases of `var`
- Removes any curly brackets that are inside of a switch statement
- Adds curlybrackets to some (most??) if statements that do not have them
- Fixes some of the auto-generated enums so they are readable
- Improves general readability

was bored, so i decided to fix some shit code

Note: please ignore the first three commits, dont have a clue why they got pushed aswell


Meqolo:
nice, github just decided to kill my merge conflict fixes

AdenFlorian:
I think we're fine if we just fix things as we go, only changing lines of code that we actually need to touch to fix a bug or implement a feature. For big changes like this, it's best to create an issue first and then get the go ahead from the maintainers before implementing.

PR #518, opened 2020-09-08
Display Waypoint ETA in UTC 

### LINKED ISSUE #458: [BUG] Waypoint ETA on ND is displaying in local time not UTC.
**Mod Version**

cef39ab (master)

**Describe the bug**

When the pfd displays the ETA to the next way-point on the flight-plan, it displays the time in the local time corresponding to the position of the aircraft instead of UTC. I checked with one of our irl A320 pilots, and he says that he believes that those ETA's should display in UTC. It doesn't make much sense for the ETA display local time when the chrono in the aircraft displays UTC.

**To Reproduce**

1. Start a flight with a flight-plan in the MCDU
2. Takeoff from your departure airport and engage the A/P and proceed with your flight normally
3. Note that the next way-point ETA in the top right of the right PFD display is indicated in the local time and not in utc (picture below for reference).

**Expected behavior**

The ETA should display in UTC not the aircraft's local time.

**Actual behavior**

The ETA displays in local time.

**References**

image

Note the ETA circled in red on the left, and the chrono displaying the correct UTC time on the right. 

**Additional context**

UncleClapton:
Fixes #458

**Summary of Changes**

* Waypoint ETA displays in UTC
* Shared files used by the MFD are now prefixed with `A32NX_` to prevent any possible conflict with the original or other mod files.
    * @AdenFlorian said this was probably a good idea, so I did it. 😆 

Before:

After:

**Additional context**


UncleClapton:
I moved this back to draft because I identified a problem when messing about.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A32NX_NDInfo.js
```diff
@@ -0,0 +1,531 @@
...
+                              this.waypointDistance.textContent = this.currentWaypointDistance.toFixed(1);
+                          else
+                              this.waypointDistance.textContent = this.currentWaypointDistance.toFixed(0);
+                      }
+                  }
+                  if ((_eta != this.currentWaypointTimeETA) || _force) {
+                      this.currentWaypointTimeETA = _eta;
+                      const utcETA = (_eta + SimVar.GetGlobalVarValue("TIME ZONE OFFSET", "seconds")) % 86400;
+                      if (this.waypointTime != null) {
+                          const hours = Math.floor(utcETA / 3600).toString().padStart(2, "0");
```
veikkos, line 199: Please consider keeping the implementation as it was; first calculate the value (original code as a reference)
```
var hours = Math.floor(localETA / 3600);
```
as that value is not tied to the UI representation. Only then create the presentation format when needed
```
this.waypointTime.textContent = hours.toString().padStart(2, "0")...
```
This way the actual value is re-usable and it's clear the what line does what. Now value calculation is mixed up with the presentation.

Just my 2 cents.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A32NX_NDInfo.js
```diff
@@ -0,0 +1,531 @@
...
+                      if (this.waypointDistance != null) {
+                          if (this.currentWaypointDistance < 10000)
+                              this.waypointDistance.textContent = this.currentWaypointDistance.toFixed(1);
+                          else
+                              this.waypointDistance.textContent = this.currentWaypointDistance.toFixed(0);
+                      }
+                  }
+                  if ((_eta != this.currentWaypointTimeETA) || _force) {
+                      this.currentWaypointTimeETA = _eta;
+                      const utcETA = (_eta + SimVar.GetGlobalVarValue("TIME ZONE OFFSET", "seconds")) % 86400;
```
veikkos, line 197: `const utcETA = (_eta + SimVar.GetGlobalVarValue("TIME ZONE OFFSET", "seconds") + 86400) % 86400;` should do as then the value never goes negative.

UncleClapton: Why would the value ever go negative?

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A32NX_NDInfo.js
UncleClapton: yeahh thats a fair point. reverted.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A32NX_NDInfo.js
UncleClapton: I ended up doing an entirely different approach that works better. I didn't think the new approach was possible since, at the time when I was first testing, it looked like waypoint ETE wasn't reporting correctly. I was just wrong and looking at the wrong simvar 😆

AdenFlorian:
very nice fix, thanks for this!

AdenFlorian:
tested locally and looks good!

PR #519, opened 2020-09-08
Fix Landing gear Indicator text illumination

### LINKED ISSUE #198: Landing Gear UNLK light color
The landing gear unlock light is gray
UNLK_GRAY

The expected is to be red when operating the gear and gray when idle
image

UncleClapton:
Fixes #198 (which was closed by mistake)

**Summary of Changes**

Fix `UNLK` text so it properly illuminates. This PR also replaces the default landing gear light template so it's easier to understand and maintain.

Before:

After:

After (at night):

**Additional context**

Funny enough the real fix here was to not do whatever Asobo was doing to calculate if this light should be on. Something about setting a simvar in the same expression block screwed things up.


JackZ974:
Nice!
Can you fix the Ldg gear lever that should be able to move on the ground (it's electromechanically locked in the down position while on the ground) too?
And the red arrow that should be way more red when lighted, and still show the white contour of the arrow even when lighted.

Thx

AdenFlorian (approved):

PR #544, opened 2020-09-09
Changes .editorconfig to use tab

Meqolo:

**Additional context**


Meqolo:
closing because i need to make a new fork, my fork is broken

PR #545, opened 2020-09-09
Changes .editorconfig indent_style to tab

Meqolo:

**Additional context**


Meqolo:
im so confused, why wont this remove my commits that arent even related to this, i literally created a new fork and its still here

lousybyte:
I can already see the pitchforks, it's worse than a religious war.

Meqolo:
redoing this... again to fix commit history

PR #546, opened 2020-09-09
Change .editorconfig indent style to tab

Meqolo:


Benjozork:
Spaces were chosen because of their use in the base code and the unwillingness to reformat large portions of code at once.

Meqolo:
We shouldnt have to conform to bad practices that are being used, thats the exact opposite of what this project stands for and aims for.

AdenFlorian:
I think it's best to be consistent, and existing code is spaces (some are tabs, so just have your editor match the file that you are editing)

PR #554, opened 2020-09-09
Change acceleration altitude check from AGL altitude to MSL altitude

### LINKED ISSUE #552: [BUG] THR RED/ACC altitudes should be MSL instead of AGL
**Mod Version**

master

**Describe the bug**

On the PERF page of the MCDU, on page Take Off, the THR RED/ACC altitudes are currently assumed to be AGL altitudes when deciding the altitude at which the plane will ask for the throttle lever to be reduced to 'Climb', and the altitude at which the plane will switch from the Take Off state to the Climb state. According to feedback from an IRL A320 pilot, these altitudes should be entered as MSL altitudes. The thrust reduction altitude is already changed to be read as MSL by changes made in issue #397. The acceleration altitude, however, is still being read as AGL.

**Expected behavior**

THR RED/ACC altitudes to be used as MSL altitudes like in the real aeroplane.

**Actual behavior**

THR RED/ACC altitudes used as AGL altitudes.

RikvanD1997:
Change acceleration altitude check from AGL altitude to MSL altitude

Fixes #552 

**Summary of Changes**

Plane will now look at the barometer altitude to decide when to go into climb phase, rather than the altitude from ground. This is based on feedback from an IRL A320 pilot.


RikvanD1997:
Will recreate this because I should have really made a branch in my fork. Whoops.

PR #555, opened 2020-09-09
Change acceleration altitude check from AGL altitude to MSL altitude

### LINKED ISSUE #552: [BUG] THR RED/ACC altitudes should be MSL instead of AGL

RikvanD1997:
Change acceleration altitude check from AGL altitude to MSL altitude

Fixes #552

**Summary of Changes**

Plane will now look at the barometer altitude to decide when to go into climb phase, rather than the altitude from ground. This is based on feedback from an IRL A320 pilot.


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -705,8 +705,8 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
 
         //Changes to climb phase when acceleration altitude is reached
         if (this.currentFlightPhase === FlightPhase.FLIGHT_PHASE_TAKEOFF && airSpeed > 80) {
-            const agl = Simplane.getAltitudeAboveGround();
-            if (agl > (this.accelerationAltitude || this.thrustReductionAltitude || 1500)) {
+            const msl = Simplane.getAltitude();
+            if (msl > (this.accelerationAltitude || this.thrustReductionAltitude || 1500)) {
```
tyler58546, line 709: If we are using MSL, the default `1500` will not work at airports higher than 1500 ft. This needs to be changed to airport elevation + 1500. We should also check to make sure acceleration altitude and thrust reduction altitude are above the airport elevation.

tyler58546 (changes requested):

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
RikvanD1997: Agreed. Only thing I'm not sure on right now is where the default thr red/acc is set. Could you perhaps point me in the correct direction for that? Then I will work on it as soon as I can.

RikvanD1997: Okay I just realized I interpreted this completely wrong probably. I was assuming you were talking about the default value on the MCDU page. While I still think that should be changed, and I'd be open to change it, I'll at least change the default value in the check when I get out of work tomorrow.

MCDU change might be better in a different issue anyways, since it seems there are still some discrepancies if I look at something like [http://www.a320dp.com/A320_DP/nav-flight-management/sys-14.5.3.html](http://www.a320dp.com/A320_DP/nav-flight-management/sys-14.5.3.html)

veikkos: Should the default be tied to airport? Or just current ground level?

```
        if (this.currentFlightPhase === FlightPhase.FLIGHT_PHASE_TAKEOFF && airSpeed > 80) {
            const agl = Simplane.getAltitudeAboveGround();
            const msl = Simplane.getAltitude();
            const heightAboveSeaLevel = msl - agl;
            if (msl > (this.accelerationAltitude || this.thrustReductionAltitude || (1500 + heightAboveSeaLevel))) {
                this.currentFlightPhase = FlightPhase.FLIGHT_PHASE_CLIMB;
            }
        }
```

Or maybe something else. :)

veikkos: On a second note, I don't think you can even have both as unset, i.e. null. Game automatically sets one of them and you cannot clear it as far as I can tell, so the third condition does not ever affect in practise. Doesn't hurt to have something there though.

RikvanD1997: Yeah exactly. There is actually a simvar containing the current MSL altitude of the ground, so I'm going to use that. Commit coming up in a bit after I've tested it.

tyler58546: I think we should use the airport altitude, not the current ground altitude.

RikvanD1997: Good point. I got the information I need for it now. I'll change it here and in #585

RikvanD1997: I kept the ground altitude in there, but only if there is no origin airport selected in the CDU.

tyler58546: If the ground isn't flat, that could cause problems. I think it would be better if we used the initial ground altitude after takeoff.

veikkos: I wouldn't spend too much time on the default value. As said, it will likely never be executed as you cannot unset both THR RED/ACC, you always have one (or soon two, like in real aircraft). It would be more important to get the actual fix in place.

RikvanD1997: You might not always have those values after I implement my other PR. If you start a flight without a destination selected, your origin airport on the INIT A page will be empty, and then so will the THR RED/ACC values. Without a default value this could mean that you get stuck in Take Off mode. This will be a small change anyways.

veikkos: Thanks for clarification.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -690,14 +689,33 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
         if (this.currentFlightPhase <= 2 && SimVar.GetSimVarValue("L:A32NX_Preflight_Complete", "Bool") == 1) {
             if (!(highestThrottleDetent == ThrottleMode.TOGA || highestThrottleDetent == ThrottleMode.FLEX_MCT) && SimVar.GetSimVarValue("RADIO HEIGHT", "Feet") < 100) {
                 SimVar.SetSimVarValue("L:A32NX_Preflight_Complete", "Bool", 0);
+                this.climbTransitionGroundAltitude = null;
             }
         }
-
+
```
veikkos, line 695: Minor but extra new lines.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -690,14 +689,33 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
-
+        
         //Changes to climb phase when acceleration altitude is reached
         if (this.currentFlightPhase === FlightPhase.FLIGHT_PHASE_TAKEOFF && airSpeed > 80) {
-            const agl = Simplane.getAltitudeAboveGround();
-            if (agl > (this.accelerationAltitude || this.thrustReductionAltitude || 1500)) {
+            const planeAltitudeMsl = Simplane.getAltitude();
+            let accelerationAltitudeMsl = (this.accelerationAltitude || this.thrustReductionAltitude);
+            
+            if(!accelerationAltitudeMsl) {
```
veikkos, line 701: Minor: `if (!a [...]` space missing, same line below.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -690,14 +689,33 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
         //Changes to climb phase when acceleration altitude is reached
         if (this.currentFlightPhase === FlightPhase.FLIGHT_PHASE_TAKEOFF && airSpeed > 80) {
-            const agl = Simplane.getAltitudeAboveGround();
-            if (agl > (this.accelerationAltitude || this.thrustReductionAltitude || 1500)) {
+            const planeAltitudeMsl = Simplane.getAltitude();
+            let accelerationAltitudeMsl = (this.accelerationAltitude || this.thrustReductionAltitude);
+            
+            if(!accelerationAltitudeMsl) {     
+                if(!this.climbTransitionGroundAltitude) {
+                    let airport = this.flightPlanManager.getOrigin();
```
veikkos, line 703: `const airport` since value is not changed.

veikkos:
Minor comments. Works as intended. Hopefully @tyler58546 could take a look soon. 👍

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
RikvanD1997: Fixed in next commit

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
RikvanD1997: Fixed in next commit

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
RikvanD1997: Fixed in next commit

RikvanD1997:
Small change coming in the next commit: Using `getOrigin().altitudeFP` instead of `getOrigin().infos.coordinates.alt`. `getOrigin().infos.coordinates.alt` is only filled when loading a flightplan in the create flight menu, and won't be filled if you manually change the origin airport in the INIT A screen on the MCDU. `getOrigin().altitudeFP` will always return an accurate altitude of the origin airport. Currently testing it.

veikkos (approved):

tyler58546 (changes requested):
Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.

RikvanD1997:
> Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.

I remember once seeing a behaviour where it goes to climb by default at 5000ft. I'll dig through the code to see if I can find it.

veikkos:
> 
> 
> Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.

It happens with master too, right? I don't think it's this change. 🤔

RikvanD1997:
> Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.

I tested 3 different scenarios at KJAC, and couldn't reproduce this problem in any of them. I tested this with the latest master with this change. I'll merge master back into here now. Perhaps that was the problem.

No ACC alt: https://streamable.com/7lanpn
KJAC ACC alt: https://streamable.com/wywswy
Manually changed ACC alt: https://streamable.com/myhr3y

veikkos:
> 
> 
> Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.

Just making sure, so this "LVR CLB" is the problem you are refering to? So the computer wants you yo set throttle to CL even if the CLB mode is just armed (BLUE below SRS).

kuva

RikvanD1997:
> > Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.
> 
> Just making sure, so this "LVR CLB" is the problem you are refering to? So the computer wants you yo set throttle to CL even if the CLB mode is just armed (BLUE below SRS).
> 
> kuva

If that's it, that would be unrelated to this change. That message pops up in both take off and climb modes. Something might be going wrong there in an altitude check, but not anything this change would affect.

veikkos:
> 
> 
> > > Tested at KJAC. No matter what I put in the MCDU it switched to climb mode immediately.
> > 
> > 
> > Just making sure, so this "LVR CLB" is the problem you are refering to? So the computer wants you yo set throttle to CL even if the CLB mode is just armed (BLUE below SRS).
> > kuva
> 
> If that's it, that would be unrelated to this change. That message pops up in both take off and climb modes. Something might be going wrong there in an altitude check, but not anything this change would affect.

Yep. I'm looking into this issue but it's related to FMA and happens with master too.

veikkos:
It seems that the issue is because `Simplane.getThrustReductionAltitude()` at `IsActive_LVRCLB` still returns 1500 at `IsActive_LVRCLB`. Despite what has been set to mcdu. This PR is about acceleration altitude and I think the issue with "LVR CLB" should be fixed separately. I'll keep on looking into it.

veikkos:
Issue is that we should change the thrust reduction and acceleration altitude through MCDU's API so that the change propagates to Simplane.

Fix is in `UpdateThrRedAccFromOrigin`:

```
-        mcdu.thrustReductionAltitude = thrRedAccAltitude;
-        mcdu.accelerationAltitude = thrRedAccAltitude;
+        const thrRedAccAltitudeString = thrRedAccAltitude.toFixed(1);
+        mcdu.trySetThrustReductionAccelerationAltitude(`${thrRedAccAltitudeString}/${thrRedAccAltitudeString}`);
```

You can integrate it to this PR even though it's unrelated but at least allows proper testing for this change as well.

RikvanD1997:
> Issue is that we should change the thrust reduction and acceleration altitude through MCDU's API so that the change propagates to Simplane.
> 
> Fix is in `UpdateThrRedAccFromOrigin` like:
> 
> ```
> -        mcdu.thrustReductionAltitude = thrRedAccAltitude;
> -        mcdu.accelerationAltitude = thrRedAccAltitude;
> +        if (thrRedAccAltitude) {
> +            const thrRedAccAltitudeString = thrRedAccAltitude.toFixed(1);
> +            mcdu.trySetThrustReductionAccelerationAltitude(`${thrRedAccAltitudeString}/${thrRedAccAltitudeString}`);
> +        }
> ```
> 
> But you cannot clear it through `trySetThrustReductionAccelerationAltitude` as far as I can tell, so not sure how to handle case where value is unknown.

Found the SimVars that the sim uses to keep track of the thrust reduction and acceleration altitude. In the next commit these will be set together with the variables in the MCDU, instead of using `trySetThrustReductionAccelerationAltitude` which is designed for user input. The default values of the SimVars are '0', so setting these to 0 and the MCDU variables to undefined will make this work as intended. Commit incoming after my last tests.

RikvanD1997:
Scenarios tested:

- No altitudes entered: https://streamable.com/9rmylo
Plane will never ask for LVR CLB and will transition to climb phase 1500 ft above ground level.
- Thrust reduction altitude entered: https://streamable.com/3nz8y0
Plane will both ask for LVR CLB and transition to climb phase at the specified thrust reduction altitude.
- Both altitudes set, to seperate values: https://streamable.com/5sxouu
Plane will ask for LVR CLB at the thrust reduction altitude, and will transition to climb phase at the acceleration altitude.
- KJAC entered as origin airport, but taking off from VNKT (~2000 ft lower): https://streamable.com/z2tmg2
Plane will both ask for LVR CLB and transition to climb phase at the altitude automatically entered on the PERF screen by selecting the origin airport.

veikkos:
Very nice!

RikvanD1997:
@tyler58546 Could you have another look at this? Not trying to be annoying, just pinging this in case it disappeared in your notifs.

RikvanD1997:
I just royally messed with this branch by accident. I'm such an idiot. Will create a new PR for this at a later date when other issues are solved...

veikkos:
For future reference, you can just move back the branch reference where it was and then force push. No need to create new branch or PR.

PR #557, opened 2020-09-10
Allow for settable V-speeds, add better auto-calculation UX

### LINKED ISSUE #281: Unable to manually enter constraints and V-speeds
MCDU does not allow you to manually enter altitude and airspeed constraints for the route. Also when entering a flight plan in manually (not making one in the MSFS world map) no altitudes or airspeeds appear, it is just appears as ---/". Same with V-speeds, not able to enter them in manually and it appears that they only change when the flap lever changes flap settings rather than when entering a flap setting on the MCDU.

UncleClapton:
Partially resolves #281

**Summary of Changes**
* VSpeeds are now manually changeable.
* VSpeeds can be automatically calculated by clicking a speed option with an empty input, and pressing it again to confirm the result. (see video)
  * By default, selecting the speed with an empty input would result in a new auto-calculated speed, but I changed it for the sake of a better user experience. In real life these values cannot be calculated automatically by the aircraft, so there is no reference to work off of.
* VSpeed auto-calculation prefers to look at MCDU's flap setting over the current flaps of the plane
  * By default it ignored flap settings.

Video: https://imgur.com/a/0K2ODGu

**Additional context**

This partially resolves #281. Since these are two very separate problems being reported under one issue, I'm preferring to split up the work.


UncleClapton:
right.. as per the editorconfig some of these files were comitted as crlf line endings. this caused merge conflicts with another recently merged PR. Will try to fix.

UncleClapton:
Fixed and ready for review!

wpine215 (approved):

PR #563, opened 2020-09-10
Checklists: less INOP again

Venorcis:
**Summary of Changes**
In the checklists, removed INOP from the Vspeed settings and Predictive Windshear System. 

Also moved EXT PWR (ON IF AVAIL) before the APU fire test & start, which makes more sense to me and seems to be the suggested flow by 320 Sim Pilot (see https://youtu.be/Zsk72luNZvc?t=180).


DoToAdventures (approved):
Whoops! This is what happens when you quickly want to fix something in the minute it gets reported and instantly push it! You're correct, of course.

PR #577, opened 2020-09-10
Fixed fuel display on ECAM as discussed in #283

### LINKED ISSUE #283: FOB/fuel on board in single digits is wrong

Th3Falc0n:
Fixes #283

**Summary of Changes**

The values shown on the upper ECAM and lower ECAM (FUEL page) are modified according to the constraints discussed in issue #283.
They now have a precision of 20kg for fuel remaining and 10kg for fuel used.


AdenFlorian (approved):
tested and LGTM!

PR #585, opened 2020-09-10
[CDU] Automatically fill in thrust reduction altitude and acceleration altitude based on runway height

### LINKED ISSUE #578: [BUG] THR RED/ACC altitudes not automatically set
**Mod Version**

master

**Describe the bug**

In the real airplane the THR RED/ACC altitudes on the PERF Take Off page would be automatically set to runway level + an amount specified by the company. This would be set either when the plane powers on with a flight plan, or when the FROM/TO airports in the INIT A screen of the MCDU change.

**To Reproduce**

1. Start a flight from a runway or from C&D
2. Look at the PERF Take Off page

**Expected behavior**

THR RED/ACC should be set to the runway altitude + a specified amount

**Actual behavior**

THR RED/ACC set to 1500/---

**Additional context**

Since MFS seems to use 1500 as the default value for the thrust reduction altitude, I say we use that as the default company value for now.

RikvanD1997:
Fixes #578 

**Summary of Changes**

On plane power on, so either when powering on the plane from C&D or when entering the cockpit when starting the flight from the runway, the current MSL altitude of the ground is determined. That altitude + 1500 will be set into the thrust reduction altitude and the acceleration altitude on the PERF Take Off page, like it would in the real plane.

Before: 
image

After:
image

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
@@ -93,7 +93,17 @@ class CDUInitPage {
             let value = mcdu.inOut;
             mcdu.clearUserInput();
             mcdu.tryUpdateFromTo(value, (result) => {
-                if (result) {
+                if (result) {                    
+                    let thrRedAccAltitude;
+                    
+                    const origin = mcdu.flightPlanManager.getOrigin();
+                    if (origin) {
+                        thrRedAccAltitude = origin.altitudeinFP + 1500;
+                    }
+                    
+                    mcdu.thrustReductionAltitude = thrRedAccAltitude;
+                    mcdu.accelerationAltitude = thrRedAccAltitude;
+                    
                     CDUAvailableFlightPlanPage.ShowPage(mcdu);
                 }
             });
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
@@ -94,15 +94,7 @@ class CDUInitPage {
             mcdu.clearUserInput();
             mcdu.tryUpdateFromTo(value, (result) => {
                 if (result) {                    
-                    let thrRedAccAltitude;
-                    
-                    const origin = mcdu.flightPlanManager.getOrigin();
-                    if (origin) {
-                        thrRedAccAltitude = origin.altitudeinFP + 1500;
-                    }
-                    
-                    mcdu.thrustReductionAltitude = thrRedAccAltitude;
-                    mcdu.accelerationAltitude = thrRedAccAltitude;
+                    CDUPerformancePage.UpdateThrRedAccFromOrigin(mcdu);
                     
                     CDUAvailableFlightPlanPage.ShowPage(mcdu);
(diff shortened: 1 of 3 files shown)


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
```diff
@@ -93,7 +93,17 @@ class CDUInitPage {
...
-                if (result) {
+                if (result) {                    
+                    let thrRedAccAltitude;
+                    
+                    const origin = mcdu.flightPlanManager.getOrigin();
+                    if (origin) {
+                        thrRedAccAltitude = origin.altitudeinFP + 1500;
+                    }
+                    
+                    mcdu.thrustReductionAltitude = thrRedAccAltitude;
```
veikkos, line 104: Wouldn't it be better to have

```
// let thrRedAccAltitude; // removed

const origin = mcdu.flightPlanManager.getOrigin();
if (origin && origin.altitudeinFP) {
    const thrRedAccAltitude = origin.altitudeinFP + 1500;
    mcdu.thrustReductionAltitude = thrRedAccAltitude;
    mcdu.accelerationAltitude = thrRedAccAltitude;
}
```

because otherwise "NULL" is set to `thrustReductionAltitude ` and `accelerationAltitude` if origin doesn't exist? Then leave the values as they were. Also make sure `altitudeinFP` exists.

RikvanD1997: This behaviour is intentional. When the CDU doesn't know what the origin airport is, or what the altitude of that airport is, THR RED/ACC are supposed to be empty.

Source:
image

veikkos: Thanks for the clarification. Then that part is good.

Would be good to check availability of "altitudeinFP".

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -46,6 +46,18 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
         }
         this._onModeManagedHeading();
         this._onModeManagedAltitude();
+
+        let thrRedAccAltitude;
```
veikkos, line 50: It's a matter of style, but this can be written with less lines and without mutable variable. And then encapsulated into a function in `A320_Neo_CDU_InitPage.js` to avoid copy-paste.

Like
```
static updateThrRedAcc(mcdu) {
    const origin = mcdu.flightPlanManager.getOrigin();
    const thrRedAccAltitude = origin && origin.altitudeinFP ? origin.altitudeinFP + 1500 : undefined;

    mcdu.thrustReductionAltitude = thrRedAccAltitude;
    mcdu.accelerationAltitude = thrRedAccAltitude;
}
static ShowPage1(mcdu) { // existing code
[...]
```
and then call `CDUInitPage.updateThrRedAcc(this);` from this file and `CDUInitPage.updateThrRedAcc(mcdu);` from file above.

veikkos (changes requested):

veikkos (approved):
You can commit several files at the same time btw. ;) Changes look good!

RikvanD1997:
> You can commit several files at the same time btw. ;) Changes look good!

Yeah I know, just something I'm used to when working on relatively small changes :P
Cheers.

PR #587, opened 2020-09-10
Fix MCDU green dot speed

Evaske:
**Summary of Changes**
This fixes the green dot speed that is displayed on the PERF TAKEOFF and APPROACH pages of the MCDU. It utilises the formula from Airbus of:
`(gross weight in thousandths * 2) + 85kts`

It previously referenced random numbers added to the aircraft weight.


tyler58546:
Does this change the green dot speed shown on the airspeed tape?

Evaske:
> Does this change the green dot speed shown on the airspeed tape?

No. That is coming from a different function.

Benjozork:
LGTM

ImenesFBW:
Looks good to me! Tried different load-outs and din't run into any problems. Thanks !

wpine215 (approved):

PR #607, opened 2020-09-11
ECAM CRZ page performance optimization

Jargendas:
**Summary of Changes**

@UncleClapton Notified me that the ECAM CRUISE page I wrote apparently causes higher draw times and more frame dips when active. I originally didn't really optimize it for performance in any way.
So I did exactly that: The page is now doing as little as possible during each update call and only redraws the page if there are actually new values to draw.
This made the code a bit more spaghetti-like, but I think it is still bearable. I will make that better another time.
The page should look the same as before but run faster. 

However, as my system is much more limited by the GPU than by the CPU, I cannot really test how much it actually improved the situation, so I hope @UncleClapton can verify it.

Sidenote: I also adapted the recent changes from the fuel page while I was at it (adds mod 10 to the consumption values), assuming the cruise page probably behaves the same.


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_CRZ.js
```diff
@@ -67,67 +86,102 @@ var A320_Neo_LowerECAM_CRZ;
...
+            const leftConsumption = SimVar.GetSimVarValue("GENERAL ENG FUEL USED SINCE START:1", "gallon") * this.unitFactor * 0.001;
+            const rightConsumption = SimVar.GetSimVarValue("GENERAL ENG FUEL USED SINCE START:2", "gallon") * this.unitFactor * 0.001;
+
+            const leftConsumptionShown  = fastToFixed(leftConsumption - (leftConsumption % 10), 0);
+            const rightConsumptionShown = fastToFixed(rightConsumption - (rightConsumption % 10), 0);
 
-            this.FuelUsedTotal.textContent = Math.round(leftConsumption)+Math.round(rightConsumption);
-            this.FuelUsedLeft.textContent = Math.round(leftConsumption);
-            this.FuelUsedRight.textContent = Math.round(rightConsumption);
+            if ((leftConsumptionShown != this.LeftConsumptionDisplayed) || (rightConsumptionShown != this.RightConsumptionDisplayed)) {
```
Benjozork, line 95: I would change the names, "shown" and "displayed" are confusingly similar

UncleClapton: Asobo's pattern for value diffing uses the parameter pattern: `last<VariableName>`. Might be best to follow the same pattern so it's consistent

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_CRZ.js
```diff
@@ -67,67 +86,102 @@ var A320_Neo_LowerECAM_CRZ;
...
+                this.RightConsumptionDisplayed = rightConsumptionShown;
+            }
 
             // Oil
             let value = SimVar.GetSimVarValue("ENG OIL QUANTITY:1", "percent") * 0.01 * 25;
-            this.OilQuantityLeft.textContent = value.toFixed(1).split(".", 2)[0] + ".";
-            this.OilQuantityLeftDecimal.textContent = value.toFixed(1).split(".", 2)[1];
+            let valueShown = fastToFixed(value, 1);
+            if (valueShown != this.OilQuantityLeftDisplayed) {
+                const decimalSplit = value.toFixed(1).split(".", 2);
```
UncleClapton, line 109: fun fact for the future. You can use destructuring to give array values names.

```
const [firstValue, secondValue] = value.toFixed(1).split(".", 2);
```

UncleClapton: Also maybe `fastToFixed()` should be used instead? I'm not sure how much faster it is but couldn't hurt.

Jargendas: I tried to use it where possible, but it has a huge flaw: When rounding to .0, it cuts the decimal. Really stupid behavior if you ask me.
I figured baking in checks to catch that is worse than just using toFixed. But now I have a weird mix of both functions and I don't really like it. I'll think about it again.

Jargendas: > fun fact for the future. You can use destructuring to give array values names.
> 
> 
> 
> ```
> 
> const [firstValue, secondValue] = value.toFixed(1).split(".", 2);
> 
> ```

Uuhh thanks, I didn't know!

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_CRZ.js
Jargendas: Okay, I will do that

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_CRZ.js
UncleClapton: yeaaaa I just noticed that bit about fastToFixed a couple hours ago... thats no good 🙃
WELP oh well.

wpine215 (approved):

PR #626, opened 2020-09-12
Set TCAS switch to TA/RA during takeoff and flight init phases

wpine215:
Fixes #[issue_no]

**Summary of Changes**

Set TCAS switch to TA/RA by default when spawning on runway, cruise, approach, or final. STBY otherwise.


wpine215:
Simple fix, tested and works fine, so I'll go ahead and merge it.

Issue #610, opened 2020-09-12
[REQUEST] IRS INIT> Button on MCDU INIT Page


Zyonix007:
The current MCDU does not have a ~~align IRS~~ IRS INIT button

**Describe the feature you'd like to see implemented**

The MCDU should have a ~~Align IRS~~ IRS INIT button when the IRS is set to nav

**References**
 
toliss

Benjozork:
note to anyone who wants to work on this : This is not a real A320 feature, but a QoL shortcut.

Zyonix007:
It is a real thing. It's like pos init on Boeing

Benjozork:
> It is a real thing. It's like pos init on Boeing

No, ADIRS takes time to align. POS INIT is for setting the initial position and IRS will still take time to align.

Zyonix007:
Correct Airbus has a button called align IRS that functions like pos unit.
See here https://youtu.be/o3--ALDKLUc

UncleClapton:
For quick visual reference of page design

![](https://clapton.dev/u/2009/9a3386f9.png)

Externoak:
I'm currently working on a IRS INIT page.

Current status:
![image](https://user-images.githubusercontent.com/17742637/93017547-e8221200-f5c9-11ea-9a90-9821ae8eb316.png)

Externoak:
Mentioning issue https://github.com/flybywiresim/a32nx/issues/401

UncleClapton:
Looks great!

AdenFlorian:
trying to figure out if this issue is for the QoL feature, or the IRS INIT page...

AdenFlorian:
@Zyonix007 please provide a picture of this "align IRS button" that you're talking about.

Zyonix007:
IRS init is what i mean

Zyonix007:
[image: image.png]

Externoak:
@AdenFlorian I think previous versions of airbus had a Align IRS button (not 100% sure since i'm no pilot) check my comment https://github.com/flybywiresim/a32nx/pull/629#issuecomment-691631469 first reference.

AdenFlorian:
@Externoak that reference is for the old CRTs, this video shows the newer LCDs which show IRS INIT: https://www.youtube.com/watch?v=o3--ALDKLUc&feature=youtu.be

AdenFlorian:
@Zyonix007 the image didn't upload

Externoak:
[...]

Issue #398, opened 2020-09-05
[REQUEST] Quick IRS align


Alexr03:
**Is your feature request related to a problem? Please describe.**
As per the request in PR #196 this issue is being created.

**Describe the features you'd like to see implemented**
When we get a EFB integrated into the aircraft, we should allow a way to set the speed of IRS alignment. This is the "standard" in most payware level aircrafts in other sims, they allow a way to change the IRS align speed.

**Additional context**
N/A

NathanInnes:
PR #196

AdenFlorian:
We can make a hidden MCDU CHEATS page, that should not be able to be stumbled upon by players, and we can add something for quick align in there.

madmonkeyfpv:
> We can make a hidden MCDU CHEATS page, that should not be able to be stumbled upon by players, and we can add something for quick align in there.

Is this happening and when?

ntlgr:
Would love to see that soon. With the amount of trouble we have with sim bugs and CTDs, having to wait 10 minutes to restart a failed flight is way too much for me.

wpine215:
Implemented.

madmonkeyfpv:
implemented is excellent news, any guide on how we actually do it?

scott-vincent:
I'd like to know how to do this too. I agree with all the FS2020 bugs it's a pain to have to wait 7 minutes for alignment each time.

ThatRedMelon:
In the dev version, it is in MCDU Menu->Options->Align Time. There are 3 options: real, fast and instant, corresponding to align times of 7 minutes+, 90 seconds, and instant. It's not in stable currently.

scott-vincent:
Works great. Thank you.

PR #629, opened 2020-09-12
Added "Align IRS" message to MCDU for quick IRS alignment

Externoak:
Issue: https://github.com/flybywiresim/a32nx/issues/610

**Summary of Changes**
Added a Align IRS option inside MCDU to quickly align IRS.
Option will only appear if IRS is actively aligning or isn't already aligned.
Also will clear ECAM IRS align message.

image

**Additional context**

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_InitPage.js
@@ -53,7 +53,7 @@ class CDUInitPage {
         let alignCheck = SimVar.GetSimVarValue("L:A320_Neo_ADIRS_IN_ALIGN", "Boolean");
         let alignOption = ""
         if (alignCheck) {
-            alignOption = "ALIGN IRS"
+            alignOption = "ALIGN IRS>"
         }
         let flightNo = SimVar.GetSimVarValue("ATC FLIGHT NUMBER", "string", "FMC");
         if (!flightNo) {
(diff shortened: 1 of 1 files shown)


Benjozork:
Hmm. I think we should discusss whether we want that feature or not in the first place.

oliverpope03:
> Hmm. I think we should discusss whether we want that feature or not in the first place.

I agree also i recommend if it is wanted that it is "ALIGN IRS>" like the wind below it to show the button is to activate it

Externoak:
I agree as mentioned in https://github.com/flybywiresim/a32nx/issues/610 it is a QoL improvement. 
Also as said in https://github.com/flybywiresim/a32nx/issues/398 most payware level aircraft have a way to quickly align the IRS, if this method is to be used or another one is to be discussed.

Externoak:
Now shown as:

image

Benjozork:
> I agree as mentioned in #610 it is a QoL improvement.

What I mean is that those discussions need to take place in the community in a broader sense, because some people wanted this as a function of the EFB.

wpine215:
The ToLiss A319/A321 features something similar. However, with a key difference. Instead of the "ALIGN IRS>" button instantly completing the alignment like the current behaviour of this PR, it could instead act as a prerequisite for the completion of alignment - the departure ICAO must be entered and the "ALIGN IRS>" button must be pressed during alignment, or else it won't complete. It should not instantly complete alignment or clear any relevant ECAM memos.

Externoak:
@wpine215 
Should have this "ALIGN IRS>" button pressed during alignment like so [reference](https://youtu.be/uqKeSO6msDk?t=576) or as linked in https://github.com/flybywiresim/a32nx/issues/610#issuecomment-691621359 [reference](https://www.youtube.com/watch?v=o3--ALDKLUc&feature=youtu.be) as a "IRS INIT" page?

Meqolo:
I think it would be best if you had it say "QUICK IRS ALIGN" or something of the sort, as its currently quite vague

AdenFlorian:
I think any QoL features should be on a hidden MCDU page that you wouldn't be able to get to without knowing how.

AdenFlorian:
Closing this as we originally decided to wait for an EFB before adding cheats. We can add them to the MCDU, but it should be in a way where players won't stumble upon it. Let's continue discussion in #398

PR #646, opened 2020-09-13
NO DH Fix

devHazz:
**Summary of Changes**
NO DH will only display on the PFD when a DH is given/DH = 0. You can enter NO for no DH.

image
image
image
image

**Additional context**
It's gonna throw Invalid Entry, I'll try and push a new fix where it clears it but for now, it should work like normal.

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
@@ -591,6 +591,10 @@ class CDUPerformancePage {
         mcdu.onRightInput[2] = () => {
             let value = mcdu.inOut;
             mcdu.clearUserInput();
+            if(value == 'NO' || 'no') {
+                mcdu.setPerfApprDH(0 || '0');
+                CDUPerformancePage.ShowAPPRPage(mcdu);
+            }
             if (mcdu.setPerfApprDH(value)) {
                 CDUPerformancePage.ShowAPPRPage(mcdu);
             }
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
@@ -588,12 +588,16 @@ class CDUPerformancePage {
         if (isFinite(mcdu.perfApprDH)) {
             dhCell = mcdu.perfApprDH.toFixed(0);
         }
+        if(mcdu.perfApprDH == -1) {
+            dhCell = "NO";
+        }
         mcdu.onRightInput[2] = () => {
             let value = mcdu.inOut;
             mcdu.clearUserInput();
             if(value == 'NO' || 'no') {
-                mcdu.setPerfApprDH(0 || '0');
+                mcdu.setPerfApprDH(-1);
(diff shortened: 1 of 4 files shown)


PCGamesF1Tech:
Our procedure is to explicitly enter NO as value to omit the auto callout "hundred above".
If you enter "0" the auto callout will come at 100ft RA which interferes with our (Cpts) "Continue" call.

UncleClapton:
IMO we shouldn't be piecemealing issues. Others might feel differently but if this is just a partial fix to the same problem, then it should go back to the draft state until the full fix is complete

devHazz:
It may look like I was peacemealing it but it was my mistake. I did have a final which I pushed but then would forget something etc.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
```diff
@@ -588,9 +588,17 @@ class CDUPerformancePage {
         if (isFinite(mcdu.perfApprDH)) {
             dhCell = mcdu.perfApprDH.toFixed(0);
         }
+        if(mcdu.perfApprDH == -1) {
+            dhCell = "NO";
+        }
         mcdu.onRightInput[2] = () => {
             let value = mcdu.inOut;
             mcdu.clearUserInput();
+            if(value == 'NO' || 'no') {
```
davidwalschots, line 597: Assuming this is plain JavaScript:
```suggestion
            if(value == 'NO' || value == 'no') {
```

veikkos: `value.toLowerCase() === 'no'` would be the choice for many pros. Triple equal sign is the way to go for added explicitness.

davidwalschots: Most definitely usage of `===` makes most sense. Having no experience with any of development within this project, I'm assuming `mcdu.inOut` represents the value entered through the keyboard by the pilot (or alternatively the value displayed as feedback to the pilot)? If so, then I'm assuming characters are either always lowercase or uppercase and therefore a single equality expression would be enough and no `toLowerCase` would be needed.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
```diff
@@ -588,9 +588,17 @@ class CDUPerformancePage {
...
+                mcdu.setPerfApprDH(-1);
+                CDUPerformancePage.ShowAPPRPage(mcdu);
+                mcdu.clearUserInput();
```
davidwalschots, line 600: **question**: Do any of the function calls above this one set input?
**suggestion**: If not, then remove this call to `clearUserInput`, because it was already cleared earlier on line 596.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
AdenFlorian: yes, please confirm if rest of code is checking upper or lowercase, and then do that.

AdenFlorian:
Closing stale PR. @devHazz Please reopen with the changes from latest master and with your final changes (you made it sound like you had more changes to push but haven't pushed yet). I tried looking into the changes to see what needed fixing but couldn't find a simple solution.

PR #647, opened 2020-09-13
Added Dummy AudioSwitching knob on top overhead

oliverpope03:
**Summary of Changes**

Init value at Norm as stated by Jayzee
Switch should remain in Norm unless abnormal ops

**Additional context**


camilo625:
Works as described

PR #650, opened 2020-09-13
Update default screen brightness in C&D

### LINKED ISSUE #592: [REQUEST] Screens and panel lights should be dimmed, not completely off, at startup.

Our airline SOP and other ones that I new about never leave the screens completely off or dark when leaving or securing the aircraft.  It should be dim downed to lowest setting, where you still can see that the screens and panels are on/lit, and then left in that state for the next crew.

(Should also prevent new user of our mod to think that there installation is broken with  dark screens at startup.

**References**

Real life SOP

**Additional context**

UncleClapton:
Fixes #592

**Summary of Changes**

Sets default screen brightness in C&D start to 0.1 (first setting)

**Additional context**

See conversation in #592 on the authenticity of this decision.

I feel like even if we wanted to interpret a "cold and dark" state to mean something akin to "this plane has been in storage for a month so everything is just off", the amount of complaints we've gotten from users thinking the screens are broken is a problem on it's own. Just my 2 cents.

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/apron.FLT
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/apron.FLT
@@ -169,13 +169,13 @@ Potentiometer.7=0
 Potentiometer.14=0
 Potentiometer.15=0
 Potentiometer.16=0
-Potentiometer.17=0
-Potentiometer.18=0
-Potentiometer.19=0
-Potentiometer.20=0
-Potentiometer.21=0
-Potentiometer.22=0
-Potentiometer.23=0
+Potentiometer.17=0.1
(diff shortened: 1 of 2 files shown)


CoffeePanda0:
Just tested, can confirm all working thanks!

AdenFlorian:
@veikkos why the thumbs down?

AdenFlorian (approved):
code looks good!

veikkos:
> 
> 
> @veikkos why the thumbs down?
> 
> I'm hearing that majority of SOPs leave them at dim and not off, so it's not wrong, right?

Then I stand corrected and it's fine. I got the impression that this change is compromising realism because people don't know how to use the plane. In case SOPs tend to leave them dim, it's thumbs up.

PR #652, opened 2020-09-13
Fixed flight phase stuck in takeoff

### LINKED ISSUE #562: On Subsequent flights aircraft will not cycle to the CLB phase [BUG]
**Mod Version**

Latest

**Describe the bug**

If you do 2 consecutive flights, on the second flight -  the aircraft is stuck in the takeoff phase.

**To Reproduce**

1.Fly to an airport land, and take off to another airport in the same session.

**Expected behavior**

It should not be stuck in the takeoff phase.

**Actual behavior**

**References**

**Additional context**

tyler58546:
Fixes #562 

**Summary of Changes**

* Fixed flight phase not changing to takeoff on a second flight without reloading the aircraft


flightsimsimon:
Tested all work as expected however LS light is on even when switch is off

flightsimsimon:
Conducted a flight which was planned on msfs them when landed entered a new plan without restarting and this time flight had no issues moving to cruise so not stick in take off mode

PR #659, opened 2020-09-13
Add more enum options, fix error in types.d.ts

lukecologne:
**Summary of Changes**

* Fixed error in the typescript file
* Added more enum options

I found the file where those enums are defined, as well as the Simplane functions (if that was not already known). For Steam installs, it's located in `Steam\steamapps\common\MicrosoftFlightSimulator\Packages\fs-base-ui\html_ui\JS` in `SimPlane.js`. There are many more functions in Simplane, if you want I can look through them and maybe add a few more interesting ones for convenience

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -42,6 +42,12 @@ declare global {
         FLIGHT_PHASE_GOAROUND
     }
 
+    enum AutopilotMode {
+        MANAGED,
+        SELECTED,
+        HOLD
+    }
+
     enum ThrottleMode {
         UNKNOWN,
         REVERSE,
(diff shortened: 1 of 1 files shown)


lukecologne:
Oh, and SimVar is also defined in that directory, I'll se what I find there

Benjozork:
Nice. It doesn't seem like we use `FLIGHT_PHASE_PREFLIGHT` or `FLIGHT_PHASE_TAXI` yet. Could be useful to actually implement them instead of starting on `FLIGHT_PHASE_TAKEOFF`.

lukecologne:
Hm. According to my IDE, Preflight is actually used

lukecologne:
SimVar seems to have some extra features, for example batch reading of Simvars, and also got a method for getting and setting GameVars, and a method for getting (no setting) GlobalVars.

AdenFlorian (approved):

Issue #660, opened 2020-09-14
[REQUEST]F/CTL ECAM Page


tareksabet:
**Is your feature request related to a problem? Please describe.**
Yes

In the F/CTL Page on the lower ECAM is appeared in green Prior Engine Start 
![Microsoft Flight Simulator 9_14_2020 12_51_36 AM](https://user-images.githubusercontent.com/70240330/93032048-7bd9f980-f62f-11ea-846c-5107580c2cc0.png)

**Describe the feature you'd like to see implemented**
In the F/CTL page on lower ECAM should appear in amber 

**References**

![image0](https://user-images.githubusercontent.com/70240330/93031991-200f7080-f62f-11ea-95b6-e55cbf112976.jpg)

**Additional context**

Benjozork:
I will look into this.

lukecologne:
It displays in amber when hydraulic pressure is insufficient. Since we do not have a hydraulics system currently, We would need to "fake" it by simply turning it amber when engines are stopped

JackZ974:
Better to wait till the HYD system is implemented IMHO. It's in the works anyways

wpine215:
Fixed with merged PR.

PR #662, opened 2020-09-14
feat(ecam/f-ctl): make controls amber when both engines dead

Benjozork:
Partially takes care of #660

**Summary of Changes**

The F/CTL ECAM page will now have controls amber when both engines are not started.

This is temporary behaviour until we have proper modeling of the hydraulic systems.

image
image

**Additional context**

N/A


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_LowerECAM_FTCL.html
```diff
@@ -1,7 +1,10 @@
-<link rel="stylesheet" href="./A320_Neo_LowerECAM_FTCL.css" />
+<link rel="stylesheet" href="./A320_Neo_LowerECAM_FTCL.css?v=44" />
```
tyler58546, line 1: You forgot to remove this, also you don't need it for CSS.

Benjozork:
Note: the behaviour in the first screenshot is incorrect, it is fixed to require both engines off in the lastest commit.

PR #675, opened 2020-09-14
[MCDU] Fix takeoff speed to V2+10

### LINKED ISSUE #643: [BUG] SRS does not maintain V2+10
**Mod Version**

master

**Describe the bug**

SRS should command a pitch that maintains V2+10 on takeoff, or the speed when TOGA was set on go around. 

**To Reproduce**

1. Set TOGA/FLEX thrust for takeoff. 
2. SRS commands acceleration to a speed far above V2+10. 

**Expected behavior**

Pitch bar should command V2+10. 

**Actual behavior**

See above. 

**References**

http://www.a320dp.com/A320_DP/nav-autoflight/sys-13.3.16.html

**Additional context**

Until the go around SRS issue is fixed, SRS won't work for go around. It should work in that case too.

veikkos:
Fixes #643 

**Summary of Changes**

Aircraft should aim for V2+10 at takeoff until climb phase is achieved. Fix includes only CDU part. PFD support seems to be lacking and is to be fixed separately.

**Note**: For me at least MSFS doesn't do great job keeping the speed but at least it's better than the old +190kt target which is clearly wrong.

**Additional context**

[Discord ref](https://discordapp.com/channels/738864299392630914/739223532269076561/755091076238409760)
[From Issue](http://www.a320dp.com/A320_DP/nav-autoflight/sys-13.3.16.html)

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
@@ -678,7 +678,8 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
             }
             if (this.currentFlightPhase === FlightPhase.FLIGHT_PHASE_TAKEOFF) {
                 if (this.isAirspeedManaged()) {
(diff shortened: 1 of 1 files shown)


AdenFlorian (approved):
code looks fine, ready to test!

FlightSimmer7700:
It held 160knots which seems quite reasonable if its supposed to be V2 +10 and then at 1500ft it went up to 185knots so it works

PR #684, opened 2020-09-14
Better input validation and parsing for Flaps/THS setting in MCDU

### LINKED ISSUE #674: [BUG] MCDU > PERF TAKE OFF > FLAPS/THS input not limited, Vspeeds implausible
**Mod Version**

master

**Describe the bug**

Possible bug: The MCDU > PERF > TAKE OFF page allows entering values for the FLAPS setting which are not in {0, 1, 2, 3, 4}.
Clearly, this is low priority.

**To Reproduce**

1. Go to the TAKE OFF PERF page on the MCDU
2. Enter a value > 4 (preferably 99) for the FLAPS.
3. Press v1 twice to set v1 to an automatically calculate value.

**Expected behavior**

Flaps should not be allowed to be set to values outside of {0, 1, 2, 3, 4}. Attempting to do so should result in an invalid entry error.
If however values larger 4 are deemed okay (I am lacking reference material), V speed calculation should not be possible.

**Actual behavior**

Settings larger 4 are accepted and Vspeed calculation is attempted.

**References**

In-game screenshot of the issue:
Image of issue

Snippet from A320 non-neo reference material:
A320 non-neo refeence

**Additional context**

In A320 reference material I found that this info is only for reference and no input limits are given. So, it is not a big issue in terms of realism. However, currently the setting is used for the QOL feature that allows automatic calcilation of Vspeeds. With wrong flap settings, the Vspeeds are also implausible. So I think the input on the flap page should be limited.

Related: No input is accepted for the /THS part, but I get that it is just not implemented, yet.

UncleClapton:
Fixes #674 

**Summary of Changes**

* Add better input validation for FLAP/THS.
    * Flap setting is limited to 1-3
    * THS now limited to a max of 2.5
* Add better parsing for FLAP/THS.
    * Flap is now optional input. (e.g. `/UP0.5` is a valid input now)
    * THS can be written with UP/DN at either the start OR end of the input.
    * THS number can be written as `0.1`, `.1`, or even in plain whole numbers like `1`
* Fix CHG Code to read `A20N` instead of `A2N` to satiate @tareksabet's OCD.

**Additional context**


sharksfan166269:
Everything works as intended, except one thing I'm not entirely sure on - if you enter /UP0.0 or /DN0.0 in the THS space it doesn't accept it which may or not be incorrect. I checked this behavior on the FSLabs A320, any entering of either /UP0.0 or /DN0.0 accepts and shows as /UP0.0 in the MCDU (even when entering /DN0.0). This could be a software difference between the CEO and NEO however. If I was a good tester, I'd know for sure, but I thought I'd just update you that everything else apart from that works as intended for me.

UncleClapton:
@sharksfan166269 ah yeah I'm missing an `=` sign there. my bad, will correct

UncleClapton:
Okay fixed and ready for re-review

PR #688, opened 2020-09-15
Fix #679 APU Fire Test "INOP"

### LINKED ISSUE #679: [BUG] APU firetest button still has an INOP mouseover
**Mod Version**
master

**Describe the bug**
The APU fire test button on the overhead panel gives an '(Inop.)' tooltip, whilst it actually works. Because of this the cursor is also wrong (a regular cursor instead of a hand).

NathanInnes:
Fixes #679 

**Summary of Changes**

APU Fire Test button no longer says INOP.

image

**Additional context**

Discord username: Sabes#8668


oliverpope03:
The APU fire test button wasn't "INOP" the part of the larger button that unlocks it to be pulled was "INOP" and its click box was weird and went half over the APU button 😄

PR #701, opened 2020-09-16
[ATC] Added background lighting and realistic size (update A320Neo_ATC.css)

JackZ974:
Modified the css background for realistic background lighting at night, reduced the size of the screen and the size of the digits as well as per real ATC unit.
No change on the code

Fixes #[issue_no]

**Summary of Changes**

Before: 
image
After:
image
Daylight:
image
Real instrument (sourec @Party_Grunt):
image

**Additional context**


camilo625:
I am getting a slightly different result on the TCAS font position: 
image

JackZ974:
Thanks for pointing the error. To modified lines were missing. Should be OK now.

UncleClapton:
Looks really good. I love it!

JackZ974:
> 
> 
> Looks really good. I love it!

Devil is in the details ;-)

Coralec:
image
The text is now centered properly. However the background lighting is oriented up and left. Also I assume this fix is only for the squawk code indicator as the frequency indicators are unchanged with this update.

Really like this change by the way, great touch.

JackZ974:
1-The display seem to be shifted to the left because your example displays 1200. The number1 has less footprint than a 6 for example, so it is normal and expected. Try with 4444 or 6666 and you will see what I mean.
2- I don't have the same display as yours. Are you using the new font?

Carter633:
Microsoft Flight Simulator 21_09_2020 14_12_59
Microsoft Flight Simulator 21_09_2020 14_13_03
Microsoft Flight Simulator 21_09_2020 14_13_31
Microsoft Flight Simulator 21_09_2020 14_13_36
Microsoft Flight Simulator 21_09_2020 14_14_22
Microsoft Flight Simulator 21_09_2020 14_14_41
Microsoft Flight Simulator 21_09_2020 14_14_45
Microsoft Flight Simulator 21_09_2020 14_14_50
Microsoft Flight Simulator 21_09_2020 14_12_52

As Jack was mentioning above, the alignment seems fine. The actual text window has a larger bezel on the right hand side so it may be why it 'looks misaligned' but I think the backlighting is brilliant! Seems to work fine for me. Tested with all dome and pedestal lights off, and also tested with a few on dim, and in addition to testing at night, tested at dusk/dawn too and seems fine in all instances!

Carter633 (approved):

AdenFlorian:
that looks off, purple box needs to go right and down, there's more purple above the numbers than below.

AdenFlorian:
fixed positioning. There was new font in this PR, so someone can add that in a future PR, thanks!
image

PR #705, opened 2020-09-16
[BUGFIX] Prevent ECAM auto page switching when page is selected

lukecologne:
**Summary of Changes**
* Fixes issue where automatic ECAM page selection would override the manually selected page.

**Additional context**
* ~~Maybe a problem: A page triggered by a failure will not override the selected page, I don't know if this is expected behaviour or not~~
* ~~Problem: When Engine mode selector is not in normal, the engine page would come up, but will not go away, somewhere it overrides the currently selected page.~~ Should be fixed now

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.js
@@ -68,6 +68,7 @@ class A320_Neo_EICAS extends Airliners.BaseEICAS {
         this.ApuAboveThresholdTimer = -1; // MODIFIED
         this.MainEngineStarterOffTimer = -1;
         this.CrzCondTimer = 60;
+        this.PrevFailPage = -1;
 
         this.topSelfTestDiv = this.querySelector("#TopSelfTest");
         this.topSelfTestTimer = -1;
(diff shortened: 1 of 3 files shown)


lukecologne:
Ok, the switching happens in the XML template for the Engine mode selector. We need to override it

lukecologne:
Done. I'm not really experienced with this XML switch stuff, so if I should change something, let me know.

AdenFlorian:
> Maybe a problem: A page triggered by a failure will not override the selected page, I don't know if this is expected behaviour or not

Please find out if it is or isn't. I recommend asking in #ecam

lukecologne:
According to the FCOM, it should replace the selected page
grafik
I'll come up with a solution

lukecologne:
Alright, should now work as expected

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.js
```diff
@@ -244,7 +245,7 @@ class A320_Neo_EICAS extends Airliners.BaseEICAS {
             }
         } else if ((ToPowerSet || !Simplane.getIsGrounded()) && !crzCond && this.minPageIndexWhenUnselected <= 2) {
             this.pageNameWhenUnselected = "ENG";
-        } else if (crzCond && !(isGearExtended && altitude < 16000) && this.minPageIndexWhenUnselected <= 3) {
```
AdenFlorian, line 247: why was this removed and not the other instances of it?

AdenFlorian:
talked with @lukecologne , I think I get it better now, they will also revert the change that made the failure pages change the selected button.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.js
lukecologne: Should I revert this as well?

lukecologne:
Alright, this should do it. There is an issue with the XMLVAR_ECAM_CURRENT_PAGE rapidly switching values after the user has cleared all memos with the CLR button, which causes some weird flickering on the ECAM button light. This has to be an issue somewhere else though, because I don't set the Simvar in my code.

camilo625:
Works great!

camilo625:
One cuestion, if you are seeing the automatic page, and switch ignition ON, you will see the engine page, but if you turn ignition off it will stay on the engine page (basically it switches your automatic page), its that normal?

lukecologne:
Yes: 
grafik

lukecologne:
See last item

AdenFlorian:
@lukecologne can you confirm whether the rapid switching issue that you mentioned exists on master or not?

lukecologne:
I'll rebase onto master, and see if the issue still exists

lukecologne:
No, seems to look fine now.

Issue #719, opened 2020-09-16
Porting aircraft changes from 1.8.3.0


UncleClapton:
This issue is to track changes made to files we have touched which were updated by the latest patch.

Double check the comments to ensure what you want to work on isn't taken by someone else already. Please comment on what you are working on before beginning a branch for it.

- [x] `PFD` instruments, AKA the reason @JackZ974 drinks (#770)
    - [x] `AirspeedIndicator.js` (#770)
    - [x] `AltimeterIndicator.js` (#770)
    - [x] `AttitudeIndicator.js` (#770)
    - [x] `ILSIndicator.js` (#770)
- [x] `FlightDirector=True` in `Climb`, `approach`, `cruise`, `final`, `taxi`, and `runway` FLTs. (#720)
    - [x] Manual fix we applied for this should be removed! (#771)
- [x] New var: `A320_NEO_CDU_START_PAGE=35` in `final.FLT`. (#720)
- [x] MANY `flight_model.cfg` changes. See Diff. (#720)
- [x] `A320_NEO_INTERIOR.xml` (#720)
    - [x] Asobo_GT_Update has new conditions. See Diff. (#720)
    - [x] `Overhead_Audio_Panel` add new override params. (#720)
    - [x] New tooltips on `AP` buttons and `ATHR`. (#720)
    - [x] Rename `ASOBO_AIRBUS_Switch_Altitude_Increment_Template` to  `ASOBO_AUTOPILOT_Switch_Altitude_Increment_Template` (#720)
- [x] `panel/panel.xml` (#720)
    - [x] `<AlwaysUpdate>True</AlwaysUpdate>` on CDU and FCU Insturments (#720)
- [x] `systems.cfg` (#720)
    - [x] Some light definitions got changed. @Snapmatics's changes might have made this obsolete though? (#720)
    - [x] Rename `autothrottle_disarm_on_idle` -> `autothrottle_managed_by_plane ` (#720)
- [x] MCDU (Various code pages changed. Hard to summarize) (#721)
    - [x] `A320_Neo_CDU_AvailableArrivalsPage.js` (#721)
    - [x] `A320_Neo_CDU_FlightPlanPage.js` (#721)
    - [x] `A320_Neo_CDU_InitPage.js` (#721)
    - [x] `A320_Neo_CDU_MainDisplay.js` (#721)
    - [x] `A320_Neo_CDU_PerformancePage.js` (#721)
- [x] `A320_Neo_UpperECAM.js` (#731)
[...]

PR #731, opened 2020-09-17
Port Upper ECAM changes from 1.8.3

lukecologne:
**Summary of Changes**

* Update UpperEcam.js with new changes (#719)
* Update `types.d.ts` with Simplane functions

**Additional context**

* Flaps are functionally the same, except for a caching feature in `Simplane.getFlapsHandleIndex()`, which doesn't seem to cause any issues. Also can be overridden if necessary.
* Changes to the throttle mode do not seem to break anything after a short testing.


lukecologne:
Actually, there may be a bit of an issue with the blue handle position indicator lagging behind a bit. Will test

nistei:
@lukecologne The lag is clearly noticeable and completely gone when overriding the cache.

UncleClapton (approved):

tyler58546 (approved):

lukecologne:
Alright, if no one has anything against it, I'll push the no cache version

UncleClapton (approved):

PR #736, opened 2020-09-17
fix(pfd/fma): merge changes from asobo-1.8.3.0

Benjozork:
See #719

**Summary of Changes**

Merges changes in `Airbus_FMA.js` from `asobo-1.8.3.0` branch.

N/A

**Additional context**

N/A

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/Airbus_FMA.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/Airbus_FMA.js
@@ -412,7 +412,9 @@ var Airbus_FMA;
                             this.setRowHighlightStyle(0, Airbus_FMA.HIGHLIGHT_STYLE.OPEN_BOTTOM);
                             this.setRowHighlightStyle(1, Airbus_FMA.HIGHLIGHT_STYLE.OPEN_TOP);
                             this.setRowText(0, "MAN", Airbus_FMA.MODE_STATE.STATUS);
-                            this.setRowMultiText(1, "FLX", Airbus_FMA.MODE_STATE.STATUS, Airbus_FMA.CurrentPlaneState.flexTemperature.toFixed(0), Airbus_FMA.MODE_STATE.ARMED);
+                            let temperatureText = Airbus_FMA.CurrentPlaneState.flexTemperature >= 0 ? "+" : "-";
+                            temperatureText += Airbus_FMA.CurrentPlaneState.flexTemperature.toFixed(0);
+                            this.setRowMultiText(1, "FLX", Airbus_FMA.MODE_STATE.STATUS, temperatureText, Airbus_FMA.MODE_STATE.ARMED);
                             break;
                         }
                     case Column1.ROW_1_2_STATE.MANMCT:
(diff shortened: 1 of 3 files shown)


Benjozork:
I left some variable names unclear because I am not sure what Asobo is checking here now.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/Airbus_FMA.js
```diff
@@ -1446,7 +1449,9 @@ var Airbus_FMA;
             return false;
         }
         GetModeState_APPNAV() {
-            if (Simplane.getAutoPilotAPPRHold() && Simplane.getAutoPilotGlideslopeHold() && Simplane.getAutoPilotApproachType() != 4) {
+            const approachNotLoadedOrSomething = !Simplane.getAutoPilotApproachLoaded() || Simplane.getAutoPilotApproachType() !== 4;
```
UncleClapton, line 1452: wat...

UncleClapton:
Honestly some of this might be completely obsolete by our own fixes. I would love to get @Curtis-VL to check over this to make sure it's all correct

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/Airbus_FMA.js
Benjozork: See the coment above @UncleClapton

Benjozork:
@UncleClapton Yeah I opened this so that we can have something to test and / or comment on. I do not think any of what they fixed is actually useful, but I opened the PR regardless so that more eyes can go over this.

UncleClapton:
There are also the smaller changes to PFD.css and PFD.js. They should probably be a part of this PR.

Benjozork:
@UncleClapton changes to `A320_Neo_PFD.css` and `A320_Neo_PFD.js` have been added to the PR.

JackZ974:
> 
> 
> There are also the smaller changes to PFD.css and PFD.js. They should probably be a part of this PR.

The only change in PFD.js applies to the horizon which is updated as well. I think this renders the "Fix Horizon" mod obsolete

UncleClapton:
Posting comparison of old and new PFDs

before (pre 1.8 patch, stock plane):


after (Post patch, + mod, + this PR):


Without this PR the positions of each element on the PFD is about the same location. The main thing the CSS changes in this PR is the position of the instruments on the PFD.

UncleClapton:
Tested pretty extensively. Looks like everything is working fine after these changes.

JackZ974:
Why is the Box around MAN TOGA only partial?

Benjozork:
Don't think it was, the only changes are to position.

UncleClapton:
@JackZ974 Box should go all around "MAN TOGA" right?

lukecologne:
I checked out the commit before this PR and the issue is also present there.

UncleClapton:
Come to think of it, yeah this has been around long before the patch.

JackZ974:
> @JackZ974 Box should go all around "MAN TOGA" right?
yes I think so

JackZ974:
> 
> 
> I have that too, not sure if it was introduced by this PR, or if it was there before

guess it was there before

UncleClapton:
TOGA is probably just missing a border style. easy enough fix. Will make (or look for) a ticket on it

PR #739, opened 2020-09-17
Fix unable to disable mach speed mode

lukecologne:
**Summary of Changes**

* Fixes the issue after the 1.8.3 update where after switching to mach mode it was impossible to go back to normal mode

Example: 

This was from cold and dark, but it happens when spawned on the runway as well. Latest master

With the fix


tyler58546:
The button works fine for me without this fix.

NathanInnes:
Works as intended.

RikvanD1997:
I had this issue yesterday after getting to cruise after the FMA switched to 'MACH' mode. I was not on the latest master build though, rather one from around 5 days ago.

jacf182:
I had the same problem with the default A320. I think it's not a problem with this mod but with the official code.

lukecologne:
Yes, pretty sure it is, but this fixes it. @tyler58546 are you sure? I know of at least two other people who have this bug, not including the ones that are in this thread

DominikTV:
I had this issue too after I used the "Jump to" option (was using the mod)

lukecologne:
See updated main post for a gif

UncleClapton:
Confirmed working

ErikNL80:
I have tested in the latest master branch. Flying at FL280. Speed switched to Mach. Cannot get it back to knots. See the image below:
[img]https://i.postimg.cc/PJx7Pww6/Naamloos.png[/img]

lukecologne:
See this: 
grafik
It might be intended that when it switches to mach automatically, you cannot go back to speed mode. Some A320 pilot should confirm this though

ErikNL80:
When descending, speed changes back to knots, and then it works again... But still, I am wondering if this is correct behaviour.

NathanInnes:
This seems to be broken again. Can't switch from Mach to speed.

lukecologne:
On the ground? or while cruising? because the only thing I fixed was being unable to switch on the ground

NathanInnes:
Mid flight at cruise

DesertRick13:
I'm a 
Mach button
Mach at altitude
IAS at altitude


747 guy, not A 320, but with a similar problem. Button is broken. The workaround is to configure a button in the menu

Benjozork:
@DesertRick13 We cannot help you with 747 problems here.

DesertRick13:
I wasn’t looking for help, I was showing a workaround to a problem that is common to both the 747 and the A 320.
Thank you

PR #744, opened 2020-09-19
Temporary fix for engine 1 fuel flow being cut off during APU shutdown

wpine215:
Fixes: engine 1 shutdown issue

**Summary of Changes**

Engine #1 no longer shuts down when APU master is turned off without fuel in center tank.

### DIFF
--- a/A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
+++ b/A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
@@ -6,11 +6,30 @@ class A32NX_APU {
         console.log('A32NX_APU init');
         SimVar.SetSimVarValue("L:APU_FLAP_OPEN", "Percent", 0);
         this.lastAPUBleedState = -1;
+
+        // FIX FOR 1.8.3 UPDATE
+        SimVar.SetSimVarValue("L:A32NX_FUEL_FIX_APPLIED", "Enum", -1);
     }
     update(_deltaTime) {
-        const currentAPUMasterState = SimVar.GetSimVarValue("FUELSYSTEM VALVE SWITCH:8", "Bool");
+        const currentAPUMasterState = SimVar.GetSimVarValue("A:FUELSYSTEM VALVE SWITCH:8", "Bool");
         const apuFlapOpenPercent = SimVar.GetSimVarValue("L:APU_FLAP_OPEN", "Percent");
         const APUPctRPM = SimVar.GetSimVarValue("APU PCT RPM", "percent");
+
+        //////////////////////////
+        // FIX FOR 1.8.3 UPDATE //
+        //////////////////////////
+        if (currentAPUMasterState === 1 && APUPctRPM > 10 && APUPctRPM < 90) {
+            SimVar.SetSimVarValue("L:A32NX_FUEL_FIX_APPLIED", "Enum", 0);
+        }
+        const center_qty = SimVar.GetSimVarValue("FUEL TANK CENTER QUANTITY", "Gallons");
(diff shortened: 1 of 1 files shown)


Benjozork:
## DISCLAIMER

This is an EXPERIMENTAL FIX. We need your feedback and that's why we make it public before it's tested, but PLEASE do not yell at us if the fix does not work 100% !

AdenFlorian (approved):
Code looks fine, ready for testing

sharksfan166269:
Did two real quick start-ups from cold and dark and then did a circuit at Copenhagen, one with the stable build, one with this fix, the fix, based on one flight with it, seems to be working as designed. When the APU shut off, engine 1 did not shut down, whereas it did with the stable version. This could obviously change during a longer flight or something, but I just wanted to say, initially it looks good.

A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
```diff
@@ -6,12 +6,31 @@ class A32NX_APU {
...
+        // FIX FOR 1.8.3 UPDATE
+        SimVar.SetSimVarValue("L:A32NX_FUEL_FIX_APPLIED", "Enum", -1);
     }
     update(_deltaTime) {
-        const currentAPUMasterState = SimVar.GetSimVarValue("FUELSYSTEM VALVE SWITCH:8", "Bool");
+        const currentAPUMasterState = SimVar.GetSimVarValue("A:FUELSYSTEM VALVE SWITCH:8", "Bool");
         const apuFlapOpenPercent = SimVar.GetSimVarValue("L:APU_FLAP_OPEN", "Percent");
         const APUPctRPM = SimVar.GetSimVarValue("APU PCT RPM", "percent");
 
+        //////////////////////////
```
ymnoor21, line 18: Wouldn't it be better to check the `center_qty` first and only if it's below 40 gallons you can apply the fix?

A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
```diff
@@ -6,12 +6,31 @@ class A32NX_APU {
         console.log('A32NX_APU init');
         SimVar.SetSimVarValue("L:APU_FLAP_OPEN", "Percent", 0);
         this.lastAPUBleedState = -1;
+
+        // FIX FOR 1.8.3 UPDATE
+        SimVar.SetSimVarValue("L:A32NX_FUEL_FIX_APPLIED", "Enum", -1);
```
ymnoor21, line 11: I am not familiar with the SimVar, so please ignore if it's not a valid question. Why not use boolean instead of Enum type here?

AdenFlorian:
Tested and seems to work to me!

A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
Benjozork: I believe that is what we are doing

A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
Benjozork: The fix simvar is a 3-state value.

AdenFlorian: i agree boolean would be better, but its a temp fix and will be removed soon hopefully

A32NX/html_ui/Pages/A32NX_Core/A32NX_APU.js
ymnoor21: Yes, but all those expressions are moot if you find out that center_qty is greater than 40. So you should check center_qty first as it seems like the primary condition for this fix.

Issue #632, opened 2020-09-13
[BUG] Meta-issue for brake temperature calculation


Benjozork:
**Mod Version**

brake-temp-improvements (8308cdba)

**Describe the bug**

 - [x] Move core logic to A32NX_Core (PR ~~#631~~ #749)
 - [x] Move "brakes hot" logic to A32NX_Core
 - [x] Fix brake temperature not being affected by autobrake application (Issue #624, PR ~~#631~~ #749)
 - [x] Make heat-up a function of wheel RPM in addition to brake position (PR #749)
 - [x] Calculate heat-up and cool-down separately for different wheel sets (PR #749)
 - [x] Use actual heat transfer coefficient math (PR #749)
 - [x] Use actual ambient temperature, not 45C (PR #749)
 - [x] Only show in increments of 5 (PR #749)
 - [x] Clamp displayed temperature above 0 (PR #786)

**To Reproduce**

N/A

**Expected behavior**

N/A

**Actual behavior**

N/A

**References**

TODO

**Additional context**

N/A

Benjozork:
@lukecologne, let's discuss here !

Do you think it would be interesting to use actual heat transfer / friction equations with data about the materials used in the brakes ?

lukecologne:
Unfortunately I don't think it's that easy to model friction heating accurately. Heat transfer between the brakes shouldn't be too difficult if we have the material it's made of (carbon I think?). And the convective cooling to the air I also wouldn't know an equation off the top of my head, but I think that would be simpler than the friction heating. Radiative energy transfer can be neglected I think.

Benjozork:
Yeah I think that sounds good. Do you have any prototype code for this ?

lukecologne:
I had something basic, but my PC died yesterday, currently on my laptop, but I don't have it here. Maybe I'll have it up later today or tomorrow

lukecologne:
Another question, is there even heat transfer between the two brakes on one wheel? Or is each of the 4 systems disconnected?

UncleClapton:
@Benjozork 2 problems I've identified:

1. Applying parking break stops all temp decay, even when the aircraft is stationary. This isn't really all that realistic.
2. Brake temperature decay is based on `deltaTime`, so low screen refresh rate slows decay significantly. This MAY have been helped by moving the logic to `a32nxcore`, but not completely alleviate it. (see #529 for more info) I may have a universal solution for core modules coming up soon though.

lukecologne:
[...]

PR #631, opened 2020-09-13
fix(core/break-temp): move and improve brake temp logic

Benjozork:
**Summary of Changes**

 - Moved brake temperature logic to A32NX_core
 - Make brake temperature function of brake position (closes #624)
 - Increased brake temperature heat up rate

**Screenshots (if necessary)**

N/A

**Additional context**

N/A

PR #749, opened 2020-09-19
Brake temp improvements

lukecologne:
Fixes #[issue_no]

**Summary of Changes**

* Move brake temp logic into A32NX core
* Improve heatup and cooldown formulas
  * Make heatup dependent on amount of brakes applied, and wheel RPM
  * Cooldown is now based on the temperature difference to the ambient temperature
  * Cooldown is also now a function of airspeed, and if the gear is deployed or not.

**Additional context**

This is a "continuation"/replacement of #631 by @Benjozork.
**Additional testing of the cooldown and heatup speed is required and probably needs to be tuned**
See #632 for more discussion

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.html
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.html
@@ -80,6 +80,6 @@
 <script type="text/html" import-script="/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_ECAMGauge.js"></script>
 <script type="text/html" import-script="/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A32NX_LocalVarUpdater.js"></script>
 <script type="text/html" import-script="/Pages/A32NX_Core/A32NX_APU.js"></script>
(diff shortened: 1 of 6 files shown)


lukecologne:
Alright should be good to test for a start. There are few things I'd still like to do though, but those will be only visual

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A320_Neo_EICAS.html
```diff
@@ -81,4 +81,5 @@
 <script type="text/html" import-script="/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/A32NX_LocalVarUpdater.js"></script>
 <script type="text/html" import-script="/Pages/A32NX_Core/A32NX_APU.js"></script>
 <script type="text/html" import-script="/Pages/A32NX_Core/A32NX_Core.js"></script>
+<script type="text/html" import-script="/Pages/A32NX_Core/A32NX_BrakeTemp.js"></script>
```
AdenFlorian, line 84: shouldn't this go above the line that imports A32NX_Core.js?

AdenFlorian: maybe it still works, but i think it would look nicer above it, to better show that it gets called from the core and not the other way around

AdenFlorian (approved):
LGTM! Ready to test!

hansmeller:
I tested it and everything works just fine.

PR #753, opened 2020-09-19
Engine fan rotation

lukecologne:
**Summary of Changes**

* Creates custom fan animation templates that use N1 instead of some other wrong RPM. 
  * Because of that, startup and shutdown are now animated properly
  * Windmilling is also now a thing

https://streamable.com/5uj76f

**Additional context**

* Startup and shutdown are way too quick in my opinion, but that is being worked on already
* There is a problem where at startup and shutdown, the animation stops before the engine spools up (see video), that is an issue with the game, I just take the N1 value as given to us


camilo625:
The animation seems to be working but it only lasts a second before transitioning to the blurry texture. Seems different from your video. 
Example: https://streamable.com/ynhsjy

lukecologne:
Have you run build.py? If you haven't it probably uses the original files

camilo625:
> Have you run build.py? If you haven't it probably uses the original files

I downloaded the hole branch and just copy/paste the A32NX folder, the same way i did for all other PRs

lukecologne:
That only works if no new files have been added in that PR. If they are new files, you will need to run build.py locally

camilo625:
> That only works if no new files have been added in that PR. If they are new files, you will need to run build.py locally. If that PR has been merged, the CI will do that for you, that's why you can just download master

Alright, will try that, thanks!

sharksfan166269:
Windmilling looks real cool on this, that works as designed. However I noticed a few things, so, the engine shutdown is rather quick, from spinning at idle, to not moving (with no winds), took about 10 seconds, and it didn't carry any momentum, so it appeared rather abrupt. This might be an issue with the engine shutdown speed, but not sure. Also, when the engine is windmilling and the engine startup sequence is initiated, the windmilling stops abruptly (I believe it's not supposed to stop windmilling, the N1 fan just keeps keeps the momentum going into the startup), about 2 seconds, and it's much more noticeable when the winds are high and windmilling is pretty fast. Also, after engine shutdown, windmilling does not seem affect the engine, until a few seconds after shutdown is fully complete, regardless of the wind. Just to clarify, I did also run build.py prior to running MSFS and didn't notice blurry textures as mentioned above, but this is good stuff, just with a few issues :)

lukecologne:
Yes, i know. See additional context on the original post. nothing i can do about all these things unfortunately, that's a default issue. The startup and shutdown times will hopefully be changed in the future, so that quick stopping will hopefully be fixed soon. Unfortunately, i don't think we can do much about the stopping between windmilling and shutdown/startup.

sharksfan166269:
> 
> 
> Yes, i know. See additional context on the original post. nothing i can do about all these things unfortunately, that's a default issue. The startup and shutdown times will hopefully be changed by us in the future, so that quick stopping will hopefully be fixed soon. Unfortunately, i don't think we can do much about the stopping between windmilling and shutdown/startup.

Ah okay, this is why you read the whole post first lol. Okay, apart from that, works as designed 👍

JackZ974:
Nice work.
From your video, the counterclockwise rotation of the fan is weird and seems to be incorrect IMHO. 
Seems that after the fan picks up speed, the fan then "turns" clockwise (which is the correct rotation direction)  again, which adds to the weirdness. 

As for the windmilling, the rotation speed should be much lower IMO, as the fan blades are quite large, and even if made out of carbon fiber, it takes quite a strong wind to make them start turning IRL. 
I would say that 80% of the time the fan will stay stationnary when engines are shutdown, except for some residual rotation, but in any case of windmilling, it will be very slow and decaying after a while in that case.

lukecologne:
1. Not sure what you mean with counterclockwise rotation? The fan always turns in the same direction, clockwise. Maybe it just looks that way when it's spinning fast.
2. The windmilling speed is given by the sim, I can just lower the speed in total, so if I lower the windmilling speed, I also have to lower the speed at which the blades rotate when starting.

JackZ974:
From your video at the beggining of the engine start, the fan appears to turn to the left (anticlockwise) then, as it speeds up, the animation makes the fans seems to be turning to the right (clockwise)

lukecologne:
Oh, that. That is just a visual artifact, it appears to be rotating backwars because it is rotating quickly. I'll see if I can find a video if it in RL

A32NX/ModelBehaviorDefs/Asobo/Exterior.xml
```diff
@@ -1695,4 +1695,100 @@
...
+    </Template>
+    <Template Name="A32NX_ENGINE_Turbine_Slow_Visibility_Template">
+    <UseTemplate Name="ASOBO_GT_Visibility">
+        <VISIBILITY_CODE>
+            (A:TURB ENG N1:#ID#, Percent) #MIN_N1_PCT_FOR_SLOW# &gt;
+            (A:TURB ENG N1:#ID#, Percent) #MIN_N1_PCT_FOR_BLUR# &lt; and
+        </VISIBILITY_CODE>
+    </UseTemplate>
+    </Template>
+    <Template Name="A32NX_ENGINE_Turbine_Blurred_Visibility_Template">
```
AdenFlorian, line 1779: Where is this template getting used?

A32NX/ModelBehaviorDefs/Asobo/Exterior.xml
```diff
@@ -1695,4 +1695,100 @@
...
+        <UseTemplate Name="ASOBO_GT_Visibility">
+            <VISIBILITY_CODE>(A:TURB ENG N1:#ID#, Percent) #MIN_N1_PCT_FOR_BLUR# &gt;</VISIBILITY_CODE>
+        </UseTemplate>
+    </Template>
+        <!-- Base visibility template for turbines -->
+    <Template Name="A32NX_ENGINE_Turbine_Visibility_Base_Template">
+        <Condition NotEmpty="NODE_ID">
+            <True>
+                <Component ID="#NODE_ID#" Node="#NODE_ID#">
+                    <UseTemplate Name="A32NX_ENGINE_Turbine_#STATE#_Visibility_Template">
```
lukecologne, line 1789: Here

lukecologne: the state gets replaced with either "blur", "slow", or "still"

lukecologne: I know, it's pretty convoluted, but that's how asobo did it, and I didn't really want to invent a whole new template

AdenFlorian (approved):
LGTM!

lukecologne:
@AdenFlorian You said yesterday in your stream that you were worried people might edit the existing Asobo shared templates I copied in in this PR. I might have a solution for that, I'm testing if it works right now.

lukecologne:
Ok, it works. I'll push it now

lukecologne:
I basically added my custom template to a new file, instead of the shared Asobo file, and included that in the model XML. Maybe it would be a good idea to do that to all our other changed templates as well?

AdenFlorian (approved):

NathanInnes:
Sitting at gate with 2kt wind and it's  spinning, surely this wouldn't happen?

lukecologne:
that's what the sim provides, and if it make it slower, startup and shutdown would look worse imo.

PR #757, opened 2020-09-19
fix(ecam/ewd): do not show REV and thrust limit in reverse

Benjozork:
**Summary of Changes**

 - Do not show "REV" and thrust limit % when thrust reversers are active

image

**Additional context**

re. Ares#3394 on discord and F-COM

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
@@ -1026,7 +1026,7 @@ var A320_Neo_UpperECAM;
                                                         return SimVar.GetSimVarValue("L:A32NX_TO_CONFIG_NORMAL", "Bool") == 1;
                                                     })
                                                  ],);
-            
+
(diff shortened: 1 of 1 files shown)


AdenFlorian (approved):
ready to test

Coralec:
I believe this is a result of the proper performance of the thrust limit % indicator, but if the aircraft is spawned on the runway and the thrust has not yet been set to FLX or TO/GA for takeoff then the thrust limit will be stuck at 100%. When this is the case, then the thrsut limit % indicator will remain even if reverse thrust is set. To test this just spawn on the runway and set reverse thrust before ever beginning your takeoff roll. I do not know if this is working as intended.

Besides this however, the thrust limit % indicator does disappear when reverse thrust is set. So this is working properly. I'm just wondering about it being stuck on at 100% before takeoff.

Coralec:
image
Here is what I am looking at. As you can see the thrust limit % is stuck at 100% (as I think it would be for takeoff). But as a result it remains illuminated even if reverse thrust is set.

Coralec:
image
Here is the same ecam after I moved the thrust lever up to FLX and then engaged reverse thrust. As you can see the thrust limit % indicator is now gone.

Benjozork:
@Coralec yeah that's an issue I am aware of but I didn't address it with this PR.

Coralec:
Then this pull seems to be working as intended. Obviously there is no reason to be setting reverse thrust before takeoff so this isn't really a big issue anyway.

PR #720, opened 2020-09-16
port CFG and XML file changes from 1.8.3.0

UncleClapton:
**Summary of Changes**

Update FLTs, Update Flight Model, Update Systems.cfg, update XMLs

**Additional Context**

Us in Voice chat have started chunking the work out to certain sections. This is just the first in a series of PRs to bring the codebase up to date. QA testing these changes are going to be diffiuclt, so we might need to just rely on code review for now, and QA the whole project once completed.

PR #762, opened 2020-09-19
Fix incorrect slat position after 1.8.3

### LINKED ISSUE #751: [BUG] Slat position incorrect
**Mod Version**

Newest Master

**Describe the bug**

Slats are in a wrong position in every flaps config

**To Reproduce**

1. Spawn on runway
2. Look at upper ECAM

**Expected behavior**

grafik

**Actual behavior**

grafik
grafik
grafik
grafik

**References**

**Additional context**

* Does not occur in c1b040ad5ddb0eaa1fbf225fbdd8048eb29c379d
* I think this actually represents the slat position, so the indicator isn't wrong, but the actual slat position is.

lukecologne:
Fixes #751

**Summary of Changes**

* Fix incorrect Slat position introduced by #720 after update 1.8.3

#720 also changed slats lift scalar, should be tested

Source: 
grafik
Before: 
grafik


AdenFlorian (approved):
ready to test!

Coralec:
Tested. The slat position is now working properly.

lukecologne:
Everyone who is testing this: make sure to also check the lift caused by the slats/flaps, because Asobo changed the slat lift/drag scalars as well, and I am just setting the slat postion back to normal, while leaving the changed scalars. Maybe we will have to reset those as well.

AdenFlorian:
@lukecologne how would one test the lift?

lukecologne:
I don't know. It's especially a problem since they changed the flaps scalars as well, so I think the only way to test the slats indivdually may be config 1. I was hoping a IRL pilot could comment on the "feel"

Coralec:
Any issues with the lift performance of the slats/flaps would really be a separate issue. We can keep an eye out for issues with lift performance in future testing but at least this pull fixes the indicators.

AdenFlorian:
I had Ares take a look at this PR, and he's not even sure how he wouold test it, so I'll go ahead and merge it. There's so much wrong with how the plane flies right now, that I think we have to wait until a lot of things are fixed and then we'll start doing flight model passes.

DABUS320:
I have no way to test this issue... I cannot tell you if it feels right since its a virtual game.

PR #767, opened 2020-09-20
Improved autobrake logic

### LINKED ISSUE #255: Braking while taxiing turn autobrake off
If you break for traffic while taxiing autobrakes are turned off.

Venorcis:
Fixes #255

**Summary of Changes**
Set the autobrakes_disabled_on_braking to 0 in the systems.cfg, which prevents the autobrakes from disarming when manually braking. IMO this is a sane default that fixes the taxi-for-takeoff issues right now, and if the autobrakes do need to disabled when braking manually (e.g. after landing), that's probably easier to code with this default.


Snapmatics:
I think it will not get disabled until you reach 60 kts and above. Not sure.

Snapmatics:
Anyways, I am about to test it right now.

Snapmatics:
Pressing the brakes, I am at 100+ kts and it's not activating.
image

Venorcis:
> Pressing the brakes, I am at 100+ kts and it's not activating.
> image

I'm not sure what you're testing here. This is probably unchanged behavior wrt. vanilla and/or master? I made sure they don't DISable when you're taxing for takeoff, i.e. fixing #255.

Snapmatics:
How am I suppose to cancel my take off if something happens? The Auto Brakes won't work while taking off.

Snapmatics:
> > Pressing the brakes, I am at 100+ kts and it's not activating.
> 
> I'm not sure what you're testing here. This is probably unchanged behavior wrt. vanilla and/or master? I made sure the autobrakes don't DISable when you're taxiing for takeoff, i.e. fixing #255.

_Autobrakes should only activate above 72kts and spoilers armed_

A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
```diff
@@ -151,7 +151,7 @@ wing_dihedral = 7 ; Dihedral angle Lambda (DEGREES)
 wing_incidence = 0 ; Wing incidence (DEGREES)
 wing_twist = -2 ; Wing twist epsilon (DEGREES)
 oswald_efficiency_factor = 0.75 ; Wing Oswald efficiency factor e (non dimensional)
-wing_winglets_flag = 0 ; Has winglets true/false
```
UncleClapton, line 154: This should probably be it's own issue. We should really investigate the effect changes to the flight model have extensively before deciding to adopt the change.

UncleClapton (changes requested):

UncleClapton:
> > > Pressing the brakes, I am at 100+ kts and it's not activating.
> > 
> > 
> > I'm not sure what you're testing here. This is probably unchanged behavior wrt. vanilla and/or master? I made sure the autobrakes don't DISable when you're taxiing for takeoff, i.e. fixing #255.
> 
> _Autobrakes should only activate above 72kts and spoilers armed_

This is indeed a different problem from what this PR addresses. This flag shouldn't affect autobrake activation on rejected takeoff.

KonPed:
Hello everyone,

Just tested that and by tapping the brakes while taxing again it disables the autobrakes which i had set to max for RTO.
No Taxi
No Taxi
Taxing
taxi

A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
Venorcis: Good point. I’ll revert it here and put it in a separate pull for further testing.

lukecologne:
I can confirm the issue that autobrakes do not engage at RTO in latest master. I might have a fix ready

NathanInnes:
I wouldn't set AUTOBRAKE disabled on parking to 0. Reason being is it doesnt disable once decel is complete and you take over with manual braking.

A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
Venorcis: Done @UncleClapton :)

Venorcis:
> Hello everyone,
> 
> Just tested that and by tapping the brakes while taxing again it disables the autobrakes which i had set to max for RTO.
> No Taxi
> No Taxi
> Taxing
> taxi

It works for me, so I'm not sure what the difference is. I'll do some more testing

Venorcis:
> I wouldn't set AUTOBRAKE disabled on braking to 0. Reason being is it doesnt disable once decel is complete and you take over with manual braking.

I actually didn't notice any change on landing, I think the autobrake ending on landing triggers automatically. However, even if it doesn't, I still think this default is better (probably easier to add some code for landing than to override the game disabling it when taxiing). I'll do some more testing though.

NathanInnes:
I was messing a lot with it last night before this PR was put through, it outright disables you from disabling the AUTOBRK with toe brakes. Not the best solution, I don't imaging it's too difficult to make a good fix. I will look into it

UncleClapton:
It looks like there's drawbacks to this approach. We might want to try to tackle this in JS instead, overriding the default asobo behavior. @Venorcis Would you be interested in trying to address this in JS instead?

Venorcis:
> It looks like there's drawbacks to this approach. We might want to try to tackle this in JS instead, overriding the default asobo behavior. @Venorcis Would you be interested in trying to address this in JS instead?

Yeah I could give it a try, but maybe @NathanInnes is close to a solution already?

NathanInnes:
No not close, you wanna give it a shot @Venorcis?

Venorcis:
> No not close, you wanna give it a shot @Venorcis?

Done. I think I'll include the autobrakes actually triggering when aborting takeoff too.

Venorcis:
Ok nvm, I need to rethink this, the BRAKE POSITIONS already change with the autobrakes themselves...

lukecologne:
grafik
This is the autobrake logic. You can see that autobrake activation depends on ground spoiler deployment, and has a delay when not set on the MAX setting. Now, since the brake position Simvar can set, maybe it would be worth implementing our own autobrake logic, with the proper decelerations, and just disabling autobrake in the config files?

Venorcis:
> Ok nvm, I need to rethink this, the BRAKE POSITIONS already change with the autobrakes themselves...

Ok, this is all not as easy as it first seemed. There don't seem to be separate (accessible) variables for checking the footbrakes vs. the autobrakes; just the brake positions in general (which apply in both cases). Moreover, a template similar to the "AUTOPILOT_Update_Autopilot_Match_Simvar" will have to be created to make sure the enabled autobrake button's state changes, otherwise it will just overrule the change to the Simvar each second. I'm wondering now if fixing this the other way around (i.e. re-setting the XMLVAR_Autobrakes_Level after it automatically resets when taxiing) would work.

Triggering the autobrakes for an aborted take-off is actually relatively easy to realise though...

Benjozork:
Hello, please take note that we have started including a CHANGELOG.md file that requires every PR to add a change-log line that describes its changes. As your PR is not merged yet, we kindly ask that you add this information following the included directives in the file.

Disclaimer: This is a copy-and-pasted message.

AdenFlorian:
please reopen when ready for review again.

PR #764, opened 2020-09-20
Fix #333 No Gear Up on Ground

NathanInnes:
Fixes #333 

**Summary of Changes**

Gear lever is no longer able to come up when on ground.

**Screenshots (if necessary)**

**Additional context**

Discord username (if different from GitHub): Sabes#8668

PR #769, opened 2020-09-20
[STEERING] Changes in rudder steering speeds authority and max steering angle (Update flight_model.cfg)

JackZ974:
Changed the speeds and angle of steering authority to be closer to the real ones, according to FCOM values. The full range of speed authority cannot be tweaked correctly though in the absence of a real Nose Wheel Steering system independent logic from the rudder steering system. So the following tradeoff:
-1 Modified the speed limit for gear extension : 260 kts
-2 Modified the max speed for full deflection as per NWS settings (20 kts)
Note: The NWS is supposed to be still operative until 80 kts but since this is not independent, the decrease in steering authority is too slow, so the threshold is set to 40 kts, as this is the rudder max deflection speed, in order to avoid too tight turns at high speed during taxi.
-3 The min deflection rudder at 40kts has been set to 6 degrees which is the max rudder authority angle if using the rudder only. This angle should decrease with speed as well (from 6degrees max  down to 0 degrees at 130 kts) but since NW and rudder steering cannot be separated for now this decrease from 6 to 0 with speed increase cannot happen. (see proposed curve in red)
This PR needs feedback from users for further tweaking

Fixes #[issue_no]
Follow up of #764 (gear not up on gnd) that it includes already
**Summary of Changes**

image

**Additional context**

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
@@ -33,7 +33,7 @@ gear_locked_on_ground = 1 ; Defines whether or not the landing gear handle is lo
 gear_locked_above_speed = 439 ; Defines the speed at which the landing gear handle becomes locked in the up position. (-1 = Disabled) ===>260 kts or 439 ft/sec (was -1)
 max_speed_full_steering = 33.7 ; Defines the speed under which the full angle of steering is available (in feet/second).==> 20 kts or 33.7 ft/sec (was 8)
 max_speed_decreasing_steering = 67.5 ; Defines the speed above which the angle of steering stops decreasing (in feet/second). ==> 40 kts or 67.5 ft/sec (was 50)
-min_available_steering_angle_pct = 0.08 ; Defines the percentage of steering which will always be available even above max_speed_decreasing_steering (in percent over 100)
+min_available_steering_angle_pct = 0.08 ; Defines the percentage of steering which will always be available even above max_speed_decreasing_steering (in percent over 100) ===> 6 degrees (was 0.2 or 15 degrees)
 max_speed_full_steering_castering = 20 ; Defines the speed under which the full angle of steering is available for free castering wheels (in feet/second).
 max_speed_decreasing_steering_castering = 40 ; Defines the speed above which the angle of steering stops decreasing for free castering wheels  (in feet/second).
 min_castering_angle = 0.05236 ; Defines the minimum angle a free castering wheel can take (in radians).
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
@@ -33,7 +33,7 @@ gear_locked_on_ground = 1 ; Defines whether or not the landing gear handle is lo
 gear_locked_above_speed = 439 ; Defines the speed at which the landing gear handle becomes locked in the up position. (-1 = Disabled) ===>260 kts or 439 ft/sec (was -1)
 max_speed_full_steering = 33.7 ; Defines the speed under which the full angle of steering is available (in feet/second).==> 20 kts or 33.7 ft/sec (was 8)
 max_speed_decreasing_steering = 67.5 ; Defines the speed above which the angle of steering stops decreasing (in feet/second). ==> 40 kts or 67.5 ft/sec (was 50)
-min_available_steering_angle_pct = 0.08 ; Defines the percentage of steering which will always be available even above max_speed_decreasing_steering (in percent over 100) ===> 6 degrees (was 0.2 or 15 degrees)
+min_available_steering_angle_pct = 0.08 ; Defines the percentage of steering which will always be available even above max_speed_decreasing_steering (in percent over 100) ===> 6 degrees or 0.08% of 75 degrees max deflection (was 0.2 or 15 degrees)
 max_speed_full_steering_castering = 20 ; Defines the speed under which the full angle of steering is available for free castering wheels (in feet/second).
 max_speed_decreasing_steering_castering = 40 ; Defines the speed above which the angle of steering stops decreasing for free castering wheels  (in feet/second).
 min_castering_angle = 0.05236 ; Defines the minimum angle a free castering wheel can take (in radians).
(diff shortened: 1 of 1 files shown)


UncleClapton (approved):
LGTM! ready for testing

Benjozork:
## Testing report

 - [ ] Modified the speed limit for gear extension : 260 kts - Does not seem to work
 - [x] Modified the max speed for full deflection as per NWS settings (20 kts) - Seems to work
 - [ ] The min deflection rudder at 40kts has been set to 6 degree - Not sure how to test

AdenFlorian:
@JackZ974 did the gear level lock thing work when you tested it?

Coralec:
I can still deploy the gear at 300 and even 350kts. The limit for gear extension is not working.

JackZ974:
Hmm seems that the gear speed limit value is not in ft/sec, like the others parameters but in kts. I changed again the limit from 439 to 260
Sorry for the trouble, but my PC is packed and I can no longer test it in the sim.

lukecologne:
Is the gear lever actually locked in the up position, like when on ground, or does the gear just not deploy? If it is not locked, maybe it would be better to edit the XML so that you can move the lever, but it doesn't deploy

JackZ974:
> 
> 
> Is the gear lever actually locked in the up position, like when on ground, or does the gear just not deploy? If it is not locked, maybe it would be better to edit the XML so that you can move the lever, but it doesn't deploy

TBH, i don't know. I have never tested this issue IRL (this is a case of being fired!), and in the real sim we don't have the time to play with such things during our sessions. 
The FCOM only specifies that the Gear cannot be put into the UP position when on ground but doesn't say anything about the DOWN position. From what I understand the valve is electrically isolated when speed is above 260kts, so  I think that the lever can still be operated down and left down, but if the speed is >260kts nothing happens, until the speed goes below this upper limit.

lukecologne:
I'll se what I can do with the XML if you want me to. We can also just lock in the up position, that is up to you, I don't know what would the most realistic solution

Coralec:
The gear now properly locks at 260kts. The gear lever becomes immovable which is probably not realistic. I spoke with one of our other irl A320 pilots and he also thinks that is probably not realistic. It is up to the devs whether they want to implement that change in its current state.

I will do some more testing on the NWS and rudder deflection and post my results shortly.

JackZ974:
Thanks @Coralec for your feedback. Glad to see that the change of Unit solved the problem of Gear down speed limit the comment didn't specify the unit expected. 
@lukecologne if you want to have a go at the XML route, feel free to do so. Anyways we are not supposed to move the lever above the VLO speed, aren't we?

JackZ974:
Talking about the gear, i noticed an unrealistic behaviour in the wheel page. 
The "In transition" red triangles are not supposed to flash IRL

lukecologne:
I was changing something on the wheel page anyway, I'll change that as well

Coralec:
I've tested the NWS inhibition at >80kts and that is working great. I'm not sure what you mean with the 3rd change to rudder deflection. If that change is just made to make the nose wheel control less touchy then I think it is working well. But I'm not sure what to be testing with that one.

AdenFlorian:
I tried it out in the flight factor a320, when going over 260 kts, you can still move the lever down, but the gear won't deploy, do doors wont even open. It displays L/G CTL in amber in the middle of the WHEEL page, and theLDG lights says UNLK in red.

If you leave the gear lever down, and then slow down, immediately after going under 260, the gear will deploy.

There's more things that happen on the upper ecam if you leave the lever down for a while, but that doesnt need to be in this ticket.

When it's not practical to get an IRL answer, I would go with how existing payware does it. 

image

AdenFlorian:
If you agree @JackZ974 , you can remove the gear lock changes out of this PR.

JackZ974:
So your findings confirm what I intuitively thought when asked. Good! The FCOM is somewhat scarce of informations in some sectors, and it's difficult to get all the answers from there.
> If you agree @JackZ974 , you can remove the gear lock changes out of this PR.
No problem, as it is not realistic. An XML based solution is evidently better suited, though I'll leave this better solution to expert hands.

JackZ974:
> 
> 
> I've tested the NWS inhibition at >80kts and that is working great. I'm not sure what you mean with the 3rd change to rudder deflection. If that change is just made to make the nose wheel control less touchy then I think it is working well. But I'm not sure what to be testing with that one.

the rudder steering authority should decrease from +/-6degrees at or below 40 kts down to 0 degrees at 130 kts, since at this speed the rudder surface control is way more efficient.
Problem is if I implement this solution the rudder steering maximum limit has to be set at +/-6 degres instead of +/-75 degrees when using the Tiller, as there is no separate system for the Tiller in MSFS. 
Such a small angle is impossible to use for taxi and parking, so I implemented an hybrid solution where the rudder/NWS is able to steer form 75 at low speeds down to 6 degrees above 40 kts, even though the rudder travel should decrease from 6 to 0
This should improve the steering control at intermediate speeds but is still too sensitive at high speed.
Unless a complete NWS system is created, not much we can do at the moment.

AdenFlorian:
sounds like its an improvement, we can adjust later if needed

Villason:
> > I've tested the NWS inhibition at >80kts and that is working great. I'm not sure what you mean with the 3rd change to rudder deflection. If that change is just made to make the nose wheel control less touchy then I think it is working well. But I'm not sure what to be testing with that one.
> 
> According to the FCOM, the rudder steering authority should decrease from a fixed +/-6degrees at or below 40 kts down to 0 degrees at 130 kts, since at this speed the rudder surface control is way more efficient.
> Problem is if I implement this solution the rudder steering maximum limit has to be set at +/-6 degres instead of +/-75 degrees when using the Tiller, as there is no separate system for the Tiller in MSFS.
> Such a small angle is impossible to use for taxi and parking, so I implemented an hybrid solution where the rudder/NWS is able to steer from +/-75degrees at low speeds, down to 6 degrees above 40 kts, even though the rudder travel should continue to decrease from 6 to 0 at 130 kts (see the proposed red curve).
> This should improve the steering control on runways at low and intermediate speeds but might be still too sensitive at high speed.
> Unless a complete NWS system is created, not much we can do at the moment.
> 
> So if you feel aftyer testing that the steering control is still too sensitive at high speeds above say 100 kts, either the speed threshold of 40kt or the minimum angle of 6 degrees can be further reduced, but to the detriment of better control at intermediate speeds.

Given the constraints we have, I think we need to have the cut-out to 6 degrees at 80 knots, not at 40. That way we be closer to real life steering performance with the compromise of using pedals inputs, but at least we could taxi properly.

PR #770, opened 2020-09-20
port Asobo changes from 1.8.3.0 for PFD Instruments

UncleClapton:
**Summary of Changes**

* Updates instrument files (where appropriate)
* Removes a ton of code for other aircraft. All of these files are loaded independently, and not used by the other aircraft. There is no need for us to maintain and hold onto this code really. 
* Introduce `A32NX_Avionics`, a package that holds customized Avionics classes.


Benjozork:
Testing this on request of @UncleClapton

Benjozork:
LGTM

UncleClapton:
To QA: this one should be easy to test for the most part. Just make sure the PFD Altimeter, speedometer, and ILS overlay work as expected.

hansmeller12345:
Everything works as it should! Good work!

PR #777, opened 2020-09-20
Make the ADIRS align timer respect the in-game time.

robteix:
**Summary of Changes**

The ADIRS timer was using real world time to update its timer, which means it completely disregarded sim rate. This patch makes the timer take acceleration of the simrate in account. In essence, the timer now follows the same time as the aircraft clock.

As a tiny optimization bonus, we only calculate the delta when ADIRS is in alignment.

**Additional context**

The original code used `Date.now()` which returns the current time from the computer running the simulation. With this patch, we'll instead use the simvar `ZULU TIME` which holds the time _in sim_. We use Zulu Time instead of the sim's local time to avoid having to deal with things like timezone changes (i.e. daylight time.)

There are other similar real world timers instead of in-game time elsewhere in the code (e.g. the 40s self-test timer, and the flight phase reset) and we could also change those in subsequent PRs if desired.

Discord username (if different from GitHub): GrinningLlama

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
@@ -881,12 +881,6 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
     }
     updateADIRS() {
 
-        //Get the time since last update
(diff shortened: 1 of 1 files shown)


DoToAdventures (approved):
LGTM, needs testing

AdenFlorian:
we don't want to use zulu time for these kinds of things, because we would get incorrect behavior when changing the in game time, like if you turned it back an hour. We're working on a global solution which will take the sim rate into account. For the future, i recommend creating an issue before starting work on something, to give people a chance to give their input earlier.

PR #779, opened 2020-09-20
Make GPS and Position Monitor CDU pages dynamic, cosmetics/realism

lhoenig:
**Summary of Changes**

This PR makes the GPS and Position Monitor pages of the CDU Data section dynamic, meaning they will continuously update to display the current values. It also includes realism and cosmetics improvements for these pages.

**Additional context**

lukas_#6140

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
@@ -1,5 +1,5 @@
 class CDUGPSMonitor {
-    static ShowPage(mcdu) {
+    static ShowPage(mcdu, merit1, merit2, sat1, sat2) {
         let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
         let TTRK = SimVar.GetSimVarValue("GPS GROUND MAGNETIC TRACK", "radians") || "000";
         let GROUNDSPEED = SimVar.GetSimVarValue("GPS GROUND SPEED", "Meters per second") || "0";
@@ -12,7 +12,20 @@ class CDUGPSMonitor {
 
         var UTC = `${hours.toString().padStart(2, "0") || "00"}:${minutes.toString().padStart(2, "0") || "00"}:${seconds.toString().padStart(2, "0") || "00"}`
 
+        if (typeof merit1 == 'undefined') {
+            merit1 = Math.floor(Math.random() * 10) + 40;
+            merit2 = Math.floor(Math.random() * 10) + 40;
+            sat1 = Math.floor(Math.random()*5) + 8;
+            sat2 = Math.floor(Math.random()*5) + 8;
+        }
+
         mcdu.clearDisplay()
+        // ideally, this would update with the same frequency (is it known?) as the A320 GPS
+        mcdu.refreshPageCallback = () => {
+            CDUGPSMonitor.ShowPage(mcdu, merit1, merit2, sat1, sat2);
+        };
+        SimVar.SetSimVarValue("L:FMC_UPDATE_CURRENT_PAGE", "number", 1);
+
         mcdu.setTemplate([
             ["GPS MONITOR"],
             ["GPS1 POSITION"],
@@ -20,13 +33,13 @@ class CDUGPSMonitor {
             ["TTRK", "GS", "UTC"],
             [`${Math.round(TTRK)}[color]green`, `${Math.round(GROUNDSPEED)}[color]green`, `${UTC}[color]green`],
             ["MERIT", "MODE/SAT", "GPS ALT"],
-            [`${Math.floor(Math.random() * 10) + 40}FT[color]green`, `NAV/${Math.floor(Math.random()*5) + 8}[color]green`, `${Math.round(ALTITUDE)}[color]green`],
+            [`${merit1}FT[color]green`, `NAV/${sat1}[color]green`, `${Math.round(ALTITUDE)}[color]green`],
             ["GPS2 POSITION"],
             [`${currPos}[color]green`],
             ["TTRK", "GS", "UTC"],
             [`${Math.round(TTRK)}[color]green`, `${Math.round(GROUNDSPEED)}[color]green`, `${UTC}[color]green`],
             ["MERIT", "MODE/SAT", "GPS ALT"],
-            [`${Math.floor(Math.random() * 10) + 40}FT[color]green`, `NAV/${Math.floor(Math.random()*5) + 8}[color]green`, `${Math.round(ALTITUDE)}[color]green`]
+            [`${merit2}FT[color]green`, `NAV/${sat2}[color]green`, `${Math.round(ALTITUDE)}[color]green`]
         ])
     }
 }
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
@@ -2,10 +2,16 @@ class CDUGPSMonitor {
     static ShowPage(mcdu, merit1, merit2, sat1, sat2) {
         let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), 
                                   SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
-        let currPosSplit = currPos.split("N")
+        if (currPos.includes("N")) {
+            var currPosSplit = currPos.split("N")
+            var sep = "N/"
+        } else {
+            var currPosSplit = currPos.split("S")
+            var sep = "S/"
+        }
         let latStr = currPosSplit[0]
         let lonStr = currPosSplit[1]
-        currPos = latStr + "N/" + lonStr
+        currPos = latStr + sep + lonStr
         let TTRK = SimVar.GetSimVarValue("GPS GROUND MAGNETIC TRACK", "radians") || "000";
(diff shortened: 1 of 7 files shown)


DoToAdventures:
Hello, can you include screenshots as well as references used for these changes?

lhoenig:
Btw, the same trick I figured out to allow dynamic CDU pages would also be needed in A320_Neo_CDU_NavRadioPage.js, I tested it to no ill effect but I cannot see why that section should even be dynamic, so not including it in the PR for now.

lhoenig:
This is what the Position Monitor page looks like after the changes.
Before, it would say "MIXIRS" without the blank on the frozen page.
Also I added the "at HHMM" part to the title of the frozen page. Both changes according to http://www.a320dp.com/A320_DP/nav-flight-management/sys-14.7.2.html and http://www.a320dp.com/A320_DP/nav-flight-management/sys-14.7.2B.html
I hope that page is legit?
Oh, and I made the line with the three NAV values green on the unfrozen page, too. It was white before. Also per the same reference. The GPS section is visually unchanged, but also updates continuously. I couldnt't actually find a source saying explicitly they both do the updating, but it highly makes sense to me, also given the existence of the "Freeze" button.
sn1
sn2

DoToAdventures:
Looks good, however I'm not sure about the source used. The A320DP site is based on CEOs, not NEOs and there might be MCDU changes there as well as the source is quite dated, we'll need an IRL pilot to confirm these changes, I think.

camilo625:
Works as described

DoToAdventures (approved):
Ares confirmed this looks correct as it does in the real aircraft.

lhoenig:
Nice! In the commit I just added, I did the same thing for the IRS monitor pages, making them dynamic. I also added the frozen page, for which the button already existed. As a source I again only have the A320DP site (http://www.a320dp.com/A320_DP/nav-flight-management/sys-14.7.3A.html), but logically it should really be the same situation here.
Screenshot of new frozen page attached.
irs

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_IRSStatusFrozen.js
```diff
@@ -0,0 +1,45 @@
+class CDUIRSStatusFrozen {
+    static ShowPage(mcdu, index) {
+        mcdu.clearDisplay()
+        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"),
+                                  SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        let GROUNDSPEED = SimVar.GetSimVarValue("GPS GROUND SPEED", "Meters per second") || "0";
+        let THDG = SimVar.GetSimVarValue("GPS GROUND TRUE HEADING", "radians") || "000";
+        let TTRK = SimVar.GetSimVarValue("GPS GROUND MAGNETIC TRACK", "radians") || "000";
+        let MHDG = SimVar.GetSimVarValue("GPS GROUND TRUE TRACK", "radians") || "000";
+        let WIND_DIR = SimVar.GetSimVarValue("AMBIENT WIND DIRECTION", "Degrees") || "000";
```
DoToAdventures, line 10: Getting ambient wind seems to cause the windindicator to jump around **a lot**. Perhaps implementation via aircraft wind is wiser? See the implementation of this on the ND in A32NX_NDInfo.js where simplane values are used instead of simvars.

lukecologne:
``` javascript
function getWindDirection() {
        var angle = SimVar.GetSimVarValue("AMBIENT WIND DIRECTION", "Degrees");
        return angle;
    }
    Simplane.getWindDirection = getWindDirection;
```
This is the Simplane function for getting the wind direction.
Simplane is most of the time (not always) just a wrapper for SimVar to make your life easier, it should be exactly the same in this case.

DoToAdventures: Mh, I agree with you but I don't really understand where the discrepancy between the two values comes from then. The severity of this however seems to depend on the exact weather conditions, it's more noticable with stronger winds. 

https://streamable.com/0iozx7

DoToAdventures (changes requested):
After some talking with Ares, on the Position Monitor page FMGC needs to be changed too FMS1 and FMS2, the rest varies on the specific systems used.
unknown

lhoenig:
Changed  FMGC to FMS as requested.

Also changed the Lat Long strings to fit the picture, see commit.

I also found out the reason the wind direction is displayed differently on the ND, they use some elaborate logic to smooth the angle, see
 `A32NX_NDInfo.js:setWind(_windAngle, _windStrength, _planeAngle, _force = false)`
which gets called in the update function using 
`this.setWind(Math.round(Simplane.getWindDirection()), Math.round(Simplane.getWindStrength()), Simplane.getHeadingMagnetic());`
Should I integrate that logic into the monitor pages?

DoToAdventures:
Tested, looking good! 
I'd suggest integrating the smoothing if possible, yes.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
```diff
@@ -1,6 +1,11 @@
 class CDUGPSMonitor {
-    static ShowPage(mcdu) {
-        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+    static ShowPage(mcdu, merit1, merit2, sat1, sat2) {
+        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), 
+                                  SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        let currPosSplit = currPos.split("N")
```
Externoak, line 5: It is me or N is not a constant value?
It could be N or S depending on the position of the plane in the world.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_IRSStatus.js
```diff
@@ -1,7 +1,16 @@
...
     static ShowPage(mcdu, index) {
         mcdu.clearDisplay()
-        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        mcdu.refreshPageCallback = () => {
+            CDUIRSStatus.ShowPage(mcdu, index);
+        }
+        SimVar.SetSimVarValue("L:FMC_UPDATE_CURRENT_PAGE", "number", 1);
+        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"),
+                                  SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        let currPosSplit = currPos.split("N")
```
Externoak, line 10: Same as above.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PositionMonitorPage.js
```diff
@@ -1,19 +1,28 @@
...
     static ShowPage(mcdu) {
         mcdu.clearDisplay();
-        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"), SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        mcdu.refreshPageCallback = () => {
+            CDUPositionMonitorPage.ShowPage(mcdu);
+        };
+        SimVar.SetSimVarValue("L:FMC_UPDATE_CURRENT_PAGE", "number", 1);
+        let currPos = new LatLong(SimVar.GetSimVarValue("GPS POSITION LAT", "degree latitude"),
+                                  SimVar.GetSimVarValue("GPS POSITION LON", "degree longitude")).toShortDegreeString();
+        let currPosSplit = currPos.split("N")
```
Externoak, line 10: Same as above

Externoak (changes requested):
It is me or N is not a constant value?
It could be N or S depending on the position of the plane in the world.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_GPSMonitor.js
Externoak, line 5: It is me or N is not a constant value?
It could be N or S depending on the position of the plane in the world.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_IRSStatus.js
lhoenig: Oops, correct! I'll fix that later today!

lhoenig: Resolved

lhoenig:
Correct handling of souther hemisphere position added and wind direction smoothing like on the ND implemented.
The smoothing depends on the delta time, which depends on the framerate, which can be slightly different for instruments.
That means the wind direction values on the MCDU will sometimes have a very slight deviation from the ND values. But it is like 1 degree for half a second.
It could only be made perfect if I could read the value directly from the ND, but I'm not sure how to do this. Does this suffice? I suggest you test it to see the effect for yourself. Its much better than before nonetheless.

DoToAdventures (approved):

DoToAdventures:
Much better!

PR #794, opened 2020-09-21
Set Emer Exit light to dummy

benjwri:
**Summary of Changes**

Changed the type of the Emergency Exit Status light to be 'Dummy', so it does not appear clickable, as it is not a push button

Discord username (if different from GitHub): Benjwri#2516


AdenFlorian:
im not sure what this is actually changing, could you give more details on what changes from the user perspective?

NathanInnes:
image
He means this button is no longer clickable with this PR.

Tested and works as stated.

PR #800, opened 2020-09-22
fix ground cursor

### LINKED ISSUE #793: [BUG] Ground side stick position is no longer visible on PFD.
**Mod Version**

Master

**Describe the bug**

Ground side stick position is no longer visible on PFD.

**To Reproduce**

1. Load in on the runway
2. look at PFD
3.

**Expected behavior**

The side stick position should be indicated on the PFD

**Actual behavior**

Side stick order indication should be displayed on the PFD during ground ops.

**References**

image
image

**Additional context**

AdenFlorian:
Fixes #793

**Summary of Changes**

Old code was tied to eng starter, which isn't tied directly to the switches exactly, so now have it using a more robust simvar that is tied to the engine switches.

**Additional context**


Carter633 (approved):
Can't see anything working incorrectly

DoToAdventures (approved):
This is the proper simvar to be used. Looks good, works good.

PR #930, opened 2020-09-27
Engine rotation speed fix

lukecologne:
**Summary of Changes**

* Reduces engine rotation animation speed (Forgot to multiply with 0.01 in the original PR, percentages and all)
* Adjust the blur mesh thresholds accordingly

**References**

* See for example @JackZ974's comment on #753 

**Additional context**

* Please let me know how it looks now, maybe if it's too slow. Especially shutdown might look weird now, since the fan stops way too fast at the moment, but that is being worked on.

<!-- You may optionally provide your discord username, so that we may contact you directly about the issue. -


NathanInnes (approved):
All looks good, tested and looks way better than before.

NathanInnes:
Got told it looks good from QA.

PR #911, opened 2020-09-26
Update Door ECAM page

wpine215:
**Summary of Changes**

- Minor updates to fuselage SVG
- New cockpit oxygen display (top right) updated for Neo variants
- "CKPT OXY" turns amber if Crew Oxy Supply pushbutton on overhead is off
- Updated SVG for door open indicators
- Over-wing emergency exits are now always armed

**Screenshots (if necessary)**

FlightSimulator_7RrXfIitEd

**Additional context**

Discord username (if different from GitHub):

PR #937, opened 2020-09-28
Patch #911 Lower ECAM DOOR Page

NathanInnes:
Patches #911 

**Summary of Changes**

Changed Amber colour on ECAM DOOR Page from Yellow to Amber.

Before:
image

After:
image

**References**

Feedback from @Ares1378 A320 Pilot

**Additional context**

Discord username (if different from GitHub): Sabes#8668

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -12,3 +12,4 @@
 5. [CDU] Fixed a crash when trying to replace a waypoint in the flight plan - @lousybyte (lousybyte)
 6. [EXTERIOR] Fixed issue in engine rotation animation that made it rotate too quickly - @lukecologne (lukecologne)
 7. [PFD] Remove code related to unrelevant aircraft - @1Revenger1 (Avery Black)
+1. [ECAM] Lower ECAM DOOR Page Colour Fix - @nathaninnes (Nathan Innes)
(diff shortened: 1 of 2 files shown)


DoToAdventures (approved):
Tested, looks good to me.

DoToAdventures:
Are we absolutely sure this is correct? I asked and looked around, this appears to be white in all the references I could find.

komp1821:
> Are we absolutely sure this is correct? I asked and looked around, this appears to be white in all the references I could find.

Yes and no.

Older A320 only have one oxygen bottle. That's why there's only one pressure indication. Apart from that, this is still valid.

oxy

DoToAdventures:
So shouldn't the oxy indication be white when the pressure indicator is green according to this?

waredrive:
FYI: ukflyer posted a video in https://github.com/flybywiresim/a32nx/issues/924# which also shows the amber OXY text you are talking about (https://youtu.be/Mo7A3CuVb8g?t=148)

NathanInnes:
If you have a look at the video @waredrive provided you can see that it's amber not yellow. No abnormal message in the Airbus is yellow.

When normal it turns white as seen here
image
Ref. ipadecam.co.uk

This PR wasn't to fix that page. Just to fix the colour of the text

NathanInnes:
> So shouldn't the oxy indication be white when the pressure indicator is green according to this?

The amber displays when CREW SUPPLY is off which was the condition in my screenshot. It becomes amber with other conditions also but that is currently not simulated.

DoToAdventures:
Yeah of course the amber is correct, just wondered why it was amber in the first place and not white but you're correct of course. The logic will have to be adressed seperately. Ready to merge then, imo.

Ares1378:
@DoToAdventures it is only displayed in white when PSI is >=500 PSI I.E. the crew oxygen valve is open.
image

UncleClapton (approved):

PR #990, opened 2020-09-30
Removed MSFS Update 1.8.3.0 Fuel Patch

NathanInnes:
**Summary of Changes**

Removed fuel patch that was implemented by Iceman™ as it's not longer needed.

**References**

**Additional context**


lousybyte:
Tested and looks good, waited about 10m after APU off and both engines were still running.

lousybyte (approved):

Issue #984, opened 2020-09-30
[BUG] Pitch too high 


EDKL1980:
seems that the pitch is now slightly high on normal cruise flight level with normal speed abs light wind.

With Mach .78 the pitch was > than 10%

![1003E2B3-CC2E-41EE-88E0-C05801772D1B](https://user-images.githubusercontent.com/72131688/94669052-0d748700-0311-11eb-965e-db907410e4d4.jpeg)
![00A9080F-797A-49A3-836F-94FEA5DE4758](https://user-images.githubusercontent.com/72131688/94669066-11a0a480-0311-11eb-8f3f-42689242eb57.jpeg)
![314EDA4C-E1EF-4E0A-8249-D2DFF737FF51](https://user-images.githubusercontent.com/72131688/94669082-149b9500-0311-11eb-96bc-391733de718a.jpeg)
![A657F09D-CC0E-4A95-82A5-3D31A2540E06](https://user-images.githubusercontent.com/72131688/94669089-16fdef00-0311-11eb-9a28-f3888918473d.jpeg)

**Mod Version**

**Describe the bug**

**To Reproduce**

1.
2.
3.

**Expected behavior**

**Actual behavior**

**References**

**Additional context**

Was this working before/when did the issue start occurring?

Is this a problem in the vanilla unmodded game?

Discord username (if different from GitHub):

dmajestic80:
I have noticed this too especially when on approach and established on the ILS. I even tested with a very light aircraft (~3000kg fuel and 0 pax) The nose pitch is unusually high around 7 degrees. Removed the mod and didn't have this issue.

PCGamesF1Tech:
Wind should be irrelevant, you are just moving inside the airmass (calm or 100kt makes no difference, just changes your ground speed, depending on direction).
Pitch should be around 2.5° nose up and power around 80% (depending on weight)

EDKL1980:
Weight was balanced with loading 90% and around 10 tons of fuel balanced left and right with empty Center

PCGamesF1Tech:
If within (MAC) limits no problem.
Light weight: 2.5°, some power
Medium weight: 2.5°, more power
Heavy weight: 2.5°, even more power

EDKL1980:
Also experienced this issue during final approach. Flaps full and 140-145kts the pitch is too high. I always need to speed up to 150-160 which is definitively too fast for the final with flaps full

apolinski:
[...]

PR #1029, opened 2020-10-01
Flightmodel hotfix

### LINKED ISSUE #1014: [BUG] Takeoff and landing stall speeds are faster
**Mod Version**
Stable 0.3.1

**Describe the bug**
During takeoff, before 0.3.1 I was about to take off at about 130 knots very easily and smoothly. Now, if I take off at 130 knots my main landing gear doesn't leave the ground and when it does it almost always stalls. I have to reach speeds of about 160 to smoothly take off. Furthermore, the regular approach speed that I use is about 140 knots. At 140 knots now, I am not able to even raise the nose, the plane just falls out of the sky. This is a problem that I have seen a few other people on the discord complain about.

**To Reproduce**

1. Either take off or get on an approach
2.Use your regular takeoff/approach speeds
3. Observe the strange behaviour

**Expected behavior**

You are supposed to be able to take off at about 130 knots and approach at about 140 knots without problems.

**Actual behavior**

At 130 knots the main landing gear doesn't even lift off the ground. At 140 knots on approach you can't even control your pitch anymore.

**Additional context**

It was working fine with the 0.3.0 release AFTER the MSFS update. While I had no screens, I could see my speed in external view and everything worked as it should. When I downloaded 0.3.1, this problem started happening.

I haven't checked, but as said above, this didn't happen on 0.3.0 AFTER the update but only after I downloaded 0.3.1

Boris#9134

DoToAdventures:
Fixes #1014 

**Summary of Changes**

This brings most of the flightmodel.cfg back in line with Asobo's version while keeping the changes we made to ground behaviour,  MMO/VMO as well as slat positions

**Additional context**

The actual inflight behaviour should now be identical with the default plane again, with all the changes Asobo has made to it so far. To me this feels good, but I'm no expert on how it should feel and behave.
This PR is mostly just there to fix #1014 (as I view it as critical) until we figure out which changes to the flightmodel are causing the behaviour described in the issue.

Discord username (if different from GitHub): DoTo#5564


hotshotp:
Just a question, what kind of ground behaviour did you guys change?

DoToAdventures:
https://github.com/flybywiresim/a32nx/commits/master/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg looks like mostly nosewheel steering as well as gear extension stuff to me.

AK-777:
Thank you for your work guys but i think the default Asobo's flight model numbers are also wrong you cannot have the same peformance of A320neo by this i tried to customize flight model and it really become more stable but my problem that i'm not sure that my customize numbers are matching the real peformance numbers of Airbus data but i'm sure that the flight model file need more work in all sections of Elevator / Alerones / Rudders / Trim / etc ..

DoToAdventures:
I'm aware they need work, this is just supposed to be a hotfix to fix wrong behaviour at speeds below 150 knots.

joniatus:
And it does fix that issue for now. Thank you!!!

lousybyte:
Any impact on #984, #1034 or unrelated?

DoToAdventures:
Haven't tested, it might have. If there's flight model issues in the mod that don't exist in vanilla, this will 90% fix them

Venorcis:
> Haven't tested, it might have. If there's flight model issues in the mod that don't exist in vanilla, this will 90% fix them

https://github.com/flybywiresim/a32nx/issues/1034#issuecomment-702604116 this seems to suggest that it will :) important to merge this quickly I think...

ImenesFBW:
Name: Imenes#8739
Date of testing: 2/10/2020
Version of the sim: 1930
PR Tested: #1029

Testing technique: Try-out

1. Loaded the plane on final, and checked low speed behaviour

Overall Rating: Good
Conclusions of the testing: Working as documented! Thank you!

PR #1085, opened 2020-10-04
Revert "Hotfix to make av fan loop sound playing in external view"

hotshotp:
Reverts flybywiresim/a32nx#1070

This actually made it so every time you switch into the external view and back inside the sounds take about 5 seconds to start up again, very abruptly god damnit wwise. For now I guess the fan sounds will have to play on the outside. I have been trying to fix this for a while and people have been reporting this issue over and over again so I guess this is a pretty big issue. I would like to revert it to what it was before which made it so it played on the outside but no abrupt change that makes you jump in your seat 5 seconds after switching your view.

sorry about so many PRs in the recent days, there is a bug which i try to fix which causes another bug, you know the cycle :(


hotshotp:
I accidentally found a fix to the bug lol

PR #1142, opened 2020-10-06
Revert flap value changes to Asobo values

### LINKED ISSUE #1053: [BUG] A320 Last flight model dosent work good.
**Mod Version**
Master (Downloaded 4 hours ago)

**Describe the bug**
when i am in flaps 1 the nose of the airplane goes down while aircraft continue climbing, its just not realistic.

**To Reproduce**
1. Take off Flaps 1+F
2. Didnt gain speed with flaps 1
3. retract Flaps 1 gain speed and altitude but its not real!
4. Approach same issue, was decending to 4000, and hit Flaps 1 and flare to 5,300.
5. imposible to land.

**Expected behavior**
Just reverted it like it was before, that flight model dosent works as it should, please check all the PR good... make and exhaustive test.

**References**
FlightSimulator_z292sTbmna

Was this working before/when did the issue start occurring? Yes, much better!

Is this a problem in the vanilla unmodded game? no

DoToAdventures:
Fixes #1053 (Hopefully)

**Summary of Changes**

Reverts the flap values in the flightmodel.cfg to the Asobo default to fix the flap lift / angle of attack issues with flaps extended at low speeds

Discord username (if different from GitHub): DoTo#5564


lukecologne:
If we really need to implement these changes, then the least we can do is actually make the extension of slat position 2 and 3 the same: 
grafik
Also, we would need to change the upper ECAM flaps indicator positions, since they depend on flaps angle and will be wrong.

DoToAdventures:
Mhh.. You're correct there. Perhaps a different way to make the flightmodel behave properly again would be better...

DoToAdventures:
Closed for now  until I can find a better way to fix the flightmodel

PR #1159, opened 2020-10-07
Added 3d sound emitters, cargo door, gear extension/retraction, mech horn, evac horn, and emer cabin call

### LINKED ISSUE #1113: [BUG] GPU Packs Audio delay after switching view
**Mod Version**

Latest Master Branch

**Describe the bug**

When you can hear the GPU air conditioning and switch view, the sound stops and there is a 10 second delay or so until the pax sounds start again

**To Reproduce**

1. Spawn cold and dark
2. Turn on GPU and batteries
3. Switch view to external and then back to cockpit
4. Listen and the air conditioning sound does not return for a few seconds

**Expected behavior**

The air conditioning sound in the cockpit is continuous

**Actual behavior**

After switching views, there is a delay until the sound starts again

**References**

**Additional context**

Latest master

No

Discord username (if different from GitHub): obama#6262

hotshotp:
Fixes #1113 

## Summary of Changes
Firstly, this fixes the issue of the avionics fans taking a few seconds to start up in internal view after switching to external and back.
Then, this changes the avionics fans from the previous ones to be higher quality. Then it adds extra sounds for the battery relay, the cargo doors, gear extension and retraction, emer cabin call, mech horn, and evacuation horn. 
Finally, I have made every sound have a 3D position around the cockpit, so it is an "emitter". This will make the entire soundscape more alive.

Note: Evac horn shutoff button works only when holding it down as it only toggles the simvar when pressed. This can be easily fixed in the future by having that button have to "toggle" the simvar rather than "holding" it.

## References

I have used real videos, samples, and took feedback from many IRL pilots and engineers to make sure the sounds are realistic.

## Additional context

There is one annoying issue that I cannot fix. This is, when you spawn on the runway or on approach all the startup sounds play in a sequence. There is no way that I could find to remove it, but I don't think this is that bad of an issue as it is only for about 10 seconds.

Another tiny issue, the CIDS chimes (the 2-6 chimes you hear in the cockpit after connecting battery power), these are played after only one battery is connected. While this is annoying, I cannot fix that until the new electrical system is added and it has new simvars for individual batteries being on.

Last and final issue, when you spam external power over and over the avionics fan self test sound starts from the middle. That is just how Wwise works and I cannot actually fix that, so just don't spam the ext power button :)

Boris#9134


wuperguy (approved):

Benjozork:
Force-pushed onto the branch to fix an error after disucssing with @hotshotp on discord

hotshotp:
had some issues will open another one

PR #1163, opened 2020-10-07
[PFD/MFD] Hotfix - Optimised PFD/MFD impact on framerate

### LINKED ISSUE #1110: Performance decrease of around 20% in FPS and more stutter in latest master[BUG]
**Mod Version**

Master downloaded 05/10/2020

**Describe the bug**
Assuming it is the separate FO and CP screens implementation, the performance has decreased a lot and personally I don't think the feature gain is worth it unless there is a workaround. When shared cockpit comes out then for sure. 

**To Reproduce**
Fly any flight and see the performance loss, even when the PFD and ND screens on the FO side are all the way off.

**Expected behavior**
Maybe make it such that if you turn of the FO screens that it does not transmit data. I would expect this

**Actual behavior**
In normal configuration and all screens turned to on, the FPS loss is too much. As mentioned above, if the FO and CP screens are set to OFF and not just dimmed, the performance should be like before.

2hwk:
Fixes #1110

## Summary of Changes
* Optimisation - reduce calling of document -> can simply be init once, stored and index for display retrieved as object vars. 
* Escape out of all display update code when the display's brightness knob is turned off - Display will not update when brightness is 0. This is good for ~5ms of reduction in CoherentGTdraw times when both the right-side PFD and MFD are disabled, which should help in CPU limited scenarios. 
* Added check for querySelectorAll() being valid to prevent issues with the MFD selection knob.

## Screenshots

Before:

Current Master

After: 
2 Displays

With Right MFD + PFD off: 
1 Display

## Additional context
image

Majority of users will see no difference, if you were having stutters and frame drops before this may help. 

Discord username (if different from GitHub):2hwk#1022


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A320_Neo_MFD.js
```diff
@@ -402,10 +396,17 @@ class A320_Neo_MFD_Compass extends NavSystemElement {
         this.svg = this.gps.getChildById("Compass");
         this.svg.aircraft = Aircraft.A320_NEO;
         this.svg.gps = this.gps;
+        const url = document.getElementsByTagName("a320-neo-mfd-element")[0].getAttribute("url");
+        this.disp_index = parseInt(url.substring(url.length - 1));
     }
     onEnter() {
     }
     onUpdate(_deltaTime) {
+        const pot_index = this.disp_index == 1 ? 89 : 91;
```
Venorcis, line 405: might be good to make the pot_index a property like disp_index too, to avoid this duplication?

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A320_Neo_MFD.js
```diff
@@ -402,10 +396,17 @@ class A320_Neo_MFD_Compass extends NavSystemElement {
...
+        const currentKnobValue = SimVar.GetSimVarValue("LIGHT POTENTIOMETER:" + pot_index, "number");
```
Venorcis, line 406: might be good to make this a function (getCurrentKnobValue or something)

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/A320_Neo_PFD.js
```diff
@@ -325,7 +350,7 @@ class A320_Neo_PFD_Attitude extends NavSystemElement {
             }
 
             this.hsi.setAttribute("slip_skid", Simplane.getInclinometer().toString());
-            this.hsi.setAttribute("radio_altitude", Simplane.getAltitudeAboveGround(true).toString());
+            this.hsi.setAttribute("radio_altitude", Simplane.getAltitudeAboveGround().toString());
```
Venorcis, line 353: was this 'true' removed here intentionally?

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A320_Neo_MFD.js
```diff
@@ -577,10 +578,17 @@ class A320_Neo_MFD_NDInfo extends NavSystemElement {
         this.allSymbols.push(this.ndInfo.querySelector("#WXRTURB"));
         this.allSymbols.push(this.ndInfo.querySelector("#TURB"));
         this.allSymbols.push(this.ndInfo.querySelector("#MAP"));
+        const url = document.getElementsByTagName("a320-neo-mfd-element")[0].getAttribute("url");
```
Venorcis, line 581: also good to put this in a function I guess, to avoid the duplication

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/PFD/A320_Neo_PFD.js
2hwk: No, adding it back

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/MFD/A320_Neo_MFD.js
2hwk: Done

Venorcis (approved):

Venorcis:
Name: Venorcis
Date of testing: Today (07/10/2020)
Version of the sim: 1.9.3.0
PR Tested: #1163 

Steps you've made to test the Pull Request: Spawned at EHAM, IFR flight to EDLP.

No issues found, stutter seems to be fixed, especially with the co-pilot's screens turned off.

Overall Rating: Good
Conclusions of the testing: Works as described!

lukecologne (changes requested):
This breaks the self test screen when switching the PFD/ND off with the potentiometer. Self test when powering up the aircraft works fine.

danbw9:
> This breaks the self test screen when switching the PFD/ND off with the potentiometer. Self test when powering up the aircraft works fine.

The self test screen should not keep popping up. Only when powering up the plane

lukecologne:
> 
> 
> > This breaks the self test screen when switching the PFD/ND off with the potentiometer. Self test when powering up the aircraft works fine.
> 
> The self test screen should not keep popping up. Only when powering up the plane

The DUs go into self test after being disabled for around 10 seconds, but that needs to be adressed in another PR. The point is, this change is breaking previous behaviour.

2hwk:
Old behavior has been restored

lukecologne (approved):

Venorcis (approved):

starvingjavi:
Name: Javi#2557
Date of testing: Today (07/10/2020)
Version of the sim: 1.9.3.0
PR Tested: #1163

Testing technique: Did a full flight starting from Cold and Dark

Steps you've made to test the Pull Request: Spawned at SAVT, IFR flight to SAZS.

Stutter seems better, specially when turning the FO screens. When everything is on small stutter are apparent but not like before. GTDraw is now doing better. None of these are doing as good as before the split of screens, but this is an improvement

Overall Rating: Good
Conclusions of the testing: Works as described

danbw9:
> > > This breaks the self test screen when switching the PFD/ND off with the potentiometer. Self test when powering up the aircraft works fine.
> > 
> > 
> > The self test screen should not keep popping up. Only when powering up the plane
> 
> The DUs go into self test after being disabled for around 10 seconds, but that needs to be adressed in another PR. The point is, this change is breaking previous behaviour.

I am not sure where you get that info from.

danbw9:
I can confirm that the performance seems to be a little bit improved but only marginally. Not quite like before.

lukecologne:
> 
> 
> > > > This breaks the self test screen when switching the PFD/ND off with the potentiometer. Self test when powering up the aircraft works fine.
> > > 
> > > 
> > > The self test screen should not keep popping up. Only when powering up the plane
> > 
> > 
> > The DUs go into self test after being disabled for around 10 seconds, but that needs to be adressed in another PR. The point is, this change is breaking previous behaviour.
> 
> I am not sure where you get that info from.

I have sent it to you in discord, but I will do it here as well: https://discordapp.com/channels/738864299392630914/754851312306487326/758257387432968222
That is from an engineer, komp#1821

Artax04:
i tried the lastest master, i got only a low improvement, still unplayable to me. Too much less fps and too much higher frametime than default plane.
Was perfect until few days ago :(

jooglestumpe:
Same for me, even when turning of F/O PFD and ND. Thanks to the mod team for trying to get a fix so fast, anyway! It would be greatly appreciated if you looked into this further as it would be such a shame to not be able to enjoy all those other great PRs coming up every day :)

myselfaaryan:
I can confirm the same issue @Artax04 💻

c1be:
Still pretty bad, a lot of micro stutters, around 10 fps lower, even with screens off. This didn't really improve much, maybe slightly.

Artax04:
> Still pretty bad, a lot of micro stutters, around 10 fps lower, even with screens off. This didn't really improve much, maybe slightly.

Yeah, the strange thing is that also with boths screens off it goes a lot worst than a week ago. The frametime is crazy inside cockpit right now.

c1be:
> > Still pretty bad, a lot of micro stutters, around 10 fps lower, even with screens off. This didn't really improve much, maybe slightly.
> 
> 
> 
> Yeah, the strange thing is that also with both screens off it goes a lot worst than a week ago. The frametime is crazy inside cockpit right now.

It's stuttering like crazy, totally not worth it. I use only stable version from now on, and i hope this doesn't get in the next one.

myselfaaryan:
Is it possible to get an acknowledgement from the dev team that this is gonna be fixed? 😄

PR #1339, opened 2020-10-14
Turned down wind noise

hotshotp:
Fixes #[issue_no]

## Summary of Changes

When I made my engine noises they were loud so i naturally turned up my wind sound. But here when I reverted the old engine sounds the wind is now too loud. So i reverted it to what it was before

## Additional context

Boris#9134


hotshotp:
made a mistake

Issue #1375, opened 2020-10-15
[BUG] Plane can't reach normal cruising altitude


marci1229:
**Mod Version**
 **Release version 0.4.0** 

 **built: 2020-10-15T01:38:50+00:00**
**ref: refs/heads/master
sha: a72a977f0b5c74eab9396bab64cb16f6cc4c8120
actor: Benjozork
event_name: push** 

**Describe the bug**
**The plane can't climb and can't reach FL380. Weird climbing and cruising.**

**To Reproduce**

1. Start an IFR flight with the plane.
2. Try to climb FL380 with Live Weather.
3. Plane will stops at about FL300 and can't reach higher levels. The cruising angle will be weird like when you in stall statement.

**Expected behavior**
**Normal easily climb with normal cruising and speed.** 

**Actual behavior**
 **Weird climbing and cruising.** 

**References**
**None.** 

**Additional context**
**You can see cruising on the picture.**

Was this working before/when did the issue start occurring? Yes.

Is this a problem in the vanilla unmodded game?
**Modded with liveries and REX Weather Force only.**

Discord username (if different from GitHub): marci1229#1646
![Weird Cruising](https://user-images.githubusercontent.com/72943374/96174600-d2d23780-0f29-11eb-9fc2-11bdbbc7bef8.png)

Benjozork:
Pitch angle reported in #984

Benjozork:
Is #1377 the same problem you are facing for altitude ?

ntlgr:
Make sure your plane is not all frozen. If so, turn on wing & engine a.ice

marci1229:
> Make sure your plane is not all frozen. If so, turn on wing & engine a.ice. Judging from your screenshot, it is.

It's okay but the ice set to visual only.

marci1229:
> Is #1377 the same problem you are facing for altitude ?

No it is a different bug or something. Also with approaches the pitch is too much but now while cruising. I can't even reach FL330. The plane can't handle it.

MisterChocker:
I haven't experienced this in https://github.com/flybywiresim/a32nx/pull/1295 maybe my changes to aero resolve this.

MisterChocker:
Also what was your GW?

ntlgr:
> > Make sure your plane is not all frozen. If so, turn on wing & engine a.ice. Judging from your screenshot, it is.
> 
> It's okay but the ice set to visual only.

I can assure you that the plane can reach FL330+.
[...]

Issue #1315, opened 2020-10-14
[REQUEST] New speeds for managed CLB, CRZ and DES (MDCU)


Watsi01:
**Is your feature request related to a problem? Please describe.**
Messed up managed speeds in MCDU. See here:
Issue "[BUG] Managed speeds are a kind of a mess" [https://github.com/flybywiresim/a32nx/issues/1080](https://github.com/flybywiresim/a32nx/issues/1080)

**Describe the feature you'd like to see implemented**
Since I'm actually to stupid to file a proper pull request, I do it this way. I tried to make the three flight phases a bit smoother regarding the according managed speeds and the transition between the phases. Maybe someone of you PR pro's can add the change at some time. So I only changed the managed speeds for the flight phases (CLB, CRZ and DES). Didn't touch any flap speeds, speed restrictions or else. 

Inside A32NX folder:
1. go to "...\html_ui\Pages\VCockpit\Instruments\Airliners\A320_Neo\FMC"
2. open the "A32NX_FMCMainDisplay.js" and change

 _getCrzManagedSpeed() {
        let dCI = this.costIndex / 999;
        dCI = dCI * dCI;
        const flapsHandleIndex = SimVar.GetSimVarValue("FLAPS HANDLE INDEX", "Number");
        if (flapsHandleIndex != 0) {
            return this.getFlapSpeed();
        }
        let speed = **290** * (1 - dCI) + 310 * dCI;
        if (SimVar.GetSimVarValue("PLANE ALTITUDE", "feets") < 10000) {
            speed = Math.min(speed, 250);
        }
        return speed;
    }
    getDesManagedSpeed() {
        const dCI = this.costIndex / 999;
        const flapsHandleIndex = Simplane.getFlapsHandleIndex();
        if (flapsHandleIndex != 0) {
            return this.getFlapSpeed();
        }
        let speed = **288** * (1 - dCI) + 260 * dCI;
        if (SimVar.GetSimVarValue("PLANE ALTITUDE", "feets") < 10000) {
            speed = Math.min(speed, 250);
        }_

This makes the cruise a tiny bit faster (according to the cost index) and the managed descent speed is not ~240kts from top down anymore, but around 285-290kts (speed limit below 10.000 still kept). I didn't touch the CLB speed here, cause - for some reason - it doesn't do anything (even if there are numbers for CLB around this section).

3. for CLB speeds you have to go to:
[...]

PR #1295, opened 2020-10-13
Improved engine performance and flight dynamics

MisterChocker:
Fixes #210
Fixes #359
Fixes #824
Fixes #850
Fixes #917
Fixes #974
Fixes #984
Fixes #1034
Fixes #1043
Fixes #1166
Fixes #1173
Fixes #1204
Fixes #1255 
Fixes #1375
Fixes #1377 
Fixes #1442 (VApp calculation)
(are all about climb rate/cruise attitude/not able to reach cruise/stalls/fuel consumption)

## Summary of Changes

- Engine (redesigned performance table)
  - Taxiing thrust adjusted
_enables taxiing on idle thrust and also impacts aircraft performance in flight e.g. approach or decent_
  - N1 on Approach will match real life on about 40 to 45%
  - Climb performance reworked
  - Cruise N1 adjusted
  - Fuel burn during cruise adjusted
-  Aerodynamics
   - Flaps
     - Increased lift coef
_prevents early stalling of the aircraft on landing_
_improves the overall feeling of the aircraft in flight_
     - Decreased drag coef
_deceleration on approach now slower_
     - Increased pitch momentum
_pitch now changes noticeable when flaps position changes_
   - Speedbrakes / Spoiler
     - Decreased drag coef
    - Landing gear
      - Increased drag coef
   - Elevator
     - Increased momentum coef
_makes aircraft (more) responsive_
- MCDU
_The following speeds are now calculated based on charts below_
  - VLS
  - VApp
  - F (for approach)
  - S (for approach)
- PFD
  - F and S speeds for approach are shown based on charts below
  - The Alpha Max and Alpha Prot has been tweaked as best as possible but is not correct.
_This is an issue to be solved by the future flight model_

## References

https://www.youtube.com/watch?v=cv41hm4JEB8&feature=youtu.be&t=100
https://www.youtube.com/watch?v=FODhSJucbK0
https://www.youtube.com/watch?v=rhI5XGwlnaI
image
image
image
image
image
image
image

## Additional context

Discord username (if different from GitHub): DerL30N#3751

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build.py will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, slick on the **Artifacts** drop down and click the **A32NX** link

PR #1389, opened 2020-10-16
New speeds for managed CLB, CRZ and DES flight phases

### LINKED ISSUE #1080: [BUG] Managed speeds are a kind of a mess
**Mod Version**
from 0.3.1 to current master

**Describe the bug**
The whole managed speed is a bit a mess. Climbing with 320kts managed, automatic switch to mach, but when CRZ mach is reached still holds the kts, then suddenly going back to 285kts (managed CRZ). The descent goes with managed ~240kts (it should be CRZ mach first and then switch to according kts ~280 or higher).  Even if you switch to selected speed prior to open descent, the FCU shows mach, but it still will descent with knots (speed overshoots speed bug).

**To Reproduce**

1. Plan a full flight and start it
2. Don't fiddle with the phase speeds in MCDU (except takeoff speeds)
3. Fly and watch the speeds, during different flight phases
4. Try to switch to selected speed prior to descent from CRZ alt. Use mach speed first and watch the actual speed overshooting the mach speed bug.

**Expected behavior**
see above

**Actual behavior**
see above

**References**

**Additional context**
I think it started with version 0.3.1. Could be also a 1.9.3 vanilla issue. I don't know.

### LINKED ISSUE #1358: Managed descent speed defaulted to 240kts [BUG]
**Mod Version**
 

built: 2020-10-15T01:38:50+00:00
ref: refs/heads/master
sha: a72a977f0b5c74eab9396bab64cb16f6cc4c8120
actor: Benjozork
event_name: push

**Describe the bug**

Not sure if a bug but the default managed descent speed is 240 kts and cannot be changed. Moments after initiating descent the aircraft will decrease to 240 kts (the managed speed within the PERF DES page) This seems very slow, I'd expect this to be around the 280 kts.

**To Reproduce**

1. Load any normal flight and try to change default 240kt managed speed in PERF DES page
2.
3.

**Expected behavior**

Entering A Managed Speed; As long as the Descent Phase is not active, the PERF DES page may be used to insert either a Speed or a Mach number or both to replace the FM computed ECON Managed descent speed.The  FMGS  then  uses  your  manually  inserted  Speed/Mach  Managed  Speed  instead  of  its  automatic  ECON Managed speed for computing the descent profile and determining the descent Managed speed profile.  Note that even though you have manually modified the FM computed ECON descent speed, itis still considered by the FM to be Managed Speed –it’s just different from what the FM automatically computed.When the system switches to the Descent  Phase, it sets the  Managed target speed to the entered speed.   From there,  you  can  only  modify  the  speed  by  using  the  FCU  selector  knob  only  (i.e.  Selected  Speed).    Once  in  the Descent Phase, the pilot cannot change the Descent Managed Speed again.

**Actual behavior**

Unable to changed managed speed which is defaulted to 240kts

**References**

**Additional context**

Obviosuly a workaround is to use selected speed but this is not realistic

Was this working before/when did the issue start occurring? Has always been an issue

Is this a problem in the vanilla unmodded game? Unsure

Watsi01:
Fix #1080
Fix #1358 
Request #1315  

Just trying my second PR.

## Summary of Changes
The first edit makes the cruise a tiny bit faster (according to the cost index) and the managed descent speed is not ~240kts from top down anymore, but around 285-290kts (speed limit below 10.000 still kept). I didn't touch the CLB speed here, cause - for some reason - it doesn't do anything (even if there are numbers for CLB around this section).

The second edit (A32NX_FMCMainDisplay.js) reduces the climb speed from ~320 to ~290kts. I'm not sure about this number. I did it that way, because the Asobo AP seems to handle mach speeds incorrectly. Even if the speed bug is switched to mach, the AP seems to still hold KIAS. So the 290kts here are mainly to smooth the transition between CLB and CRZ. Suggestions of other / more realistic values are appreciated for the future (when AP is working correctly). 

Any other speeds (flap speeds, restrictions, etc.) were not touched.

## References
No references since this is mainly to smooth the transitions between flight phases and have a work around for the rudimentary Asobo AP. 

## Additional context
I think these values are within an normal A320(neo) flightphase profile. If a real pilot thinks, my numbers a of the hook, please give feedback.

EDIT - to be more specific on the problem with AP, managed speeds and mach transitions: I think (someone correct me please, if I'm wrong), that the automatic switch from KIAS to mach in the AP is hardcoded at .78 mach. But many times (depending on cost index and CRZ FL) the plane won't reach .78 up to CRZ FL. **And** the managed CRZ mach is way lower than .78. Now in real life (real pilots?), the AP would switch to mach if CRZ mach is reached during climb - for example .76 or .74 or whatever. Here, since the .78 is hardcoded, the plane would like to always climb with .78 (when reached). Which could be a hard job for the engines (could be related to this issue #1375). And the transition between CLB mach and (way lower) CRZ mach is overboarded. With my lower managed speed numbers for climb (see above), the plane climbs with less speed (the managed KIAS actually) and the plane doesn't want to always reach .78 mach at climb. Also this way the transition from CLB to CRZ mach is much smoother.

Edit: also the autopilot seems to follow KIAS over KCAS: see here #1422. 

Discord username (if different from GitHub): RogePete


Watsi01:
@Benjozork Hi, I made it from scratch again (this time on my PC and not via the online editor which broke it). So everything seems to be ok and ready for testing.

Thanks again for your help!!

Carter633 (approved):
Name: British Avgeek YT -- Carter633 #1436
Date of testing: Current date 12/10/2020
Version of the sim: 1.9.5.0
PR Tested: #1389

1. Spawned on stand Cold and Dark for full flight
2. Took off, monitored CLB speeds - seemed good, no issues noticed there
3. Monitored CRZ - activated well and speed worked as described too, no issues noticed for me there
4. Monitored DES during descent phase of flight, managed speed increased as described until FL100 then reduced to below 250kts as it should below 10,000. 

Overall Rating: Good
Conclusions of the testing: No issues found, seemed good - descent phase much better now with better speeds

Watsi01:
@Carter633 Thanks for testing. :-)

Carter633:
Pleasure - if it helps further I ran a second test that finished just now - same criteria - full flight from cold and dark again this time Tenerife to Madeira on Vatsim.Same result as the first test... really good and no issues found.

JackZ974:
Descent speed and Cruise Speed in managed mode are normally related to Cost Index. Is your mi fixation is taking the CI into account?

Mariolink123:
> Descent speed and Cruise Speed in managed mode are normally related to Cost Index. Is your modification taking the CI into account?
> And Mach/Knots transition is related to altitude, not phase of flight. It is called Crossover altitude and varies with IAS. On the A320 it is around FL280/300
> 6B347C9C-DC55-4F00-93AB-C48780D576E3

`let dCI = this.costIndex / 999;
        dCI = dCI * dCI;`
`let speed = 290 * (1 - dCI) + 330 * dCI;`

Yeah, it does take CI into account.

Watsi01:
Yeah, it does effect the speeds. But only to a minimum effect (regarding the formulas). As far as I can see on the table, my CRZ numbers are pretty close to it. The crossover is a problem though, since the AP doesn't handle it correctly (see PR). But it only effects it in a way that you have to switch to mach (and back to KIAS during DES) manually. The speeds should still be ok.

nobodyn0se:
Yup, the SPD/MACH auto switch at crossover altitude is broken. It must be manually switched both ways. These speeds make the transition smooth.

FlyingCarpet7:
Short review from IRL pilot.

Tested on a flight from A-B. 
Cost index 20, TOW 65 T , FL370

Climb:
Managed speed 290/M.78 is a normal value during climb.
My in flight perf calculator gives me 295/.78.

Cruise:
Got M .783
Calc M .78

Descend
Got IAS 285
250 at FL100

PR works fine!

We have a lot of other problems like:
CI changes in flight does not appear to effect the speeds in flight
Can not see the CRZ speeds on performance page
Crossover IAS - MACH - IAS is wrong
Engine performance to weak in climb (but know they work on it)
Wrong speeds all over the scale during approach.

Benjozork:
@FlyingCarpet7 can you clarify if those problems arise from the changes in this PR if you're stating already existing problems in the game ?

FlyingCarpet7:
> @FlyingCarpet7 can you clarify if those problems arise from the changes in this PR if you're stating already existing problems in the game ?

PR works fine and did not create any other issues what I can see. The other things I mentioned are already known bugs in the sim.

donstim:
> 
> `let dCI = this.costIndex / 999; dCI = dCI * dCI;`
> `let speed = 290 * (1 - dCI) + 330 * dCI;`
> 
> Yeah, it does take CI into account.

**Note:  These comments are directed more toward possible follow on work to better address the CI effect on speeds rather than toward this PR in particular.** (For what it's worth, I'm in favor of this PR as is, but I"m no AQ tester nor am I a pilot.)

Did some flight testing to check to see if the above equation is the sole determinant for ECON (managed) speed, and it doesn't seem so. According to the above equation, at CI=0, speed should be 290 and for CI=999, speed should be 330. In my flight test, I got the following speeds placed in the MCDU during the CLB/CRZ/DES phases"
CI =1 (wouldn't accept 0): 320/285/240
CI = 999: 330/310/260

@FlyingCarpet7 The Airbus "Getting to Grips with Cost Index" document shows climb speeds of 308 KCAS for CI=0 up to 340 KCAS for CI=100 for the A320 ceo with CFM56 engines at 75 T. Is this not representative of the CEO with LEAP engines? Sounds like not if your performance calculator is giving you 290 KCAS for CI=20. 
For cruise, ti looks closer -- for the A320/V2500 at 65 T, CI 20 gives M0.777 or so.
For descent, it gives M.779/278 for CI 20 A320/CFM56 at 60 T

Was hoping this document could be used to come up with a mod for the CI effect on speed. It seems like it would at least be better than what is in there now, but maybe it could be further tweaked by using your calculator. (This is all idle speculation on my part as I am not a programmer, so I don't know how feasible that would be.)

Watsi01:
@donstim Yes, I also thought about bringing the CI more to an impact on speeds. But it will take some time to figure out a forumula which works for all cases. And I'm not sure if I'm able to do that (but I also didn't dive very deep into this matter so far).

FlyingCarpet7:
@donstim 
I had a look at the Airbus document to. It is a bit old, at least the one I got is from 1998, but it should not be that big difference in the tables.

The climb and descents speeds are to high in the Airbus graph , both If I compare to real life and my performance calculator .
I fly V2500 and Leap engines.
It is strange. I can not really find out why it should not match.
For eg a climb a speed of 308 at CI 0, even if it is near MTOW is way to high.

Same CI in the NEO generally gives a slower speed than in a CEO.

Note:
If we later would like to add the wind effect I found this:

ECON Mach will be corrected in relation to the wind as follows:  
+0.5 point of Mach for 50 knots headwind.
-0.5 point of Mach for 50 knots tailwind.

St54Kevin:
I just did a quick test flight together with the #1295 PR. Everything went as described above. Nothing major happened also I didnt notice any wrong things. Everything felt natural and reasonable. Im looking forward to the next Speed tweaks. We already spoke about details in Discord (S, F Speed, VAPP, VLS etc.) This PR is good to go i think

Benjozork:
I think we can go ahead and merge this.

donstim:
@FlyingCarpet7 I agree that although the document is a bit dated, I would think that the basic relationships would still hold. I wonder if they had a re-think on their climb and descent speeds since that time period?

When you say the same CI in a NEO generally gives a lower speed than in the CEO, are you speaking from personal experience, or are you assuming the speeds in the document are correct for a CEO?

One other thing that confused me in the document is that for the A320, they only showed cost indexes up to 100, calling that "high." For the A300-600, they only went to 60, and for the A310, A330, and A340, they went to 200. Perhaps cost index was normalized differently back then and the maximum CI input in the MCDU was not 999? Another tidbit that might support that is Figure 22 on page 39 (43 of the PDF file) shows an ECON cruise speed of 0.80 up to at least FL330. Max CI speed is supposedly limited to MMO - 0.02, so that doesn't leave much room for any speed increase between CI 100 and CI 999.

I saw that general "rule" for wind affect on speed. That would probably be the easiest part of this to incorporate at this point!

St54Kevin:
Above CI 99 an airbus takes ALT into consideration. So if you put like 250 in it calculates a new optimal Cruise ALT which is faster (Wind, or higher TAS etc.)

FlyingCarpet7:
@donstim @Watsi01 
> When you say the same CI in a NEO generally gives a lower speed than in the CEO, are you speaking from personal experience, or are you assuming the speeds in the document are correct for a CEO?

The same CI gives a lower speed in the NEO compared to the CEO, especially in climb and descent. That is based on experience and also by verifying in my EFB Performance calculator. I fly both NEO Leap and CEO V2500.

Maybe we should make our own table or graph? I can run various weights and numbers in my calculator and we can then interpolate the rest? I could help you with that if you help me to set up what we need?
I get ECON speeds for climb and cruise, but the calculator does not give me an ECON speed for descent.

donstim:
> Above CI 99 an airbus takes ALT into consideration. So if you put like 250 in it calculates a new optimal Cruise ALT which is faster (Wind, or higher TAS etc.) while up to 99 its only a speed modification

Isn't it there just isn't much of a change in OPT ALT for lower CI's? I thought it always used CI in the OPT ALT determination?

Watsi01:
>
> Maybe we should make our own table or graph? I can run various weights and numbers in my calculator and we can then interpolate the rest? I could help you with that if you help me to set up what we need?
> I get ECON speeds for climb and cruise, but the calculator does not give me an ECON speed for descent.

@FlyingCarpet7 That would be very helpful. Thanks for the offer. What could be done is (just as a first suggestion):
Keep the aircraft config steady, Same GW and also same FL. I would say FL350 and GW 68t? Our other / more average numbers?

Then calculate different CI with the EFB: CI 0, 10, 20, 30, 40, 50, ... See what CLB CRZ and DES speeds will come out. And then I would check out, if I can make that into a formula.

donstim:
It depends on how linear the relationships are at different points in the envelope and how many points you are willing to run! I would think that we would probably end up with a table rather than an equation since there would potentially be a different curve of cruise speed for each CI based on Fig 22 of the Airbus document.

I wonder why the calculator doesn't do descent speeds?

To scope the issue, there are 1000 possible CIs (0 through 999). Should probably concentrate on more likely choices, which I assume would be less than CI 100, so we should try for more resolution there (like maybe every 20 CI as @Watsi01 suggested), then less resolution at higher CIs (maybe every 100CI up to 999?). For a particular set of calcs, if you're not seeing much difference between different CIs, then maybe open up the interval. 

The speed will depend on the CI, weight, altitude. temperature,  and wind. I think we can discard looking at wind and just use the simple correction stated in the Airbus document that you noted in a previous comment. For weight, you might want to consider doing at least a min weight, a max weight, and a mid weight. For temp, maybe ISA and ISA+10C? For altitude, probably need to look at doing a number of likely initial cruise altitudes and seeing how much variation there is to see how many altitudes need to be done. Maybe start with 29,000, 35,000, and 39,000?

As you can see, the permutations can get big real fast! Maybe you need a computer guru to hack your calculator and retrieve the equations it is using!

FlyingCarpet7:
ISA variation does not affect the speed at optimum crz altitude, so that could be left out. 
ISA variation does affect max recommended alt, and N1/fuelflow. 

Depending on how this should be programmed and how the data will be created (pilled from a tablet graph, formula ) I can help out to give some numbers if/when needed.
Probably start with Weight and CI as input and generate a speed. 
Then adjust it for wind.

Watsi01:
@FlyingCarpet7 @donstim Yes, it could be more complicated than my first thought. I'm not sure if CIs above 100 are really used by someone. What is CI 800? 1200kts CLB and mach 5.90 CRZ? ;-)

The sources do differ, regarding the CIs commonly used on different airlines. Mainly, I read about common CI from 1 zu 30. Whereas with Boeing planes, I read about CIs of 60 - 100. But Boeing surely calculates CI differently.

If it is not a linear curve, as donstim assumed (and I think you could be right), it will get more complicated and I may need some help, figuring out how to calculate this. Because that exceeds my brain (in the current status). 

@FlyingCarpet7  Maybe you could put some initial number into a table - for a first impression on where this is going to end.

FlyingCarpet7:
Yes, I can try.
The graph is not linear when it comes to CI vs speed. Lower range CI changes make a buffer impact then higher.
We use CI 50 as max if we are running late. 
Normal CI from flightplan is around 6-10.
Would be nice to have a table that can interpolate by it self and fill in the blanks.. 
I only got google sheets.. Don’t know if another excel program is available. Will have a look.

donstim:
CI 999 is constrained by MMO -0.02, or 0.80. Although it may not get used often (or ever?), it CAN be entered, so a result needs to be provided. Cost index is basically a fraction. So, you can normalize it however you want to. You can make it have values from 0 to 1, or 0 to 10,000. According the doc below, Boeing planes vary from 0-200 on 737 classics to 0-9999 on the 777.

Values can also differ depending on whether you are using metric units (kg/min) or imperial units (100 lb/hr). which is 1.32 times bigger. 
[Boeing cost index explained.pdf](https://github.com/flybywiresim/a32nx/files/5413812/Boeing.cost.index.explained.pdf)


I have Excel if it will help.

FlyingCarpet7:
@donstim @Watsi01 
I started with a Cost index table just to get the feel for it.
Did two graphs as well to get the look of how weight and flightlevel are affecting the speeds with different costindex.
If this could be sorted with a super formula it would be better. 
You can try this link to my Google sheet:

https://docs.google.com/spreadsheets/d/1Q4g-p3BUtblrTS29WeJ804EWFhPgLXraR5K5NboHrhU/edit?usp=sharing

Oberstumfuher:
Does anyone have or know where I can download A32NX 1.1.1? I downloaded an MSFS 2020 version 1.7.42, as I do not have the option to "return" the update, I want to use the old one until it is fixed. Right now I used the DEV built version: 2020-10-21T12: 39: 06 + 00: 00, everything worked perfect, great takeoff, great climb, very smooth, smooth cruise ... until .... it was time to land, you already know what happens, so I would like to know, or if someone still has the version of A32NX 1.1.1 that could be made available.

Watsi01:
@FlyingCarpet7 @donstim 
Ok. First, thanks. This is not linear by all means. The problem is: the MCDU doesn't show any mach speeds. In the according files, I didn't find any basic mach speed specification. Even not for CRZ. This speed is also calculated in kts. Although (meanwhile) I'm not sure anymore, if I had any influence in the final CRZ mach speeds, with my changes. The DES and CLB, yes. But mach CRZ?  I'm not sure.

So I'm not able to compare these numbers with any present code / function, cause the code I found was calculated with kts. I'm still searching for any code I oversaw. Wherever it could be. Also the function right now is not considering any weight or altitude. It just takes a CI calculates some basic values with it.

`getCrzManagedSpeed() {
        let dCI = this.costIndex / 999;
        dCI = dCI * dCI;
        const flapsHandleIndex = SimVar.GetSimVarValue("FLAPS HANDLE INDEX", "Number");
        if (flapsHandleIndex != 0) {
            return this.getFlapSpeed();
        }
        let speed = 290 * (1 - dCI) + 310 * dCI;
        if (SimVar.GetSimVarValue("PLANE ALTITUDE", "feets") < 10000) {
            speed = Math.min(speed, 250);
        }
        return speed;`

It could be, that this function was actually made for CI values in the higher area (up to 999). Cause lower values (<100) don't bring much effect to the final CRZ speed.

But maybe that doesn't matter. This has to be a total new function to make it work anyway (considering weight and alt). And the new one could return a mach number instead.  But I don't know if the sim wants / can handle mach numbers from the MCDU right now (have to check the SDK). 

How is the Perf page CRZ displayed in the real thing? Is there a mach number only?


--BTW: I'm off for the weekend (work related), but I'll be back on sunday--

donstim:
Sorry, I haven't been able to get back to this yet. Will do so tomorrow!

Watsi01:
Hi @FlyingCarpet7 and @donstim, I'm back.

So here are the news so far:
It's hard to get these numbers into a function. With some of the curves ( see graph), you kinda could. With some not.

Example: FL330, 65t, CI20 (top line)
via Excel, you can add a trendline with a third degree polynomial, then you get a formula like this:

y = 0,0008x3 - 0,0113x2 + 0,0529x + 0,6973  (x=FL, y=CRZSpeed)

This works quiet well.

But if you take CI 10 (see yellow line) instead of 20, it gets harder. A fourth grade plynomial is used there. 

y = 0,0014x4 - 0,0204x3 + 0,0972x2 + 0,7348

Mach-FL



It seems, there's no ONE function which could handle them all. And they could lack precision in some cases. So what can we do?

Option 1: building a table with a bunch of numbers, and the MCDU just "looks" for the correct values. This of course would mean to calculate any CI, FL and GW with a pro tool (like an EFB) before, and write them down.

Option 2: Take a big drill, and check out what's inside FlyingCarpets EFB (I would prefer this!!!) xD

Option 3: TBA

donstim:
Sorry @Watsi01 and @FlyingCarpet7 I've been trying to improve the flight model, and it's taking a lot longer than I thought it would.

I would vote for option 2 as well! Otherwise, I have been leaning towards option 1.

FlyingCarpet7:
Well, it’s not a good idea to drill into my perf calculator..
We use this: https://www.navblue.aero/product/flysmart-plus/

Every calculation I make takes almost 30 seconds so it is very time consuming. 

For the PROG page we also need to calculate max recommended FL ( should be easy because I think it’s only weight and temp that’s influence) and Optimum FL (I bit trickier)

From the Cost Index calculations we see that not much happens with the speed curves at CI above 50 so it is the lower regions that are more interesting.

I wonder how FSLabs , Tollis and Flightfactor solves this..

If we go for the tables solution I can calculate a few values here and there but we need to autocomplete the tables with interpolation.
I don think we can make tables for all CI and weight combinations , so we need the FMGS to also be able to interpolate between different tables.

PR #1434, opened 2020-10-17
Fix knob animation value not being accessible

lukecologne:
## Summary of Changes

* Change the FBW knob template to be able to set the animation variable type to LVars, at @ThePilot916's request


lukecologne:
Will merge since this is a small change. No functional changes.

PR #1451, opened 2020-10-19
[FIX] Hide blurred spinner mesh

lukecologne:
## Summary of Changes

* Fix the engine rotation templates in order to hide the blur spinner mesh
* This causes the spinner spiral to properly show up when the "medium blur" mesh is shown (see videos)
* Show the "Slow" blur mesh now all the time, this is a workaround to the fact that we cannot add a spinner spiral to the fast blur mesh, and should be removed once we can do this.

## Screenshots (if necessary)

Before: https://streamable.com/iiidxt
After: https://streamable.com/6vvu1p
After, with the slow blurred mesh showing all the time: https://streamable.com/au7de2


Pleasure0102:
Think it is better ! Still only a soloution until we can change the model i would say, but a good compromise for now.

lukecologne:
What I can do is just use the "SLOW" blur mesh all the time. But I don't know if this will look wrong or not, I'll have to try it out

lukecologne:
> 
> 
> Think it is better ! Still only a soloution until we can change the model i would say, but a good compromise for now.

We would only need to change the UV of the "BLUR" spinner, and the add a spiral to it's texture.

lukecologne:
I updated the PR comment with a potential change, where the slow blur mesh would be shown even at high RPM. Let me know what you prefer

lukecologne:
Alright, I asked in the discord which version looks better, and they agreed the version with the "Slow" blur mesh looked best. This is still only a temporary solution until we can fix the UVs for the "blur" spinner and add a spiral to it.

PR #1342, opened 2020-10-14
Custom MCDU Font

tyler58546:
## Summary of Changes

* Added custom MCDU font

## Screenshots (if necessary)

Before:
image
After:
image
image
image

## References

image
image

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build.py will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, slick on the **Artifacts** drop down and click the **A32NX** link

PR #1463, opened 2020-10-20
Update ECAMFontRegular.ttf to have a proper slash in the zero character as the real plane has.

parlock:
Updated font to have a proper slash in the zero character as the real plane has.

## Summary of Changes
Update the ECAM font to have a proper slash in the zero character as the real plane has in its displays.


Benjozork:
Hello, can you include reference for that change ? I cannot find a picture of the ECAM with the 0 character having a slash.

parlock:
Video of real training on the MCDU of the A320.  Noticed the slashed zero character.  There are tons more available to verify...

https://www.youtube.com/watch?v=hzknF9qIUm4

Benjozork:
Right now all the displays use the same font. Someone is already working on a custom MCDU font, but we do not know whether it will be integrated since we haven't decided if we are modelling a Thales or Honeywell MCDU.

You might wanna talk with @wpine215 and @tyler58546 about fonts - they're the main people working on them for the MCDU.

Lucky38i:
not to downplay your contribution but it might be a little mute because of #1342

parlock:
I'd delete it, but it won't let me... sigh!!!

PR #1526, opened 2020-10-24
[Fix] Remove glow from tiller

### LINKED ISSUE #1350: [BUG] Modelling - Steering tiller glowing disconnect button
**Mod Version**

Dev version:
built: 2020-10-14T20:06:03+00:00
ref: refs/heads/master
sha: 3906ea82ef00d54f85387430444c583238144621
actor: Benjozork
event_name: push

**Describe the bug**

The steering tiller disconnect button glows at night. There's no ambient light coming from it, so I don't think this is intentional.

**To Reproduce**

1. Load up at night
2. Apply electrical power (batteries?)
3. Increase main panel and pedestal integral lighting
4. Look at steering tiller

**Expected behavior**

No glowing parts

**Actual behavior**

Steering tiller disconnect button glows.

**References**

No battery power:
image

Battery 1 and 2 on:
image

Main panel and pedestal integral lighting:
image
image
I don't think the button should be glowing...

**Additional context**

Unsure

Unsure

Eefoe#0570

lukecologne:
Fixes #1350

## Summary of Changes

* Removes the glow from the steering tillers.
* This also removes the text integrated lighting, since the button model and text are in the same mesh. I don't know if the text should have integrated lighting anyways, so it would be nice if someone could comment on that.

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/model/A320_NEO_INTERIOR.xml
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/model/A320_NEO_INTERIOR.xml
@@ -713,13 +713,16 @@
                 <MAX_STEERING_VELOCITY>30</MAX_STEERING_VELOCITY>
             </UseTemplate>
 
+
             <UseTemplate Name="ASOBO_GT_Component_Emissive_Gauge">
                 <NODE_ID>PUSH_HANDLE_LEFT_YOKE</NODE_ID>
-                <POTENTIOMETER>85</POTENTIOMETER>
+                <!--<POTENTIOMETER>85</POTENTIOMETER>-->
+                <EMISSIVE_CODE>0</EMISSIVE_CODE>
             </UseTemplate>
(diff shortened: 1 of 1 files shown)


Pleasure0102:
@lukecologne what you mean by model and text in the same mesh ? AFAIK the text is as decal, not the same mesh.

lukecologne:
You can look at the model, normally it is like this, yes, but not on the tiller. That is why they made it so the whole tiller is glowing

Pleasure0102:
yeah, i actually checke myself again, you are correct.. oh my god we really have to fix that when we can edit the model, but i think for now disabling glow is better than the weird glowing tiller model

DarkOfNova:
> yeah, i actually checke myself again, you are correct.. oh my god we really have to fix that when we can edit the model, but i think for now disabling glow is better than the weird glowing tiller model

Agreed as it's inop anyway, no point in being an eye sore

lukecologne:
Sure, i'll merge it then.

PR #1548, opened 2020-10-26
[GPWS] Add minimums and 100above logic

linkeleven:
Fixes #1461 [REQUEST] decision height/minimum callouts
Fixes #808 [BUG]Minimum Call 

## Summary of Changes

- Added logic for minimum and 100 above callout
- MDA will trigger the callouts based on indicated altitude (barometric altimeter)
- DH will trigger callouts based on altitude above ground (radio altimeter)

## Additional context

Discord username (if different from GitHub):

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build.py will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, slick on the **Artifacts** drop down and click the **A32NX** link

PR #1564, opened 2020-10-27
[MCDU] MDA/DH inputs

### LINKED ISSUE #1551: [BUG] [MCDU] MDA and DH input
version: master

**Describe the bug**

1. Currently booth MDA and DH can be entered at the same time.
2. Values entered in either MDA or DH cant be cleared with CLR or NO

**To Reproduce**

1. Enter MDA and DH on the approach phase page in the MCDU
mcdu_MDA_DH

2. Try to clear the MDA or DH with CLR or entering NO
mcdu_MDA_CLR_1
mcdu_MDA_NO_1
mcdu_MDA_CLR_2

**Expected behavior**

1. Only one value can be entered at the same time
2. If a value for MDA is entered the entry for DH should be cleared and vice versa
3.Entering CLR or NO should clean the entry

**Actual behavior**

**References**

grafik

**Additional context**

paul92ilm:
Fixes #1551 

## Summary of Changes
- Add ability to clear MDA/DH
- Add "NO DH"  logic (input NO, NODH or NO DH if ILS approach selected)
- Add DH range (0 to 700)
- Fix : only one value (MDA or DH) can be entered at the same time

## Screenshots (if necessary)
Before :
image

## References
image

## Additional context
Discord username (if different from GitHub): Lussion


Benjozork:
Thank you for your first PR!

Benjozork (approved):

lousybyte:
Any reason for new functions? They are already defined here and only used on the PERF page from what I can tell: [https://github.com/flybywiresim/a32nx/blob/87324ef670e164a028fd698dbdd95396f9c52943/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js#L1581-L1600](https://github.com/flybywiresim/a32nx/blob/87324ef670e164a028fd698dbdd95396f9c52943/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js#L1581-L1600)

beheh:
Which document is that reference from? I seem to remember some kind of logic that forbids entering a standalone "0" as DH and instead requires entering "NO", and I'd like to check the reference for that.

linkeleven:
@beheh , @lukecologne gave the reference picture

paul92ilm:
> Which document is that reference from? I seem to remember some kind of logic that forbids entering a standalone "0" as DH and instead requires entering "NO", and I'd like to check the reference for that.

"NO DH" added for NO, NO DH and NODH input in 3R

paul92ilm:
image
image

beheh:
Oh thanks, I'm blind, it was right there all along :-) That matches what I remember.

beheh:
I believe the motivation for the explicit "NO DH" input is that 0 cannot be inserted accidentally. I see nothing in the code that prevents a 0, or even a negative number from being inserted. Maybe we could double-check this against Allowed MCDU Data Values, or from RL pilot feedback?

paul92ilm:
Agree with you for the negative DH that makes no sense.
But I don't have any references for that point and for the 0 input in DH.

This PR fully fixes the original issue #1551, maybe open a new issue when we found references for 2 others points ?

beheh:
Yeah, sounds good.

beheh:
Actually I just searched a little more and I found this referenced in an FCOM under DATA FORMAT LIST:

image

So I assume 0 is fine actually because there's a difference between a decision height of 0 and no decision height, so I don't think there's a bug here. Nonetheless, you might interested in the official bounds here (0-700). Otherwise I'll just a file a new issue once this is merged.

It also sounds like NO is only available when an ILS approach is selected, but I don't think the current behavior encodes that. Any little step towards the correct implementation is good though, so you might as well keep it as is.

Keep up the good work!

paul92ilm:
Thank you for reference, range added

I also think that there is a huge work to do on MDA/MDH/Baro (2R) and DH/Radio (3R) logics

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
```diff
@@ -558,18 +558,25 @@ class CDUPerformancePage {
...
             mcdu.clearUserInput();
-            if (mcdu.setPerfApprMDA(value)) {
+            if (mcdu.setPerfApprMDA(value) && mcdu.setPerfApprDH(FMCMainDisplay.clrValue)) {
                 CDUPerformancePage.ShowAPPRPage(mcdu);
             }
         };
         let dhCell = "[ ]";
         if (isFinite(mcdu.perfApprDH)) {
-            dhCell = mcdu.perfApprDH.toFixed(0);
+            if (mcdu.perfApprDH >= 0 && mcdu.perfApprDH <= 700) {
```
lousybyte, line 567: I think this check can be moved in the `setPerfApprDH(s)` function, since you want to check the values on input instead of on display.

paul92ilm: Fully agree 👍
Check moved to `setPerfApprDH(s)`
Thank you

danbw9:
Name: danbw9#7028
Date of testing: Current date 28/10/2020
Version of the sim: 1.9.5.0
PR Tested: #1564
Object of test: Checking correct behaviour with MDA/DH inputs

Testing technique: Test two phases of flight. Spawn C&D and plan as if doing a flight and check APP PERF page for the above described features. Then spawned on an approach to UACC and checked any differences to the input behaviour´.

Attempt the scenarios listed in the above PR
- checking a DH of 800 does not work
- checking only MDA or DH can be displayed
- writing '0' and 'NO' into DH all fine.
- able to clear any inputs


Flight notes
Startup: Everything normal. 

Taxi/Takeoff : N/A

Cruise: N/A

Descent/Landing : Everything normal. 

Shutdown: N/A

Issues (If Any): None
Overall Rating: Good
Conclusions of the testing: LGTM. Thank you for your work. Not sure if inputting '0' is suitable. Another IRL pilot could comment on that. As is mentioned above -  'NO DH' is only for ILS approach with a CAT IIIB approach, so a condition could be created for that.

beheh:
We've confirmed in Discord that 0 is appropriate and means something different, and it's as per the Honewell documentation linked above:

image

danbw9:
@beheh Thanks. Across two operators I have never seen that before. In practice it is probably never used. Just like CAT IIIC.

paul92ilm:
I'm not sure that ILS CATIIIB is implemented yet.
Maybe open an issue for that particular point after PR ?
This PR is needed for Minimums Logic #1548 and NO DH wasn't part of the initial issue #1551

lousybyte (approved):

linkeleven:
Name: linkeleven#2557
Date of testing: Current date 30/10/2020
Version of the sim: 1.10.7
PR Tested: #1564
Object of test: Checking correct behaviour with MDA/DH inputs

Testing technique: 1 flight from C&D, 5 flights spawned at runway

Attempt the scenarios listed in the above PR

- Add ability to clear MDA/DH 
// could clear MDA/DH at any phases of flight
- Add "NO DH" logic (input NO, NODH or NO DH if ILS approach selected) 
// could only enter NO, NODH, NO DH if an ils approach was loaded in the MCDU
- Add DH range (0 to 700) 
// could only enter numbers from 0-700 in DH
- Fix : only one value (MDA or DH) can be entered at the same time 
// entering a number in the empty Field clears the other one

Taxi/Takeoff : N/A

Cruise: N/A

Descent/Landing : Everything normal.

Shutdown: N/A

Issues (If Any): 
- entering a MDA <= 0, will still show NO DH in the PFD when in approach phase

Additional note:
- As discussed with me, DH is set to -1 when cleared, NO, NODH or NO DH. this is a preparation for the minimums callout logic

Overall Rating: Good

Conclusions of the testing: everything as described works perfectly fine, just the one issue i stated above

PR #1588, opened 2020-10-28
[MCDU] Changing cost index to a 0-99 range

### LINKED ISSUE #345: [REQUEST] Cost index does not seem to do anything
When I change the cost index for the flight the cruise is always at 285 knots. I can only change the cost index through the INIT page and when I do the cruise speed does not change. You can also not change the cost index on the performance pages on the MCDU. The cost index does, however, change the descent speed anywhere from 240-242. A lower cost index should result in a slower climb speed, cruise speed, and descent speed. The opposite is true for a higher cost index. I have no idea how much it should change by, just that it should change.

**Describe the features you'd like to see implemented**
I would like to see cost index be simulated. Speed can be controlled manually for now so it is not a big deal, but it would be nice to have.

**Additional context**
You guys are doing a great job with this mod, keep it up 👍

Watsi01:
Fixes #345 

## Summary of Changes
Changed the cost index, so 0-99 inputs in the MCDU are creating an according impact on CLB, CRZ and DES flightphase speeds.

## Additional context
I kept the vanilla function to calculate the cost index as it was before. I only changed the range of the given cost index. Before, the function headed for ranges between 0-999, whereas cost indexes in the A320 are more in a range of 0-99 (Pilots confirm?). Therefore an input of a cost range of 20 didn't make any difference from a cost range of 30 (examples). Now it does.

Maybe some fine tuning of the basic speeds of the flightphases has to be done. Have to check the impact on higher values (like 80).

Discord username (if different from GitHub): RogePete

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
@@ -229,7 +229,7 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
         if (flapsHandleIndex != 0) {
             return Math.min(maxSpeed, this.getFlapSpeed());
         }
-        let dCI = this.costIndex / 999;
+        let dCI = this.costIndex / 99;
         dCI = dCI * dCI;
         let speed = 290 * (1 - dCI) + 330 * dCI;
         if (SimVar.GetSimVarValue("PLANE ALTITUDE", "feets") < 10000) {
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js
@@ -993,7 +993,7 @@ class FMCMainDisplay extends BaseAirliners {
         return false;
     }
     getClbManagedSpeed() {
-        const dCI = this.costIndex / 999;
+        const dCI = this.costIndex / 99;
         const flapsHandleIndex = Simplane.getFlapsHandleIndex();
         if (flapsHandleIndex != 0) {
             return this.getFlapSpeed();
@@ -1005,7 +1005,7 @@ class FMCMainDisplay extends BaseAirliners {
         return speed;
     }
     getCrzManagedSpeed() {
-        let dCI = this.costIndex / 999;
+        let dCI = this.costIndex / 99;
         dCI = dCI * dCI;
         const flapsHandleIndex = SimVar.GetSimVarValue("FLAPS HANDLE INDEX", "Number");
         if (flapsHandleIndex != 0) {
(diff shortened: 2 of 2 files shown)


lukecologne:
Cost index is 0-999 on the honeywell. 
grafik

Watsi01:
@lukecologne I won't contradict that. But any Pilot I spoke to said, they are in an input range below 30 (some below 10). And with the vanilla function, this has no impact. So I wonder what's the right way to do it.

lousybyte:
You also need to change this line as well if this goes through:
[https://github.com/flybywiresim/a32nx/blob/68da95724d8e2f546f3857040bc7719473a0b570/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js#L389](https://github.com/flybywiresim/a32nx/blob/68da95724d8e2f546f3857040bc7719473a0b570/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FMC/A32NX_FMCMainDisplay.js#L389)

Also not sure if this fixes the issue mentioned, since technically the CI is working as intended, and the other half of that issue is regarding input not working as mentioned, which is not fixed in this PR.

donstim:
It really doesn't matter what CI pilots typical use. CIs from 0 to 999 are possible in the real airplane, so they should also be possible here. Note:  Last time I tried to enter a 0 CI, it wouldn't take it. Need to make sure it will take a 0 input for CI.

Watsi01:
So I close this, since it's not welcome. Thanks to all.

lousybyte:
> It really doesn't matter what CI pilots typical use. CIs from 0 to 999 are possible in the real airplane, so they should also be possible here. Note: Last time I tried to enter a 0 CI, it wouldn't take it. Need to make sure it will take a 0 input for CI.

0 is the default value - if it should be or not, that's a different issue - but by default the sim uses the lowest managed speeds based on CI 0.

Watsi01:
@donstim By the way: the Honeywell, providing a range from 0-999 doesn't necessarily mean, that Airbus is using it. If Airbus is calculating with CIs like this

airbusci

with a range of 0-999, it would have no impact on speeds.

lousybyte:
Those numbers are just an example, it says above that the formula is being scaled based on input range.

donstim:
> > It really doesn't matter what CI pilots typical use. CIs from 0 to 999 are possible in the real airplane, so they should also be possible here. Note: Last time I tried to enter a 0 CI, it wouldn't take it. Need to make sure it will take a 0 input for CI.
> 
> 0 is the default value - if it should be or not, that's a different issue - but by default the sim uses the lowest managed speeds based on CI 0.

You're telling me that a climb speed of 320 kts is for a CI of 0?

St54Kevin:
I can also confirm inputs over 99 are possible. The plane then includes flight level into the calculation. So up to 99 its changes speeds, over that it changes cruise level, because FL390 might not be the fastest.

lousybyte:
@donstim currently above FL100 it's 290kts with CI 0 and 330kts with CI 999 during CLB
@St54Kevin nothing has changed, it's more of a discussion if it should change and why

St54Kevin:
> @donstim currently above FL100 it's 290kts with CI 0 and 330kts with CI 999 during CLB
> @St54Kevin nothing has changed, it's more of a discussion if it should change and why

Oh I thought  it got already merged because its under closed PRs

donstim:
I realize this is closed, but just for the record, you should still be able to input a CI of 0, regardless of whether or not that is chosen to the the default for this mod.

PR #1597, opened 2020-10-29
[CDU] Duplicate waypoint names page rework

### LINKED ISSUE #1480: MCDU - Waypoint distance in the list of duplicate waypoints.
It is a request

When there are more waypoints with same names in the database, and you want to insert into the flight plan, the mcdu will give you a list of waypoints with gps coordinates. On a real MCDU on the right side the distance of the waypoint should be presented to that it is obvious to select the closest (proper) one.

**References**
https://www.youtube.com/watch?v=c7UJlJMJdfk&feature=youtu.be&t=591&ab_channel=320SimPilot
This guy is a real A320 pilot and clearly says that on the right side normally there is the distance from your position.

mcdubefore-1
mcduafter-1

**Additional context**
N/A

BreakerHUN#3986

pessip:
Fixes #1480

## Summary of Changes

I reworked the duplicate waypoint page to make it look more like the real one. I added distance to waypoint and navaid frequencies (on airports it just uses the first frequency in the frequencies array). I removed the waypoint type. Visually I redisigned the page with columns and added colours.

## Screenshots (if necessary)

Before:
before
After:
after

## Additional context

Normally the number of nautical miles is in green and NM in white but that doesn't seem to be possible with the sim.

Discord username: pessimoo#6055


wpine215:
Is it possible to remove the "LAT/LONG" and "FREQ" titles for all but the top small-text row, to match the reference?

pessip:
@wpine215 I'll do that

MisterChocker:
Does the screenshot above shows the before or after version?

pessip:
@MisterChocker I added a before screenshot. That was the after version

MisterChocker:
Thanks, could you add the * in front of the indicator as well?

MisterChocker:
Am I the only one seeing the L in the last row at ENO? Is this a bug or feature?

pessip:
@MisterChocker Yeah I thought that was wierd also but I haven't changed anything related to that so it's probably just normal behaviour.

MisterChocker:
seems odd, you could check the name tho before inserting them to the list.

MisterChocker:
Would also be nice to remove LAT/LONG and FREQ for every row except the the most top.

pessip:
I have confirmed that this works on the newest version and merged the changes from master.

beheh:
After #1830 lands you should be able to split the label color here into green/white for distance/nm, as that does currently not match the reference.

Benjozork:
Hello, please rework to fit #1830

beheh:
#1830 has landed - after rebasing you can now do something like this:

```javascript
"{green}" + distance + "{end}nm"
```

...to get your split colors for e.g. `180nm`.

beheh:
You should still implement sorting the waypoints by distance like in reality:

> The DUPLICATE NAMES page normally displays  the waypoints starting with the one closest to the aircraft.

image

The latter part ("is not displayed when two waypoints with the same identifier") could be handled in a different place as this page doesn't even need to load in that case, but also does not have to happen as part of this PR.

pessip:
Great I'll work on trying to sort them today. And now that that works I van also add the colors!

pessip:
What should be displayed if there is no frequency associated with the waypoint?

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU.html
```diff
@@ -1,132 +1,132 @@
-﻿<link rel="stylesheet" href="A320_Neo_CDU.css" />
```
beheh, line 1: I think your commit messed up the line endings here. You probably will want to revert this file.

wpine215:
Tested, LGTM!

PR #1785, opened 2020-11-07
Add fs-base-ui typedefs

davidwalschots:
## Summary of Changes
Add TypeScript type definitions for a lot of fs-base-ui types to help us develop correct code quicker and easier.

_Some notes_:

As the Revealing Module Pattern is used a lot in Asobo's code, many of the definitions contain namespaces, which are best suited to represent this pattern.

Given that `SimVar.GetSimVarValue` may return `null`, a lot of the `SimPlane` functions that use `SimVar.GetSimVarValue` may also return `null`. Only functions that cannot return `null` (e.g. because they use `Math.max` which never returns `null`) have an output type without `null`.

## Additional context

**Note that QA doesn't apply to this PR, as the PR doesn't contain any code that runs inside the sim.**


types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1101 @@
...
+         * strToBool("false");
+         * strToBool("unicorns and rainbows");
+         */
+        function strToBool(value: string): boolean;
+
+        /**
+         * Filters the given input element with the given filter. When the filter returns true, the value is accepted,
+         * when it returns false, the previous input element's value is set.
+         */
+        function setInputFilter(inputElement: HTMLInputElement | HTMLTextAreaElement, inputFilter: (value: string) => boolean);
```
lars-reimann, line 274: The return type of setInputFilter is not specified (void).

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+         * Gets the VR speed up during and before takeoff, -1 after.
+         */
+        function getV2Airspeed(): Knots | -1 | null;
+        function getREFAirspeed(): Knots | null;
+        function getVXAirspeed(): Knots | null;
+        function getFMCGreenDotSpeed(): number | null;
+        function getFMCApprGreenDotSpeed(): number | null;
+        function getGreenDotSpeed(): Knots | null;
+        function getCrossoverSpeedFactor(cas, mach: Mach): number | null;
+        function getFlapsLimitSpeed(aicraft: any, flapIndex: number): Knots | null;
```
lars-reimann, line 250: aicraft -> aircraft

types/fs-base-ui/html_ui/JS/Types.d.ts
```diff
@@ -0,0 +1,196 @@
...
+         * Parses a string into a LatLong or LatLongAlt
+         * @param str A string formatted as either "0.0,0.0" or "0.0,0.0,0.0".
+         * When the string contains spaces around the numbers, those are ignored.
+         * @returns An instance of LatLong or LatLongAlt, depending on the number of values in the string. Null when
+         * the string doesn't contain a comma.
+         */
+        static fromStringFloat(str: string): LatLong | LatLongAlt | null;
+    }
+
+    class LatLongAlt {
```
lars-reimann, line 52: Would it be correct to declare this as a subclass of LatLong? Then it would not be necessary to write `LatLong | LatLongAlt` for Avionics.Utils.computeDistance, Avionics.Utils.computeGreatCircleHeading, Avionics.Utils.computeGreatCircleDistance, etc.

davidwalschots: It is not a subtype in Asobo's code, which is why I haven't declared it as such. I did consider writing `type LatLongAltitude = LatLong | LatLongAlt` but decided against it, because I feel it is confusing to developers.

types/fs-base-ui/html_ui/JS/Avionics.d.ts
```diff
@@ -0,0 +1,198 @@
...
+            static InlineSvg(url: string, imageElement: InnerHTML): void;
+
+            static computeDashLine(startX: number, startY: number, length: number, steps: number, width: number, color: string): SVGElement;
+        }
+
+        class SVGGraduation {
+        }
+
+        class Scroller {
+            constructor(nbItems: number, increment: number, canBeNegative: boolean,
```
lars-reimann, line 81: `canBeNegative`, `moduloValue` and `notched` are optional.

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+            OnListenerRegistered: (name: any) => void;
+            onRegister(name: string, viewListener: ViewListener): void;
+            getListenerByName(name: string): ViewListener;
+            onUnregister(name: string, viewListener: ViewListener, force?: boolean): void;
+        }
+
+        let g_ViewListenersMgr: ViewListenerMgr;
+    }
+
+    function RegisterViewListenerT<T>(name: string, callback: (() => void) | void, type: new() => T,
```
lars-reimann, line 615: `callback` is optional.

types/fs-base-ui/html_ui/JS/dataStorage.d.ts
```diff
@@ -0,0 +1,22 @@
...
+        deleteData(key: string): any;
+    }
+
+    interface DataStorageSearchData {
+        key: any;
+        data: any;
+    }
+
+    function OnDataStorageReady(): void;
+    function GetDataStorage(): DataStorage | void;
```
lars-reimann, line 15: Can also return `null`.

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+
+        /**
+         * Returns the remainder of one number divided by another.
+         */
+        function Modulo(num: number, mod: number): number;
+
+        /**
+         * Returns the number with as many "0" prefixes as needed to get to a string of the given length.
+         */
+        function pad(num: number, length: number): string;
```
lars-reimann, line 183: Would also for other types for `num` than number, e. g. if `num` was a string.

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+         * Generates a lorem ipsum text of the given length.
+         */
+        function generateLorem(length: number): string;
+
+        /**
+         * Removes "bitch", "shit", and "asshole" from the given string.
+         */
+        function filterProfanity(value: string): string;
+
+        function Translate(key: string): string;
```
lars-reimann, line 328: Can also return `null`.

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+         * @param value The numeric value to add zeros to.
+         * @param nbDigits The minimum number of digits before the decimal separator.
+         * @param pointFixed The minimum number of digits after the decimal separator.
+         */
+        function leadingZeros(value: number, nbDigits: number, pointFixed?: number): string;
+
+        /**
+         * Counts the number of digits after the decimal separator.
+         */
+        function countDecimals(value: number): number;
```
lars-reimann, line 394: Would also work if `value` already was, say, a string.

types/fs-base-ui/html_ui/JS/simvar.d.ts
```diff
@@ -0,0 +1,114 @@
...
+type FuelLevelsVarUnit = "FuelLevels";
+type GlassCockpitSettingsVarUnit = "GlassCockpitSettings";
+
+type SimVarBatchType = "string" | "number";
+
+declare global {
+    namespace SimVar {
+        class SimVarValue {
+            constructor(name?: string, unit?: VarUnit, type?: string);
+            name?: string;
```
lars-reimann, line 29: Should `__type` be listed here, since it's also listed for `RangeDisplay` and other displays?

types/fs-base-ui/html_ui/JS/simvar.d.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        function SetSimVarValue(name: string, unit: NumberVarUnit, value: number, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: TextVarUnit, value: string, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: LatLongAltVarUnit, value: LatLongAlt, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: LatLongAltPBHVarUnit, value: LatLongAltPBH, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: PitchBankHeadingVarUnit, value: PitchBankHeading, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: PID_STRUCTVarUnit, value: PID_STRUCT, dataSource?: string): Promise<void>;
+        function SetSimVarValue(name: string, unit: XYZVarUnit, value: XYZ, dataSource?: string): Promise<void>;
+
+        function GetGlobalVarValue(name: string, unit: any): any | null;
```
lars-reimann, line 96: Should `unit` be a string here as well?

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        /**
+         * Gets the VR speed up during and before takeoff, -1 after.
+         */
+        function getV2Airspeed(): Knots | -1 | null;
+        function getREFAirspeed(): Knots | null;
+        function getVXAirspeed(): Knots | null;
+        function getFMCGreenDotSpeed(): number | null;
+        function getFMCApprGreenDotSpeed(): number | null;
+        function getGreenDotSpeed(): Knots | null;
+        function getCrossoverSpeedFactor(cas, mach: Mach): number | null;
```
lars-reimann, line 249: `cas` has no type.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        function getAutoPilotSelectedAirspeedHoldValue(): Knots | null;
+        function getAutoPilotManagedAirspeedHoldValue(): Knots | null;
+        function getAutoPilotMachModeActive(): boolean | null;
+        function getAutoPilotMachHoldValue(): number | null;
+        function getAutoPilotSelectedMachHoldValue(): number | null;
+        function getAutoPilotManagedMachHoldValue(): number | null;
+        function getAutoPilotHeadingManaged(): boolean;
+        function getAutoPilotHeadingSelected(): boolean;
+        function getAutoPilotHeadingLockActive(): boolean | null;
+        function getAutoPilotHeadingLockValue(radians?: boolean): Radians | Degrees | null;
```
lars-reimann, line 288: Maybe this can be restricted with boolean literal types:
```ts
        function getAutoPilotHeadingLockValue(radians?: true): Radians | null;
        function getAutoPilotHeadingLockValue(radians?: false): Degrees | null;
```

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        function getAutoPilotSelectedHeadingLockValue(radians?: boolean): Radians | Degrees | null;
```
lars-reimann, line 289: Same as above.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        function getAutoPilotAltitudeManaged(): boolean;
+        function getAutoPilotAltitudeSelected(): boolean;
+        function getAutoPilotAltitudeArmed(): boolean | null;
+        function getAutoPilotAltitudeLockActive(): boolean | null;
+        function getAutoPilotFLCActive(): boolean;
+        /**
+         * @param units Default = feet.
+         */
+        function getAutoPilotAltitudeLockValue(units?: string): Feet | null;
```
lars-reimann, line 298: Could also return something other than feet if we change the unit. Maybe just say it returns `number | null`?

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        /**
+         * @param units Default = feet.
+         */
+        function getAutoPilotSelectedAltitudeLockValue(units? : string): Feet | null;
```
lars-reimann, line 302: Same as above.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        /**
+         * @param units Default = feet.
+         */
+        function getAutoPilotDisplayedAltitudeLockValue(units?: string): Feet | null;
```
lars-reimann, line 306: Same as above.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        function getGroundReference(): Feet | null;
+        function getTurnRate(): RadiansPerSecond | null;
+        function getHeadingMagnetic(): Heading | null;
+        function getPitch(): Degrees | null;
+        function getBank(): Degrees | null;
+        function getFlightDirectorPitch(): Degrees | null;
+        function getFlightDirectorBank(): Degrees | null;
+        function getIsGrounded(): boolean;
+        function getAltitudeAboveGround(forceSimVarCall?: boolean): Feet;
+        function getCrossoverAltitude(cas, mach: Mach): Feet | null;
```
lars-reimann, line 383: `cas` has no type.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
```diff
@@ -0,0 +1,421 @@
...
+        function getFuelPercent(): Percent;
+        function getFuelQuantity(): Gallons;
+        function getTotalFuel(): Kilograms | null;
+        function getFuelUsed(engineIndex: number): Kilograms | null;
+        function getCompassAngle(): Radians | null;
+        function getPressureValue(): InchesOfMercury | null;
+        function getPressureValue(units?: "inches of mercury"): InchesOfMercury | null;
+        function getPressureValue(units?: "millibar"): Millibar | null;
+        function getPressureSelectedUnits(): "inches of mercury" | "millibar";
+        function getPressureSelectedMode(aircraft: Aircraft): "QFE" | "QNH" | "STD" | "";
```
lars-reimann, line 406: Should the other (unused) `aircraft` parameters also have type `Aircraft`?

lars-reimann (commented):
Looks extremely thorough and precise overall and should be a massive boon for further development.

types/fs-base-ui/html_ui/JS/Avionics.d.ts
```diff
@@ -0,0 +1,198 @@
...
+            arc?: HTMLElement;
+            arcRadius?: number;
+            arcSize?: number;
+            arcColor?: string;
+
+            init(arcName: string, radius: number, size: number, color: string): void;
+            setPercent(percent: number): void;
+            translate(x: number, y: number): void;
+            rotate(angle: number): void;
+            get svg(): HTMLElement;
```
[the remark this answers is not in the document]
davidwalschots: For the two comments regarding `SVGPathElement`, I'm indeed changing them to `SVGPathElement` (good catch) but keeping them optional/`undefined` as well, as the value of `arc` is only set in an init function.

types/fs-base-ui/html_ui/JS/Avionics.d.ts
```diff
@@ -0,0 +1,198 @@
...
+        class Curve {
+            add(time: number, value: any): void;
+            evaluate(time: number): number | string;
+            debugLog(): void;
+        }
+
+        class DictionaryItem {
+        }
+
+        class Dictionary {
```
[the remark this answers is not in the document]
davidwalschots: The problem with this `Dictionary` type is that it isn't a dictionary. A dictionary in JavaScript would in fact be an object, where the object's properties would be the key and thus have either type `string | number` and the object's value would then be any type. This would be far more efficient. I.e.:
```typescript
let innerObj = {};

function set(key: string | number, value: any) {
    innerObj[key] = value;
}

function get(key: string | number): any {
    return innerObj[key];
}
```

Nothing in the code tells me it is limited to a particular type of key or value so any is best. **I'd definitely not use it in our code.**

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+
+    enum KeyNavigationDirection {
+        KeyNavigation_None = 0,
+        KeyNavigation_Horizontal = 1,
+        KeyNavigation_Vertical = 2,
+        KeyNavigation_Grid = 3
+    }
+
+    class UIElement extends HTMLElement {
+        static getUIElement(elem: UIElement): UIElement | null;
```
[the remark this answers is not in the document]
davidwalschots: I've made it `null`, because I'm certain it will return `null` and not `undefined`. `type void = null | undefined`.

types/fs-base-ui/html_ui/JS/common.d.ts
```diff
@@ -0,0 +1,1102 @@
...
+            onEventToAllSubscribers: (eventName: string, ...args: any[]) => void;
+            onGlobalEvent(eventName: string, ...args: any[]): void;
+            off(name: string, callback: (...args: any[]) => void, context: any): void;
+            on(name: string, callback: (...args: any[]) => void, context: any): void;
+            trigger(name: string, ...args: any[]): void;
+            triggerToAllSubscribers(event: any, ...args: any[]): void;
+        }
+
+        class ViewListenerMgr {
+            OnListenerRegistered: (name: any) => void;
```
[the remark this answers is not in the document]
davidwalschots: Yeah. I came to the same conclusion. For now I can't declare the type. What I've done now is put the Asobo fs-base-ui code in a git repository to compare it on every release and see where the interface changes are. Maybe they'll fix it eventually? ;)

types/fs-base-ui/html_ui/JS/common.d.ts
davidwalschots: I have declared it optional with the `void` type. An optional parameter before a non-optional parameter cannot be defined in TypeScript (and I doubt it's legal in JavaScript).

types/fs-base-ui/html_ui/JS/dataStorage.d.ts
davidwalschots: `type void = null | undefined;`

types/fs-base-ui/html_ui/JS/common.d.ts
davidwalschots: I'll add `string` to the types. I don't think adding any other type would be in line with the "thought" behind this function.

types/fs-base-ui/html_ui/JS/common.d.ts
davidwalschots: Same as my earlier comment: I'll add string to the types. I don't think adding any other type would be in line with the "thought" behind this function.

types/fs-base-ui/html_ui/JS/simvar.d.ts
davidwalschots: I'll remove it on `RangeDisplay` and the others. I was in doubt for a while, but I believe any `_` prefixed variable or `m_` variable was intended to be private.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
davidwalschots: Using the JS code, I was unable to determine the type. Any suggestions?

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
davidwalschots: I think it's a valuable idea and thus wrote:

```typescript
function getAutoPilotHeadingLockValue(): Radians | null;
function getAutoPilotHeadingLockValue(radians: true): Radians | null;
function getAutoPilotHeadingLockValue(radians: false): Degrees | null;
```

However, sadly only the first two are correctly "identified" by WebStorm. When I pass false it will still point to the second (true) overload. Thus I've decided to stick to the current version for now.

types/fs-base-ui/html_ui/JS/Avionics.d.ts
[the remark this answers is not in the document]
lars-reimann: Yes, definitely. Also, the only addition of `Dictionary` over a proper [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), which could have arbitrary keys and still has constant time access, would be the `changed` property. If `changed` is indeed useful (and therefore this Dictionary class), we could just replace the `items` array with a Map and implement `set`, `get`, `remove` and `exists` in terms of this. But we aren't using `Dictionary` anyway.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
davidwalschots: Using the JS code, I was unable to determine the type. Any suggestions?

types/fs-base-ui/html_ui/JS/common.d.ts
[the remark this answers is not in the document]
lars-reimann: Sorry, I was talking about the type of the parameter `elem` not the return value. In `Name_Z` you write `static isValid(a: Name_Z | void): boolean;` when there is a check `if (a) ...` in the function and here we also have
```ts
        if (elem) {
            // ...
        }
```
But it's not really important anyway.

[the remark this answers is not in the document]
davidwalschots: Got it, thanks. You're right. I'm changing it! 👍

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
lars-reimann: I'd guess that `cas` would be the calibrated airspeed, which would be measured in Knots.

types/fs-base-ui/html_ui/JS/SimPlane.d.ts
lars-reimann: I'd guess that cas is the calibrated airspeed, which would be measured in Knots.

davidwalschots:
> 
> 
> Looks extremely thorough and precise overall and should be a massive boon for further development.
> 
> (Also sorry for the split review comments from earlier, I didn't know I was able to group them together).

Thank you for your thorough review! I hope that unlike me you won't be dreaming about types in the coming days.

PR #1839, opened 2020-11-09
LocalStorage SimVar distributor

davidwalschots:
## Summary of Changes

Reading SimVars inside JS causes performance issues since it does a blocking engine call. Thus far, we have considered SimVars as the only mechanism to share data between various displays in the cockpit. My intent is to create a mechanism that distributes SimVars by reading them in one component and then making it available to all other components through localStorage. After finishing this work, I'll measure the performance to see if it does indeed provide a noticeable performance improvement.

**Any comments on how to best measure performance are greatly appreciated.**

## Additional context

### DIFF
--- a/A32NX/html_ui/Pages/A32NX_Core/SimVarPerformance.js
+++ b/A32NX/html_ui/Pages/A32NX_Core/SimVarPerformance.js
@@ -1,0 +1,53 @@
+(function () {
+    const originalGetSimVarValue = SimVar.GetSimVarValue;
+
+    // Uncomment one of the lines below to switch to that form of SimVar usage.
+    // It will activate 10 seconds after the panel comes "online", so that any startup logic that must affect
+    // the sim's SimVars can still run.
+
+    // setTimeout(useInMemoryStorage, 10000);
+    // setTimeout(useLocalStorage, 10000);
+
+    function useInMemoryStorage() {
+        const simvarValues = {};
+        SimVar.GetSimVarValue = function (name, unit, dataSource = "") {
+            if (simvarValues[name] !== undefined) {
+                return simvarValues[name];
+            } else {
+                const value = originalGetSimVarValue(name, unit, dataSource);
+                simvarValues[name] = value;
+
+                return value;
+            }
+        };
+
+        SimVar.SetSimVarValue = function SetSimVarValue(name, unit, value, dataSource = "") {
+            simvarValues[name] = value;
+        };
+    }
+
+    function useLocalStorage() {
+        SimVar.GetSimVarValue = function (name, unit, dataSource = "") {
+            const item = localStorage.getItem(name);
+            if (item !== null) {
+                switch (unit) {
+                    case "Text":
+                    case "string":
+                        return item;
+                    default:
+                        return Number(item);
+                }
+                return item;
+            } else {
+                const value = originalGetSimVarValue(name, unit, dataSource);
+                localStorage.setItem(name, value);
+
+                return value;
+            }
+        };
+
+        SimVar.SetSimVarValue = function SetSimVarValue(name, unit, value, dataSource = "") {
+            localStorage.setItem(name, value);
+        };
+    }
+})();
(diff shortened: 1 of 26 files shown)


wpine215:
Just a thought - we're not sure if localstorage uses memory then writes to disk on exit, or reads/writes to disk directly no matter what. If it's the latter, there could be a possibility of performance issues as well. 

Would be good looking into a central class in A32NX_Core which every display calls/inherits from, that way it can all be done with javascript and memory access.

lukecologne:
How about sessionStorage? That sounds like it wouldn't write on disk.

davidwalschots:
> Would be good looking into a central class in A32NX_Core which every display calls/inherits from, that way it can all be done with javascript and memory access.

The way I understood things work thus far is that each A320 screen is a separate browser window, which is why sharing variables hasn't been done yet. However, while working on ts definitions I've seen some code related to window parents. If we're lucky we might be able to access variables of different "windows" through `window.opener`. I'll take a look at that as well and also consider @lukecologne's `sessionStorage` suggestion.

davidwalschots:
@wpine215 @lukecologne Each screen definitely doesn't have access to other screens. There is no parent iframe we can access. `sessionStorage` only affects the screen itself and is not shared between screens. `localStorage` therefore seems the only feasible option.

I decided to first do some measurements to determine what the impact of SimVar calls is on my FPS. My system is an AMD Ryzen 5 2600, MSI GeForce GTX 1660 Ti, 32 GB memory at 3000 Mhz, Intel 660p NVMe, running 1920x1080. I am generally limited by my CPU and also was during this test.

At Sedona rwy 22, mid day with clear skies:
- With SimVar 29-31 fps.
- With in-memory SimVar 35-37 fps.
- With local storage SimVar 34-36 fps.

To replace `SimVar` usage, I have overwritten the `GetSimVarValue` and `SetSimVarValue` in every screen 10 seconds after initialising the screen (to allow for any startup sequence which should affect the actual sim's SimVars to occur). Refer to the code in this draft at the moment of writing this comment for the implementation and feel free to run it locally.

Currently it isn't yet clear what percentage of `SimVar` calls we can actually replace with a `localStorage` implementation. If this is very low then it might not be worthwhile to do.

wpine215:
@davidwalschots Any updates on the viability of including this in the code, and where would be the best place to put it?

davidwalschots:
@wpine215 I will close this for now. I haven't been able to work on it and currently for me personally the Rust systems side is more interesting. If anyone else wants to pick this up they should feel free to do so. Given the amount of changes that are still going to happen I consider this optimisation premature, especially considering that any form of caching might lead to hard to find bugs in timing of calculations if not done right.

Some of the changes that might be coming that will affect this work or make it obsolete are:
- A lot of the logic shall be moved to Wasm, this should reduce JS `SimVar` reads quite a bit.
- Asobo might improve `SimVar` read speed, or provide asynchronous ways of reading.
- As far as I understood from @devsnek there are talks with Asobo on having some "bus" for communicating between screens should we still need that.

Should someone still want to implement this, then consider:
- `localStorage` probably persists over multiple simulation sessions, thus on starting a new session make sure values of a previous session don't interfere with the new one.
- Writes should always be immediate.
- Overwriting the `SimVar.GetSimVarValue` function as I did in the example isn't the best approach as it will affect any code using that function, including code that we're not maintaining at this moment. Thus, creating and using our own function which wraps `SimVar.GetSimVarValue` is best.
- The performance gains will be less than in this example. In the example 100% of the `SimVar` reads are cached. Meaning the aircraft and simulation don't work anymore. It's unclear what percentage of `SimVar` reads is ours, and what percentage is in Asobo code we don't maintain. Thus, in a situation where every read goes to cache 50% of the time, the performance gain is at most 50% of the gain described above, but likely less.
- It's possible to define a cache duration per `SimVar` on a time or frames scale, e.g. `OurSimVar.GetSimVarValue("SomeName", "bool", 20)` could read the given `SimVar` every 20 frames. However, such an implementation would make timing issues more likely so I'd not recommend it. Setting a scale per screen might be an option, as then at least all the code within that screen runs at the same frequency. 
- For the point above, do note that the screens contain logic not just pertaining to itself, but also to e.g. providing aural alerts.

PR #1830, opened 2020-11-09
[CDU] Add MCDU formatting helpers

beheh:
## Summary of Changes

In some cases, the MCDU requires multiple colors per cell. A good example is the INIT B page, where the user may edit the fuel, but not the number of minutes (which are computed). Hence the cell is split in blue and green, which is not supported out of the box by the MCDU template. There are other in-cell formatting changes

To fix this in the past pages, various pages use a little marker in their template (e.g. ` {smallFront}` and `{smallEnd}`) together with script that replaces these markers in the innerHTML with `<span>`s using the corresponding CSS classes. This is a bit ugly, especially as multiple pages need this.

As part of #1342, @tyler58546 built a little helper for the CDU code that generalized the markers and applies them everywhere by default. Unfortunately #1342 is pretty sizeable and it's not clear when this PR will land. Therefore, after a [conversation with @Lucky38i in Discord](https://discord.com/channels/738864299392630914/747622836381810739/775306984680587274) I'm pulling this into a standalone, separate PR that does nothing but introduce the formatting helper where it's possible.

This PR adjusts the pages that are already in the repository to use this new methods. I've tried searching the code base for all I could find. Please let me know if you find any others. Once this PR lands I'm happy to help out with any open PRs that want to rebase onto this (although it should be very straightforward).

All in all there are no expected user facing changes, so I do not think a CHANGELOG entry is necessary. This for developers and should simplify code in a few places and make it easier and faster to implement cells with special formatting in future.

### Notes for QA (once tagged with 'Ready to Test')

QA should verify that all MCDU pages are formatted as usual (and as expected), especially the ones touched as part of this PR:
- AIRWAYS (FPLN => LAT REV => AIRWAYS)
- IRS INIT (INIT => IRS INIT)
- INIT

## Screenshots (if necessary)

While nearly all changes are invisible, I cleaned up the IRS INIT page that was using a font that was manually defined and did not match the normal font sizes, especially after #1243. Replacing this caused the font sizes to be corrected again, and this is the only change that should be visible to players (left old, right new):

IRS INIT

## Additional context

PRs that roll their own templating or are forgoing dual formatting and would benefit from this:
- #1524
- #1597
- #1823
- #1845

PR #1342 (where this is from) would also have to be adjusted. These PRs could be rebased once this PR lands to use the new templates.

Pages we know that use split colors and would benefit from this:
- INIT B (various fuel entries)
- FUEL PRED (same)
- AIRWAYS TO (title)
- PERF page (O=/S=/F=)
[...]

PR #1845, opened 2020-11-10
[CDU] Improve LAT REV page layout

beheh:
## Summary of Changes

This PR greatly improves the look of lateral revision page (LAT REV) in the MCDU, which right now is quite a bit away from the reference documentation.

Please note that this is mostly fixes visual issues on the LAT REV page only, not any behaviours and not any of the subpages, e.g. HOLD or AIRWAYS are out of scope for this PR. Items that were non-functional are also marked in gray now, but the behaviours were never there in the first place (nothing has been removed as part of this PR).

Here's a list of all the changes:
* [x] Fix page title according to reference (colors, text size, show runway if set)
* [x] Fix coordinate format (N12.123.1 E12.123.12 => 12.123.1N/012.123.123E)
* [x] Fix position of ARRIVAL (1L => 1R)
* [x] Fix typo (WING => XING)
* [x] Show or hide items based on type of waypoint and in some cases flight phase
  * Three primary versions: departure airport, enroute waypoint, destination airport
  * ENABLE ALTN is only visible if an alternate is set on the init page, and only on enroute waypoint and destination
* [x] Change text color for items missing implementation to inop gray
* [x] Disable NEXT WPT on destination airport for now as that crashes the sim

## Screenshots (if necessary)

before/after

(note that the wrong N/E vs N/W coordinates were fixed after these screenshots were taken as part of #1867).

## References

grafik

## Additional context

The visual layout of the page should be pretty obvious and changes depending on the type of waypoint you select (departure airport/enroute waypoint/destination waypoint). However:
- `NEXT WPT`: This will be hidden on the origin airport page as per the documentation after PREFLIGHT has ended, so you may not see it if you're in the takeoff phase.
- `NEXT WPT`: Is disabled on the destination airport for now as the sim can't handle that yet and crashes
- `HOLD`: While a hold page exists, it doesn't work really yet. I'm keeping it available for now.

The subpages (HOLD/ARRIVAL/DEPARTURE/AIRWAYS) don't cleanly return to this page yet on the return button, but I'll keep that out of scope for now as it's a bit tricky.

Discord username (if different from GitHub): BehEh#1234

### DIFF
--- a/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_LateralRevisionPage.js
+++ b/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_LateralRevisionPage.js
@@ -4,57 +4,137 @@ class CDULateralRevisionPage {
         console.log(waypoint);
         mcdu.clearDisplay();
         mcdu.page.Current = mcdu.page.LateralRevisionPage;
+
+        let coordinates = "";
+        if (waypoint && waypoint.infos && waypoint.infos.coordinates) {
+            const lat = CDUInitPage.ConvertDDToDMS(waypoint.infos.coordinates['lat'], false);
+            const long = CDUInitPage.ConvertDDToDMS(waypoint.infos.coordinates['long'], true);
+            coordinates = `${lat.deg}°${lat.min}.${Math.ceil(Number(lat.sec / 100))}${lat.dir}/${long.deg}°${long.min}.${Math.ceil(Number(long.sec / 100))}${long.dir}[color]green`;
+        }
+        const isDeparture = waypoint === mcdu.flightPlanManager.getOrigin();
+        const isDestination = waypoint === mcdu.flightPlanManager.getDestination();
+        const isWaypoint = !isDeparture && !isDestination;
+
         let waypointIdent = "---";
         if (waypoint) {
             waypointIdent = waypoint.ident;
+            if (isDeparture) {
+                const departureRunway = mcdu.flightPlanManager.getDepartureRunway();
+                if (departureRunway) {
+                    waypointIdent += Avionics.Utils.formatRunway(departureRunway.designation);
+                }
+            } else if (isDestination) {
+                const approachRunway = mcdu.flightPlanManager.getApproachRunway();
+                if (approachRunway) {
+                    waypointIdent += Avionics.Utils.formatRunway(approachRunway.designation);
+                }
+            }
         }
-        let coordinates = "---";
-        console.log(waypoint);
-        if (waypoint && waypoint.infos && waypoint.infos.coordinates) {
-            coordinates = waypoint.infos.coordinates.toDegreeString();
-        }
-        let departureArrival = "";
-        if (waypoint === mcdu.flightPlanManager.getOrigin()) {
-            departureArrival = "<DEPARTURE";
+
+        let departureCell = "";
+        if (isDeparture) {
+            departureCell = "<DEPARTURE";
             mcdu.onLeftInput[0] = () => {
                 CDUAvailableDeparturesPage.ShowPage(mcdu, waypoint);
             };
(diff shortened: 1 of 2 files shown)


A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_LateralRevisionPage.js
```diff
@@ -11,35 +11,95 @@ class CDULateralRevisionPage {
         let coordinates = "---";
         console.log(waypoint);
         if (waypoint && waypoint.infos && waypoint.infos.coordinates) {
-            coordinates = waypoint.infos.coordinates.toDegreeString();
+            const lat = CDUInitPage.ConvertDDToDMS(waypoint.infos.coordinates['lat'], false);
```
beheh, line 14: I'm not super happy about pulling this from the Init page - both the Init and the InitIRS page use this so far. Any suggestion for a more centralized place where this could live in future?

Benjozork:
Hello, please rework to fit #1830

beheh:
Thanks, I've incorporated the new formatting helpers. Should be ready to go out of draft soon.

beheh:
Note to self: Double-check pseudo waypoints (DECEL)

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_LateralRevisionPage.js
lousybyte: `A320_Neo_CDU_MainDisplay.js`? Should be able to call it anywhere after with `mcdu.ConvertDDToDMS()`.

danbw9:
Name: danbw9#7028
Date of testing: 12/11/2020
Version of the sim: 1.10.11.0
PR Tested: #1845
Object of test: Test new visual changes to the LAT REV page

Testing technique: Identify the issues with the current version of LAT REV page without PR with screenshots and then compare to with PR installed - and check items in PR are all correct and have no negative effect on F-PLAN page/MCDU functionality. 

Loaded up C&D in ESGG. Normal flight from ESGG to EKCH. 

In preflight stage MCDU setup:
- checked for the various types of LAT REV pages (DEP, ARR, different types of ENR waypoints) and the distinct differences that make sense - specifically the hidden and shown features when appropriate.
- tried ENABLE ALT without ALT set in INIT A and not possible. With ALT set, it's there. 
- Typo fixed now correctly showing XING

Can confirm that compared to the reference 'AFTER screenshots' - that with the coordinates PR, the text and coordinates format for waypoints make sense. 

After takeoff in CLB and CRZ was able to make lateral revisions (the ones possible) with no issues 
- direct to en-route
- change of STAR and Approach


Issues (If Any): None
Overall Rating: Good
Conclusions of the testing: LGTM. Looks great even if it's just visual changes. I wondering if all the other core devs are happy with the grey text being there when it doesn't look realistic but indicates no functionality...someone else could share their opinions on that.  Thanks for this!

wpine215:
LGTM, but @beheh you need to update your fork with the current master - I can't seem to do it from the Github interface because of the force pushes that were made.

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/A320_Neo_CDU_LateralRevisionPage.js
beheh: Hm, it can be static though, and that might be more convenient for other places. Some kind of static helper class might be more useful. I'm just leaving this where it is now and we can pull it later if we need it in even more places.

beheh:
I don't think it's the force pushes, it's probably this setting:

grafik

I've rebased onto master and also changed that now, so you should be able to edit. Sorry for the inconvenience.

PR #1863, opened 2020-11-11
[FIX] "too low terrain" warning on takeoff

lukecologne:
## Summary of Changes

* This change fixes the wrong "Too low, Terrain" warning on takeoff
* It was caused by the fact that the radio alt is set to a non zero value while loading (in my testing around 300ft), which caused problems.

## Additional context


Tonyehrd:
I still have this issue (yes i did download the new DEV this morning). Happend to me @ EHAM.

PR #1715, opened 2020-11-02
fix FMA modes for ILS Approach when disconnecting AP. Also add new fl…

antikythera288:
Fixes #1222 
Fixes #1705
Fixes #146

## Summary of Changes

- [x] Many fixes on the several possible FMA modes.
- [x] Implementation of Goaround (including MCDU Pages and AP Modes).

Limitations:

- Re-engaging the AP with earlier armed APPR will only succeed when you are on, - or above - the glide slope (check magenta diamonds). If you are below, OP CLB will be initiated until you "De-" and Re-Activate the APPR button. This is probably a problem with how the Asobo code works internally. The GS HOLD event is driven somewhere internally and its really a mess.

- During enganged APPR Mode, you should be able to pre-select a HDG for goaround. Right now you can pre-select it (validate on PFD) but its still "dashed" (---) on the FCU

- As we dont have yet missed approach flight plans or something similar, during GOAROUND you will see HDG armed with GA TRK. The GA TRK is implemented with the WING LEVEL Event and can be exited when pulling the HDG knob or exits automatically when plane reaches thrust reduction altitude, thus enganging HDG to the synchronized current heading.

- When flying the GOAROUND by hand, the FMA's will show correct modes for it, except the FD's ... This seems to be always wrong when you disable the AP (vanilla bug) ... 

## Screenshots (if necessary)

will upload video soon from APPR Mode to GOAROUND and LAND

## References

- Help from IRL Pilot (@St54Kevin)

- http://www.a320dp.com/

- https://www.theairlinepilots.com/forumarchive/a320/a320-fma.pdf

- https://youtu.be/FsROrvv9tJs?t=6847

## Additional context

Discord username (if different from GitHub): mrmercury#8898

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build script will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, click on the **Artifacts** drop down and click the **A32NX** link

PR #1922, opened 2020-11-14
[FIX] Wrong GPWS warnings

lukecologne:
## Summary of Changes

* After the changes from #1715 by @antikythera288, go around flight phase is now implemented, so now, the "Too low terrain" warning on go-around should be no longer playing.
* "Don't sink" warnings on takeoff on some airports should also be fixed.

### DIFF
--- a/A32NX/html_ui/Pages/A32NX_Core/A32NX_GPWS.js
+++ b/A32NX/html_ui/Pages/A32NX_Core/A32NX_GPWS.js
@@ -71,12 +71,12 @@ class A32NX_GPWS {
             const Airspeed = SimVar.GetSimVarValue("AIRSPEED INDICATED", "Knots");
             const gearExtended = SimVar.GetSimVarValue("GEAR TOTAL PCT EXTENDED", "Percent") > 0.9;
 
-            this.update_maxRA(radioAlt, onGround);
+            this.update_maxRA(radioAlt, onGround, phase);
 
             this.GPWSMode1(radioAlt, vSpeed);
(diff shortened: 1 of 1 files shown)


antikythera288:
will download the artifact and give you feedback after testing! I just down know if i will be able to run the sim before Monday

danbw9:
Name: danbw9#7028
Date of testing: 14/11/2020
Version of the sim: 1.10.11.0
PR Tested: #1922
Object of test: Testing that the GPWS callouts which were existent in dev few days ago are now gone + GA phase

Testing technique: Noted previous issues with GPWS call outs issue and Go Around phase not inhibiting GPWS call outs. Using same airport (UAAA) where I experienced the issue

 - Spawned at UAAA for circuit, took off with normal rotation rate of 3 degrees a second (which would previously give GPWS error callouts) but only until 12 degrees pitch up - GPWS call out non existent this time round. 
- Flew the circuit and on final - down to an point lower than ILS minima to simulate a low approach and go around at around 100 AGL. 
- Executed manual normal go around and all was fine. 

I am currently doing another flight with the PR to test another go around but with AP on. If I have any more issues I will report back. If PR not merged tomorrow, I will try other scenarios (airports and other types of go arounds

Issues (If Any): None. 
Overall Rating: Good
Conclusions of the testing: All good, nice fix. And cool to see parts of the go around phase being made a bit more realistic. I had noticed before this PR was installed, that the 'don't sink' callout on take off had been fixed but I guess that was a temp fix for now. This one solved more than that one and the go around. Thanks for your work!

PR #2107, opened 2020-11-26
Attempts to fix AP issue

MapleNAV-TCASFAULT:
Fixes #[issue_no]

## Summary of Changes

This fix should make the bank more even between left and right, and the aircraft on AP is now more responsive during HDG select and NAV. (note: I'm not the curator of this! Imenes just told me to make this PR)
## References

Temporary fix, no reference was provided, credit goes to Idle Betazoid#944 for raising the issue!

## Additional context

Discord username (if different from GitHub): ChristopherXie#3904, Idle Betazoid#944

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/apron.FLT
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/apron.FLT
@@ -273,7 +273,7 @@ ForceDisplayUI=False
 AltitudeSlotIndex=2
 SpeedSlotIndex=2
 VSSlotIndex=2
-
+MaxBankIndex=0
 [Controls.0]
 SpoilersHandle=000.00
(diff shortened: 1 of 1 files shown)


ImenesFBW:
This fix worked with my problematic route. The plane was still a bit off ND track, but it was within the margins.

donstim:
Did not prevent airplane from descending and overspeeding on KLGA Wy 22 Jute 3 departure like the workaround of saving and loading flight plan did. although it seemed better than previously.

armankir:
Cool PR, some comments from initial testing.

My number one issue is what was described above, the aircraft is flying slightly off track, but not to a substantial margin. 

Also experienced a CTD twice. Can't say if it's related, I have been getting one or two random CTDs since the updates, but I thought it would be a good measure to include this detail.

Good work, could use some refinement I would say.

PR #2137, opened 2020-11-30
Latest changes from wasm repository

aguther:
## Summary of Changes
- includes fix for throttle issues by handling them directly
- allows configuration of throttle to map to detents correctly
- adds support of proper sidestick position display using l-vars
- description of throttle configuration added

⚠ WASM build is not fully working yet, the reading and writing of the config file will fail in the meantime.

## Additional context

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,36 @@
 # Documentation Placeholder Readme
 
 Fill me out!
+
+## Custom Fly-By-Wire
+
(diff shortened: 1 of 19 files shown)


wpine215 (approved):

aguther:
You can merge it, the custom fbw works like before, just the throttle part is broken, but that is also with the current version on the branch. When @devsnek has the fix ready, it will work.

PR #2229, opened 2020-12-04
Add default baro unit option

### LINKED ISSUE #2228: [REQUEST] Ability to change default baro unit to HPa
If we fly in HPa regions (e.g. Europe) we currently have to manually change the baro unit selector to HPa every time we load the aircraft. 

I believe we need an option in the new MCDU options page that allows users to set the default baro unit to either In Hg (default as in current Asobo build) or to HPa.

**References**
I believe that real A320 pilots that operate in HPa regions (e.g. Europe) wouldn't change the baro unit selector to In Hg when leaving the a/c, meaning that the next pilots would almost always find this in HPa when they enter the cockpit. 

Discord username: Theodore Messinezis

theomessin:
Fixes #2228

## Summary of Changes
Added a "INIT BARO" option in `A320_Neo_CDU_OPTIONS_Menu` that saves the baro option in the `CONFIG_INIT_BARO_UNIT` NXDatastore setting. Default value for `CONFIG_INIT_BARO_UNIT` is `IN HG` which is the current functionality. This can change to `HPA` ot `AUTO`. Takes effect on next A/C load.

Custom `A32NX_BaroSelector` module added which handles setting up initial baro unit. Main complexity is for the `AUTO` setting which fetches the nearest aerodrome. See more in comments below.

## Screenshots
[](https://postimg.cc/3yqB9RrM)

Discord username: Theodore Messinezis


ErikNL80:
Instead of making it an option, could this option not look at departure airport for example? Based on this location the MCDU can determine which units are used? How does this work IRL?

theomessin:
@ErikNL80 we can do it that way too I suppose, most addons I've seen simply have an option, so that's what I did. We can even combine the two and have an AUTO option that checks the departure airport, that way we're not missing out on anything.

ErikNL80:
> @ErikNL80 we can do it that way too I suppose, most addons I've seen simply have an option, so that's what I did. We can even combine the two and have an AUTO option that checks the departure airport, that way we're not missing out on anything.

That is a great solution! Did not think of that!

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_Menu.js
```diff
@@ -22,8 +23,8 @@ class CDU_OPTIONS_MainMenu {
             ["<METAR SRC", `${storedAccelAlt} FT.>[color]cyan`],
             ["AOC[color]green", "DMC SELF-TEST[color]green"],
             ["<SIGMET SRC[color]inop", `${storedDMCTestTime} SEC.>[color]cyan`],
-            ["AOC[color]green"],
-            ["<TAF SRC"],
+            ["AOC[color]green", "DEFAULT BARO[color]green"],
+            ["<TAF SRC", `${storedDefaultBaroUnit}[color]cyan`],
```
beheh, line 27: A suggestion for consistency with the other settings:
```suggestion
            ["<TAF SRC", `${storedDefaultBaroUnit}>[color]cyan`],
```

beheh:
Once we have an AUTO option I'd suggest that become the new default, as I believe it's strictly better than the Asobo default of IN HG.

theomessin:
@beheh Yeah, that's what I'm doing at the moment. I'm open to suggestions though 😄

beheh:
I think that sounds good and is appropriate here!

theomessin:
I've added an AUTO option which should work for the codes defined here: 
https://github.com/flybywiresim/a32nx/pull/2229/commits/a90e49f9de9fa27ea4e48b2e035a5f3e47cb32c7#diff-acc1c7c3d87ec7183e2af60cdfd2fe037e2ac34b03532a1335d499782755080eR3

Not a huge fan of this implementation, but it should work...

A32NX/html_ui/Pages/A32NX_Core/A32NX_BaroSelector.js
```diff
@@ -0,0 +1,50 @@
+// Used to determine if an ICAO code uses InHG or HPa.
+function AirportUsesHPA(icao) {
+    const InHgRegions = ["K", "C", "M", "P", "RJ", "RO", "TI", "TJ"];
+    for (let i = 0; i < InHgRegions.length; i++)
+        if (icao.substr(0, InHgRegions[i].length) == InHgRegions[i]) return false;
+    return true;
+}
```
Benjozork, line 7: Should start with a lowercase character and be inside the class (move inside class, call using `this` and remove `function` keyword).

A32NX/html_ui/Pages/A32NX_Core/A32NX_BaroSelector.js
```diff
@@ -0,0 +1,50 @@
+// Used to determine if an ICAO code uses InHG or HPa.
+function AirportUsesHPA(icao) {
+    const InHgRegions = ["K", "C", "M", "P", "RJ", "RO", "TI", "TJ"];
+    for (let i = 0; i < InHgRegions.length; i++)
+        if (icao.substr(0, InHgRegions[i].length) == InHgRegions[i]) return false;
+    return true;
```
beheh, line 6: Can be simplified to this: 

```suggestion
    return !InHgRegions.some(r => icao.startsWith(r));
```

A32NX/html_ui/Pages/A32NX_Core/A32NX_BaroSelector.js
```diff
@@ -0,0 +1,50 @@
...
+            this.SimVarCallbackPending = false;
+        });
+
+        this.SimVarCallbackPending = true;
+    }
+    SetBaroSelector(value) {
+        SimVar.SetSimVarValue("L:XMLVAR_Baro_Selector_HPA_1", "bool", value);
+        this.BaroSelectorHasBeenSet = true;
+    }
+}
```
beheh, line 50: Missing trailing new line (this can often be set in your IDE).

beheh (changes requested):

A32NX/html_ui/Pages/A32NX_Core/A32NX_BaroSelector.js
theomessin: I'd like to be able to use this function in the MCDU options page as well (A320_Neo_CDU_OPTIONS_Menu.js). It'd be nice to be able to AUTO set when the option is toggled as well. Any way of doing that without duplicating code? Had it as a function of its own to remind me of that 😉

theomessin:
`AUTO` option is broken. When ending a flight and starting a new one elsewhere, even though the `lat,long` pair is correct, the `fs9gps:NearestAirport` is stuck to the old value. I've tried adding a delay before fetching it, didn't work. Also tried fetching it twice (one after the other), also didn't work. Any ideas?

theomessin:
I've found that for a new flight (first after starting the sim), the first couple of `GetSimVarArrayValues` calls never call their callback. I've found that we can assume that this has timed-out after 100ms (at least on my machine). The latest commit will keep trying until it gets the callback called. 

I've also found that if we end the flight and spawn elsewhere (e.g. from LGAV, to KLGA) the first N callbacks will be the previous airport (in this case `LGAV`). The latest commit will get the airport code at least N+1 times to make sure it is correct (N found to be 2 experimentally - again, at least on my machine). Time to get this QA'd.

theomessin:
## Expected functionality 1
For `INIT BARO` option setting set to `IN HG`, aircraft should load in `In Hg`.


## Expected functionality 2
For `INIT BARO` option setting set to `HPA`, aircraft should load in `HPa`.

## Expected functionality 3
```js
const inchOfMercuryRegions = ["K", "C", "M", "P", "RJ", "RO", "TI", "TJ"];
```
For `INIT BARO` option set to `AUTO`: after loading the aircraft, if the nearest airport is in the ICAO code regions listed above, it should load in `In Hg`. Otherwise, the aircraft should load in `HPa`.

**Please test the `AUTO` functionality rigorously! Please test spawning in the air, ending and starting new flights, starting clean flights (restarting FS), etc. Also, please check the `In Hg` regions above, I might've left some out.**

danbw9:
Category: QA Test
Name: danbw9#7028
Date of testing: 06.12.2020
Version of the sim: 1.10.6.0
PR Tested: #2229
Object of test: Check out the new INIT BARO option functionality through MCDU

Testing technique: Try out multiple flights and vary the startpoint

Spawn C&D at EKCH. MCDU MENU > OPTIONS > INIT BARO > set to AUTO (then pressed the blue 'RETURN' - to make it work)
Checked the following scenarios to spawn in
- ESGG (spawn on approach then reset + spawn C&D)
- RJTT (spawn C&D + restart flight)
- EKCH (spawn on approach)
- KIAD (spawn C&D + on RWY)

In all the above spawn points, the AUTO function worked and correctly indicated the pressure in the units used in that region. 

Checked that it does not interfere with weather being dynamically injected (pressure changes).

I just noticed a small bug in the start of testing this PR (see conclusion) which if someone else could have a look at, that would be great.

Flight notes (If required):

Issues (If Any): None
Overall Rating: Good
Conclusions of the testing: LGTM. A pretty convenient feature to have especially if you don't know altimetry units around the world. As I mentioned above, I noticed a small bug where I could not get this feature to work unless I selected 'AUTO' and then clicked 'RETURN' - only then would this feature work for me, whereas the other options on there work instantly without having to return the page. Thanks for your work!

theomessin:
Hi @danbw9, thanks for testing this! I've got a question regarding your "RETURN" bug. Pressing `RETURN` on the CDU (i.e. pressing L6) shouldn't do anything, no matter what the setting for the `INIT BARO` is. This setting only takes effect when the aircraft is booting up (i.e. during a new flight or when restarting a flight, etc.)

Restarting the aircraft can be emulated by restarting the CDU using the devtools by pressing the gray R button in the pop-up of the CDU (as the core is currently tied to the CDU). Here's a video showing what I mean:

https://streamable.com/mnubw9

Changing the `INIT BARO` option and hitting `RETURN` (i.e. L6) doesn't do anything - which is the expected functionality. Resetting the CDU using the gray R in the debug window will force the core to restart and the new option to take effect. Can you confirm this is what you're seeing on your end too?

Awemeter:
Category: QA Tester
Name: Awemeter#7919
Date of testing: 12/6/2020
Version of the sim: 1.11.6.0
PR Tested: #2229 
Object of test: MCDU/QNH

Testing technique: Tried out
1. Spawned into multiple airports in many countries with auto mode on. Also spawned in air.
2. Spawned into multiple airports with hpa on in multiple countries. Also spawned in air.
3. Spawned into multiple airports in different countries with hg mode on. Also spawned in air.
4.
5. 

Flight notes (If required):

Issues (If Any): 
Overall Rating: Good
Conclusions of the testing: Works properly after all my test always spawned in with the right QNH unit.

theomessin:
@Benjozork shall I get rid of the `console.logs` before you merge?

A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_Menu.js
```diff
@@ -93,5 +94,19 @@ class CDU_OPTIONS_MainMenu {
...
+        mcdu.rightInputDelay[3] = () => {
+            return mcdu.getDelaySwitchPage();
+        };
+        mcdu.onRightInput[3] = (value) => {
+            if (value != "") {
+                mcdu.showErrorMessage("NOT ALLOWED");
+            } else {
+                // We'll go from AUTO -> HPA -> IN HG -> AUTO.
+                const newInitBaroUnit = storedInitBaroUnit == "AUTO" ? "HPA" :
+                    storedInitBaroUnit == "HPA" ? "IN HG" : "AUTO";
```
MisterChocker, line 106:
```suggestion
                const newInitBaroUnit = storedInitBaroUnit === "AUTO" ? "HPA" :
                    storedInitBaroUnit === "HPA" ? "IN HG" : "AUTO";
```

MisterChocker (changes requested):

Benjozork:
> 
> 
> @Benjozork shall I get rid of the `console.log`s before you merge?

Yes

wpine215 (approved):

PR #2242, opened 2020-12-05
TCAS improvements

lukecologne:
## Summary of Changes

This is a modification of Asobos default "TCAS"
There are many features still missing, but the conditions for which traffic is displayed should be greatly improved.
* Features missing: 
   * RA generation and display
   * TCAS messages on the ND
   * proper on ground detection for traffic aircraft
   * display of aircraft that are not live traffic
   * sounds
* Features implemented:
    * Traffic display based on display knob: THRT, ALL, ABV, BLW
    * Display of vertical speed and relative alt
    * Realistic symbols for traffic threat status: white hollow for no threat, white filled for proximate, amber for TA, red for RA (although no RA will be generated)
* This is only a temporary solution, until we find a way to move TCAS to WASM, where we could also implement the missing features (and more).

This shouldn't bring a performance hit, since it replaces asobos TCAS which should have a similar performance, but it still runs on both NDs seperately, which is an issue that can not really be solved until we can pass larger amounts of data between screens, at which point we will be able to hopefully do the same with WASM.

## References

https://www.faa.gov/documentlibrary/media/advisory_circular/tcas%20ii%20v7.1%20intro%20booklet.pdf

grafik
grafik
grafik

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -134,6 +134,7 @@
 1. [PFD] Added vR indication - @MisterChocker (Leon)
 1. [MFD] Corrected icon for the Decel point on flightplan - @LoungeFlyZ (Chris Johnson)
 1. [ND] Update ND to only show Mode Changed when mode button is used - @ilyeshammadi (Ilyes Hammadi)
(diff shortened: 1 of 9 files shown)


wpine215:
Works as described, LGTM

PR #2246, opened 2020-12-05
Fix invisible icing

### LINKED ISSUE #2177: [BUG] Nose is very up on approach using the calculated Vapp
**Mod Version**

latest master branch

**Describe the bug**

On approach, using the calculated Vapp speed, the plane's nose will go up close to 10degrees and it will trigger the stall warning. I think this is a result of [this commit ](https://github.com/flybywiresim/a32nx/commit/02da15192f41797214205a35b8ce772003f228ac)to the flight model. 

**To Reproduce**

1. Fly with the same route and the same fuel, payload as this [Simbrief report](https://drive.google.com/file/d/1yMlLsYYJp3Wf22p8F98RtkW5n9Yp4sQG/view?usp=sharing).

**Expected behavior**

The calculated approach speed shouldn't trigger the stall warning nor the plane's nose goes up close to 10 degrees. 

**Actual behavior**

The calculated approach speed is too low that causes the plane nose goes up close to 10 degree's and could trigger the stall warning.

**References**

**Additional context**

pepperoni505:
Fixes #2177 

## Summary of Changes

This restores the LOD00 model from the 1.10.11, as for some reason with 1.11.6 the icing mesh got removed from only the LOD00. Ice would still form, but the user wouldn't be able to see it unless they zoomed out to where LOD01 loaded. 

This is a temporary fix and should probably be reverted after Asobo fix this issue.

## Additional context


hansmeller:
Category: QA Tester
Name: hansmeller#8988
Date of testing: 06.12.2020
Version of the sim: 1.11.6.0
PR Tested: #2246 
Object of test: Checking that the icing works

Testing technique: Little Flight       
1. Spawned at a random airport
2. Set weather to overcast
3. Climb to FL150
4. Turned Delevoper mode on to see how the icing is working
5. Looked with the outside camera that the icing is visible

Flight notes (If required):

Issues (If Any): 
Overall Rating: good 
Conclusions of the testing: Works as described, no issues found

PR #2470, opened 2020-12-16
Add Soft Go-Around Function

MisterChocker:
## Summary of Changes

- Added MAN GA SOFT and GA SOFT FMAs
- Added "GA SOFT" Thrust Rating Mode to ECAM
- Added engine out logic
- Improved Upper ECAM Thrust Rating field (it's centered now as it should've been before)
- Added GA SOFT logic to SRS
- Added GA SOFT logic to NAV Mode

## References

https://safetyfirst.airbus.com/introduction-to-the-soft-go-around-function/

## Additional context


pareil6:
Is switching from APPR directly to GA on MCT actually correct? From that article, I think you're supposed to select TOGA first, then pull back to MCT for SGA:

> When the go-around is initiated, the flight crew sets the thrust levers to TOGA position, as usual, to trigger all the logics (approach modes disengagement, FMS FPL sequencing…), and then activates the SGA by moving back without delay the thrust lever to the FLX/MCT detent.

wpine215 (approved):

donstim:
I don't see any reference to this commanding a 2000 fpm rate-of-climb if under autopilot control or by the flight director command bars if not. Is that in here?  That is a significant piece of the soft go-around function.

Issue #2558, opened 2020-12-20
[BUG] Fuel decreasing on stand with engines and APU off


claguazza:
**Mod Version**

Latest Stable

**Describe the bug**

While on stand with less than 3000kgs of FOB, the fuel starts decreasing quite quickly and message "OUTR TK FUEL XFRD" appears steady

**To Reproduce**

1. Set up a flight from stand, cold&dark at any airport
2. Start batteries and external power
3. Load via MCDU a quantity of fuel less than 3000kgs

**Expected behavior**

Fuel should not decrease having both engines and APU off

**Actual behavior**

Message "OUTR TK FUEL XFRD" appears steady and fuel on board (FOB) starts decreasing quickly

**References**

**Additional context**

The same message "OUTR TK FUEL XFRD" appears flashing inflight when FOB drops below 3000kgs but in this case the fuel reduces normally as per engines running
Was this working before/when did the issue start occurring?
It happened from the first time I installed the latest stable version 0.5.0
Is this a problem in the vanilla unmodded game?
Not tested

Discord username (if different from GitHub):

Watsi01:
Just as an info: "OUTR TK FUEL XFRD" tells you, that the fuel in your inner wing tanks dropped below 750kg (each). Then fuel from the outer wing tanks will automatically be pumped to the inner tanks.

claguazza:
Thanks for your info. I was familiar with that message since I have flown the 320 in real life. Good to share that with everybody. What is weird is the behaviour on ground before starting engines and apu. Hopefully it can be fixed somehow

MisterChocker:
duplicate of #2083

Issue #1111, opened 2020-10-05
[BUG] Plane nose is extremely pointed down


lukasrz:
**Mod Version**

0.3.1 from 03 october 2020

**Describe the bug**

Did you noticed, that during normal descend, or little faster ~2000ft/min, plane nose i extremely pointed down? 

**To Reproduce**

1.
2.
3.

**Expected behavior**

**Actual behavior**

**References**

It looks like this, or even more.
![a32](https://user-images.githubusercontent.com/70447248/95088262-025f9380-0723-11eb-9396-63ca2fcab518.jpg)

**Additional context**

Was this working before/when did the issue start occurring?

Is this a problem in the vanilla unmodded game?

Discord username (if different from GitHub):

Ares1378:
I can confirm this. I will explain this more in detail. It seems the flaps are generating too much lift? The aircraft on approach is -15 NOSE down pitch MAX down pitch in normal law to obtain -1000fpm descent. This is incorrect. The aircrafts normal pitch attitude on approach is 2 to 2.5 nose up.

theersink:
Others reported similar. Look at the end of #1053 for possible fix. Flightmodel essentially needs to be fixed by Asobo, other than the ground handling after 1.9.3 the mod uses the default.

Benjozork:
Please respect the issue template. Will close until that is done.

PR #2583, opened 2020-12-21
Fix outer tank fuel transfer

donstim:
- Fixes outer fuel tank valve opening and closing during flight. Once it is opened it should remain open.
- Also results in "OUTR TK XFRD" ECAM message remaining on once it is triggered (as it should be)
- Adjusts the triggering level of inner tank fuel to cause the outer to inner transfer valves to open to match the FCOM

## Summary of Changes
1. Changes the triggering inner tank fuel level for opening the outer tank transfer valve
2. Keeps the outer tank transfer valve open once it has been opened

## References
Screenshot (457)

## Additional context
According to the FCOM, the triggering value for the outer tank transfer valve to open is when the inner tank fuel level reaches 750 kg/16650 lb.  In MSFS, the trigger is determined by the number of gallons in the tank. Since MSFS uses 6.7 lb/gal, the triggering level should be 246 gal. It is currently set to 250, so this PR changes it to 246.

Currently, the transfer valves close again when the inner tank fuel triggering level is exceeded. What happens is that when the transfer is triggered, a transfer occurs that causes the inner tank level to go up to slightly above the triggering level, which closes the valve. Pretty soon thereafter, fuel burned by the engines triggers the valve to open again, and this process repeats ad infinitum. Evidence of the valve opening and closing is giving by the ECAM "OUTR TK XFRD" message that appears, disappears, re-appears, etc.

In the real airplane, as shown in the reference FCOM material, as well as comments by our A320neo pilots, once the outer tank fuel transfer valve opens, it stays open until the next refueling operation. To simulate the closing of the valves at refueling (in case follow-on flights are taken without restarting the simulator), I made the trigger for valve closure to be 475 gallons. This way if nearly all the fuel from an outer tank is transferred to an inner without the engines burning any (for some reason), (228 gal transferred at the triggering level of 246 gal gives 226+246=474) the valve will not close. It will only close if the airplane is refueled, which would be signified by putting **more** than 475 gallons into the inner tanks.

Note:  This PR does not address issue #2558. What is happening there, is that if the trigger level for the outer tank transfer valve to open is met with the engines off, the transfer occurs from the outer tanks, but it does not increase the inner tank fuel level. I have not been able to determine why that is happening, since both the output from the outer tanks and the input to the inner tanks are set up to be gravity fed in the same manner, so if one works without the engines being on, the other should to.

Note 2: We should recommend always starting a flight with at least 500 gallons (3350 lb/1520 kg) in each center tank to ensure that the transfer valve is closed. (This will also prevent issue #2558 from being a problem.) Also, if much less than the initial triggering value of 246 lb is initially put in the inner tanks, the transfer valve may not open at all, and no fuel will transfer from the outer tanks. 

Discord username (if different from GitHub):donbikes#4084


sidnov:
Category: QA Tester
Name: sidnov#8337
Date of testing: 21/12/2020
Version of the sim: 1.11.7.0
PR Tested: #2583
Object of test: Fixed outer tank fuel transfer trigger
Testing technique: Runway spawn and circuit flying

1. Spawned at VABB RWY27 for departure, with 641kg each on Left and Right Inner tanks, 692kg each on Outer Tanks, 297kg on Center Tank. No message on ECAM related to Fuel transfer (Checked)
2. Took off for a circuit flight
3. After takeoff, maintained stable flight as suggested at 5000ft
4. 3 minutes into flight, OUTR TK XFRD alert on ECAM and the same was verified with inner tank fuel increasing.
5. Landed at VABB after circuit.

Issues (If Any): Works as expected
Overall Rating: Good
Conclusions of the testing: The alert and fuel transfer trigger works as expected, screenshot after transfer started is attached. LGTM

Outer Fuel XFRD

beheh:
@donstim Great job with this PR and providing references! I have one suggestion: in addition to your manual refueling detection (to detect the user adding fuel in air) we could also close the valve if we detect the refueling truck being connected, like the `REFUELG` ECAM message:

https://github.com/flybywiresim/a32nx/blob/3cc1593326720076d018f0f9be878bbf46ff6ab3/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js#L980-L983

We would have to ensure though it doesn't repetitively cycle between open/closed during refueling though, if you refuel when below the minimum amount in the outer fuel tanks.

lukecologne (approved):

tjburni:
Category: QA Tester
Name: [Z+1]tjburni#0409
Date of testing: Current date 21.12.2020
Version of the sim: 1.11.7
PR Tested: #2583
Object of test:outer tank fuel transfer

Testing technique:Full C&D
1. Did a full C&D IFR
2. Observed upper and lower ecam for the fixes
3. landed

Flight notes (If required):I saw the ecam alert and saw

Issues (If Any): none
Overall Rating: Good
Conclusions of the testing:I saw the ecam alert and saw the fuel transter. Works good.Thanks

donstim:
> @donstim Great job with this PR and providing references! I have one suggestion: in addition to your manual refueling detection (to detect the user adding fuel in air) we could also close the valve if we detect the refueling truck ground service being connected, like the `REFUELG` ECAM message:
> 
> https://github.com/flybywiresim/a32nx/blob/3cc1593326720076d018f0f9be878bbf46ff6ab3/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js#L980-L983
> 
> We would have to ensure though it doesn't repetitively cycle between open/closed during refueling though, if you refuel when below the minimum amount in the outer fuel tanks.

That's a great idea, but I'm not sure how to get around the issue you bring up -- that opening the refueling panel would close the valve, but the valve would repetitively recycle between open/closed during refueling until more than 246 gal of fuel is added to the inner tanks. As it is, the PR closes the valve whenever the inner tank fuel value reaches 475 gal, regardless of when, where, or how it happens, so it does cover the refueling truck ground service as well.

If you (or someone else) can figure out how to prevent the repetitive cycling between open/closed if the valve is first closed when the refueling panel is opened, have at it!

beheh:
If we can somehow access the simvar in the fuel code to close the valve when the refueling truck is connected, could we not just also check that the fuel truck is *not* connected when we do the check for opening it? I'm not sure what capabilities the template language gives us.

Road-to-Hell9:
Category: QA Tester
Name: Road_to_Hell9 #1111
Date of testing: 21/12/2020
Version of the sim: Newes 21/12/2020
PR Tested: #2583
Object of test: Fix outer tank fuel transfer

Testing technique: Description of how you tested the Pull Request
1.Normal cold start
2.Entered flight plan
3.Entered ZFW, CG and Block fuel
4.Modified The Alt airport fuel
5.

Flight notes (If required):
The Transfer works

Issues (If Any): The Fuel predictions just disappear
Overall Rating: To be improved
Conclusions of the testing: The Fuel predictions have to be repaired (No sure if it's your fault)

donstim:
> If we can somehow access the simvar in the fuel code to close the valve when the refueling truck is connected, could we not just also check that the fuel truck is _not_ connected when we do the check for opening it?
> 
> I'm not sure what capabilities the template language gives us.

Not quite sure how that would work. The triggers I'm using to open and close it are actually in the flight_model.cfg file. I'm not sure how those triggers would interact with using the fueling simvar to close the valve via code when the simvar used to show the "REFUELG" message is displayed. If there is a known trigger variable for the refuel panel being opened that I can use in the flight_model.cfg file, I would think I could make it work, but the SDK doesn't cover that yet as far as I can tell.

wpine215 (dismissed):
Looks like there's a conflict with the changelog

donstim:
> Looks like there's a conflict with the changelog

Hopefully, I've fixed it now?

lukecologne:
You might need to look at if this still works after #2611 and the sim update

donstim:
> You might need to look at if this still works after #2611 and the sim update

Thanks for the heads up.  Added the #2611 changes to it in the version on my computer and it works fine.

Benjozork (approved):
LGTM

PR #2625, opened 2020-12-22
Fix Issue #2586: AUTOBREAK MAX turn off while taxing/breaking before T/O

### LINKED ISSUE #2586: AUTOBREAK MAX turn off while taxing/breaking before T/O
**Mod Version**

MASTER (FBW)

**Describe the bug**

when i prepare the A/C for Takeoff after engine start and press "autobreak MAX", the preselected autobreak turnes off while taxing to the runway. Eighter when breaking or only for no reasons. it only be active for few secounds.

**To Reproduce**

1. prepare the Aircraft from Cold and dark for takeoff
2. set autobreak MAX for takeoff
3.

**Expected behavior**

autobreak MAX shoud stay on/active until Takeoff 

**Actual behavior**

autobreak MAX turn off, eighter when breaking on the way to the runway, or by itself after few secounds

**References**

-

**Additional context**

-
worked befor 0.5.0
no

geoffda:
Change #2510 which was converting RPN to infix was applied incorrectly to the autobrake code.

Fixes #2586

## Summary of Changes
One of the if conditions in the autobrake code was ported incorrectly when it was converted from RPN to infix.

## Additional context

geoffda

## Testing instructions
Autobrake switch behavior should work as follows:
1) When starting on the runway, max should be automatically set.
2) Once set to max, the settings should stay on max while taxiing and braking.
3) After takeoff with autobrakes set to max, autobrake setting should be turned off within 10 seconds of retracting the landing gear.
4) In flight, setting autobrakes max should not work. If autobrakes were set to either low or medium before attempting to select max, those settings should be preserved.
5) Upon landing, autobrakes should engage as per setting. Once manual brakes are applied, the autobrake setting should turn off.
6) At that point (with the aircraft on the ground) any autobrakes setting should be allowed.

### DIFF
--- a/src/behavior/src/A32NX_Interior_Misc.xml
+++ b/src/behavior/src/A32NX_Interior_Misc.xml
@@ -183,6 +183,7 @@
                             alias onGroundWaitCounter = (L:A32NX_ON_GROUND_WAIT_COUNTER, number);
                             alias disarmAutobrakesOnLgRetract = (L:A32NX_DISARM_AUTOBRAKES_ON_LG_RETRACT, bool);
                             alias autobrakesActivated = (L:A32NX_AUTOBRAKES_ACTIVATED, bool);
+                            alias autobrakesActive = (A:AUTOBRAKES ACTIVE, bool);
 
                             if onGround and (A:AUTO BRAKE SWITCH CB, number) == 4 {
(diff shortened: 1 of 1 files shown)


geoffda:
Looks like @pareil6 submitted a fix underneath me.

PR #2963, opened 2021-01-09
fix: correct cost index imported from SimBrief API

mattsains:
## Summary of Changes
I changed the simbrief flight plan import to multiply the cost index by 10, because in the mod's ATSU, the cost index ranges from 0-1000, whereas simbrief ranges from 0-100.

### Simbrief flight plan for testing:
Screenshot (69)
### Before
Screenshot (68)

### After
Screenshot (70)

## Additional context

Discord username (if different from GitHub): mattsains#8270

## Testing instructions

1. Create a flight plan on simbrief for the A320. No settings are really important, but take a note of the cost index.
2. Start the sim with the A320 cold and dark. Power on the avionics
3. Click "Request" on the init page. The cost index can be viewed from the same page and should be a value between 0 and 1000.


Benjozork (approved):
LGTM

pareil6:
MCDU cost indexes are actually 0-999, so I think this will fail on a SimBrief CI of 100 - could someone test?

MisterChocker:
Well this is not really how the ci works. the range 0-99 if I recall correctly works similar to what simbrief intends and the higher ranges actually have a more unique vertical profile.

mattsains:
If there is a formula I can reference to do the conversion, I’m happy to change it

Watsi01:
Yes, that's how I learned it. Above 100 the CI effects mosty climb or descent speed for special profiles. Right now, the CI is calculated with a very small effect to the speeds. Because of the default Asobo formula (still present). Asobo has changed the formular with some update (last one or the one before) for the vanilla A320. But we havn't implemented this new calculation so far. I'm not sure if the new formula wouldn't produce very bizarre flightphase speeds with the A32NX.

mattsains:
@MisterChocker it sounds like you're saying that not mapping to the upper 900 CI values is actually the intended behaviour, because the lower 100 have the same meaning as simbrief, whereas the higher values represent more "exotic" profiles. Is that accurate?

What @Watsi01 mentioned is interesting because I assumed that the CI values weren't doing anything because they had ten times less dynamic range as expected in the Honeywell system we model. Sounds like it's just that cost index doesn't really do anything in the current version

MisterChocker:
> @MisterChocker it sounds like you're saying that not mapping to the upper 900 CI values is actually the intended behaviour, because the lower 100 have the same meaning as simbrief, whereas the higher values represent more "exotic" profiles. Is that accurate?
> 

Yes there is a difference. I am not very familiar with the topic and haven't yet looked up the details.

> What @Watsi01 mentioned is interesting because I assumed that the CI values weren't doing anything because they had ten times less dynamic range as expected in the Honeywell system we model. Sounds like it's just that cost index doesn't really do anything in the current version

Yes currently there is no simulation of cost index.

Watsi01:
@mattsains It does something, but veeery little. If you are interested in the calculation, for the climb you can find it here:
...\Packages\Community\A32NX\html_ui\Pages\VCockpit\Instruments\Airliners\A320_Neo\CDU\A320_Neo_CDU_MainDisplay.js
Look for "getClbManagedSpeed"

for CRZ and DES you find it here:
...\Packages\Community\aircraft-A32NX\html_ui\Pages\VCockpit\Instruments\Airliners\A320_Neo\FMC\A32NX_FMCMainDisplay.js
Look for "getCrzManagedSpeed" and "getDesManagedSpeed"

You can also check the "new" code (actually it's just a little difference but with impact) from Asobo for the vanilla A320 in the according files from the default A320 (under Packages\Official\...).

The problem is, that this is (for vanilla and A32NX) a linear calculation, which would mean, that high cost indexes above 100 (e.g. 500) would produce odd flightphase speeds. That's why we kept it with a low impact. If we want to calculate the CI totaly realistic, we would need a very complicated calculation or a table with all CIs, weights, CRZlevels, etc. to look for the according speeds. Or we could make it simple and limit the CI to 100 and below and stay away from the special profiles.

You can also read this discussion for further information:
https://github.com/flybywiresim/a32nx/pull/1389

beheh:
I'm slightly doubtful of this change. While I've never seen Simbrief spit out a Cost Index above 100, that doesn't mean that Simbrief's 99 means 990 in the airbus. In the past I've used the Cost Index from the OFP exactly as given without multiplying by ten in other mods (ToLiss), and the speeds and times the aircraft flew were very accurate compared to the Simbrief predictions (ground speed within 2 knots usually at cruise).

As an independent issue our CI calculation is all over the place right now so it's bad to cross-check, but a certain CI means a very specific ratio. We should check how the flight time evolves when manually setting the cost index in Simbrief.

If we want to emulate the behavior of our other mods I would suggest we take the exact CI and route from Simbrief, plug it into our FMC and tweak our prediction model to closely match speeds and times to what Simbrief expects. This requires us to have the prediction model in the first place (probably including VNAV) and a well-tuned Simbrief aircraft definition.

tareksabet:
so quick question here, so if I want to simulate a flight from a low cost airline with around of 0 to 10 CI how can I do that and import it from Simbrief ?

Watsi01:
I wouldn't use a CI of 0 right now. Use at least CI1. Overall, as said, the CI is not simulated accurately yet. The impact in speeds is there, but very minimal. We have to wait, if and how we are able to figure this out.

mattstertx:
Category: QA Tester/Trial
Name:  Mattstertx#7410
Date of testing: Current date 18/01/2021
Version of the sim: 1.12.13.0
PR Tested: #2963
Tier of testing: 2
Changes to observe:
I changed the simbrief flight plan import to multiply the cost index by 10, because in the mod's ATSU, the cost index ranges from 0-1000, whereas simbrief ranges from 0-100.

Testing technique: Loaded F-Plan
1. Loaded C&D at KDFW
2. Imported Simbrief F-Plan
3. Compared Simbrief CI to A32NX CI
4.
5.

Media: Microsoft Flight Simulator Screenshot 2021 01 18 - 23 47 59 40
Microsoft Flight Simulator Screenshot 2021 01 18 - 23 53 22 82

Issues (If Any): N/A


Overall Rating: Good
Conclusions of the testing: Works as intended did two tests first one CI on Simbrief 36 in sim 360, Second test Simbrief 26 in sim 260.

tracernz:
[YouTube training video on MCDU](https://youtu.be/hzknF9qIUm4?t=245)
> We enter the cost index. The cost index is decided by the dispatch. **It is usually around 35 to 50, but it can go from 1 to 999.**

I don't think this change is correct.

mattsains:
It's clear that this PR isn't the intended behaviour, so I'm going to close it.

Issue #1211, opened 2020-10-08
[BUG] RMP does not work with Logitech radio panel


tooldrmr2009:
**Mod Version**
 latest flyby wire mod downloaded using the installer

**Describe the bug**

Radio stack in the Airbus a320 does not work with the logitech flight radio panel.  The a320 without your mod works with the logitech radio panel and the panel works in various other aircraft.

**To Reproduce**

1. Happens all the time
2. No certain time frame
3. Logitech radio panel inop with the a320nx

**Expected behavior**

When switching radio frequencies it'll show in the a320 it's being switched and when clicking the standby button it'll switch from active to standby and it does not do this.

**Actual behavior**
. Changing the logitech frequencies leaves the a320 frequencies alone and doesn't show they've been changed

**References**

**Additional context**

Was this working before/when did the issue start occurring?

Is this a problem in the vanilla unmodded game?

Discord username (if different from GitHub):

ElSrJuez:
Please check if this may have a common root cause with https://github.com/flybywiresim/a32nx/issues/1079
If so, consider merging.

tooldrmr2009:
I believe it does share similarities however I am using logitech software to provide the drivers for the MSFS2020 it works in all other aircraft except the a320 thats why I believe its an issue directly with the a320 either flybywire or the a320 in general.

Benjozork:
> either flybywire or the a320 in general.

Can you please confirm this ?

tooldrmr2009:
I will remove the fly by wire mod and see if it works tonight and post up my findings!

tooldrmr2009:
I have removed flybywire and ran the default MSFS2020 A320 and the panel still does not work so its not a Flybywire mod, but maybe it can still be fixed by you guys?  My panel works with other aircraft, 747, 787 ect so i know its not my panel or settings/drivers.

carl-lol:
I've got the same problem

Benjozork:
Please state your add-on version.

carl-lol:
Version 0.5.3, the latest stable version

Benjozork:
Can you try with the development version? Should work better.

carl-lol:
ok, is there a great difference between those version for me as a player?

carl-lol:
thx now it works

davidwalschots:
@tooldrmr2009 Are you still having this problem?

MisterChocker:
Closing since no response from op.

PR #2980, opened 2021-01-10
feat(rmp): separate and rewrite radio management panels

### LINKED ISSUE #1079: [BUG] COM Radios do not update display
**Mod Version** 3.1

**Describe the bug**

The A320 doesn't update the COM radio frequencies on the Virtual Cockpit display when using SimConnect Events to change the frequency value or when switching from Active/Standby

**To Reproduce**

1. Launch a flight
2. Update the radio frequencies using SimConnect events (i.e., Axis and Ohs)

**Expected behavior**

The A320 VC Radio display should maintain display of current frequencies (active and standby) on all updates.

**Actual behavior**

3.The Neo Radio display doesnt show the change
4. By using other apps to read radio variables (i.e., vPilot) the updated frequencies show
5. Other default Asobo aircraft update display just fine

**References**

https://forums.flightsimulator.com/t/logitech-switch-panel/173895/12

**Additional context**

This is important for interacting with the cockpit via SimConnect

No

Yes

ElSrJuez

### LINKED ISSUE #1792: [REQUEST] Working ACP on co-pilot side
This is a feature request for a new feature. 

Create a working ACP (communication frequencies) for the co-pilot side. (Screens currently display right information as selected on pilot side, only buttons and inputs do not work). turning knob for selecting frequencies and button to swap active and stand-by frequencies. 

**References**
Exactly the same as on the pilot side. 

**Additional context**
No additional context.

### LINKED ISSUE #2002: [REQUEST] Radio Panel - Possible to speed up / accelerate radio tuning?

In a VATSIM environment, it can take a long time to change radio frequencies using the tuner knobs.  Is there a way to make it scroll through the frequencies faster, or ideally, could you have it start scrolling through frequencies slowly, but then accelerate the longer you hold the mouse button down?

**References**

**Additional context**

Discord username (if different from GitHub):  zzyzxuk

### LINKED ISSUE #2644: [BUG] Radio frequency always 124.850
**Mod Version**

Master (23 dec)

**Describe the bug**

The vanilla A320 received a fix for this bug, there the readio frequency is automatically updated when you use ATC. With the FBW mod, frequency always says 124.850.

**To Reproduce**

1. Start a flight with the A320nx
2. Monitor radio frequency
3. Use ATC to change frequency
4. Notice they always stay at 124.850
5. Test the same procedure in the vanilla version, there the issue is solved

**Expected behavior**

I would expect the radio frequencies to be updated visually when changing it by using ATC

**Actual behavior**

The radio frequency always stays at 124.850

**References**

**Additional context**

Occured after the bug was introduced in the vanilla version.

No

### LINKED ISSUE #2709: [BUG] COM frequency does not update with in-game button function
**Mod Version**
0.5.1 Stable & Dev
{
  "built": "2020-12-25T10:12:36+00:00",
  "ref": "",
  "sha": "",
  "actor": "Benjamin Dupont <benjozorkfr@gmail.com>",
  "event_name": "manual"
}

**Describe the bug**
The standby frequency does not change, if you push a button mapped to increase or decrease the frequency. Neither whole, nor fract nor fract,carry mode works. The frequency is also not updated in ATC menu if you switch for crosscheck (displays what was the standby frequency before), so it's not an issue with the displays.

**To Reproduce**
1. Launch FBW A32NX
2. Map any button to Increase[Decrease] COM1 (Whole[Fract][Fract,Carry])
3. Press the buttons
4. See nothing happening

**Expected behavior**
Standby Frequency should increase[decrease] by xx1.xxx on whole, increase[decrease] by xxx.x25 on fract, and increase[decrease] by xxx.x25 with increasing[decreasing] xx1.xxx when applicable on fract, carry

**Actual behavior**
Nothing happening

**Additional context**
Might be related to #1079 

unknown

No (as long as you are not stuck with 124.850)

Foxtrot Sierra#6420

theomessin:
Fixes #1079 Fixes #1792 Fixes #2002 Fixes #2644 Fixes #2709 

## Summary of Changes
- [X] Complete rewrite of the RMP displays in React
- [X] Complete rewrite of the RMP behaviours (uses `rnp`)
- [X] Separate radio navigation panels for left and right
- [X] Correct VHF1 / VHF2 / VHF3 RMP implementation
- [X] Continuous SimVar syncing (buttons now work!)
- [X] Separate standby frequencies for each RMP
- [X] Reasonable default frequencies per state

## Feature preview (video)
https://youtu.be/voBVPn4HvmU

## References
[Publicly available FCOM for simulator A320](https://www.smartcockpit.com/docs/A320-Communications.pdf) - data was crosschecked with a real Neo FCOM.
Everything was double-checked with real A320 pilots.

## Additional context
Discord username (if different from GitHub): Theodore Messinezis

## Testing instructions
Radio management panels were completely rewritten. Both need to be fully tested **together and separately**. This PR only implements VHF1 / VHF2 / VHF3. Remote radio nav aid tuning will be a separate PR.. VHF3 acars data mode has not been implemented yet.


theomessin:
Preview of the frequency selector knob with acceleration:
https://streamable.com/a4l11l

theomessin:
Here's another preview of the two radio management panels rewritten from scratch:
https://streamable.com/tlv1dp

VHF1, VHF2, VHF3, HF1, HF2 currently all work as expected (of course HF is just a dummy).
Ignore the initial HF frequencies, I just need to initialise the `L:VARS`.
The knobs use the `AccelerationKnob` from the FCU rewrite and works just fine 😎 

Many thanks to @tyler58546 for splitting the RMPs by applying some model voodoo magic.

src/instruments/src/RMP/AcceleratedKnob.mjs
```diff
@@ -0,0 +1,34 @@
...
+export class AcceleratedKnob {
+    constructor(timeout = 300, increment = 0.20) {
+        this._currentSpeed = 0;
+        this._timeout = timeout;
+        this._increment = increment;
+        this._previousTimestamp = 0;
+    }
+
+    increase() {
+        if (this._currentSpeed < 1 || (Date.now() - this._previousTimestamp) > this._timeout) this._currentSpeed = 1;
```
beheh, line 18: So generally in a Javascript world Date.now() is fine. However, because we're in a Sim here in future, imagine a possible replay mode in future that replays events, but the user chooses 0.5x as the replay speed, or 2x. While that is a hypothetical, I'm generally hesitant of referencing the out-of-sim real-world time from inside our Javascript, especially as the sim already passes us a generic deltaTime (which we have access for example in React with the appropriate hook).

I don't have a concrete suggestion here, especially as it would require a significant rewrite, but I would just like to point out that referencing real-world time should not be done as lightly in the sim as on a website.

src/instruments/src/RMP/index.jsx
```diff
@@ -0,0 +1,30 @@
+import ReactDOM from 'react-dom';
+import { renderTarget } from '../util.mjs';
+import './style.scss';
+import { useState } from 'react/cjs/react.production.min.js';
+import { RadioManagementPanelPair } from './RadioManagementPanel.mjs';
+
+
+const radioManagementPanelPair = new RadioManagementPanelPair();
```
beheh, line 8: I'm generally not stoken about globals like this that are used by React code, but kept outside of React. They can make testing hard in future when we have it and where usually you can just introduce props around or clearly observe React state, these live outside of the tree making both those use cases harder than they need to be.

An alternative pattern could be to keep the class instance in React state itself:

```
const [rmp, _] = useState(() => new RadioManagementPanelPair());
```

src/instruments/src/RMP/RadioManagementPanel.mjs
```diff
@@ -0,0 +1,80 @@
...
+
+    registerCallbacks(renderCallback) {
+        this.renderCallback = renderCallback;
+        this.left.registerCallbacks(() => this.renderRequested());
+        this.right.registerCallbacks(() => this.renderRequested());
+    }
+
+    renderRequested() {
+        // We need to use setTimeout because SimVars do not update instantly, unfortunately...
+        if (this.renderCallback) setTimeout(() => this.renderCallback(this.frequencies), 50);
```
beheh, line 71: So this seems like a *really* common problem to have. We change something, yet we don't see the SimVar immediately changing. I'm aware of at least one similar issue in the MCDU on the RAD NAV page.

This indicates to me, that we'd want a generic solution to this - ideally also a first-class way to do it in React. For example, I could imagine this living in it's own helper class or function to be used anywhere, and a dedicated React hook with the following signature for usage in React code:

```
const {value: mySimVar, notify: notifyMySimVar} = useObserveSimVar("L:MY_SIMVAR");

useInteractionEvent("A32NX_CHANGE_MY_SIMVAR", () => {
  updateSimvar(); // do some stuff, like increment or decrement
  notifyMySimVar(); // instruct our observe hook to poll the simvar in 50ms
});
```

src/instruments/src/RMP/RadioManagementPanel.mjs
```diff
@@ -0,0 +1,80 @@
...
+        if (this.requestRender) this.requestRender();
+    }
+
+    modeButtonPressed(mode) {
+        this.mode = mode;
+        if (this.requestRender) this.requestRender();
+    }
+
+    registerCallbacks(requestRender) {
+        this.requestRender = requestRender;
```
beheh, line 40: For readability, I would just avoid thinking about renders here at all, also from a naming scheme. How about calling this `registerCallbacks(changeListener)`? Then it's clearer that this just a class that observes events and simvars and notifies a callback, rather than something to do with React.

beheh (commented):
Hey, while I see this is still in draft I wanted to share my thoughts so far.

I think in it's a nice improvement over the existing code. However, the new code doesn't really utilize React at all. Instead it implements a sensible structure for classes and then just directly hooks a single React component up to it. This is fine, especially for the RMP where not a lot of renders will probably happen, but especially for any of the larger components or screens in future (DCDU, ECAM, MCDU), this is not a great pattern because at that point we're barely benefiting from what React gives us.

In a fully React-centric implementation, I could imagine the following components:
- A component for the active frequencies, that just displays a frequency prop
- A component for the standby frequencies, that displays a frequency prop but also installs event listeners to react to the knob events and dispatch the change event props
- A frequency pair component wrapping an Active and a Standby frequency, that has an event listener to react to the swap event, monitors the frequency SimVars, and passes the current frequencies to it's frequency children, and in addition a change callback to the standby child
- Finally a parent component, that composes the frequency pairs you want, but otherwise doesn't really do anything

With this structure we can isolate functionality to exactly where it's needed. If the standby frequency detects the knob, it can inform the Frequency Pair about the change. The Frequency Pair can then update the SimVar, and it can also schedule a check for itself to get the updated SimVar value. It can also keep the requested SimVar in it's state, so it can immediately update the Standby component with the requested value, without having to wait for the SimVar to update.

The procedure for a frequency swap is very similar: Upon detecting it, the Frequency Pair component would just update the SimVars, and similarly it could also already optimistically update some local state to immediately update the visuals, without waiting for the SimVar to update.

Thinking one step ahead then, we could come up with a generic SimVar hook we could use in future, where it's `[simVar, setSimVar] = useSimvar("L:MY_SIMVAR")` that changes the SimVar, but also immediately remembers the new value, so there is no functional delay to the return value. It could also incorporate periodic checks to the SimVar itself, in case a user changes it through an external application like SimConnect. But the nice thing is, by making this a hook, we can change the implementation in one place, and all components would benefit.

This writeup is probably a bit confusing because it describes a completely different implementation, but just to be absolute clear I think the code itself is perfectly fine (outside of some easy fixable things like calling hooks from an non-React class).

If we are able though to write this in a more React-centric way, and we manage to abstract some of the problems that we're facing here (like SimVars not updating immediately) we can make our lives a lot easier in future, and benefit more from React, especially if we have to start thinking about performance and how to bail out of re-renders as high up as possible.

LoungeFlyZ:
Moving discussion to this PR from a discord discussion. 

I noticed that I cant bind keys to the the COM up and down bindings (COM_RADIO_WHOLE_INC etc...) at all in the a320 (both the nx or standard). These events are used by things like the Logitech radio panel (& my homebrew arduino radio tuner). 

It would be GREAT if the new radio panel supported the ability to use these events as that would allow people to bind keys to them too.  Currently binding a key to tune the freq doesn't seem possible in the a320 as its not listening to the standard events. 

There are some issues that this seems related to: #2709, #1079 and #1211

Supporting COM_RADIO_WHOLE_INC, COM_RADIO_WHOLE_DEC, COM_RADIO_FRACT_INC, COM_RADIO_FRACT_DEC, COM_STBY_RADIO_SWAP (+ the equivalents for COM2 and COM3) would solve these issues. 

I went poking around the code a while back to see if I could do this, but I am not familiar enough with how events are wired up etc... and then i found your PR :)

St54Kevin:
@theomessin Is this being worked on currently? What's the status?

theomessin:
@St54Kevin yup! Coming soon™ - just had life get in the way. Should have another PR (which will be closer to being ready) some time this week.

LoungeFlyZ:
@St54Kevin do you think you will get the event support for COM_RADIO_WHOLE_INC, COM_RADIO_WHOLE_DEC, COM_RADIO_FRACT_INC, COM_RADIO_FRACT_DEC, COM_STBY_RADIO_SWAP events in?  

Put me down for testing this out.  I am dying not having external radio control in the a320.

St54Kevin:
> @St54Kevin do you think you will get the event support for COM_RADIO_WHOLE_INC, COM_RADIO_WHOLE_DEC, COM_RADIO_FRACT_INC, COM_RADIO_FRACT_DEC, COM_STBY_RADIO_SWAP events in?
> 
> Put me down for testing this out. I am dying not having external radio control in the a320.

I dont know what that means :D

gleclere:
@St54Kevin  COM_RADIO_WHOLE_INC, COM_RADIO_WHOLE_DEC, COM_RADIO_FRACT_INC, COM_RADIO_FRACT_DEC, COM_STBY_RADIO_SWAP are simconnect events allowing external app/devices to interact with the radio.

@LoungeFlyZ if you can't wait, you can download the A32NX version of the PR  [#3157](https://github.com/flybywiresim/a32nx/pull/3157). It works perfectly!

LoungeFlyZ:
@St54Kevin have a read of my comment above from 13 days ago it explains it. 

Supporting those events means people with external controls for radio will be able to use them in the a32nx.

St54Kevin:
> @St54Kevin have a read of my comment above from 13 days ago it explains it.
> 
> Supporting those events means people with external controls for radio will be able to use them in the a32nx.

Yes but Im not the creator of this PR so I dont know what I should do with that information :D

theomessin:
@LoungeFlyZ as long as these events update the default FS simvars for the frequencies they will work.

LoungeFlyZ:
@theomessin ok. that sounds promising. previously in the a32nx this didn't work (see https://github.com/flybywiresim/a32nx/issues/1079#issuecomment-729278305 for details). 

I thought, perhaps wrongly, that the a32nx would need to listen for those sim events and update the freq accordingly.

theomessin:
The latest commit includes a bunch of libraries for hooking to SimVars with React. 
The RMP itself was rewritten. The following things need to be done before this PR is ready:

- [ ] Rewrite behaviours XML.
- [ ] VHF1 / VHF2 / VHF3 switching.
- [ ] Minimum and Maximum Frequency.
- [ ] Fix frequency flicker when using knobs due to periodic refresh.

As discussed on Discord, it's probably best if this PR gets merged with full functionality for VHF and another PR be submitted to extend with dummy HF and NAV functionalities.

Benjozork:
cc. @RichardPilbery for the simvar framework

src/instruments/src/RMP/AcceleratedKnob.mjs
theomessin: More research is required to see how this would perform in edge cases (time acceleration, playback, etc.)
I think this is good enough for now.

src/instruments/src/RMP/RadioManagementPanel.mjs
theomessin: See `Framework` folder.

theomessin:
This PR is now ready for review and testing. 
Check top post for more details.
Linting will eventually be fixed...

src/instruments/src/RMP/Components/RadioManagementPanel.jsx
```diff
@@ -0,0 +1,136 @@
...
+    // If the panel is not powered, render two <svg> for spacing.
+    if (!powered) return (<span><svg/><svg/></span>);
+
+    if (mode === "vhf1" || mode === "vhf2" || mode === "vhf3")
+        return renderVhfDisplays(frequencyVariables, mode, props);
+
+    // If we get here, something's gone wrong, we'll just render 888.888
+    return (<span>
+        <SevenSegmentDisplay lightsTest={true}/>
+        <SevenSegmentDisplay lightsTest={true}/>
```
Benjozork, line 134: Should `lightsTest` be constantly `true` ?

wpine215 (approved):

Benjozork (changes requested):

Benjozork (approved):
LGTM

wpine215:
Works like a charm, even with vPilot.

LoungeFlyZ:
This has fixed the issues with COM_RADIO_WHOLE_INC, COM_RADIO_WHOLE_DEC, COM_RADIO_FRACT_INC, COM_RADIO_FRACT_DEC, COM_STBY_RADIO_SWAP events.  

Thanks a lot for all the work on this!

PR #1580, opened 2020-10-28
Improvements to Flightmodel, PFD Speedband and calculations for various speeds

MisterChocker:
Fixes #2027

## Known Issues
Slats are currently not being simulated correctly by the engine. Therefore we had to find compromises, see [Stall Speeds and Pitch Angles](https://github.com/flybywiresim/a32nx/files/5579383/Stall.Speeds.and.Pitch.Angles.pdf) for reference.

## Summary of Changes

- Flightmodel
   - Reworked Coefficient Lift vs Angle of Attack table.
   - Reworked FLAPS
   - Reworked Pitch for all phases of flight.
   - Adjusted Drag/Thrust balancing

- PFD
   - VLS, Vaprot, Vamax and Vs are now calculated by us.
   - Visual improvements to Vls and Vaprot
      - Increased stroke size, tweaked gap-sizes
   - added mach correction
   - changed green dot calculation
   - v1 disappears now after liftoff
   - GD only visible in clean config

- Moved speed calculations to Core
   - Vs
   - Vamax
   - Vaprot
   - Vls
   - F
   - S
   - GD
   - added simple calculation for V2 based of #1777

## Screenshots (if necessary)

| Takeoff | Approach (FLAPS3, Speedmode Managed) | Cruise |
| ----- | ----- | ----- |
| image | image | image |

| GD before | GD after |
| ----- | ----- |
| image | image |

| At cruise level and cruise speed |  |
| ----- | ----- |
| image |  |

## References

[vs and vls speeds](https://github.com/flybywiresim/a32nx/files/5590480/A320_Stall_and_VLS_speeds_rev.pdf)
### -251:
[320NEO_2.zip](https://github.com/flybywiresim/a32nx/files/5469381/320NEO_2.zip)
[A320-251_flight_performance.zip](https://github.com/flybywiresim/a32nx/files/5469382/A320-251_flight_performance.zip)
image
image
image
image

### -214:
[crz_A320-214.txt](https://github.com/flybywiresim/a32nx/files/5469384/crz_A320-214.txt)
[dsc_A320-214_1.txt](https://github.com/flybywiresim/a32nx/files/5469385/dsc_A320-214_1.txt)
[clb_A320-214_1.txt](https://github.com/flybywiresim/a32nx/files/5469386/clb_A320-214_1.txt)

## Additional context

Optimizing drag/thrust ratio to work with current engine model.

Discord username (if different from GitHub): 
- @donstim donbikes#4084
- @MisterChocker DerL30N#3751

[...]

PR #2993, opened 2021-01-11
[Flight Model] Improve rotation and flare characteristics

### LINKED ISSUE #2817: [BUG] Rotation/lift off is too quick
**Mod Version**

Latest custom FBW dev verison.

**Describe the bug**

While the flight model is much better with the custom FBW, the plane "jumps" off the ground very quickly right as rotation starts. Additionally, the rotation of the nose seems too quick with too little sidestick input required. 

**To Reproduce**

1. Do a normal takeoff in the custom FBW version.
2.
3.

**Expected behavior**

Nose slowly raises and eventually, airplane lifts off.

**Actual behavior**

Nose quickly raises and airplane lifts off almost immediately. 

**References**

Comparison with FSLabs. Same conditions, weight, and CG. Note that the FSLabs requires sidestick input up to about 10 degrees to get the nose off the ground whereas the FBW starts rotating with about 5 degrees of sidestick deflection. Also note that the FBW is at 60 feet RA just when the FSLabs achieves a positive rate of climb.

https://www.youtube.com/watch?v=BWKbU_9yIuE&feature=youtu.be

**Additional context**

Always an issue.
Yes

MapleNAV-TCASFAULT:
Fixes #2817

## Summary of Changes

This flight model change should improve the rotation and flare, needing more input to cause the aircraft to rotate around the pitch axis.

## Screenshots (if necessary)

https://user-images.githubusercontent.com/72575144/104204226-07577180-53fb-11eb-8894-82021d53de8d.mp4

## References

references are based of from PR #1580

## Additional context

Discord username (if different from GitHub): ChristopherXie#3904 (Ma_ple [Z-5])

## Testing instructions

NOTE: aircraft will feel a lot different along the lateral axis
NOTE: due to the changes of induced drag, fuel consumption will be slightly higher.
- Do takeoffs and landings with different weights, 70t, 60t, 50t
- Make sure that on rotation, 1/3 to 1/2 up sidestick deflection is needed to get the aircraft rotating at ~3 degrees/s. (in-game sidestick)
- Make sure that on takeoff roll, when the sidestick is neutral by 100 that the aircraft doesn't start rotating by itself.
- Make sure that at different weights, this PR maintains relatively the same flight perf integrity of PR #1580
97865422-77b28a00-1d0a-11eb-8897-493893a3a2fc
97865436-7d0fd480-1d0a-11eb-9819-8e65dcb42677
97865444-81d48880-1d0a-11eb-939b-3c31d90f41ca
97865455-8600a600-1d0a-11eb-9aed-8ccfbf17437d

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/flight_model.cfg
@@ -300,10 +300,10 @@ lift_coef_at_drag_zero_flaps = 0.10000
 [FLIGHT_TUNING]
 cruise_lift_scalar = 1
 parasite_drag_scalar = 1
-induced_drag_scalar = 1
+induced_drag_scalar = 2
 flap_induced_drag_scalar = 1
-elevator_effectiveness = 1
-elevator_maxangle_scalar = 1
+elevator_effectiveness = 0.5
+elevator_maxangle_scalar = 0.25
 aileron_effectiveness = 1
 rudder_effectiveness = 0.8
 rudder_maxangle_scalar = 0.78


donstim:
Probably the first test that needs to be run after making any change to either the elevator effectiveness or its maximum deflection is to ensure there is still enough elevator authority to perform a proper stall maneuver and that the stall speeds are not affected. I ran some checks with flaps up and at all flap settings, and the proposed changes significantly changes the stall behavior of the airplane (with the stall protection system disabled).  Except for possibly CONF 1+F, there is no longer enough elevator authority to reach the actual stall angle-of-attack. The airplane just starts descending without any identifiable stall. Without being able to identify an actual stall point, it is impossible to verify that the speed at which the stall occurs is unaffected. And because of the increased in induced drag, which is highest at high angles of attack, the speed is rapidly reduced below the normal stalling speed, but the airplane remains in a descent rather than recovering.

I have not thoroughly examined other aspects yet because this finding appears to be a significantly detrimental side effect. But I must point out that the decision to use the induced drag scalar to double the induced drag will double the induced drag in all configurations, not just the CONF 1+F takeoff configuration that you apparently based this on. The higher the lift coefficient, the greater the impact. So, for example, heavier weights will generally be more affected than lighter weights (when comparing at a constant speed), and higher flap deflections will be affected more than lower flap deflections. Since the drag level of the current flight model was determined and verified through extensive testing and comparison to reference performance data more extensive than that provided above as well as A320NEO pilot evaluations, this pr should go through the same process (assuming that the significant issue identified in the first paragraph above can be resolved).

Snapmatics:
<3

aguther:
Beside the tests @donstim has already mentioned, I want to add tests with the custom fbw system (and the upcoming custom AP system as well).

MisterChocker:
I am pretty sure that the changes will have a bad impact on our custom fbw system.

PR #1444, opened 2020-10-18
Issue 596 decel wrong light

jokey2k:
Fixes #596

## Summary of Changes
Ensures the right DECEL light comes up during autobrake usage

## Screenshots (if necessary)
Before
before

After
after

## References
Lots of mentions by 320 sim pilot, for example
https://youtu.be/32q0xHybKmA?t=1440
"Decel light should show up medium 'cause that's what we use"

## Additional context
Discord username (if different from GitHub): Jokey | Markus#0001

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build.py will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, slick on the **Artifacts** drop down and click the **A32NX** link

PR #3173, opened 2021-01-25
fix: show button labels when unlit

tyler58546:
## Summary of Changes

- Fixed button labels not showing when unlit
- Fixed white dots illuminating when ANN LT TEST is active
## Screenshots (if necessary)

Before:
image
image
After:
(it's very subtle so you will need to zoom in to see)
image
image

## References

image
image

## Testing instructions

- Check that all buttons have visible labels when unlit
- Ensure that all buttons still work normally when pressed and when ANN LT TEST is active


MisterChocker:
Are we able to improve the overall readability? e.g. using a higher contrast / increasing alpha

DarkOfNova:
Decel should be made invisible again when unlit. It was done for a reason based upon pilot and engineer feedback a while back. Not so sure about the 'UNLK' but that might be the same case.

MisterChocker:
> Decel should be made invisible again when unlit ([#1444](https://github.com/flybywiresim/a32nx/pull/1444)). It was done for a reason based upon pilot and engineer feedback a while back. Not so sure about the 'UNLK' but that might be the same case.

He is right.
image

DarkOfNova:
According to that pic UNLK should also be invisible

davidfrp:
Looking at the reference photos, it would seem that only the lower parts of the buttons are somewhat visible when unlit.
The upper half doesn't seem to be visible on any of buttons, expect for the ones with two dots.

tyler58546:
image
In this picture, the top half of the button is just slightly visible. I think we should make the texture darker for the top half of the buttons instead of hiding it.

MisterChocker:
Is this a duplicate of #3842?

tyler58546:
> 
> 
> Is this a duplicate of #3842?

Yes

PR #3270, opened 2021-02-01
refactor: introduce new SimVar hooks for React

### LINKED ISSUE #3374: [BUG] RMP active frequencies no longer sync when RMP1 and RMP2 are both set to VHF1
**Mod Version**
 "built": "2021-02-06T14:57:17+00:00",
  "ref": "master",
  "sha": "cb2f6379eb0fc00ee72c592b194b065b2b07b33d",
  "actor": "ko-zone",
  "event_name": "manual"

**Describe the bug**
When both radio panels are set to VHF1, the active frequencies do not sync.  Syncing was a feature of the RMP rework using react and was verified working

**To Reproduce**

1. Start C&D, turn on power
2. Set both RMPs to VHF1
3. Change the active frequency on either panel and observe if they sync

**Expected behavior**
When you change the active frequency on one RMP it should sync the active frequency to the other  RMP

**Actual behavior**
It is no longer syncing active frequencies

**References**

**Additional context**

Yes was working after RMP rework using react.  Not sure if it's related to the RMP fix as I haven't flown for a couple days

N/A

rmp

### LINKED ISSUE #3408: [BUG] RMP causing CTD (possible memory leak)
{
  "built": "2021-02-07T18:49:42+00:00",
  "ref": "fbw",
  "sha": "84192108aa63445a1930ae3fca5cd7d5392a6476",
  "actor": "aguther",
  "event_name": "manual"
}

**Describe the bug**
The third time the simulator CTD, it happened between 1H and 30 minutes to 2H of the flight. When looking to the left (where the EFB is located) I received extremely low frames (1-3) After, I rested my camera and the frames came back. I did this a few times and it repeated. 2 minutes after in default cockpit camera, the simulator crashed.

**To Reproduce**
1. Load a flight (my flights were from KSFO-KLAX, KPHX-KLAX, KBOS-CYYZ)
2. Near the 1 hour and 45-minute mark look towards the EFB, and identify any frame drops
3. Reset camera position and wait a few minutes to see if the simulator crashes

**Expected behavior**
Normal flight without a CTD, able to look at the direction and use the EFB.

**Actual behavior**
While looking/interacting with the EFB the simulator CTD a few minutes after

**References**

None, as I did not expect the simulator to crash

**Additional context**
I'm assuming that the EFB is causing a memory leak

The simulator did not crash for a long time since the EFB was introduced

No

Discord username (if different from GitHub): Darlack#3281

theomessin:
Fixes #3374 Fixes #3408 

## Summary of Changes

This pull request introduces the new SimVar utilities for React co-authored by @theomessin and @beheh.

It introduces the following 3 new hooks:
- **`useSimVar`**
  The basic version. You'll want this 90% of the time.
- **`useGlobalVar`**
  The same, but for global variables (like `ZULU TIME`).
- **`useInteractionSimVar`**  
  The optimized version. If you emit an H event anyway whenever it changes (e.g. a knob), this is for you.
- **`useSplitSimvar`**
  Use this if you're accessing a SimVar that is set using K events (mostly simulator SimVars). Not used often.

Here's a few examples:
```typescript
// the return value is the value itself and a setter, similar to useState
const [v1, setV1] = useSimVar('L:AIRLINER_V1_SPEED', 'Knots');

// only refresh the SimVar every 500ms (unless this SimVar is lower elsewhere)
const [lightsTest] = useSimVar('L:XMLVAR_LTS_Test', 'Bool', 500);

// access a GlobalVar which is refreshed at least every 100ms
const time = useGlobalVar('ZULU TIME', 'seconds', 100);

// the XML updates the SimVar and emits an H event, so we can use the optimized version
const [toggleSwitch, setToggleSwitch] = useInteractionSimVar(
  'L:A32NX_RMP_L_TOGGLE_SWITCH', 'bool',
  'A32NX_RMP_L_TOGGLE_SWITCH'
);

// read the SimVar 'COM STANDBY FREQUENCY:2', and set it through 'K:COM_2_RADIO_SET_HZ'
const [frequencyTwo, setFrequencyTwo] = useSplitSimVar(
  'COM STANDBY FREQUENCY:2', 'Hz',
  'K:COM_2_RADIO_SET_HZ', 'Hz'
);
```

We've also decided to write in Typescript, so we've moved the `tsconfig.json` and had to make a tiny change to `rollup.config.js`.

Other than that, we've refactored the RMP to use the new hooks as a live example of how these new SimVars work and to ensure their API is useable.

## Additional context

Future work outside of the scope of this PR:
- Add a `<RefreshRate>` component to set a default refresh rate in parts of the tree (rather than defaulting to 0)
- Add a way to guarantee SimVar updates, like a number on a screen only changing every 500ms (note that maxInterval is only best-effort for now)

## Testing instructions

- Test that the Radio Management Panels work as they should.
- VHF1, VHF2, VHF3 are split and should fully work.
- All states should load with correct frequencies.
- Ensure that any previously reported RMP bugs have been resolve (#3302, #3362, #3374) , 
- Test whether there are any performance issues that need to be addressed (reactivity, fps).
- Make sure to perform long flights.
- [!] The RMP Nav modes do not work, this is expected.
- [!] The RMP LED lights do not light up when in lights test mode, this will be a separate fix PR.


A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/panel/panel.cfg
```diff
@@ -83,11 +83,11 @@ texture=$BRK
...
 
 [VCockpit11]
-size_mm=170,256
-pixel_size=170,256
+size_mm=1280,1280
+pixel_size=1280,1280
 texture=$FDW
 
-htmlgauge00=generated/RMP/template.html, 0,0,170,256
+htmlgauge00=generated/RMP/template.html, 0,0,1280,1280
```
theomessin, line 90: This is just for debugging, should **not** be part of the final PR.

theomessin:
@beheh, we need to have a chat regarding deltaTime and using Date.now / setInterval. Related to this:
https://github.com/flybywiresim/a32nx/pull/811

Other than that, I refactored the RMP plus some minor things, also fixed SplitSimVar not instantly updating because we were caching two different keys for read/write.

src/instruments/src/RMP/Components/VhfRadioPanel.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+const useStandbyVhfFrequency = (side: string, transciever: number) => {
+    // Use custom SimVars for abnormal standby frequency.
+    // Allows true-to-life independent standby frequencies per RMP.
+    // @todo, if we ever add a third RMP (e.g. C), update this.
+    if (
+        (side === 'L' && transciever !== 1) ||
+        (side === 'R' && transciever !== 2)
+    ) {
+        return useSimVar(`L:A32NX_RMP_${side}_VHF${transciever}_STANDBY_FREQUENCY`, 'Hz', 100);
```
beheh, line 35: Fix rules of hooks.

beheh:
As  this has been co-authored by both @theomessin I obviously approve of this PR, but I'd like another developer that was not involved in the design to review this and give their formal approval.

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+        for (const [key, intervals] of Object.entries(listeners.current)) {
+            // First, let's check if there are any listeners at all
+            if (!intervals.length) {
+                continue;
+            }
+
+            // The refresh time is given by the *smallest* maximum update
+            // interval.
+            const threshold = Math.min(...intervals);
+            const lastUpdatedAgo = (cache[key] ? cache[key].lastUpdatedAgo || 0 : 0) + deltaTime;
```
davidwalschots, line 98: **suggestion**: It seems deleting a SimVar removes the key from the cache which would make checking `cache[key]` existence obsolete. Consider removing this check.

**suggestion**: Similarly to squeeze out the last bit of performance. One could say that on adding a SimVar the `lastUpdatedAgo` should be set to 0, so you can remove `||`.

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+
+        for (const [key, intervals] of Object.entries(listeners.current)) {
+            // First, let's check if there are any listeners at all
+            if (!intervals.length) {
+                continue;
+            }
+
+            // The refresh time is given by the *smallest* maximum update
+            // interval.
+            const threshold = Math.min(...intervals);
```
davidwalschots, line 97: **suggestion**: A suggestion I make on the basis that this PR is very much about performance: consider only calculating the threshold for a given key while adding or removing an interval for that key. Effectively this means keeping an array of intervals and one properly `thresholdInterval` (or however you want to call it); and then using `thresholdInterval` here.

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+                // need to fetch the latest value from the simulator and store
+                // it.
+                const [name, rawUnit] = key.split('/');
+                const unit = normalizeUnitName(rawUnit as UnitName);
+                stateUpdates[key] = {
+                    value: name.startsWith('_GLOBAL_')
+                        ? SimVar.GetGlobalVarValue(name.substr(8), unit)
+                        : SimVar.GetSimVarValue(name, unit),
+                    lastUpdatedAgo: lastUpdatedAgo % threshold,
+                };
```
davidwalschots, line 111: **suggestion**: Again if performance is the absolute goal: consider storing (1) the normalised unit, (2) whether it's global or not, and (3) the name in a separate object alongside the key, such that there is no need to parse it every time.

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+    /**
+     * This function will be called by the useSimVar hook through the context
+     * and notifies us that there is one listener less for this specific SimVar
+     * and unit combination.
+     */
+    const unregister: UnregisterSimVar = (name, unit, maxStaleness, global): void => {
+        const key = getKey(name, unit, global);
+        const old = listeners.current[key];
+        if (!Array.isArray(old) || old.length === 0) {
+            throw new Error('Attempted to unregisterHook with no known listener');
```
davidwalschots, line 207: **thought**: This is a matter of style. I tend to choose idempotency over throwing errors. I'm making the assumption that in general the developer working with these hooks will themselves not be unregistering things (it is a concern of the hooks framework you built). Thus, it would be confusing if internal errors are thrown due to a programming error the developer didn't make.

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+            // - one of the parameters below (name, unit, maxStaleness) is about
+            //   to change, or
+            // - the component is about to unmount
+            // In these cases, we want to unregister our current parameters from
+            // the SimVar provider, that we again access through the context.
+            contextValue.unregister(name, unit, maxStaleness, true);
+        };
+    }, [name, unit, maxStaleness]);
+
+    return contextValue.retrieve(name, unit, false,true);
```
davidwalschots, line 323:
```suggestion
    return contextValue.retrieve(name, unit, false, true);
```

src/instruments/src/RMP/Components/BaseRadioPanels.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { VhfRadioPanel } from './VhfRadioPanel';
+import { RadioPanelDisplay } from './RadioPanelDisplay';
+import { useSimVar, useInteractionSimVar } from '../../Common/simVars';
+import { useInteractionEvent } from '../../Common/hooks';
+
+/**
+ *
```
davidwalschots, line 8: These empty comments might fool the computer! But humans are less easily fooled! 🤣

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+ *
+ * @returns {[*, (function(*): void)]}
+ *
+ * @see {@link useSplitSimVar} if your SimVar is set through a K event
+ * @see {@link useInteractionSimVar} if you emit an H event whenever your SimVar changes
+ * @see {@link useGlobalVar} if you have a Global Var instead
+ */
+export const useSimVar = (
+    name: string,
+    unit: UnitName,
```
davidwalschots, line 269: **suggestion**: Instead of the `normalizeUnitName` function. Consider limiting the valid unit names to a list of unit names, such that we'll get some consistency in how things they are written throughout our code.

src/instruments/src/RMP/Components/RadioPanelDisplay.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import { useSimVar } from '../../Common/simVars';
+
+/**
+ *
```
davidwalschots, line 5: 🤖

src/instruments/src/RMP/Components/VhfRadioPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useSplitSimVar } from '../../Common/simVars';
+import { RadioPanelDisplay } from './RadioPanelDisplay';
+import { useInteractionEvent } from '../../Common/hooks';
+
+/**
+ *
+ */
+interface Props {
+    side: string,
+    transciever: number,
```
davidwalschots, line 12: **typo**: `transciever` should be written as `transceiver`.

src/instruments/src/RMP/Components/VhfRadioPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+/**
+ *
+ */
+const useStandbyVhfFrequency = (side: string, transciever: number) => {
+    let variableReadName = `COM STANDBY FREQUENCY:${transciever}`;
+    let variableWriteName = `K:COM${transciever === 1 ? '' : transciever}_STBY_RADIO_SET_HZ`;
+
+    // Use custom SimVars for abnormal standby frequency.
+    // Allows true-to-life independent standby frequencies per RMP.
+    // @todo, if we ever add a third RMP (e.g. C), update this.
```
davidwalschots, line 33: **thought**: I'm not sure if I'd add a `@todo` for this. As IDEs will show these in the "todo list" whilst they're not really something to do now.

davidwalschots (commented):
Looks excellent overall! I have some minor suggestions for your consideration.

src/behavior/src/A32NX_Interior_RMP.xml
```diff
@@ -21,13 +21,10 @@
...
-                    alias toggleSwitch = (L:A32NX_RMP_#SIDE#_TOGGLE_SWITCH, number);
-                    toggleSwitch * 100
-                </ANIM_CODE>
-                <LEFT_SINGLE_CODE type="rnp">
-                    alias toggleSwitch = (L:A32NX_RMP_#SIDE#_TOGGLE_SWITCH, bool);
-                    toggleSwitch = !toggleSwitch;
+                <ANIM_CODE>(L:A32NX_RMP_#SIDE#_TOGGLE_SWITCH, number) 100 *</ANIM_CODE>
+                <LEFT_SINGLE_CODE>
+                    (L:A32NX_RMP_#SIDE#_TOGGLE_SWITCH, bool) ! (>L:A32NX_RMP_#SIDE#_TOGGLE_SWITCH, bool)
+                    (>H:A32NX_RMP_#SIDE#_TOGGLE_SWITCH)
```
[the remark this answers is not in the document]
theomessin: You are correct. However, you can use `>` if your XML is under `src` because the build step with replace it with `&gt;` 😄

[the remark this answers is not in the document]
WUBingfeng1984: I know no

src/instruments/src/Common/simVars.tsx
beheh: The framework is basically doing reference counting here, and there's an assumption that that is always correct as enforced by useEffect. If we degister more than we register SimVars may stop updating, or we may keep updating SimVars we don't care about if we forget to unregister.

So if we lose track of our count here, that's a really bad thing and it can lead to errors that are subtle or hard-to-debug. I'd rather treat this scenario of an obvious miscount as a critical failure in the system and thus raise an error, rather than just ignoring the situation and leave debugging up to the developer. I think I would like to keep it as-is.

But I agree there is some personal preference involved, the code is written in a slightly defensive manner here.

davidwalschots (approved):

beheh:
Don't merge yet, I will incorporate one or two of @davidwalschots other suggestions. Thanks for the review!

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+    update: UpdateSimVar,
+    register: RegisterSimVar,
+    unregister: UnregisterSimVar,
+}>({
+    retrieve: errorCallback,
+    update: errorCallback,
+    register: errorCallback,
+    unregister: errorCallback,
+});
+const {Provider: InternalProvider}  = context;
```
Benjozork, line 60:
```suggestion

const { Provider: InternalProvider } = context;
```

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+/**
+ * This component provides the basic functionality for the useSimVar hooks.
+ * By keeping the last known SimVar values inside this provider, we're
+ * effectively caching each SimVar, so that there is no additional overhead when
+ * using multiple useSimVar hooks for the same SimVar.
+ * For improved performance, this component will only trigger renders when the
+ * "update" custom event is emitted through an instrument.
+ */
+const SimVarProvider: React.FC = ({ children }) => {
+    const listeners = React.useRef<Record<string, number[]>>({});
```
Benjozork, line 80: Why don't we import `useRef` ?

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+ * @see {@link useSimVar} if you're trying to access a SimVar instead
+ */
+export const useGlobalVar = (
+    name: string,
+    unit: UnitName,
+    maxStaleness: number = 0,
+): SimVarValue => {
+    const contextValue = React.useContext(context);
+
+    React.useEffect(() => {
```
Benjozork, line 305: Why don't we import `useEffect` ?

src/instruments/src/Common/simVars.tsx
```diff
@@ -0,0 +1,455 @@
...
+
+/**
+ * This is an internal hook that exposes the internal value for a SimVar only.
+ * You will usually want to useSimVar instead. Don't use this unless you know
+ * what you're doing and writing your own hook.
+ */
+export const useSimVarValue = (name: string, unit: UnitName, maxStaleness: number): SimVarValue => {
+    const contextValue = React.useContext(context);
+
+    React.useEffect(() => {
```
Benjozork, line 420: Same.

Benjozork (changes requested):

src/instruments/src/Common/simVars.tsx
beheh: I do agree, however as part of this PR I don't want to enforce this, it's not even correctly typed yet (unit name just being a placeholder types as `any` for now). I think this will follow when we eventually clean up the typings.

src/instruments/src/Common/simVars.tsx
beheh: I think there will be barely any measurable impact for this, for now I'd prefer to just be safe here in case it's undefined, and rather optimize it later. Right now I'd like to land this and get people used to the API, and then we may take a pass later again for small things like this.  I'll refer back to this though.

src/instruments/src/Common/simVars.tsx
beheh: Similar to above, I think for now I'm more worried about messing this up at runtime, I think that's an optimization here we can land later. Your feedback is very much appreciated, I'm sure we'll do another pass for things like this in future.

Benjozork (approved):
LGTM

Benjozork:
Category: Development Team
Name: someperson#4953
Date of testing: 13/02/21
Version of the sim: 1.12.13
PR Tested: #3270
Tier of testing: 2
Changes to observe: Intrroduction of new SimVar framework for React instruments

**THIS REPORT IS STRICTLY FOR PERFORMANCE**

Testing technique: Flight from KIAD to TNCM

Flight notes (When done a full flight): Performance remained stable throughout flight

Issues (If Any): None

Overall Rating: Good
Conclusions of the testing: This report is only for performance. I did not notice any performance drops in a 3.5hr flight.

armankir:
Category: Contributor
Name: Arman#5297
Date of testing: 13/02/21
Version of the sim: 1.12.13
PR Tested: #3270
Tier of testing: 2
Changes to observe: Introduction of new SimVar framework for React instruments

Testing technique: Flight from KIAD to TNCM

Flight notes (When done a full flight): NO performance issues noticed

Issues (If Any): None

Overall Rating: Good
Conclusions of the testing: This report is only for performance. I did not notice any performance drops in a 3.5hr flight.

GlitchedYT:
Category: QA Tester
Name: Glitched#0740
Date of testing: Current date 2/13/2021
Version of the sim: 1.12.13.0
PR Tested: #3270
Tier of testing: 2
Changes to observe:This pull request introduces the new SimVar utilities for React
(Insert the Summary of changes, in order to remind what is to test)

Testing technique: Description of how you tested the Pull Request
1. Spawn C&D at KIAD
2. Loaded a flight from simbrief to TNCM
3. performed the flight and observed issues in performance while messing with RMP
4. completed flight

Flight notes (When done a full flight): n/a

Steps to achieve the issue

Overall Rating: Good, finally no issues with performance. Can enjoy long haul flights now
Conclusions of the testing: Make a small descriptive to areas where it needs to be improved and recommendations.

Benjozork:
Seems like this does not fix #3362.

Waiting to see if we should merge anyway.

theomessin:
> Seems like this does not fix #3362.
> 
> Waiting to see if we should merge anyway.

Let's merge anyway, 3362 needs more info. I haven't been able to reproduce it.

PR #3346, opened 2021-02-05
build: cargo workspace at the project root

davidwalschots:
## Summary of Changes
By having the cargo workspace at the project root, one can develop in
our whole codebase at the same time. Earlier one needed to have two
instances of e.g. vscode open to work on both HTML/JS and Rust.

## Additional context

## Testing instructions
No testing necessary.


ZigTag (approved):
LGTM

Benjozork:
Is this not really a vscode issue ?

davidwalschots:
@Benjozork I don't know what other editors expect. I'm assuming they also don't like having the Cargo workspace defined in a subsubsub directory. Basically this is equal to putting `package.json`, `.editorconfig`, `jsconfig.json`, and others at the root; such that editors can pick it up.

Benjozork:
It works in my editor without any issues.

I'm fine with the PR, just confused as to why it's a problem in the first place. I thought we could eventually move everything to sub directories where they belong, but if vscode can't handle that...

PR #3350, opened 2021-02-05
feat(efb): Performance Page - Takeoff and Landing Calculations

matt-winfield:
## Summary of Changes
Implements approximate takeoff and exact landing calculations into the EFB.

## Screenshots (if necessary)
image
image
image
image
image
image
image
image
image

## References
Landing distance corrections obtained from QRH - Landing Distance Without Failure.
image
image

Takeoff calculations are derived from the FCOM and stall speeds. Full derivations of the maths required are documented in docs/efb-takeoff-calculations. It's important to note that V1 is an approximation when limited by runway length, whereas V2, VR and landing distances are exact. However I have tested the velocity approximation equation used for V1 speed calculation against data recorded from tests in the simulator, and it is very close to actual values.

image

## Additional context

Discord username (if different from GitHub): TheMightyMat#0585

## Testing instructions
Ensure that computed landing distanced vary as specified in the QRH landing distance without failure tables. The runway visualiser should match the input and output data.
Ensure takeoff results vary reasonably with changes in input. Runway visualiser should result correctly.


Gian85:
will you add flex calculation?

donstim:
Probably best just to start with landing performance since the data is readily available. My initial comments:
1) This information is meant for an in-flight landing check. It's purpose is for the flightcrew to check that a landing can be safely made for the specific runway they will be using for the runway surface conditions and other factors that will be present at the time of landing. There is a different landing performance check done before takeoff, using different data and meeting different requirements, that may limit the takeoff weight (or prevent going to that destination at all).
2) It is intended that a safety margin be applied to this data. The recommended value of the safety margin is 1.15. Therefore, I recommend adjusting the calculated landing distances put out by the EFB to be multiplied by 1.15..
3) The altitude adjustment is based on pressure altitude. The pressure altitude is not provided directly on the ATIS or other weather information available to the pilot, but the QNH value from which the pressure altitude can be calculated is. Therefore, I recommend that the EFP input for altitude be changed to QNH at the destination.
4) The length of the portion of the runway that can be used for landing is the landing distance available (LDA), which can be different than the runway length (primarily due to displaced thresholds for obstacle clearance during the approach).  LDA should be used instead of runway length.
5) Of the 3 distances, 2 are for the autobrake settings of medium and low. The other distance is for maximum manual braking. This should be made clear or there will be many comments that the max autobrake setting cannot be selected for landing. Also, while these distances will be realistic distances for the real airplane, it is my understanding that the sim does not yet fully simulate autobrake braking. Hopefully, that will come at some point.
6) For some reason, not all of the different runway braking condition levels are included (specifically what are called the in-betweeners -- good-to-medium and medium-to-poor). Operators fought hard to have these included in this newly implemented process! There is no requirement to do so, but I suggest they be included.
7) I suggest grouping the inputs by category -- for example, all the atmospheric inputs should be grouped together, all the runway-related inputs, and all the airplane-related inputs.
7) I think it will be difficult for most sim pilots to have the information to fill all the input fields. I strongly suggest that automatic inputs of as much of this information as possible be enabled. For example, the runway information like LDA, runway heading, and slope should be drawn from databases of such information, while other information such as landing runway, QNH, OAT, VAPP, and weight should be shared with the CDU. I don't know if sharing of such info with the CDU is entirely realistic or not, but other similar playability issues, like autofill of takeoff speeds and ZFW/ZFWCG in the CDU have been included.
8)

donstim:
For takeoff, the method and assumptions don't resemble real world methods and parameters at all. Some of these include:
1) There is no (zero) relationship between V1 and stall speed. V1 is not a flying speed, so stall speed has no bearing on the selection of V1. Then to base whether or not the runway is long enough on whether or not the assumed relationship between V1 and stall speed exists is entirely wrong.
2) The thrust rating of an engine is the amount of thrust it can generate at sea level, zero airspeed on a test stand. It doe not include the installation losses associated with mounting it on an airplane. Net thrust varies with temperature beyond the flat rating, pressure altitude, and airspeed. It is entirely incorrect to treat it as a constant value equal to the thrust rating.
3) Takeoff performance must consider both go and stop cases, and both all-engines-operating, and with one engine failing 1 second before V1. The accelerate-go cases must allow the airplane to reach a height of 35 feet at V2 within the takeoff distance available  (TODA) and a height halfway between liftoff and 35 feet within the takeoff run available (TORA). The takeoff performance approximation in this pr does not even consider the accelerate-go case.
4) For the accelerate-stop distance (referred to as RTO in this pr), the acceleration part should be with all engines operating at TOGA thrust (or FLEX if that will be an option), and with rolling friction, but the stop (RTO) portion should be with idle thrust (after engine spindown), braking friction, and spoiler lift and drag included.
5) The coefficient of drag for a given flap configuration is a function of lift coefficient at this low a speed, and the lift coefficient is  a function of angle-of-attack. LIft and drag will be a function of the lift and drag coefficients, respectively, and the dynamic pressure. Runway friction will be a function of the normal force on the gear, which is the difference between the lift force and the airplane's weight. None of this is represented correctly in the methods used in this pr.
6) Airport elevation is only an important input parameter if obstacle clearance is going to be included as part of the takeoff performance check. If not, only airport pressure altitude (which should be calculated from QNH) and OAT are important. The same comments I made in the landing performance section regarding the source of inputs and the possibility for auto-inputs also apply here.
7) Takeoff performance accountability must also include wet runways. The method of calculating the wet runway takeoff distance differs somewhat from dry runway performance in that the effect of reverse thrust may be included, and the braking friction for a wet runway is different and is specified in the regulations.
8) Even without considering obstacle clearance, there are certain takeoff climb gradient requirements that must be met. There is no consideration to this in this pr.
9) Other requirements that must be met include maximum brake energy, and maximum tire speed limits.
10) And, what most people don't realize (and no flight sim takeoff calculator I have seen includes) is that landing performance at the destination must be checked before takeoff (using different data and assumptions from the in-flight landing data). This may result in a reduction in the maximum allowable takeoff weight (or prevent departing for that destination altogether). It would be really neat, realistic, and innovative if we included it in our takeoff perf calculator.

There needs to be a good discussion of what level of realism is expected/desired out of a takeoff performance calculator. Also, should it reflect real world performance, or should it represent the sim airplane's performance where the two might not be the same?  Adding the in-flight landing performance is pretty straightforward since the data is readily available. This is not the case for takeoff performance. I think we are either faced with either using CEO data incremented for approximate NEO performance differences as best we can estimate, doing real flight tests to generate data to put into more realistic takeoff models to generate performance, or do a more "gamified" estimate as was done here. The first two options will take a lot of time and effort, and the second one should not even be started until the anticipated new engine and flight models are finished.

A last comment is that a lot of people seem to place high importance on a takeoff performance calculator, but they don't typically have to face the possibility of having an engine failure on takeoff. So, after having written all of the comments above, maybe a viable option would be to present a "gamified" takeoff performance calculator that would identify if they could safely take off within the available takeoff distance with all engines operating (no failures) and give them a V1 from which they could stop within the available accelerate-stop distance, again with no failures, but just needing to reject the takeoff for any reason. It would be similar to what it proposed in this pr, but with a better set of numbers and consideration of the distance to takeoff (with a 115% safety margin), not just stop from V1.

matt-winfield:
>     2. It is intended that a safety margin be applied to this data. The recommended value of the safety margin is 1.15. Therefore, I recommend adjusting the calculated landing distances put out by the EFB to be multiplied by 1.15..
> 
>     3. The altitude adjustment is based on pressure altitude. The pressure altitude is not provided directly on the ATIS or other weather information available to the pilot, but the QNH value from which the pressure altitude can be calculated is. Therefore, I recommend that the EFP input for altitude be changed to QNH at the destination.
> 
>     4. The length of the portion of the runway that can be used for landing is the landing distance available (LDA), which can be different than the runway length (primarily due to displaced thresholds for obstacle clearance during the approach).  LDA should be used instead of runway length.
> 
>     5. Of the 3 distances, 2 are for the autobrake settings of medium and low. The other distance is for maximum manual braking. This should be made clear or there will be many comments that the max autobrake setting cannot be selected for landing. Also, while these distances will be realistic distances for the real airplane, it is my understanding that the sim does not yet fully simulate autobrake braking. Hopefully, that will come at some point.
> 
>     6. For some reason, not all of the different runway braking condition levels are included (specifically what are called the in-betweeners -- good-to-medium and medium-to-poor). Operators fought hard to have these included in this newly implemented process! There is no requirement to do so, but I suggest they be included.
> 
>     7. I suggest grouping the inputs by category -- for example, all the atmospheric inputs should be grouped together, all the runway-related inputs, and all the airplane-related inputs.

>     1. There is no (zero) relationship between V1 and stall speed. V1 is not a flying speed, so stall speed has no bearing on the selection of V1. Then to base whether or not the runway is long enough on whether or not the assumed relationship between V1 and stall speed exists is entirely wrong.

I've pushed some changes that address these, specifically:

- Added a 1.15 safety margin to landing distances
- Added QNH input which is incorported into the landing calculations
- Change "runway length" to "runway LDA"
- In a previous commit I updated the label to make it clearer that it's for manual braking
- Added Good-to-medium and Medium-to-poor runway conditions
- Rearranged inputs so similar data is grouped together


>     3\. The altitude adjustment is based on pressure altitude. The pressure altitude is not provided directly on the ATIS or other weather information available to the pilot, but the QNH value from which the pressure altitude can be calculated is. Therefore, I recommend that the EFP input for altitude be changed to QNH at the destination.

>     6\. Airport elevation is only an important input parameter if obstacle clearance is going to be included as part of the takeoff performance check. If not, only airport pressure altitude (which should be calculated from QNH) and OAT are important. The same comments I made in the landing performance section regarding the source of inputs and the possibility for auto-inputs also apply here.

Just a clarification on this, I believe when using QNH altitude is still required, because QNH is the pressure at sea level not the airport elevation. If we were using QFE we wouldn't need to consider altitude, but I believe the pressure altitude from QNH gives us an offset from actual altitude. Please correct me if I'm wrong about this.


>     8\. I think it will be difficult for most sim pilots to have the information to fill all the input fields. I strongly suggest that automatic inputs of as much of this information as possible be enabled. For example, the runway information like LDA, runway heading, and slope should be drawn from databases of such information, while other information such as landing runway, QNH, OAT, VAPP, and weight should be shared with the CDU. I don't know if sharing of such info with the CDU is entirely realistic or not, but other similar playability issues, like autofill of takeoff speeds and ZFW/ZFWCG in the CDU have been included.

I'd like to do this but I think it would be best done as a seperate feature, since this PR is quite large already and this isn't a requirement for the calculator to work.

matt-winfield:
>     2\. The thrust rating of an engine is the amount of thrust it can generate at sea level, zero airspeed on a test stand. It doe not include the installation losses associated with mounting it on an airplane. Net thrust varies with temperature beyond the flat rating, pressure altitude, and airspeed. It is entirely incorrect to treat it as a constant value equal to the thrust rating.

>     5\. The coefficient of drag for a given flap configuration is a function of lift coefficient at this low a speed, and the lift coefficient is  a function of angle-of-attack. LIft and drag will be a function of the lift and drag coefficients, respectively, and the dynamic pressure. Runway friction will be a function of the normal force on the gear, which is the difference between the lift force and the airplane's weight. None of this is represented correctly in the methods used in this pr.

The thrust rating was the closest value I could find to an accurate number for thrust produced, and it matches the data recorded in the simulator closely, so I think it's a good enough approximation for simulation purposes.

The coefficient of drag calculated is only an approximate value, but the calculated drag matches closely with the data recorded in the simulator, and does consider different flap configurations. The air pressure is also considered in the drag calculation.

You are correct that runway friction isn't considered. That might be something we can consider adding, however I don't think the friction force will be significant in comparison to the magnitude of the thrust/drag forces.


>     3\. Takeoff performance must consider both go and stop cases, and both all-engines-operating, and with one engine failing 1 second before V1. The accelerate-go cases must allow the airplane to reach a height of 35 feet at V2 within the takeoff distance available  (TODA) and a height halfway between liftoff and 35 feet within the takeoff run available (TORA). The takeoff performance approximation in this pr does not even consider the accelerate-go case.

This is correct, we consider accelerate-stop but not accelerate-go for V1. So this PR gives us the upper-bound for V1 but not the lower-bound (except as defined in the VMCG/VMCA table). This could be added in the future?

> There needs to be a good discussion of what level of realism is expected/desired out of a takeoff performance calculator. Also, should it reflect real world performance, or should it represent the sim airplane's performance where the two might not be the same?

I've been going off the project description of aiming to be "as close to reality as possible", so using real data when we can find it, if not fall back to simulator data. But I agree that we need to get concensus on this. I think if we say "it has to be so accurate you could use it for a real plane" then we will never get a performance calculator, but if it's "as close as we can get" that is more achievable.

matt-winfield:
>     7. Takeoff performance accountability must also include wet runways. The method of calculating the wet runway takeoff distance differs somewhat from dry runway performance in that the effect of reverse thrust may be included, and the braking friction for a wet runway is different and is specified in the regulations.

I would like to account for wet runways, but don't know what the exact numbers are for how this affects braking performance. For reverse thrust we would need to know how much thrust the engines produce in reverse.

If we can find how wet runways affect braking friction (e.g. brakes are 50% effective or similar) and reverse thrust output this will be easy to add.

matt-winfield:
> a viable option would be to present a "gamified" takeoff performance calculator that would identify if they could safely take off within the available takeoff distance with all engines operating (no failures) and give them a V1 from which they could stop within the available accelerate-stop distance, again with no failures, but just needing to reject the takeoff for any reason. It would be similar to what it proposed in this pr, but with a better set of numbers and consideration of the distance to takeoff (with a 115% safety margin), not just to be able to stop from V1.

Yeah I think a realistic target (for now) should be to produce a limited performance calculator at first, that takes into account the "big" factors to give you speeds where you can safely stop (plus a safety margin). And then gradually improve the accuracy of the speeds taking into account more factors (e.g. reverse thrust active, climb gradients, obstacle clearance etc.) over time.

donstim:
> Just a clarification on this, I believe when using QNH altitude is still required, because QNH is the pressure at sea level not the airport elevation. If we were using QFE we wouldn't need to consider altitude, but I believe the pressure altitude from QNH gives us an offset from actual altitude. Please correct me if I'm wrong about this.
> 
You're right. You need both QNH and airport elevation (not altitude) to get the airport pressure altitude.

donstim:
Once we have it sorted exactly what the takeoff perf analysis will include, if there is any computation done that needs CL, CD, thrust, and braking friction, I can help you get more realistic (sim) values for those. But be aware that there is ongoing work to completely redo the flight and engine models.

Maybe we start with the all engines operating case only like you mentioned on Discord. We would then have a V1Stop and a V1 Go (but would only display a single V1). If V1Go is higher than V1Stop (that is, there is are speeds between them where you can't go or stop within available distance), then that would be one indication that the runway is too short for your takeoff weight.

Then, maybe we could try to follow up with more complexity after that gets implemented.

arturkuma:
Hi,

I think we should reuse/modify existing components. Example:
new component <PerformanceInput /> can be replaced with <Input /> (Components/Form/Input). 

If for some reason it can't be reused, it's naming should be different like 'SimpleInput', because it should be possible to reuse it in other places, not only Performance page

matt-winfield:
> I think we should reuse/modify existing components. Example:
> new component `<PerformanceInput />` can be replaced with `<Input />` (Components/Form/Input).
> 
> If for some reason it can't be reused, it's naming should be different like 'SimpleInput', because it should be possible to reuse it in other places, not only Performance page

I've converted these new components into generic components. The select/dropdown and output display are completely new so those might be helpful elsewhere, the existing Input component styling didn't really fit for the "dense" amount of data being input on these pages, so I added the "simple" input as suggested.

matt-winfield:
I'm going to close this and split into 2 - Landing performance is pretty much ready to go so that will come first, then takeoff performance can come later with some accuracy improvements.

matt-winfield:
I've split the landing performance into PR #3581 so it can be reviewed seperately.

PR #3550, opened 2021-02-16
fix(rmp): workaround for active freq being zero

### LINKED ISSUE #3362: [BUG] Incorrect values on RMP at startup
**Mod Version**
Latest Developer Version from Installer
{
  "built": "2021-02-06T17:55:54+00:00",
  "ref": "master",
  "sha": "05c413683627fa74aea71646308fe02d60b1acc6",
  "actor": "Benjozork",
  "event_name": "manual"
}

**Describe the bug**
VHF1 on Captain Side and VHF2 on First Officer side are showing 0.00000 on Startup as the active frequency. 

**To Reproduce**

1. Turn on the batterys
2. Turn on the external Power
3. Turn on the RMP

**Expected behavior**
The RMP shows a valid active frequency.

**Actual behavior**
The active Frequency on VHF1 on Captain side should show 0.00000, the active Frequency on VHF2 on First Officer Side should show the same.

**References**
grafik

**Additional context**
On the ATC menue it shows 0.000 for COM1 and COM2. By setting another frequency on ATC menue the frequency on VHF1 and VHF2 on RMP is changing too

The issue came with one of the latest updates of the developer version yesterday

No, its not

1Revenger1:
Fixes #3362 

## Summary of Changes
Add workaround for Simvars `COM ACTIVE FREQUENCY:1` and `COM ACTIVE FREQUENCY:2` being set to 0 when loading in at the gate.

## Screenshots (if necessary)
Before:
image
After:
image

## Additional context
This issue appears to only occur when spawning in at the gate - I tried spawning in the air/approach/on the runway and they all worked fine. Other things I tried:
* I diffed the apron.FLT and runway.FLT files, and didn't find any obvious differences related to the RMP.
* Reading simvars from other instruments (ie the PFD), but `COM ACTIVE FREQUENCY` was still set to zero

I also compared the original code from before the React RMP PR was merged in, and there appears to be a [workaround for these values being set to 0](https://github.com/flybywiresim/a32nx/blob/d53d3f73245cddc50bbb82d2af1f27489188f466/A32NX/html_ui/Pages/VCockpit/Instruments/Airliners/A320_Neo/FDW/A320_Neo_FDW.js#L104-L115) from Asobo. This suggests that the issue is something deeper within the sim which can't be fixed through a FLT file (else they would've likely fixed it that way instead?)

## Testing instructions
1) Spawn at a gate
2) Enable ground power and check that the active frequency for VHF1, VHF2, and VHF3 are non zero
3) Spawn at runway
4) Check the active frequency for VHF1, VHF2, and VHF3

### DIFF
--- a/src/instruments/src/RMP/Components/RootRadioPanels.tsx
+++ b/src/instruments/src/RMP/Components/RootRadioPanels.tsx
@@ -1,0 +1,36 @@
+import React, { useEffect } from 'react';
+import { RootRadioPanel } from './BaseRadioPanels';
+import { useSimVar } from '../../Common/simVars';
+
+declare const SimVar; // this can also be replaced once /typings are available
+
+function initFrequencies() {
+    SimVar.SetSimVarValue('K:COM_RADIO_SET_HZ', 'Hz', 122800000); // Unicom
+    SimVar.SetSimVarValue('K:COM2_RADIO_SET_HZ', 'Hz', 121500000); // Guard
+    SimVar.SetSimVarValue('K:COM3_RADIO_SET_HZ', 'Hz', 131500000);
+    SimVar.SetSimVarValue('K:COM_STBY_RADIO_SET_HZ', 'Hz', 121950000);
+    SimVar.SetSimVarValue('K:COM2_STBY_RADIO_SET_HZ', 'Hz', 136975000);
+    SimVar.SetSimVarValue('K:COM3_STBY_RADIO_SET_HZ', 'Hz', 135475000);
+    SimVar.SetSimVarValue('A32NX_RMP_L_VHF2_STANDBY_FREQUENCY', 'Hz', 124275000);
+    SimVar.SetSimVarValue('A32NX_RMP_L_VHF3_STANDBY_FREQUENCY', 'Hz', 135475000);
+    SimVar.SetSimVarValue('A32NX_RMP_R_VHF1_STANDBY_FREQUENCY', 'Hz', 129675000);
+    SimVar.SetSimVarValue('A32NX_RMP_R_VHF3_STANDBY_FREQUENCY', 'Hz', 136975000);
+}
+
+export const RadioPanelsContainer = () => {
+    const [com1] = useSimVar('COM ACTIVE FREQUENCY:1', 'Hz', 1000);
+    const initFreqs = com1 === 0;
+
+    // Workaround for Asobo bug
+    // Frequency can be 0 when starting at the gate even when set in Apron.FLT
+    useEffect(() => {
+        initFrequencies();
+    }, [initFreqs]);
+
+    return (
+        <div className="rmp-wrapper">
+            <RootRadioPanel side="L" />
+            <RootRadioPanel side="R" />
+        </div>
+    );
+};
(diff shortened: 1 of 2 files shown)


Kimbyeoungjang:
change title to 
fix(rmp): Workaround for active freq being zero

ZigTag (approved):

marcman86:
Name: marcman86#4907
Date of testing: 17.2.2021
Version of the sim: 1.13.16.0
PR Tested: #3550
Tier of testing: 1
Changes to observe: RMP frequencys not zero
Testing technique: Tryout

1.) Spawned C&D at EDDP
2.) Started bat. and ext. power
3.) Checkes VHF1, VHF2 and VHF3 
4.) Spawned at EDDP Runway
5.) Started bat. and ext. power
6.) Checkes VHF1, VHF2 and VHF3

Issues: None

Overall Rating: Good

Conclusions of the testing: Works!

Hope this helps :-)

theomessin (changes requested):
You are checking whether the active frequency is zero at _every_ render. It'd be best if you pushed this to an initialise panel function which only gets called once (you can keep a hasInitialised state using hooks). In there you can just set initial values to all the frequencies, not just the currently active one (that way you can set useful frequencies, e.g. `122.8` and `121.5`).

> I also compared the original code from before the React RMP PR was merged in, and there appears to be a workaround for these values being set to 0 from Asobo. This suggests that the issue is something deeper within the sim which can't be fixed through a FLT file (else they would've likely fixed it that way instead?)

Nice catch 😄 at least we now know that this is caused by the sim and not the RMP implementation.

1Revenger1:
@theomessin with the latest commit, is this what you had in mind? I wasn't sure if I should do all frequencies or not.

I feel if we are setting at least the two active frequencies which are zero at spawn, we should probably set all of them just so that they're all in a consistent place.

wpine215:
Any update on this PR?

1Revenger1:
Oh, I'd been waiting for a response. Hadn't seen the reaction on my last comment. I'll work more on this later today.

1Revenger1:
I was doing some more testing, and I figured out a couple weird things.
1) The frequency from Apron.FTL is set immediately upon pressing "Fly now" in the main menu
2) The RMP code starts running about ~50% of the way loading
3) Asobo sets the frequency to 0 (or ~400000 according to Vpilot) about 10 seconds after the RMP code starts running.

I tried using just `useEffect(() => { //init }, [])` but that would initialize the frequencies before Asobo set gibberish.
Current solution sets the frequency once every time, or three times for every time that `initFreq` changes. This all occurs though while the plane is loading so no flickering should be seen.

DavidEGTE:
Category: QA Tester
Name: David#2177
Date of testing: 26/03/2021
Version of the sim: 1.14.6.0
PR Tested: #3550 
Tier of testing: 1

Changes to observe:
workaround for active freq being zero

Testing technique: Try-out

1. Followed PR instruction
2. Spawned C&D 
3. Spawned Runway

N/A
N/A
N/A
Steps to achieve the issue:
N/A

Overall Rating: Good
Conclusions of the testing: Works as described, GTG

joseph-tobin:
FYI the default number set 429496.... is awfully similar to the max theoretical JS array length of 4294967296 elements. This may not be helpful at all but thought I'd mention it.

davidwalschots:
This was fixed in #3980 by accident. It's unclear how this was fixed by that particular PR. My _guess_ would be that adding `electric_always_available = 1` to `systems.cfg` somehow ensured the value wouldn't reset to zero.

PR #3584, opened 2021-02-17
build(scripts): only prune if image isn't available

nistei:
## Summary of Changes
Only prune the docker system when the image isn't available which only happens if it's a first-time use or we are using a new image. The pull is done automatically by  `docker run` hence I removed the command.

## Screenshots (if necessary)
N/A

## References
N/A

## Additional context
N/A

Discord username (if different from GitHub): nistei#1362

## Testing instructions
Dev Test:
1. `docker system prune --filter label=flybywiresim=true -f`
2. Build the A32NX -> Should download image
3. Build again -> Should use existing image


tyler58546 (approved):
Works, LGTM.

PR #3600, opened 2021-02-18
fix(efb): handle 'useSimVar' usage in dev env

arturkuma:
## Summary of Changes
Mock simVar values in new useSimVar hook when using browser to work on it

Discord username (if different from GitHub): arturkuma#6639

## Testing instructions
Check if EFB works as before. Run in browser and check if it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "@tailwindcss/postcss7-compat": "^2.0.2",
     "aewx-metar-parser": "~0.10.1",
     "byte-data": "^19.0.1",
+    "lodash": "^4.17.20",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "react-leaflet": "~3.0.5",
--- a/src/instruments/src/Common/simVars.tsx
+++ b/src/instruments/src/Common/simVars.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { useContext, useEffect, useRef, useState } from "react"
 import { useInteractionEvents, useUpdate } from "./hooks"
+import { random } from 'lodash';
 
 /**
  * If the same SimVar or GlobalVar is requested in multiple places with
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "@tailwindcss/postcss7-compat": "^2.0.2",
     "aewx-metar-parser": "~0.10.1",
     "byte-data": "^19.0.1",
+    "lodash": "^4.17.20",
```
beheh, line 66: You really don't need lodash just to generate random values...

src/instruments/src/Common/simVars.tsx
```diff
@@ -31,11 +32,26 @@ const normalizeUnitName = (unit: UnitName): UnitName => {
...
+        case "string":
+            return "string";
         default:
             return unit;
     }
 }
 
+const generateMockValue = (unit: UnitName) => {
+    return ({
+        'bool': !!random(0, 1),
```
beheh, line 45: Just use Math.random instead.

beheh (changes requested):
So as far I understand is not even for any in-sim behaviour, but to enhance the out-of-sim in-browser development that only applies to the EFB itself thus far.

Outside of some basic issues with this PR, like pulling in lodash for only one function that is part of the standard library, I don't agree with the strategy used here. You're modifying the actual provider that is going to be on a very critical path performance-wise to enable development. I don't think the SimVarProvider should be modified at all here.

How about we export the InternalProvider instead, and build an entirely separate "`RandomValueProvider`" or something similar that you can use in your custom in-browser EFB? That way we don't have to make every single panel like FCU, RMP etc. have this dev logic and you can provide whatever values you want. This `RandomValueProvider` would then just install some kind of dummy setters and getters, and we can go from there.

package.json
arturkuma: It'll be added in my other PR anyway

src/instruments/src/Common/simVars.tsx
arturkuma: Math.random works completely differently

arturkuma:
Yes agree. That supposed to be hotfix, but I'll implement that in proper way in free time

src/instruments/src/Common/simVars.tsx
beheh: `Math.random()` returns a random decimal between 0 and 1, so if you want a random number between 0 and 1 just `Math.round(Math.random())`.
If you need a random number between 0 and 5000 do `Math.round(Math.random() * 5000)`. 
If you need a random number between 100 and 250 do `100 + Math.round(Math.random() * 150)`.

arturkuma: Yep I know , but what's the point of repeating that instead of using simple function . That only decreases readability

Benjozork:
Is this still being worked on ?

src/instruments/src/Common/simVars.tsx
tracernz: Actually I think this is less readable than a Math.random equivalent because it's not clear that random(0, 1) returns an int rather than a float unless you're familiar with lodash's API. On first inspection it looks like you've created a generator that almost never returns false.

PR #3583, opened 2021-02-17
feat: custom transponder font

tyler58546:
## Summary of Changes

* Added custom transponder font

## Screenshots (if necessary)

image

## References

image

## Additional context

Discord username (if different from GitHub):

## Testing instructions

Make sure the font is accurate.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build script will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, click on the **Artifacts** drop down and click the **A32NX** link

PR #3672, opened 2021-02-21
feat: brake gauge visual improvements

tyler58546:
## Summary of Changes

* Changed brake gauge font to Poppins Medium (Futura alternative)
* Adjusted position and size of text on brake gauge
* Improved brake gauge markings

## Screenshots (if necessary)

Before:
image
After:
image

## References

image
image

## Testing instructions

Make sure the changes are accurate.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build script will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, click on the **Artifacts** drop down and click the **A32NX** link

PR #3675, opened 2021-02-21
refactor: React transponder

### LINKED ISSUE #1127: [BUG] Functionality of the ATC “CLR” key
**Mod Version**

0.3.1

**Describe the bug**

With the CLR key today just one digit at a time is cleared and replaced with a dash.

**To Reproduce**

**Expected behavior**

The correct functionality of the ATC “CLR” key is as follows:
Press The CLR key one time delete the last entered number.
Press a second time deletes all entered numbers.
There should not be dashes in stead of numbers when cleared.
If not a full set of four numbers are inserted within a certain time (not sure but estimate to 5 seconds) the previous code will be displayed again.

Note: The code “2000” is a normal code to have at startup. It is often selected just prior going from Auto/On to Stand By.

**Actual behavior**

With the CLR key today just one digit at a time is cleared and replaced with a dash.

**References**

I am a IRL A320 pilot.
From FCOM:
CLR Key
The flight crew uses this key to clear the code display.
Note: As long as the four figures of the new code are not entirely written, the previous code remains.

**Additional context**

Never worked before

Yes

lukecologne:
Fixes #1127

## Summary of Changes

- [x] Reimplement the transponder in react.
- [x] Fix the issues described in #1127
- [x] Change the font to the RMP font (they are the same). This replaces #3583.
- [x] Refactor xml for the ATC panel

## Additional context

## Testing instructions

* Make sure the transponder screen functions properly
* Perform long flights to ensure no performance drops


src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    // Check if the transponder code simvar has changed. If it has, and we are not currently editing the code,
+    // set the display to the new code.
+    useEffect(() => {
+        if (currentEnteredDigits === 4) {
+            const newDigits = getDigitsFromCode(transponderCode);
+            if (newDigits.length === 4) {
+                setCurrentDigits(newDigits);
+            }
+        }
+    }, [transponderCode]);
```
beheh, line 42: This useEffect? It's just gone now! No reason to synchronize vars any more. When the SimVar changes and we're not editing it'll automatically be applied. If we're editing it will be silently updated in the background without affecting our "draft" digits, and if the "draft" is then discarded (e.g. through timeout) we automatically switch to the new version, staying fully in sync with the SimVar.

```suggestion
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        for (let i = 0; i < currentEnteredDigits; i++) {
+            text += currentDigits[i].toString();
+        }
+    }
+
+    return (
+        <svg>
+            <text x="8%" y="45%">{text}</text>
+        </svg>
+    );
```
beheh, line 123:
```suggestion
    return (
        <svg>
            <text x="8%" y="45%">{ltsTest ? '8888' : codeInDisplay.join("")}</text>
        </svg>
    );
```

This can be significantly simplified 😉

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        codeCopy = Math.floor(codeCopy / 10);
+    }
+    digits.reverse();
+    return digits;
+};
+const getBco16FromDigits = (digits: number[]): number => (digits[0] * 4096) + (digits[1] * 256) + (digits[2] * 16) + digits[3];
+
+const PoweredXpdrDisplay = () => {
+    const [currentDigits, setCurrentDigits] = useState([0, 0, 0, 0]);
+    const [currentEnteredDigits, setEnteredDigits] = useState(4);
```
beheh, line 26: So with this component I can already see the issue already on the first line.

Ultimately you're duplicating some kind of source of truth (the transponder sim var) into local state. This is not required, and is generally annoying because now suddenly you've made yourself as the programmer here responsible to synchronize things like the SimVar changing from the outside into your local state.

Your life will be a lot easier, if you just use this state variable for the edit case, with something like this:

```suggestion
    const [newDigits, setNewDigits] = useState<null | number[]>(null);
```

`null` would indicate "no edit is in progress", and then further down you can just use the SimVar itself. In fact, you could put that right here, and we've basically already got the entire display logic done:

```suggestion
    const [newDigits, setNewDigits] = useState<null | number[]>(null);
    const [transponderCode, setTransponderCode] = useSplitSimVar('TRANSPONDER CODE:1', 'number', 'K:XPNDR_SET', 'Bco16', 250);
    const codeInDisplay: number[] = newDigits !== null // if we're in the process of editing...
      ? newDigits // ...just use the WIP digits, otherwise...
      : getDigitsFromCode(transponderCode) // ...just get the digits from the SimVar
```

Now you can think of `newDigits` as the "draft version". If it exists we show it, otherwise just use the known good version. We also don't need to worry about `currentEnteredDigits` in state any more - we can just figure them out by looking at newDigits being null or not.

You'll see below how this simplifies your life! 😃

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        }
+        if (displayResetTimer > 0) {
+            setDisplayResetTimer(displayResetTimer - deltaTime / 1000);
+        }
+    });
+
+    // Reset the double press timer, if the CLR button was not pressed a second time.
+    if (doubleClrTimer < 0 && doubleClrTimer !== -1) {
+        setDoubleClrTimer(-1);
+    }
```
beheh, line 57: So putting state setters into render is notforbidden per se, but it can be weird to think about exactly when and how they trigger. As a general tip, I would suggest to just avoid this pattern if you can for now. I suggest this:

```suggestion
   useEffect(() => {
      if(doubleClrTimer < -1) {
         setDoubleClrTimer(0);
      }
   }, [doubleClrTimer]);
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    // Check if the reset timer has run up. If it has, reset the display to the previously entered transponder
+    // code, and reset the timer.
+    if (displayResetTimer < 0 && displayResetTimer !== -1) {
+        setDisplayResetTimer(-1);
+        const newDigits = getDigitsFromCode(transponderCode);
+        if (newDigits.length === 4) {
+            setCurrentDigits(newDigits);
+        }
+        setEnteredDigits(4);
+    }
```
beheh, line 72: These two could just be rewritten into a single useEffect, that hopefully reads a lot simpler:

```suggestion
    useEffect(() => {
        if(displayResetTimer === -1) {
            // timer is not running - check if it should start
            if (newDigits !== null) {
                setDisplayResetTimer(5)
            }
        }
        else {
            // timer is running - check if it has elapsed
            if (displayResetTimer < -1) {
                setNewDigits(null); // clear the entry
                setDisplayResetTimer(-1);
            }
        }
    }, [newDigits, displayResetTimer]);
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }, [transponderCode]);
+
+    // Count down the timers, if they are in use
+    useUpdate((deltaTime) => {
+        if (doubleClrTimer > 0) {
+            setDoubleClrTimer(doubleClrTimer - deltaTime / 1000);
+        }
+        if (displayResetTimer > 0) {
+            setDisplayResetTimer(displayResetTimer - deltaTime / 1000);
+        }
```
beheh, line 51: Let's ensure we don't accidentally hit -1 ourselves:

```suggestion
        if (doubleClrTimer > 0) {
            setDoubleClrTimer(timer => Math.max(timer - deltaTime / 1000, 0));
        }
        if (displayResetTimer > 0) {
            setDisplayResetTimer(timer => Math.max(displayResetTimer - deltaTime / 1000, 0));
        }
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        setDisplayResetTimer(-1);
+        const newDigits = getDigitsFromCode(transponderCode);
+        if (newDigits.length === 4) {
+            setCurrentDigits(newDigits);
+        }
+        setEnteredDigits(4);
+    }
+
+    const buttonPressHandler = (buttonId) => {
+        // Check if the pressed button is the reset button, and if ther is a digit displayed.
```
beheh, line 75: Typo.

```suggestion
        // Check if the pressed button is the reset button, and if there is a digit displayed.
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        if (buttonId === 8 && currentEnteredDigits > 0) {
```
beheh, line 76: Let's adapt this to our new variable:

```suggestion
        if (buttonId === 8 && newDigits !== null) {
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+            // If the clear button has not been pressed previously, just delete the rightmost digit.
+            if (doubleClrTimer === -1) {
+                setEnteredDigits(currentEnteredDigits - 1);
+                setDoubleClrTimer(0.2);
+                // If it has been pressed previously, clear the whole display.
+            } else {
+                setEnteredDigits(0);
```
beheh, line 83: Now prepare for the new digits:

```suggestion
                setNewDigits([]);
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        }
+        setEnteredDigits(4);
+    }
+
+    const buttonPressHandler = (buttonId) => {
+        // Check if the pressed button is the reset button, and if ther is a digit displayed.
+        if (buttonId === 8 && currentEnteredDigits > 0) {
+            // If the clear button has not been pressed previously, just delete the rightmost digit.
+            if (doubleClrTimer === -1) {
+                setEnteredDigits(currentEnteredDigits - 1);
```
beheh, line 79: We should probably handle the empty case:

```suggestion
                if(codeInDisplay.length) {
                    setNewDigits(codeInDisplay.splice(0, codeInDisplay.length - 1));
                }
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+                setDoubleClrTimer(0.2);
+                // If it has been pressed previously, clear the whole display.
+            } else {
+                setEnteredDigits(0);
+                setDoubleClrTimer(-1);
+            }
+            // If the pressed button is not the CLR button, and if not all digits have been entered yet,
+            // write to the next free digit.
+        } else if (buttonId !== 8 && currentEnteredDigits < 4) {
```
beheh, line 88: Update for new variables: Either none have been entered yet, or less than 4:

```suggestion
        } else if (buttonId !== 8 && (newDigits === null || newDigits.length < 4)) {
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+            const digitsToSet = currentDigits;
+            digitsToSet[currentEnteredDigits] = buttonId;
+            setEnteredDigits(currentEnteredDigits + 1);
+            // If 3 digits are currently entered (i.e. if this digit will fill the display), write to the simvar.
+            if (currentEnteredDigits === 3) {
+                setTransponderCode(getBco16FromDigits(currentDigits));
+                setDisplayResetTimer(-1);
+            }
```
beheh, line 96: Simplified and new variables:

```suggestion
            if(newDigits === null) {
                // currently still showing the normal transponder code - let's start with the first digit:
                setNewDigits([buttonId]);
            }
            else if (newDigits.length == 3) {
                // we are on our last digit - clear the draft and set the SimVar itself
                setNewDigits(null);
                setTransponderCode(getBco16FromDigits(newDigits)); // new digits is still available here until the next render
                setDisplayResetTimer(-1);
            }
            else {
                // we're on digit 0-3
                setNewDigits([...newDigits, buttonId]);
            }
```

beheh (commented):
Alright, I've given the main component a full-on documented rewrite, explaining the patterns I would suggest using. I haven't tested my suggestions directly in-sim, but I believe this should work or be very close to working. I'm happy to chat about this if you want.

lukecologne:
I Implemented your suggestions, and after a few small changes, it works like expected. But now I have the problem that the display flickers when you enter the last value, since the display immediately switches to the simvar, which does not seem to be updated instantly (even with maxStaleness at 0)

beheh:
Hm, so that *should* be already handled because the SimVar hooks have an internal cache they should update immediately, regardless of refresh rate. It's possible though that we poll and replace that cached, new value, with the intermediate old value, so the sequence would be:

1) setSimVar(newValue) => results in newValue
2) a sim var poll happens => results in oldValue
2) anoth sim var poll happens => results in newValue again

Could that be the case here? A higher refresh rate should temporarily fix that because it polls less often, but I believe @theomessin probably had the same issue over in the RMP, so I wonder how he solved that there. We could add some additional logic to the sim var framework where we intentionally reject the outdated sim var values from the sim for a certain time after a new value has been set.

theomessin:
I think that setting the `maxStaleness` to a non-zero value should be the recommended solution. The whole point of `maxStaleness=0` is that we are **consciously** fetching the simvar value from the sim at **every** update. I think we should just have a sensible non-zero default value as opposed to adding complexity with a cache ttl

lukecologne:
Pushed the changes right now, if you want to take a look

theomessin:
By the way @lukecologne, what setting are you using for your glass cockpit refresh rate?

lukecologne:
Medium. Just tried with a maxStaleness of 500, doesn't help

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        codeCopy = Math.floor(codeCopy / 10);
+    }
+    if (digits.length < 4) {
+        for (let i = 0; i <= 4 - digits.length; i++) {
+            digits.push(0);
+        }
+    }
+    digits.reverse();
+    return digits;
+};
```
theomessin, line 21:
```suggestion
const getDigitsFromCode = (code: number): number[] => {
    return code.toString().padStart(4, '0').split('').map((c) => parseInt(c));
};
```

src/instruments/src/ATC/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const codeInDisplay = newDigits !== null ? newDigits : getDigitsFromCode(transponderCode);
+
+    useUpdate((deltaTime) => {
+        if (doubleClrTimer > 0) {
+            setDoubleClrTimer(Math.max(doubleClrTimer - deltaTime / 1000, 0));
+        }
+        if (displayResetTimer > 0) {
+            setDisplayResetTimer(Math.max(displayResetTimer - deltaTime / 1000, 0));
+        }
+    });
```
theomessin, line 40: Might be worth having a discussion on using setTimeout for something like this. It's probably the perfect example where the code would be much cleaner and it would probably work just as well (https://coherent-labs.com/Documentation/cpp-gt/d9/d28/time_control.html).

theomessin:
Your double click state resets too quickly, I clicked clear at the same rate as in the the video and it erased two digits, not all four: https://discord.com/channels/738864299392630914/758149448399847424/813012021725495296

beheh (approved):
Looks good with the changes now!

lukecologne:
> 
> 
> Your double click state resets too quickly, I clicked clear at the same rate as in the the video and it erased two digits, not all four: https://discord.com/channels/738864299392630914/758149448399847424/813012021725495296

I will increase the time from 200ms to 300ms, hopefully that will help a bit

Samueleonard:
Category: QA Trial
Name: 1samueleonard1#0001
Date of testing: Current date 22/02/2021
Version of the sim: 1.13.16.0
PR Tested: #3675
Tier of testing: 1
Changes to observe: Verify that transponder works as it should
Testing technique: Tryout

1. load in, egkk, default livery, live time/weather
2. flick all switches, make sure they still have range of motion - they do but check notes
3. log onto vatsim as observer
4. check mode selector works with vpilot mode C - auto/ta ra - it does
5. play around with squawk code - check deleting and entering works (full clear and individual digit clear) - works as expected
6. coordinate with controller to make sure squawk code transmits to vatsim - works

N/A

image - alt rptg does not point to 'on' (probably nothing to do with this pr)

.when loading in on rwy, xpndr alt reporting is still in stby, should be auto or on
. cannot turn 'atc' selector (middle switch on the left)

Steps to achieve the issue
.spawn on runway

Overall Rating: Good

Conclusions of the testing: Make a small descriptive to areas where it needs to be improved and recommendations.
functionality wise, fine but would need testing over longer periods to make sure that there are no memory leaks/performance issues

lukecologne:
> Issues (If Any): Write any issues
> .when loading in on rwy, xpndr alt reporting is still in stby, should be auto or on
> . cannot turn 'atc' selector (middle switch on the left)
> . font seems slightly off, seems too large to me, but potentially correct, i havent checked properly
> 
> Steps to achieve the issue
> .spawn on runway
> 
> Overall Rating: Good
> 
> Conclusions of the testing: Make a small descriptive to areas where it needs to be improved and recommendations.
> functionality wise, fine but would need testing over longer periods to make sure that there are no memory leaks/performance issues

@Samueleonard Thanks for your report! The model/decal issues were not changed by this PR, it was like this before. The transponder selector (1 and 2) were not functional before, so no problem there as well. And about the font size, yes it's too large, but the problem is that the screen has wrong dimension in general, this will have to be fixed when we can edit the model. So until then I just made the font fill the screen. About the ALT RPTG, I didn't change anything in regards to default state, so it must have been like this before

BritishAvgeek:
Category: QA Tester
Name: BritishAvgeek#1436
Date of testing: 24/02/2021
Version of the sim: 1.13.16.0
PR Tested: #3672
Changes to observe: Transponder changes

Testing technique: Check transponder functionality in a full flight
1.  Spawned C&D at NZWN - set up for flight to NZQN
2.  Set up MCDU, checked XPDR - at spawn was set to 7000, changed it to multiple entries prior to stream going live, but set to 2000 during stream for VATSIM IFR on Unicom
3. Clicked all the buttons
4. Flew the full flight
5. Landed (turns out Queenstown is missing a LOT of Nav Data so went in blind and did a Flap 3 Landing)

Flight notes (When done a full flight): Route Used:
POLA5P SABDA Y622 MESIX H159 CH Y266 ELRUV ELRU4B

Media: Livestreamed here https://www.youtube.com/watch?v=dy_WWYZh4KE

Issues (If Any): None

Overall Rating: Good
Conclusions of the testing: Noted font size mention further up so ignored that, all worked nicely and had no issues with anything else in the aircraft or performance during the 1hr flight. The PR worked well for me.

FlyingCarpet7:
Tested now! Just a few adjustments to make.

Second push of CLR key in sequence will delete all numbers. The 300 ms interval should **not** be there. E.g if you CLR last number you have 7 seconds (see below) to press CLR a second time and then all numbers will be deleted.

The 5 seconds you can increase to **7 seconds**. That feels more close to the real now when I played around a bit with it. This 7 sec timer need to be **restarted** with **every** key pressed (numbers or clear key) as long as a new complete four digit code is not fully entered.

Sorry for not making this description fully clear from the beginning. Hope it is a easy fix.

lukecologne:
@FlyingCarpet7 Implemented your feedback. If you have time, could you please test again and check if it's better now?

FlyingCarpet7:
Tested again! Great!
Just found one small thing if you want to.
Lest say I want to input code 1234
By mistake I start with 125, then delete the 5 with one push of CLR. Then again by mistake insert a 6 so the entered number now is 126. If I press CLR one time now all three digits are now deleted instead of just the 6.
Tiny detail, otherwise the logic now is superb!
Pity we can not update the display size just yet.

marcman86:
Category: QA Tester
Name: marcman86#4907
Date of testing: 25.2.2021
Version of the sim: 1.13.16.0
PR Tested: #3675
Tier of testing: 2
Changes to observe: TCAS
Testing technique: Tryout

1.) Spawned C&D at EDDP
2.) Start up the bus
3.) Took off and did a small flight (1h)
4.) Played with TCAS

- Annoying that the number gets autofilled now that the numberpad works so nicely
- ATC reacts to TCAS standby and wrong number
- Performence was fine
- Font looks good

Issues: -

Overall Rating: Good

Conclusions of the testing: Works

Hope this helps :-)

St54Kevin:
Category: QA Tester
Name: St54Kevin#3583
Date of testing: 26/02/2021
Version of the sim: 1.13.16.0
PR Tested: #3675 
Tier of testing: 1
Changes to observe: Transponder
Testing technique: 

1. Loaded Up C&D, went through cockpit prep
2. checked functionality of transponder
3. Tried different way of inputs and CLR behavior

Media:
image

Issues (If Any): The font looks very weird. Way too big (I know the model is wrong). But somehow I find it to be out of place

Overall Rating: LGTM

Conclusions of the testing: Sry did not have time for a VATSIM flight, therefore just a Tier 1 test. On ground everything works as expected

beheh:
The font type and size changes were mentioned a few times now. That change was originally made in https://github.com/flybywiresim/a32nx/pull/3583 and then intentionally picked up in this PR instead.

I understand that the dimensions are wrong and we just went with filling the screen for now. However looking at the original reference I think a big part of the feel of the font is the centering with horizontal space left and right and the digits only being marginally higher than the buttons.

@lukecologne How do you feel about meeting in the middle and making the font a little smaller, but keeping it centered to retain the feel of the original RMP? I think this PR is outstanding otherwise, but I foresee a bunch of bug reports and complaints if we merge this as-is from a purely visual basis.

FlyingCarpet7:
_fix: transponder codes smaller than 1000 getting ignored when loaded_ 

Does this fix mean that we can not insert a code like 0123 or 0045?
I think all octal number combinations can be assigned by ATC?

lukecologne:
> 
> 
> _fix: transponder codes smaller than 1000 getting ignored when loaded_
> 
> Does this fix mean that we can not insert a code like 0123 or 0045?
> I think all octal number combinations can be assigned by ATC?

No, it's the reverse. Before this commit you couldn't do that, so I fixed it.

lukecologne:
Alright, slightly decreased the font size, while keeping it centered. 
grafik

St54Kevin:
> Alright, slightly decreased the font size, while keeping it centered. 
> grafik
> 

Yeah, that's an improvements. Thanks

BritishAvgeek:
Category: QA Tester
Name: BritishAvgeek#1436
Date of testing: 27/02/2021
Version of the sim: 1.13.16.0
PR Tested: #3675
Tier of testing: 2
Changes to observe: Transponder tweaks re: above, amendments to font size

Testing technique: Spawned C&D - set up and fiddled with Transponder... then spawned mid air and checked function there too
1.  Spawned C&D - checked Transponder operation - set initially to 7000, changed to 2000 and a few other "values"
2.  Pressed Clear to remove the digits one by one to check that function
3.  Pressed clear twice to clear the entire value - worked fine
4. Flew a short 50min flight into Tivat - no issues re: performance
5. After landing, I spawned in mid air to check the "live" spawn operation of it - same process as above repeated, no issues

Flight notes (When done a full flight): No issues, nothing to report, no performance degradation noticed

Media:
Microsoft Flight Simulator 27_02_2021 14_19_44
Microsoft Flight Simulator 27_02_2021 14_19_52

Issues (If Any): None

Overall Rating: Good
Conclusions of the testing: No issues found, this is the second test I have conducted on this PR in particular noting the updates/tweaks above over the last few days - all good, thanks for your work!

PR #3685, opened 2021-02-22
BUG(efb): enter username, keys always active on the plane

banzzaaiii:


Benjozork:
This is probably a mistake.

Also, that bug is known.

PR #3758, opened 2021-02-28
build(eslint): switch to @flybywiresim/eslint-config

Benjozork:
## Summary of Changes

* Uses `@flybywiresim/eslint-config`  for eslint


beheh:
So I'm not quite sure why, but airbnb-eslint suggest that you explicitly add the plugin etc. dependencies to the project itself. I'd be curious to know why they do that and why we don't need to (and use dependencies instead).

Benjozork:
@beheh in my limited testing and on CI it seems to work. I think it should be fine

PR #3780, opened 2021-03-02
fix(docs): Change CFBW download link in Readme

FoxtrotSierra6829:
## Summary of Changes
Changed CFBW download link in Readme, so it does not point to the old CDN

## Additional context
CDN was changed and old is disabled now

Foxtrot Sierra#6420

## Testing instructions
None


pepperoni505 (approved):
LGTM

PR #3782, opened 2021-03-02
feat: First hydraulic system implementation

crocket63:
This feature request is linked #2840 but not fixed yet by this PR.
This PR is a based on PR #2767, and can replace it 

## Summary of Changes

Implementation of A320neo's hydraulic system using Rust and WASM.

First thing first, many thanks to:
@davidwalschots for EXTENSIVE support
@wpine215 for launching the first stage of hydraulics (and extensive support)
@CPUBlanc for starting it
@komp1821 and @avteknysian For HUGE support on all kinds of technical details
Everyone helping me getting up to speed on all different subjects in this project

First ever PR for me, and no previous knowledge on various subjects, so be patient I'll improve lot of stuff or someone will :)

**Summary**
This is a difficult task, and can be done at any level of details. The global target for this is aiming at the extreme limit of every details that can be implemented, without needing overkill performance, or overkill details that would bring nothing to the experience. 
Further analysis on the current performance hit of this subsystem will later bring us to upgrade / downgrade details.
You won't see the internal parts of a pump moving, but you should see the landing gear swinging back and forth while retracting using only yellow electric pump and PTU providing pressure to green loop.

**Important note**
This first PR brings the first brick of hydraulics, and has NO INTENTED impact on FS2020 physics. Meaning actuators are not connected, brakes shows accordingly in pressure gauges but DO NOT control the braking force of the plane. Brake accumulator  empty and parking brake shows 0 pressure. But FS2020 would still be applying the brakes. Or if default FS2020 hydraulics are off, plane would not brake while you have actual brake pressure available from accumulator.
Of course everything will get connected later step by step.
Next PRs will focus on documentation and code refactoring if needed (and lot is needed)

**Implementation overview**
Now the fun part:
This is NOT a simulation of internal pumps parts, plungers, solenoid valves.
This is NOT a simulation of all hydraulic pipes connection solving navier stokes equations on each tube node
This is NOT a faked hydraulic simulation that increments pressures on screen until it reaches 3000 psi and decrement it when pump is off.

Every pump has a dynamic displacement logic. It takes fluid from a reservoir and throw it into its hydraulic loop. The pump regulation is for now "perfect", but can be simulated in real time basically in one line of code: this only depends on performance concerns and can evolve in the future. First implementation (and that was not easy...)  tries to model all of this at a low fixed rate of 10Hz, higher rates could enable us to add some more regulation details in the future.

[...]

PR #3833, opened 2021-03-07
feat: (ecam) Add Hydraulics ECAM page

RichardPilbery:
Fixes #[issue_no]

## Summary of Changes

Compliment to existing PR (#3782) implementing the hydraulics system on the A32NX. This is the ECAM page for the hydraulics system, written in React. Note that this PR should not be merged before #3782 

## Screenshots (if necessary)

ECAM-HYD_ELEC-PB

ecam-hyd-v4

## References

HYD-page-reference

ECAM-HYD-INFLIGHT

## Additional context

Discord username (if different from GitHub): tricky_dicky[z+0]


wpine215:
Only thing that stands out to me at the moment is that the "HYD" title seems to be positioned a bit lower than it is IRL

RichardPilbery:
> Only thing that stands out to me at the moment is that the "HYD" title seems to be positioned a bit lower than it is IRL

Yeah, but if it is moved up any higher, you lose the top of the letters. It's on my things to investigate list.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -25,25 +25,42 @@ import { SimVarProvider, useSimVar } from '../../../Common/simVars';
...
-    const [yellowFireValve] = useSimVar('L:A32NX_HYD_YELLOW_FIRE_VALVE_OPENED', 'boolean', 1000);
+    const [greenFireValve] = useSimVar('L:A32NX_HYD_GREEN_FIRE_VALVE_OPENED', 'boolean', 500);
+    const [yellowFireValve] = useSimVar('L:A32NX_HYD_YELLOW_FIRE_VALVE_OPENED', 'boolean', 500);
+
+    const [greenPumpActive] = useSimVar('L:A32NX_HYD_GREEN_EDPUMP_ACTIVE', 'boolean', 500);
+    const [yellowPumpActive] = useSimVar('L:A32NX_HYD_YELLOW_EDPUMP_ACTIVE', 'boolean', 500);
+    const [bluePumpActive] = useSimVar('L:A32NX_HYD_BLUE_EPUMP_ACTIVE', 'boolean', 500);
+
+    const [greenPumpLowPressure] = useSimVar('L:A32NX_HYD_GREEN_EDPUMP_LOW_PRESS', 'boolean', 500);
+    const [yellowPumpLowPressure] = useSimVar('L:A32NX_HYD_GREEN_EDPUMP_LOW_PRESS', 'boolean', 500);
```
crocket63: Meant YELLOW_EDPUMP here? :)

RichardPilbery: Affirm. Good spot.

RichardPilbery: Actually, this has already been fixed. Odd that it is still showing for you...

crocket63: Better this way than the other :D  Probably did a dumb click again :D

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -0,0 +1,767 @@
...
+
+    const [greenPressure] = useSimVar('L:A32NX_HYD_GREEN_PRESSURE', 'psi', 500);
+    const [yellowPressure] = useSimVar('L:A32NX_HYD_YELLOW_PRESSURE', 'psi', 500);
+    const [bluePressure] = useSimVar('L:A32NX_HYD_BLUE_PRESSURE', 'psi', 500);
+
+    const [greenPumpPBStatus] = useSimVar('L:A32NX_OVHD_HYD_ENG_1_PUMP_PB_IS_AUTO', 'boolean', 500);
+    const [yellowPumpPBStatus] = useSimVar('L:A32NX_OVHD_HYD_ENG_2_PUMP_PB_IS_AUTO', 'boolean', 500);
+    const [bluePumpPBStatus] = useSimVar('L:A32NX_OVHD_HYD_EPUMPB_PB_IS_AUTO', 'boolean', 500);
+
+    const [yellowElectricPumpStatus] = useSimVar('L:A32NX_OVHD_HYD_EPUMPY_PB_IS_AUTO', 'boolean', 500);
```
crocket63, line 40: My feeling is that pump status should directly come from the XPUMP_ACTIVE simvars. For example if any logic starts a pump without changing button status, we won't see it active.  If yellow epump is activated by cargo door operation from ground crew, it won't appear on ecam currently. Is it expected?

RichardPilbery: For the Yellow electrical pump you are right. However, the status of the push button is important too.

crocket63: Totally agree, but my point is that I already evaluate the button and compute the pump activation logic inside the WASM system. So I think there is no point doing it the same at different places UNLESS the ecam logic is different from the actual pump logic. 
Probably need to ask for this on discord :)
As an example, you should also check for blue pump override to have it showing on the ECAM when pressed, and this is already done and handled in the WASM module.  Or at least if it's done wrong e only have to update it in one place :)

wpine215 (approved):

wpine215:
Needs testing instructions then will be ready for testing.

crocket63:
> Needs testing instructions then will be ready for testing.

Is it actually testable without hydraulics PR?  In a way it should, as issues could come from hydraulics and not the ecam stuff, but not sure how to do so?

RichardPilbery:
> > Needs testing instructions then will be ready for testing.
> 
> Is it actually testable without hydraulics PR? In a way it should, as issues could come from hydraulics and not the ecam stuff, but not sure how to do so?

I don't think it is. Let's get the hydraulics PR merged and then we can test this.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -0,0 +1,788 @@
...
+                setPressures(true);
+                setPtuActive(0);
+            }
+        } else if (ptuAvailable && yellowElectricPumpStatus && greenPressure <= 2990) {
+            setPtuScenario('right-to-left');
+            setPtuActive(1);
+        } else {
+            setPtuScenario(ptuAvailable ? 'normal' : 'PTU-off');
+        }
+    }, [greenPressure, yellowPressure, yellowElectricPumpStatus]);
```
davidwalschots, line 162: **thought**: I hope @crocket63 will provide enough data to not require these calculation in the presentation layer. We should strive to keep the React side as lean as possible.

crocket63: There's actually specific ecam logic that is not aware of the PTU state. If we go full simulation that logic might be ok to be here.

crocket63:
I noticed one thing when testing: when first opening HYD page, you can see the page going from init state to actual state. Can it be kind of "refreshed" before it's actually shown?

wpine215:
#3782 has now been merged.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
crocket63: If there's a true performance issue, it wouldn't be too difficult to process the exact same logic runing here on rust side just outputting the required ecam boolean A32NX_HYD_ECAM_GREEN_AVAILABLE  A32NX_HYD_ECAM_PTU_G_2_Y_ACTIVE etc....

PR #3946, opened 2021-03-12
docs(github): remove title from issue templates

davidwalschots:
## Summary of Changes
By removing the title from the issue templates we solve the following problems:
- Duplicate information is removed (i.e. no more [BUG] prefix and Bug label).
- One can't simply hit the "Submit new issue" button, as they now must fill in a title in order to be able to submit something.
- One no longer has to first remove the "INSERT TITLE HERE" part of the title.

## Additional context


MisterChocker (approved):
LGTM

PR #4019, opened 2021-03-18
chore(eslint): turn linebreak-style rule off

davidwalschots:
## Summary of Changes
As we already have automatic CRLF handling (#1927) there's no need to have AirBNB's linebreak-style rule. We can simply turn it off and all is well.


aguther (approved):
LGTM

PR #4063, opened 2021-03-21
feat: elec system sounds, engine changes, and better ambience

hotshotp:
## Summary of Changes

I have changed many many sounds in this PR, I will list them here:

-I have changed the avionics fan sound to be more realistic using higher quality recordings from komp
-Changed the engine start sounds to have a more realistic sounding fuel introduction for a cold start (when the crank phase is skipped, it introduces fuel at a lower n2 percentage which i simulated)
-Elongated the engine shutdown sound
-Added the power transfer sounds when switching from buses
-When turning on batteries, there is a clunk sound for one battery and a quieter clunk sound for both batteries when they are on, which is then followed by the light transformer whines and the CIDS chimes - this means that all of the sounds now follow the correct electric bus
-The whines that the elec system makes are all improved too
-The cabin ambience is improved also where the cabin fans now come from around the leading edge of the wing as that is where they are located. Now the light transformers are also heard inside the cabin and there is a small whine that comes from the avionics fan that is heard if you pay attention in the front to mid cabin
-There are now pack and brake fan spool down sounds

## Additional context

Known issues:
-When the power up sequence is interrupted, like the avionics ventilation test, the sound will keep playing even when cold and dark, this is due to a wwise limitation of the sim now allowing wwise stop events to play and I can not do anything about this currently

-When the left engine is shut down the aircraft loses and gains power making many sounds at once play - this is not an issue with sounds and with the elec system - this will be fixed by @davidwalschots soon in a future PR

## Testing instructions

-Bat 1 and 2 to on and listen to the new clunks and whines
-Start APU and turn off ext pwr to listen to power transfer
-Listen to the new avionics fan sound
-Turn on APU bleed and turn it off - then go to exterior view to listen to the shutdown
-Turn brake fan on and go to exterior view to turn it off to listen to the shutdown 
-Go to front/mid cabin (showcase view 13) and listen to the improved ambience
-Start engine and listen to new fuel intro - listen to exterior for the first eng start and interior for the second eng start
-Shut down engines and listen from interior
-Move the EFB volume sliders and check if they all still work for everything such as engines and wind and exterior (make sure to test different thrust levels on the engine on exterior and interior)

### DIFF
--- a/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/sound/sound.xml
+++ b/A32NX/SimObjects/AirPlanes/Asobo_A320_NEO/sound/sound.xml
@@ -197,52 +197,57 @@
 
         <!-- COCKPIT LOOPS SOUNDS ========================================================================================== -->
 
-        <Sound WwiseData="true" WwiseEvent="TRUnit" Continuous="false" SimVar="ELECTRICAL MAIN BUS VOLTAGE" NodeName="PEDALS_LEFT" Units="VOLTS" Index="1">
-            <Range LowerBound="28" />
-            <WwiseRTPC SimVar="ELECTRICAL MAIN BUS VOLTAGE" Units="VOLTS" Index="1" RTPCName="SIMVAR_ELECTRICAL_MAIN_BUS_VOLTAGE" />
+        <Sound WwiseData="true" WwiseEvent="TRUnit" Continuous="false" LocalVar="A32NX_ELEC_AC_1_BUS_IS_POWERED" Units="Bool" Index="1" NodeName="PEDALS_LEFT">
+            <Range LowerBound="1" />
             <Requires LocalVar="A32NX_AVIONICS_STARTUP_SOUNDS_INHIBIT" Units="Bool" Index="0">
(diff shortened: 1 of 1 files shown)


brunocfalcao:
hi @aguther , just to know, was this sound pack merged into the autopilot branch? Just because I was testing the engine sound and I don´t think it´s the same as in the development branch (which is merged).

I think it would be great to merge it in case it wasn't 👍 

Thank you,
Bruno

brunocfalcao:
@aguther Hi Andreas, I don't think this PR was merged into the autopilot branch right ? I am saying this because I noticed the exterior engine start sounds are different from the DEV version compared to the FBW version. You can notice the sound difference when the fuel flow starts to happen (in the DEV version is much more realistic than in the FBW version).

Thanks in advance for the check.

Best,
Bruno

tracernz:
Known issue with the fuel flow on startup (which drives the sound that's missing). See here for known issues https://github.com/flybywiresim/a32nx/tree/autopilot/docs

brunocfalcao:
Thanks tracernz, I'll ask next time on discord, and thank for the clarification!

PR #4335, opened 2021-04-07
fix(mcdu): limit scratchpad to 22 characters

tyler58546:
Fixes #4202

## Summary of Changes

* Limited MCDU scratchpad to 22 characters

## Screenshots (if necessary)

image

## References

image
image

## Additional context

The display can fit up to 24 characters, but the last 2 are reserved for the vertical slew arrows.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #4348, opened 2021-04-08
fix(mcdu): mcdu menu options not loading

### LINKED ISSUE #4345: MCDU Menu options cannot be selected
**Mod Version**

{
  "built": "2021-04-07T23:34:58+00:00",
  "ref": "master",
  "sha": "e8e2d3f43c7627c283ee2aef1828ffe77cc62fc7",
  "actor": "MisterChocker",
  "event_name": "manual"
}

**Describe the bug**

MCDU Menu options (FMGC, ATSU, AIDS, CFDS) cannot be selected and page does not change. 

**To Reproduce**

1. Go to MCDU menu (default on startup or from pressing menu button
2. Attempt to select FMGC, ATSU, AIDS, or CFDS
3. Observe that the page you selected does not load

**Expected behavior**

Selecting the option would load that page

**Actual behavior**

The screen remains on the MCDU menu and the option selected changes to blue text with (SEL) label. You are able to change all the options to blue in this way but the page never changes. 

The FMGC pages can still be loaded with the "physical" buttons but there is no workaround for the others.

**References**

https://cdn.discordapp.com/attachments/785976111875751956/829561894079037480/20210407234126_1.jpg

**Additional context**

Maybe related to https://github.com/flybywiresim/a32nx/commit/e8e2d3f43c7627c283ee2aef1828ffe77cc62fc7

It was working earlier on 4/7/21 but stopped after an update on 4/7/21. 

No

Zero One#4399

pareil6:
Fixes #4345.

## Summary of Changes
NXFictionalMessages.emptyMessage leaves message text null, causing the scratchpad update code to error when checking it's length (from #4335).

## Additional context

Discord username (if different from GitHub): pareil6#0990

## Testing instructions
Check MCDU menu options now work.


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -780,7 +780,7 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
     }
 
     set inOut(v) {
-        if (v.length < 23) {
+        if (v === null || v.length < 23) {
```
beheh, line 783: What if v is undefined? How about also checking for typeof to catch undefined etc. as well:

```suggestion
        if (v === null || (typeof v === "string" && v.length < 23)) {
```

tracernz:
I just noticed you already opened a PR 🤣. Actually I don't think it works though because `null !== undefined`?

sidnov:
> I just noticed you already opened a PR 🤣. Actually I don't think it works though because `null !== undefined`?

Can confirm it doesn't. Tested locally 😅

pareil6:
Huh, worked when I tested it locally, not a clue what happened there...

I did realise it's not immediately obvious whether allowing undefined messages is actually sensible or not (I did this as the quick fix to get back to what we had before) - happy to close this in favour of @tracernz's if the consensus is that's cleaner.

Issue #4266, opened 2021-04-04
Fuel Used versus FOB


kutak11:
**Mod Version**

Experimental cd80138

**Describe the bug**

Fuel used displays correct Fuel comsuption during enroute (OFP planning) but remaining FOB is not correct.
Dicrepancy between initial FOB and Used fuel.  

**To Reproduce**

1. Perform normal flight and see when on cruising level that FOB plus Fuel Used is not value of fuel when you start the flight
2.
3.

**Expected behavior**

FOB + used fuel = fuel when start the engines

**Actual behavior**

discrepancy about 15%

**References**

**Additional context**

Was this working before/when did the issue start occurring?

Is this a problem in the vanilla unmodded game?

Discord username (if different from GitHub):

Taz5150:
Will check, but looks like the discrepancy could be due to unused fuel.

kutak11:
> Will check, but looks like the discrepancy could be due to unused fuel.

the total "summary" of fuel FOB+Fuel used is continously increasing during flight. For example if I load 7000kg fuel on ramp, after landing Fuel Used shows 3500kg but FOB shows 4500kg. There is somehow 1000kg more FOB than it should be. All flight is prepared through Simbrief planning. Aircraft performace profile works well on Simbrief with fuel factor bias M16. Enroute Fuel used is very accurate to OFP, the only problem is that FOB is not "burning" according that.

Taz5150:
Quick question: did you have fuel in the Center tank, or just Inner and Outer wing tanks??

kutak11:
yes center tank empty... only Inner and outer... exactly how it loads MCDU or EFB. I do nothing with MSFS loading menu.

Taz5150:
New fuel model has been pushed in the autopilot branch and is ready to be re-tested. Let me please know how it goes.

MisterChocker:
Closing since no response

PR #4390, opened 2021-04-10
fix: Fuel consumption model

### LINKED ISSUE #4194: Large fuel burn imbalance
**Mod Version**
Experimental

**Describe the bug**
Large fuel imbalance between left and right wing tanks during long flight, not consistent with differences in fuel flow to engines 1 and 2. Fuel crossfeed doesn't work to solve the problem.

**To Reproduce**
1. Take off from KJFK with 12.9 tons of fuel
2. In cruise notice that engine 2 is burning 40 kg/hour more fuel than engine 1. Multiplied by 5.5 hours, this would translate in an extra total fuel burn of 220 kg by engine 2. Notice before midway in the flight that the fuel imbalance between left and right tanks exceeds already 400 kgs. Turn on fuel crossfeed.
3. Land in KLAX with about 2 tons of fuel in the left tanks and 1 ton of fuel on the right tanks, despite fuel crossfeed on during most of the flight.

**Expected behavior**
The difference in fuel burn between engines 1 and 2 is realistic, no two engines are the same in the real world. However one would expect a fuel imbalance of about 220 kgs at landing without using crossfeed, and no fuel imbalance with crossfeed on.

**Actual behavior**
Landed with 1 ton fuel imbalance despite fuel crossfeed on.

**References**
Screenshot 2021-03-28 225743

**Additional context**

This was my first long flight with the new experimental version, and third flight overall. Before that I was using the custom fbw.

No

Discord username (if different from GitHub): predator117

Taz5150:
Fixes #4194 #4266

## Summary of Changes

This PR fixes some issues introduced with the new fuel consumption model, specifically:
- Difference between initial fuel and fuel used not matching FOB
- Heavy fuel difference/ imbalance between left and right wings

The fuel model should now behave correctly and matching aircraft fuel flow.

Disclaimer: This fuel model does not change the original fuel system. This means fuel transfer from Center to Inner Tanks is not done as in real life. This issue and a revamped Fuel System is in the works by the team,

## Additional context

## Testing instructions

Maintain level flight at a given speed and thus a stable fuel flow. Fuel flow between engine 1 and 2 could be different (imbalanced). After some time (i.e. 1 - 2 hours) check:
- Fuel Used by the left and right engine corresponds to fuel flow
- FOB is the difference between initial fuel and total fuel used


Benjozork:
Lots of commits here. Looks like a dirty branch.

Taz5150:
Yeah, been having problems with github for some days now.

PR #3545, opened 2021-02-15
fix(mcdu): Fix input issues on PERF/W&B and fuel prediction pages

felixharnstrom:
Fixes #3525
Fixes #3321
Fixes #3528

## Summary of Changes

Fixed a number of missing or incorrect unit conversions relating to fuel and payload weight conversions when using imperial units. This fixes a number of issues where entered values were either not stored correctly or not displayed correctly. The following fields are fixed:

- PERF/W&B
   -  Page 1
      - BLOCK FUEL
      - TAXI FUEL
      - TRIP FUEL
   - Page 2
     - PAYLOAD
     - FWD BAG
     - REAR BAG
- Fuel prediction pages (INIT page B and the FUEL PRED page)
   - TAXI (on INIT B)
   - RTE RSV/%
   - ALTN/TIME
   - FINAL/TIME
   - MIN DEST FOB
   - EXTRA/TIME

Max allowed block fuel value is also changed to 19046 kg to match MSFS fuel dialog and actual fuel capacity.
Also fixes an issue where TRIP FUEL input on the PERF/W&B page was incorrectly bound to 2R.
## Screenshots (if necessary)

Before             |  After
:-------------------------:|:-------------------------:
fuel_conversion_bug | fuel_conversion_fixed
payload_conversion_bug | payload_conversion_fixed
fuel_pred_bug | fuel_pred_fixed

## References

Max fuel capacity from sources in #2286, adjusted to 19046 kg to account for rounding in MSFS

## Additional context

Discord username (if different from GitHub): Jethro#4431

## Testing instructions

Test this PR with both imperial and metric units (MCDU MENU -> OPTIONS -> FMGC -> WEIGHT UNIT, make sure to reload the aircraft after changing units): 
1. Make sure that fuel and payload values on PERF/W&B (MCDU MENU -> ATSU -> AOC MENU -> PERF/W&B) are entered correctly and that the correct amount of fuel is loaded when requesting refueling
2. Enter FROM/TO and ALTN on INIT page A 
3. Verify that both fuel prediction pages (INIT page B and FUEL PRED) behave as expected when changing field values

Note that this PR is focused around input validation and unit conversions (with some corresponding improvements to the presentation on the fuel prediction pages). Please bear this in mind when reporting potential issues with calculated values or any underlying algorithms. Also note that I have attempted to make minimal changes to the PERF/W&B payload page since this will be overhauled in an upcoming PR.

## How to download the PR for QA

[...]

PR #4391, opened 2021-04-10
fix: EFB fuel tank capacities

### LINKED ISSUE #4322: Incorrect EFB fuel tank capacities
**Mod Version**

After merging of https://github.com/flybywiresim/a32nx/pull/3744

**Describe the bug**

The fuel tank capacities shown on the EFB dispatch fuel page are incorrect. They do not correspond to the capacities set in the flight model .cfg file, shown in the MSFS weight and balance window, and used by the MCDU perf/W&B page.

**To Reproduce**

1.
2.
3.

**Expected behavior**

**Actual behavior**

**References**

See https://github.com/flybywiresim/a32nx/pull/2286 and the screenshots of fuel tank capacities in different units here: https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838

**Additional context**

donstim:
Fixes fuel tank capacity values shown in EFB to match https://github.com/flybywiresim/a32nx/pull/2286 and https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838

Fixes #4322 

## Summary of Changes

 - Changes constants used for capacity of each fuel tank
 - Also changes kg to lb conversion to more closely align with what MSFS uses
 - *NEW* Changes unusable fuel quantity in each tank to zero due to MSFS bug. Includes a comment in the flight model.cfg file to change the fuel tank capacities and unusable fuel quantities back to the real values if and when the bug is fixed.
## Additional context

Discord username (if different from GitHub):donbikes#4084

## Testing instructions
Check fuel tank capacities shown vs this table for the tank capacities in U.S. Gallons: 
Screenshot (377)
Only use this reference table for the fuel capacity in U.S. Gallon. That is how it is specified in the flight model cfg file. MSFS uses its own conversions to liters, kg, and lbs. (which is different than the conversions used for that table), so use the MSFS weight and balance screen to compare capacities in those units. The weight and balance screens for full fuel for the different units can be seen here: https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838


Saschl (approved):
LGTM

tracernz (approved):

St54Kevin:
Category: QA Tester
Name: St54Kevin#3583
Date of testing: 12/04/2021
Version of the sim: 1.14.6.0
PR Tested: #4391 
Tier of testing: 1
Changes to observe: EFB fuel page values

**Steps you've made to test the Pull Request:**
1. Loaded C&D and powered up
2. Checked EFB fuel page values

**Media:**
image

**Issues (If Any):** 
- While the max values do match with the MSFS screen, the **current** fuel does not. In every tank they are different in a few kg

**Conclusions of the testing:** Pls take a look at the current fuel values.

DavidEGTE:
Category: QA Tester
Name: David#2177
Date of testing: 12/04/2021
Version of the sim: 1.14.6.0
PR Tested: #4391 
Tier of testing: 1

Changes to observe:
EFB fuel tank capacities

Testing technique: Try-out

1. C&D
2. DCT to EFB

N/A
Microsoft Flight Simulator Screenshot 2021 04 12 - 09 32 03 03

Issues (If Any): 
same as @St54Kevin 
Steps to achieve the issue:
As above

Overall Rating: To be improved
Conclusions of the testing: fuel tank capacities are incorrect

donstim:
> **Conclusions of the testing:** Pls take a look at the current fuel values.

Thanks for the test! I was only concerned about the incorrect capacities to begin with. It looked like it drew the current values from simvars so I didn't think they would be an issue. But somehow it must normalize them so that they when full they would not have been different than the incorrect capacities.  Anyway, I will look into it further. Sorry you had to find it, but I guess that's what QA testing is for!

donstim:
> Overall Rating: To be improved
> Conclusions of the testing: fuel tank capacities are incorrect

Yep, thanks for testing! It's the current values that are incorrect. The capacities are correct.

DavidEGTE:
> > Overall Rating: To be improved
> > Conclusions of the testing: fuel tank capacities are incorrect
> 
> Yep, thanks for testing! It's the current values that are incorrect. The capacities are correct.

each way you put it, it doesn't reflect EFB vs  SIM FUEL SYSTEM

donstim:
Okay, the difference between the fuel tank quantities with 100% fuel load vs the tank capacities is the amount of unusable fuel. The problem is that MSFS allows you to use the "unusable" fuel. This results in negative fuel quantities (zero minus the unusable) when the tanks are "empty," The original author of the EFB fuel page got around the problem of potentially showing negative fuel quantities by displaying the maximum of either the reported fuel quantity or zero.  A problem with this is that as you burn fuel down to zero quantity, the respective engine will continue to run for a while (burning the "negative" unusable fuel) for a while after the fuel tank quantity reaches zero.

I had attempted to deal with the MSFS allowing use of the unusable fuel by putting the usable fuel tank capacities as the total fuel tank capacities in the flight model.cfg file. (This was done in https://github.com/flybywiresim/a32nx/pull/2286). I now see that this did not fully address the issue. By allowing use of unusable fuel, MSFS was adding, rather than subtracting, the unusable fuel to the total fuel capacity. This means that, even by setting the total fuel capacity as the usable fuel capacity, MSFS is allowing you to burn more fuel than would be usable IRL. 

To fix this, I am proposing to change the unusable fuel quantity in the flight model.cfg file to zero with comments included to change the fuel tank capacities (both total and unusable) to IRL values if and when the MSFS bug is fixed. The drawback of doing this is that the weight of a full fuel tank will be less than it actually should be by the amount of unusable fuel. But I feel it is better to have the weight very slightly "off" (IRL the weight would vary anyway since fuel density varies) than to allow use of more than the IRL usable fuel quantity.

donstim:
> > > Overall Rating: To be improved
> > > Conclusions of the testing: fuel tank capacities are incorrect
> > 
> > 
> > Yep, thanks for testing! It's the current values that are incorrect. The capacities are correct.
> 
> each way you put it, it doesn't reflect EFB vs SIM FUEL SYSTEM

What?

donstim:
Ok, those issue have been fixed. I noticed one remaining one that must have been there before. It looks like it rounds off all the fuel quantities before adding them to get the total. To match the MSFS screen, it should add the unrounded quantities then round the sum. This will remain a draft until I can see how to fix that. (If anyone else wants to step in on how to do that, feel free.)

DavidEGTE:
Category: QA Tester
Name: David#2177
Date of testing: 13/04/2021
Version of the sim: 1.14.6.0
PR Tested: #4391 
Tier of testing: 1

Changes to observe:
EFB fuel tank capacities

Testing technique: Try-out

1. C&D
2. opened EFB and FUEL SYS

N/A

Microsoft Flight Simulator Screenshot 2021 04 13 - 13 16 09 22
Microsoft Flight Simulator Screenshot 2021 04 13 - 13 16 54 48
Microsoft Flight Simulator Screenshot 2021 04 13 - 13 17 40 11

N/A
Steps to achieve the issue:
N/A

Overall Rating: Good
Conclusions of the testing: Fuel is all SYNC now, GTG

donstim:
> Overall Rating: Good
> Conclusions of the testing: Fuel is all SYNC now, GTG

Thanks for testing again!

ExampleWasTaken:
I found a bug that may potentially have to do with this PR. Gonna finish my test and leave a more detailed report in a few minutes.

marcman86:
Category: QA Tester
Name: marcman86#4907
Date of testing: 16.04.2021
Version of the sim: 1.15.7.0
PR Tested: #4391
Tier of testing: 1
Changes to observe: Fuel tank capacities

Testing technique: Tryout

1.) Spawned C&D at EDDP
2.) Start AC and EFB
3.) Checked fuel capacities

Media:


Image1


Image2


Image3

Issues: - Capacities are off from the reference

Overall Rating: -

Conclusions of the testing: Overall the fuel is synced. The capacities don't match the reference. Don't know if this is a problem of the PR or MSFS unit conversions.

Hope this helps :-)

Philadelphia5105:
Category: QA Trial
Name: Discord Philadelphia#5105
Date of testing: 04/16/21
Version of the sim: 1.15.7
PR Tested: #4391 
Tier of testing: 1
Changes to observe:
Ensure that the fuel tanks have the correct amount of fuel and the EFB matches the fuel panel

Testing technique: 
1. Spawn cold and dark at gate 139 at KPHL
2. Turn on batteries, EFB, and external power
3. Set fuel to 100% using the in-game fuel panel
4. Ensure that the fuel number match in the EFB, fuel panel, and the chart above

Screenshot (193)

Only thing that is incorrect is that the kg limits don't match, but as you stated, msfs doesn't use accurate conversion. 

Overall Rating: Good
Conclusions of the testing: 
Works as intended

ExampleWasTaken:
Category: QA Trial
Name: `ExampleWasTaken#0886`
Date of testing: 16/04/2021
Version of the sim: 1.15.7
PR Tested: #4391 
Tier of testing: 1
Changes to observe:
fix efb fuel tank capacities

Testing technique:
1. Loaded C&D
2. Checked that values are as expected

Media: 
Microsoft Flight Simulator Screenshot 2021 04 16 - 16 42 30 07

Video of the fuel dropping
https://streamable.com/4tjmvy

Issues (If Any): 
1. Fuel drops all the way down to 0 while aircraft is C&D
After restarting the flight I was not able reproduce it, but I thought it might be worth mentioning.
2. Maximum values don't match [reference](https://user-images.githubusercontent.com/70166617/114285785-a6960000-9a0e-11eb-8307-f646505d1357.png)

Overall Rating: Probably GTG
Conclusions of the testing: As the first issue couldn't be reproduced it probably shouldn't be part of this PR to fix it. With regard to issue 2, that only applies if the shown values actually represent the aircraft.

donstim:
> 
> Issues: - Capacities are off from the reference
> 
> Overall Rating: -
> 
> Conclusions of the testing: Overall the fuel is synced. The capacities don't match the reference. Don't know if this is a problem of the PR or MSFS unit conversions.
> 
> Hope this helps :-)

Thanks for the test!  Yes, it helps. It brings up a new issue. You are showing that the ECAM displays incorrect fuel quantities. This EFB display does match the actual reference (see https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838) and the screenshot in @DavidEGTE's review above. I guess I'll have to raise a new issue regarding the ECAM display of fuel quantities.

donstim:
> Issues (If Any): Write any issues
> Only thing that is incorrect is that the kg limits don't match, but as you stated, msfs doesn't use accurate conversion.
> 
Could you specify exactly where you think the kg limits are not matching? From your screenshot, it looks like an exact match (after rounding).

donstim:
> Issues (If Any):
> 
> 1. Fuel drops all the way down to 0 while aircraft is C&D
>    After restarting the flight I was not able reproduce it, but I thought it might be worth mentioning.
> 2. Maximum values don't match [reference](https://user-images.githubusercontent.com/70166617/114285785-a6960000-9a0e-11eb-8307-f646505d1357.png)
> 
> Overall Rating: Probably GTG
> Conclusions of the testing: As the first issue couldn't be reproduced it probably shouldn't be part of this PR to fix it. With regard to issue 2, that only applies if the shown values actually represent the aircraft.

1. The disappearing fuel is a known issue unrelated to this mod. It happens if you have so little fuel in the inner tanks that it triggers the fuel transfer from the outer tanks. If the engines and APU are off, for some reason this fuel never makes it into the inner tanks. It just disappears. If the engines are running, it does get transferred to the inner tanks for use by the engines.
2. The reference table you quoted is only good as a reference for this for the fuel volume in U.S. Gallons. As you can see on that chart, the fuel weight in that chart is associated with a specific fuel density of 0.785 kg/l (6.551 lb/gal). In MSFS, fuel capacity is specified in U.S. Gallon and then converted to other units using MSFS internal conversion factors. For fuel density, MSFS uses the more common 6.7 lb/gal (actually 6.699999809 lb/gal) and 3.039075693 kg/l. A better reference to use for this comparison on a fuel weight basis is to simply use the MSFS weight and balance window (see https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838)

Philadelphia5105:
> Could you specify exactly where you think the kg limits are not matching? From your screenshot, it looks like an exact match (after rounding).

Could've been a little clearer, when I set the fuel to 100% the amount in the tanks didn't match your reference image

ExampleWasTaken:
> 
> 
> > Issues (If Any):
> > 
> > 1. Fuel drops all the way down to 0 while aircraft is C&D
> >    After restarting the flight I was not able reproduce it, but I thought it might be worth mentioning.
> > 2. Maximum values don't match [reference](https://user-images.githubusercontent.com/70166617/114285785-a6960000-9a0e-11eb-8307-f646505d1357.png)
> > 
> > Overall Rating: Probably GTG
> > Conclusions of the testing: As the first issue couldn't be reproduced it probably shouldn't be part of this PR to fix it. With regard to issue 2, that only applies if the shown values actually represent the aircraft.
> 
>     1. The disappearing fuel is a known issue unrelated to this mod. It happens if you have so little fuel in the inner tanks that it triggers the fuel transfer from the outer tanks. If the engines and APU are off, for some reason this fuel never makes it into the inner tanks. It just disappears. If the engines are running, it does get transferred to the inner tanks for use by the engines.
> 
>     2. The reference table you quoted is only good as a reference for this for the fuel volume in U.S. Gallons. As you can see on that chart, the fuel weight in that chart is associated with a specific fuel density of 0.785 kg/l (6.551 lb/gal). In MSFS, fuel capacity is specified in U.S. Gallon and then converted to other units using MSFS internal conversion factors. For fuel density, MSFS uses the more common 6.7 lb/gal (actually 6.699999809 lb/gal) and 3.039075693 kg/l. A better reference to use for this comparison on a fuel weight basis is to simply use the MSFS weight and balance window (see [#3545 (comment)](https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838))

Makes sense and thanks for explaining the fuel disappearing. 

Then it's GTG from me :) 👍

donstim:
> > Could you specify exactly where you think the kg limits are not matching? From your screenshot, it looks like an exact match (after rounding).
> 
> Could've been a little clearer, when I set the fuel to 100% the amount in the tanks didn't match your reference image

I guess I should have been clearer with the reference image. See my explanation about the reference in my comment above.

Philadelphia5105:
GTG from me then as well

PR #4403, opened 2021-04-11
fix: Negative N2 before engine start

Taz5150:
## Summary of Changes

This PR fixes the negative N2 values in the EWD before engine start due to the new engine model imbalance calculations.

## Additional context

## Testing instructions

- Start from a C&D situation
- Power up the aircraft so that ECAM is On
- Turn Engine Model Selector to IGN
- EWD shows N2 0% for both engines


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -2204,7 +2204,10 @@ var A320_Neo_UpperECAM;
                 var n2_i = parseFloat(0);
             }
             const name = "ENG N2 RPM:" + _engine;
-            const percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
```
lukecologne, line 2207: Please don't use `var`, use `let` instead

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -2204,7 +2204,10 @@ var A320_Neo_UpperECAM;
...
+            var percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
+			if (percent < 0.0) {
```
lukecologne, line 2208: Some indentation problems here

lukecologne (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -2204,7 +2204,10 @@ var A320_Neo_UpperECAM;
                 var n2_i = parseFloat(0);
             }
             const name = "ENG N2 RPM:" + _engine;
-            const percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
+            var percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
```
lukecologne, line 2207: Please don't use `var`, use `let` instead.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -2199,12 +2199,15 @@ var A320_Neo_UpperECAM;
...
             if (parseInt(imbalance.toString().substr(0, 1)) == _engine) {
-                var n2_i = parseFloat(imbalance.toString().substr(5, 2)) / 100;
+                let n2_i = parseFloat(imbalance.toString().substr(5, 2)) / 100;
             } else {
-                var n2_i = parseFloat(0);
+                let n2_i = parseFloat(0);
             }
             const name = "ENG N2 RPM:" + _engine;
-            const percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
+            let percent = SimVar.GetSimVarValue(name, "percent") - n2_i;
```
davidwalschots, line 2207: `n2_i` is always undefined at this point. Unlike `var` which is hoisted up, `let` is block-scoped.

davidwalschots (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
Taz5150: i stands for imbalance. Will make the changes. Thanks!

davidwalschots: Then I'd suggest renaming the variable to `n2Imbalance`.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -2199,12 +2199,15 @@ var A320_Neo_UpperECAM;
         getValue(_engine) {
             const imbalance = SimVar.GetSimVarValue("L:A32NX_ENGINE_IMBALANCE", "number");
```
davidwalschots, line 2200: What is the number format that comes out of this?
We're doing some very interesting substringing on it, so I'm curious.

Taz5150: Its a coded "word"; I use it as a single LVar to give me all engine parameter imbalances (N2, FF and EGT). First two digits sets randomly and engine to be imbalanced (either 1 or 2). Then there are three sets of double digit values to set the imbalance for N2, FF and EGT. The FADEC wasm decodes that information.

davidwalschots: Okay. I'm not sure if you already did this, but please ensure this information is added to `docs/a320-simvars.md` (and you could consider pointing to there in a comment line in this code). Alternatively consider creating a JS class which reads and decodes this variable, such that we don't even need documentation 😃.

Taz5150: Haven't added anything to the simvars.md since this is part of the experimental/ AP branch, but if needed I can start documenting the list of LVars I've created for the engine.

Regarding the JS class, I can try and add a decoding class in the future ... its just a parser and its being done in each parameter class already (N2, EGT and FF). But I can put that in the pipeline.

davidwalschots: Ah yes. The "file changes" view was hiding what came before it. I didn't see the parameter class. I think documentation is sufficient then!

oim1:
Category: QA Trial
Name: oim#0001
Date of testing: Current date 12/04/2021
Version of the sim: 1.14.6.0
PR Tested: #4403
Tier of testing: 1
Changes to observe:
EWD showing N2 0% for both engines prior to engine start instead of a negative N2 value. 

Testing technique: 
1. Loaded Cold and Dark. 
2. Powered on the plane 
3. Set Eng Mode Sel to IGN
4. Checked that EWD displays 0% N2

Media: 
n2-0percent

Issues: None

Overall Rating: Good
Conclusions of the testing: N2 was 0% prior to eng start. Works as intended. LGTM.

lukecologne:
I have a question about the imbalance: Is there a specific reason why you chose two seperate simvars for the N2 and the N2 imbalance? In my mind it would make more sense to combine them at the WASM level and then only set a single N2 Simvar per engine

Taz5150:
Yes, N2 specifically has to be done at the "visual" (JS) level since we are not setting N2, its been set internally in the Sim. So we read N2 from the Sim, and then we take away the imbalance. 

In the case of fuel flow, it is being done in the WASM since the value has a direct effect on the fuel consumption. In the case of EGT, yes I think we can implement the imbalance in the WASM.

lukecologne:
Ah yeah, I see the problem. Would there be any downside to taking the Sims N2 var, and instead of setting a LVar with the imbalance just compute the imbalance, internally add it to the sims N2 var, and then set a custom N2 Lvar? That way we wouldn't have to do this inside the display level, maintaining separation of concerns

Taz5150:
Well, N2 is kept separated as it is (just reading the SimVar would give you the real N2). Not sure also about simplicity, since we would have to create and additional LVar and what I tried with the Imbalance word was to avoid making one LVar per parameter (soon the number of parameters will increase with secondary engine variables). So not sure if we should go that way (at least for N2)

lukecologne:
We still need one additional LVar anyways (the imbalance ones). So if we replace that with just the modified N2 LVar, the count will stay the same.

PR #4367, opened 2021-04-09
fix: update runs with deltaTime of -1

davidwalschots:
## Summary of Changes
The expression `_deltaTime === -1 && !KnobChanged` causes the update to run whenever the knob has changed. As the `updateThrottler.canUpdate` function might have returned `-1`, this could cause the update to run with a `-1` delta time.

This PR corrects that behaviour and makes it easier to force an update with the correct delta time.

## Additional context

Discord username (if different from GitHub): SjotgunSjonnie#9623

## Testing instructions

The displays still update as expected:
- The navigation displays updates at a quicker rate when making a turn.
- Turning the PFD, ND and ECAM displays ON/OFF is immediate.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #4411, opened 2021-04-12
feat(du): self-test only when unpowered over 10 seconds

### LINKED ISSUE #945: Self-test only when unpowered over 10 seconds
**Mod Version**
Master (0.3)

**Describe the bug**
When reducing the brightness on the following switches to zero:
PFD, ND, Upper ECAM, Lower ECAM, right PFD, right ND.

and then raising the brightness (thereby switching the screens back on), the screens switch to a self test mode immediately. In addition, reducing (for example) the right ND, will also swap the left ND screen to auto test (and vice versa).

**To Reproduce**
1. Reduce brightness on any screen to minimum so that it switches off.
2. Increase brightness on same screen to above minimum, and the screen goes into self test mode, along with corresponding screen on the opposite side of the cockpit.

**Expected behavior**
Lowering these switches to off and then increasing the brightness knob (switching them back on) should only enter self test mode once the screen is off for 10 seconds. If less than 10 seconds, the screen should just be as it was prior to being switched off.

**Actual behaviour**
Turning the screen brightness down for any screen (ND, PFD, Upper / Lower ECAM, right ND, right PDF) switches them off, and upon switching them back on they enter self test mode immediately.

davidwalschots:
Fixes #945

## Summary of Changes
Display units run a self-test. This self-test should only be executed when the display unit was unpowered for 10 seconds or more.

Additionally this PR also fixes an issue I uncovered where turning on the chrono (shown on the MFD) with the DU off wouldn't turn the chrono on until the DU was turned on. Now the chrono runs irrespective of the DU being powered or not.

## Screenshots (if necessary)
image

## References
@komp1821 posted a video showing the behaviour which can be found [here](https://discord.com/channels/738864299392630914/754851312306487326/758257387432968222).

## Additional context

## Known issues
In some instances the original screen content still appears for a brief instant when turning the display unit on. This was already happening before this PR, and isn't fixed by this PR.

## Testing instructions
- For each display, turn it off and on to check if the self-test only runs after being off for 10 seconds or more.
- Similarly, test the same thing with power becoming available/unavailable. The easiest way to test this is at the gate by turning EXT PWR on and off.

### DIFF
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/DisplayUnit.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/DisplayUnit.js
@@ -1,40 +1,67 @@
 class DisplayUnit {
     constructor(isPoweredFn, getSelfTestTimeInSecondsFn, potentiometerId, selfTestElement) {
+        this.selfTest = new DisplayUnitSelfTest(selfTestElement, getSelfTestTimeInSecondsFn);
         this.isPowered = isPoweredFn;
-        this.getSelfTestTimeInSeconds = getSelfTestTimeInSecondsFn;
         this.potentiometerId = potentiometerId;
-        this.selfTestElement = selfTestElement;
 
-        this.turnedOnDuringThisUpdate = false;
-        this.previousUpdateBrightness = 0;
-        this.wasPoweredDuringPreviousUpdate = false;
-        this.selfTestTimerMilliseconds = 0;
+        this.previouslyOff = false;
     }
 
-    updateBeforeThrottle() {
+    isJustNowTurnedOn() {
+        return this.previouslyOff && this.isOn();
+    }
+
+    isOn() {
         const brightness = SimVar.GetSimVarValue(`LIGHT POTENTIOMETER:${this.potentiometerId}`, "number");
-        this.turnedOnDuringThisUpdate = (brightness >= 0.1 && this.previousUpdateBrightness < 0.1);
-        this.previousUpdateBrightness = brightness;
-    }
-
-    isTurnedOnDuringThisUpdate() {
-        return this.turnedOnDuringThisUpdate;
+        return brightness >= 0.1 && this.isPowered();
     }
 
     update(deltaTime) {
-        this.selfTestTimerMilliseconds -= deltaTime / 1000;
-
-        const isPowered = this.isPowered();
-        const powerStateChanged = isPowered != this.wasPoweredDuringPreviousUpdate;
-
-        if ((this.isTurnedOnDuringThisUpdate() || powerStateChanged) && isPowered) {
-            this.selfTestTimerMilliseconds = this.getSelfTestTimeInSeconds() * 1000;
+        if (this.isJustNowTurnedOn()) {
+            this.selfTest.execute();
         }
 
-        this.selfTestTimerMilliseconds -= deltaTime;
+        const isOn = this.isOn();
+        this.selfTest.update(deltaTime, isOn);
 
-        this.selfTestElement.style.visibility = this.selfTestTimerMilliseconds > 0 ? "visible" : "hidden";
+        this.previouslyOff = !isOn;
+    }
+}
 
-        this.wasPoweredDuringPreviousUpdate = isPowered;
+class DisplayUnitSelfTest {
+    constructor(element, getSelfTestTimeInSecondsFn) {
+        this.element = element;
+        this.getSelfTestTimeInSeconds = getSelfTestTimeInSecondsFn;
+
+        this.remainingTestDurationInMilliseconds = 0;
+
+        // Start with a state where turning on the display unit within 10 seconds after starting the flight
+        // will trigger the self test.
+        this.offDurationInMilliseconds = DisplayUnitSelfTest.RequiredAfterBeingOffForMilliseconds;
+    }
+
+    static get RequiredAfterBeingOffForMilliseconds() {
+        return 10000;
+    }
+
+    isRequired() {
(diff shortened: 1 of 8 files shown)


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/MFD/A320_Neo_MFD.js
```diff
@@ -162,29 +163,25 @@ class A320_Neo_MFD_MainPage extends NavSystemPage {
     }
 
     onUpdate() {
-        const currentKnobValue = SimVar.GetSimVarValue("LIGHT POTENTIOMETER:" + this.potIndex, "number");
-        if (currentKnobValue <= 0.0) {
-            this.selfTestLastKnobValue = currentKnobValue;
-            return;
-        }
```
davidwalschots, line 169: I removed this `return;` as it caused the chrono to not initialise when pressing the chrono button with the DU off. In the real aircraft the DMC updates the chrono and thus the button should also work with the DU off.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/PFD/A320_Neo_PFD.js
```diff
@@ -96,46 +95,43 @@ class A320_Neo_PFD_MainPage extends NavSystemPage {
...
             const YokeYPosition = 47.95 - (14.45 * SimVar.GetSimVarValue("YOKE Y POSITION", "Position"));
 
             this.groundCursor.style.left = YokeXPosition.toString() + "%";
             this.groundCursor.style.top = YokeYPosition.toString() + "%";
         }
 
-        const KnobChanged = (currentKnobValue >= 0.1 && this.selfTestLastKnobValue < 0.1);
-
-        _deltaTime = this.updateThrottler.canUpdate(_deltaTime);
-        if (_deltaTime === -1 && !KnobChanged) {
```
davidwalschots, line 135: Missed one case in #4367. Fixed here.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/PFD/A320_Neo_PFD.js
```diff
@@ -146,7 +142,7 @@ class A320_Neo_PFD_MainPage extends NavSystemPage {
             this.spdFlash.setAttribute("visibility", "visible");
             this.altFlash.setAttribute("visibility", "visible");
             this.vsFlash.setAttribute("visibility", "visible");
-        } else if (this.selfTestTimer > -5) {
```
davidwalschots, line 149: This condition was only ever `false` when the user had a framerate of 1 frame per 5 seconds or less. Therefore, I assumed this to be a coding error and removed the condition.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/DisplayUnit.js
```diff
@@ -0,0 +1,67 @@
...
+        this.getSelfTestTimeInSeconds = getSelfTestTimeInSecondsFn;
+
+        this.remainingTestDurationInMilliseconds = 0;
+
+        // Start with a state where turning on the display unit within 10 seconds after starting the flight
+        // will trigger the self test.
+        this.offDurationInMilliseconds = DisplayUnitSelfTest.RequiredAfterBeingOffForMilliseconds;
+    }
+
+    static get RequiredAfterBeingOffForMilliseconds() {
```
tracernz, line 43: `powerOffRideThroughTime` or something like that might be clearer, or at least relate to the physical state being modelled?

IMO the logic for determining whether the caps are fully discharged/the DU is fully dead belongs in the DisplayUnit, and the SelfTest should simply be concerned with doing the self test.

davidwalschots: > `powerOffRideThroughTime` or something like that might be clearer, or at least relate to the physical state being modelled?

I guess it might be clearer for some and less clear for others. I myself didn't know ride through time. It seems to be an electrical engineering term.

> IMO the logic for determining whether the caps are fully discharged/the DU is fully dead belongs in the DisplayUnit, and the SelfTest should simply be concerned with doing the self test.

Yes. In a previous version I actually programmed the DU to contain the duration of being off and I agree with you that from the perspective of "what object has which properties" it makes most sense to put it in the DU. However, as the self test is the only piece of the DU that cares about the duration the DU was off (at least at this moment), I decided to encapsulate it in there as to not bother the higher-level abstraction DU with it.

So now the question is, what do we prefer? 😃

tracernz: > I guess it might be clearer for some and less clear for others. I myself didn't know ride through time. It seems to be an electrical engineering term.

Fair point!

tracernz (approved):
Nice tidy-up 💯

DavidEGTE:
Category: QA Tester
Name: David#2177
Date of testing: 13/04/2021
Version of the sim: 1.14.6.0
PR Tested: #4411 
Tier of testing: 1

Changes to observe:
self-test only when unpowered over 10 seconds

Testing technique: Try-out

1. C&D + Ext. PWR
2. Tried all DU

N/A
N/A
N/A
Steps to achieve the issue:
N/A

Overall Rating: Good
Conclusions of the testing: Seems to be working fine, GTG

PR #4672, opened 2021-05-01
docs(template): remove pre-defined value -> use placeholder

MisterChocker:
## Summary of Changes

Remove pre-defined values as they allow the user not to fill in the field at all, we should use the placeholder feature instead which displays (as long as nothing has been entered) an example.

## Additional context

<!-- ## How to download the PR for QA

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_beta.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report_beta.yaml
@@ -21,14 +21,6 @@ body:
   id: build-info
   attributes:
     label: Build info
-    placeholder: |
-      {
-        "built": "2021-04-29T14:32:22+00:00"
-        "ref": "HEAD",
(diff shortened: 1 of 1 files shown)


MisterChocker:
We need to check if multi line space holders are possible

.github/ISSUE_TEMPLATE/bug_report_beta.yaml
```diff
@@ -21,6 +21,14 @@ body:
   id: build-info
   attributes:
     label: Build info
+    placeholder: |
+      {
+        "built": "2021-04-29T14:32:22+00:00"
+        "ref": "HEAD",
+        "sha": "e049d433cbb5d8c53ce083a4631626309c13c86a",
+        "actor": "David Walschots <davidwalschots@users.noreply.github.com>",
```
davidwalschots, line 29: This should be a more realistic name! I don't create many bugs. 😂

davidwalschots (approved):

PR #4734, opened 2021-05-07
refactor(ecam): convert apu page to react

brandoncabael:
Fixes #[issue_no]

## Summary of Changes

Refactored APU page in react

## Additional context

Discord username (if different from GitHub): F11killa#9855


src/instruments/src/SD/Pages/Apu/Apu.tsx
```diff
@@ -0,0 +1,136 @@
...
+            <BleedBox x={420} y={163} pressure={apuBleedPressure} apuBleedOn={apuBleedOn} />
+            <Divider x={83} y={227} />
+
+            <ApuFlapOpen x={400} y={375} apuFlapOpenPercentage={apuFlapOpenPercentage} />
+            <LowFuelPressureInd x={409} y={315} lowFuelPressure={lowFuelPressure} />
+        </EcamPage>
+    );
+};
+
+const SvgGroup = ({ x, y, children }) => <g transform={`translate(${x},${y})`}>{children}</g>;
```
Benjozork, line 44: I would avoid creating components like this tbh, MSFS is sensitive to GC lag

beheh: I think this is appropriate. We shouldn't micro-optimize this stuff away on a hunch. If this specific instance is an issue (and I don't think it is) we shouldn't be using React.

Benjozork: I don't understand what advantage this brings, in terms of readability. Why do we need an `SvgGroup` component when `<g transform={`translate(${x},${y})`}>` is a few characters longer ? It's not about optimizing everything, it's rather than this particular abstraction seems a bit silly to me, abstracting away something which is perfectly readable in the first place, and potentially introducing a deeper object structure as a possible side effect.

beheh: You've switched from arguing for performance in favor of readability. I don't have a strong opinion on what is more readable here, I can see your point. I was contesting the fact that we need to be super-mindful about every component we introduce, as that goes against the composability of React and can make working with it a lot more tedious.

Benjozork: My point is that the potential and possible reduced burden on GC is worth it considering the little improvement in readability it brings.

Of course there are other factors to consider, such as setting `x={0) y={0}` effectively creates a Coherent layer which can reduce draw calls.

davidwalschots: I introduced this particular component in the ELEC page React rewrite, which brandoncabael is basing their work on. I'm big on componentisation of things and I like to abstract from higher-level to lower-level where the higher-level abstractions leak less of the HTML/SVG things. For me not having to repeat the same thing over and over again is a good thing. For this particular situation I can see you can argue the other way, however I do hope we're not removing the componentisation advantage of working in React. That would be a shame.

As beheh noted, if object creation really has a big performance impact, then we shouldn't be using React at all. It creates loads of objects.

Benjozork: We could then add also a `layer` prop to this component to create GT layers.

Benjozork:
Is this still being worked on?

brandoncabael:
> Is this still being worked on?

Yes this is still being worked on, but some assistance with drawing the gauges would be appreciated!

Benjozork:
@ErickSharp is working on gauges for the ENGINE page

ErickSharp:
Hello, would you like my help? I just got finished on an engine page refactor. Is there something in particular you need help with? My PR includes helper components for gauges and needles if that would serve any assistance to you.

brandoncabael:
Actually yes @ErickSharp that would be much appreciated. I'll take a look, and reach out if necessary. I should have it done this weekend.

ErickSharp:
Hello, do you think we can get this pull request moving again? Are there a couple more things you need to cross off the to-do list? If so I can start contributing to your branch to get this pull request to the finish line. Just let me know :  )

brandoncabael:
Yeah @ErickSharp that would be fine. I appreciate your help.

davidwalschots:
Please consider implementing the APU GEN OFF state in this PR as well. As noted [here](https://github.com/flybywiresim/a32nx/issues/5609#issuecomment-907842908) by @lukecologne the APU GEN when pushed to OFF should display OFF on the APU page. I overlooked this at the time of building the feature.

ErickSharp:
Hello, it seems this PR has become pretty stale. Can you confirm that you are able to work on it any time soon? If not I will close this and open a new one to replace this.

ErickSharp:
For anyone viewing this pull request in the future, this pull request's work has been continued in #5969

PR #4739, opened 2021-05-07
refactor(ecam): f/ctl

davidwalschots:
## Summary of Changes
Reduce the number of lines of F/CTL from 738 to 390 by increasing reuse and using a more declarative style for some of the components.

## Known issues

- There's a logic discrepancy between the WHEEL page and the F/CTL page. When on the ground with the SPEED BRAKE on FULL, the WHEEL page displays all spoilers in the UP position, the F/CTL page only displays the inner three spoilers in the UP position. I'm going to leave fixing this outside of the PR, as it's not entirely clear which behaviour is correct (note that on the outside model all spoilers are up, but this in itself might also not be correct).

## Additional context

## Testing instructions
Check if the F/CTL page still works as it used to.


RichardPilbery:
The flight control page behaviour is correct and the wheel page will need correcting.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
```diff
@@ -1,738 +1,390 @@
...
-    const d4 = `M${x},210 l${leftorright === 'left' ? '-' : ''}8,0 l0,6 l${leftorright === 'right' ? '-' : ''}8,0`;
+    // Rudder limits
+    const [indicatedAirspeedState] = useSimVar('AIRSPEED INDICATED', 'knots', 500);
+    let maxAngleNorm = 1;
+    if (indicatedAirspeedState > 380) {
+        maxAngleNorm = 3.4 / 25;
+    } else if (indicatedAirspeedState > 160) {
+        maxAngleNorm = (69.2667 - 0.351818 * indicatedAirspeedState
+            + 0.00047 * indicatedAirspeedState ** 2) / 25;
+    }
```
davidwalschots, line 206: @aguther this piece of JS code calculates the displayed maximum deflection of the rudder on the F/CTL page. As far as I can tell it doesn't match with the maximum deflection as managed by the FBW implementation. Is this something you could expose in the form of an LVar so it can be used here?

aguther: I need to recheck if and how I implemented this, there is another limitation in the flight model as far as I know. Will check tomorrow and let you know.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
```diff
@@ -1,738 +1,390 @@
...
+            <ElevatorAxis side={side} x={0} y={5} />
+
+            <HydraulicIndicator x={side === 'left' ? -60 : 18} y={79} type={leftHydraulicSystem} />
+            <HydraulicIndicator x={side === 'left' ? -38 : 40} y={79} type={rightHydraulicSystem} />
+        </SvgGroup>
     );
 };
 
+const ElevatorAxis = ({ x, y, side }: ComponentPositionProps & ComponentSidePositionProps) => {
+    const d1 = `M0 0 l${
```
Saschl, line 373: Is there a reason for naming the constant `d1`? It’s not entirely clear to me as the content string doesn’t disclose the meaning when reading the code. Same applies for the other appearances of d1, d2 etc.

Saschl (approved):
All in all very nice and the readability is much better. Feel free to look at the suggestion/question but I wouldn’t consider it a blocker.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
```diff
@@ -1,738 +1,390 @@
...
-        let rPT = pitchTrimState / 1213.6296;
-        // Cap pitch trim at 13.5 up, 4 down
-        if (rPT > 16384.0) {
-            rPT = 16384.0;
-        } else if (rPT < -4854) {
-            rPT = -4854;
-        }
-        setRawPitchTrim(rPT);
-    }, [pitchTrimState]);
+    const hydraulicContext: HydraulicsContext = {
```
RichardPilbery, line 48: Now that the hydraulics system has been implemented, it would be good to use the actual SimVars for hydraulic system availability rather than the surrogate I implemented based on N2 state.

RichardPilbery (dismissed):

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
davidwalschots: That will/should come in a separate PR. This PR aims to rewrite the code to React.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
davidwalschots: I've not adapted the existing names in this case. I'm guessing `d` is used as that is the HTML attribute on the `path` element. 
Renaming these would likely result in names such as `lowerHorizontalPartOfWing`, `verticalStripeTowardsArrow` or who knows what other crufty name. I can do it, but if this is the only concern I'd rather avoid starting up the whole development workflow just for this 😀.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
RichardPilbery: It is written in React already. You're refactoring it, which is fine and an improvement on my original code. The SimVars and function to check for lowpressure are already in the original code (just commented out, lines 77++).

davidwalschots: I see. Sorry about that. There are by now three or four PRs related to React. Most of them are from old to React. I finished this PR's code 12 days ago and HYD came in later. Feel free to add the changes in after this PR is completed. I don't want to mix in multiple concerns in one PR.

src/instruments/src/SD/Pages/Fctl/Fctl.tsx
Saschl: Agreed :D

ExampleWasTaken:
Category: QA Testers
Name: `ExampleWasTaken#0886`
Date of testing: 22/05/2021
Version of the sim: 1.15.10.0
PR Tested: #4739 
Tier of testing: 1
Changes to observe: behavior of FCTL page after refactor

Testing technique:
1. Loaded C&D
2. Powered up to EXT PWR
3. Checked FCTL page behavior (ailerons, elevator, speedbrakes, trim, rudder)
4. Started one engine & repeated step # 3

Overall Rating: LGTM
Conclusions of the testing: All good, GTG from me. 🚀

Issue #1022, opened 2020-10-01
[BUG] One LDG light switch extends both halfway


liebbe10:
**Mod Version**

master

**Describe the bug**

If you use one LDG light switch and set it to on while leaving the other one in off both extend halfway.

**To Reproduce**

1.Power up the Acft
2.Use only one LDG light switch

**Expected behavior**

Only one of them should extend fully.

**Actual behavior**

Both extend halfway.

**References**

![Screenshot 2020-10-01 205239](https://user-images.githubusercontent.com/30324171/94851425-b9f55c80-0428-11eb-9296-754f63e5b20c.png)

**Additional context**

Was this working before/when did the issue start occurring?

Is this a problem in the vanilla unmodded game?
Probably, not checked.

Discord username (if different from GitHub):

NathanInnes:
Do you have one side retracted on the switch?

lukecologne:
Can confirm, one switch extends both halfway, and when both switches are on, both landing lights extend fully. I think the problem is that both animations have the same name. They are seperate animation when viewed in blender, but their name is the same, so both get extended at the same time

komp1821:
Confirmed. Checked it in the unmodded version as well. Same behavior.

liebbe10:
> Do you have one side retracted on the switch?

Yes one side was in retract and the other in off.
![Screenshot 2020-10-01 211438](https://user-images.githubusercontent.com/30324171/94853094-25402e00-042b-11eb-8018-6a7f513efc93.png)

PR #4026, opened 2021-03-19
feat(efb): flyPadOS v2 🚀

hiaaryan:
Fixes: #4053
Fixes: #4075
Fixes: #4575 

## Important
Remove env file from PR workflow before merge.

## Summary of Changes

flyPadOS v2. Full Layout & UI Rewrite. Special thanks to @Benjozork & @ZigTag! 

- [x] Add New Layout.
- [x] Add Navigraph Support.
- [x] Enable Remaining Settings Options.
- [x] Finishing Touches.
- [x] Remove INOP Pages.
- [x] Adjust All Pages.

## Screenshots (if necessary)

image

# Testing Instructions

Please thoroughly test every page (if you don't have a navigraph subscription don't worry about it).

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #4902, opened 2021-05-25
feat(mcdu): qol improvement - keyboard input to mcdu scratchpad

### LINKED ISSUE #4862: keyboard entry option on MCDU
### Aircraft Version

Development

### Description

I don´t know if there is already a request or idea like this.

It would be a helpful and effective feature, if we can use the keyboard to use the MCDU. I don´t know if it was the PMDG´s @ FSX but there you could click on a point @ the FMS. After this your keyboard was "connected" with the keyboad on the MCDU/FMS. after clicking again, it was back to normal use.

Just an Idea and maybe interessting.

### References (optional)

_No response_

_No response_

_No response_

2hwk:
Fixes #4862

## Summary of Changes

Click the MCDU's screen to activate keyboard input mode to the scratchpad, you may now enter letters, numbers, dots and slashes with the keyboard, rather than clicking on the buttons. Function keys also can be used to input functions (L1-L6/R1-R6)

- [x] Able to input text and numbers into the MCDU with keyboard, by clicking on the MCDU screen to put it in "focus"
- [x] Cockpit buttons are pressed with keyboard input
- [x] Visual indicator that input is being accepted by MCDU
- [x] Options for MCDU keyboard entry
- [x] Normal Sim Keyboard Events (i.e. Camera) will be disabled while MCDU is in focus.
- [x] A timer can be set to automatically lose focus from the MCDU after a set period
- ~~Arrow key functionality - Arrow keys are currently being hijacked by the options menus... thanks Coherent.~~

## Screenshots (if necessary)

[Video](https://user-images.githubusercontent.com/15316958/121357298-46ff9680-c964-11eb-999b-95cd250006cc.mp4)

## Additional context

QOL improvement.

Discord username (if different from GitHub): 2Cas #1022

## Testing instructions

Known Issues:
* Arrow keys + Esc are not captured
* CLR held (>2s) - Use Shift+Backspace instead of backspace.
* Camera is locked - hence the need for default timeout of 60s in case the screen somehow is unclickable. 
* Clicking on screen + using camera keybind keys will cause MCDU to move out of the view, as the camera is locked it is not possible to click on the screen to unfocus the MCDU - Use keyboard shortcuts CTRL+Z or press ALT to unfocus. 
* Keybindings are currently optimized for ANSI/ISO-UK keyboard layouts. ISO-DE/ISO-NORD users will notice that slash keys on their keyboards may not function as expected. This is an unfortunate limitation of the Coherent Webkit based JS Engine. The >/< (Int backslash) and # (or '/*) keys  are used as substitutes instead. Note that the numpad will work as expected regardless of keyboard layout. 

1. Ensure MCDU functions normally as a baseline
2. Attempt keyboard input without clicking on the MCDU. Nothing should happen.
3. Attempt keyboard input after clicking on MCDU. Nothing should happen (still).
4. Go into MCDU Options > Options > Realism > Keyboard Input. Select ALLOW INPUT. You will have to exit to the menu and reload the flight for this change to occur (consistent with units change from kg/lb). 
5. Click on the MCDU screen, and attempt keyboard input.. The scratchpad should be highlighted, and the title should be highlighted in cyan as well. Letters, numbers, forward slashes and periods (.) should appear on the scratchpad. Backspace should act with the same behavior as the CLR button on the mcdu. **Pressing Shift+Backspace for 1 second should function the same as holding down the CLR on the mcdu for 2 seconds.**

NOTE: Focusing the MCDU for keyboard input will stop all sim keybinds from working.  Most notably, you will not be able to move the camera. Unfocusing should return control (i.e. camera)

Invalid +/- input: -+, -+-+-+, ++-+-+
Valid +/-  input: +-, +++++-, +++-, - - - - - - -

6. Attempt to flood the MCDU and break it with keyboard input somehow. Slam the keyboard if you have to. Check that there are no inputs that would be impossible using the manual method (clicking on buttons). 
7. Check that the normal input method still works (using cockpit mcdu buttons to input into scratchpad)
8. Check that each button (on both MCDUs) moves and reacts to the keyboard. 
9. Check that all the pages are still functioning as intended. Check for visual UI glitches.
10. Check that inputting data from the scratchpad still  works as you would expect.
11. Double check that no inappropriate inputs that would not be possible with the usual method (clicking on the buttons) are possible.
12. Clicking again on the MCDU should disable keyboard input, **or press ALT or CTRL+Z**. The scratchpad should no longer be highlighted. Note: By default, there is a 60 second timeout on keyboard input. If no inputs are received in this time, the scratchpad will unfocus itself. This is intentional behaviour. 
13. Now test the aforementioned timeout function. Make an input, then wait for a minute. The scratchpad should unfocus itself. Check that keyboard input is no longer possible.
~~14. Go back to MCDU options > Options > Realism > MCDU Timeout (Seconds) and set the timeout to 0s. NONE* should appear~~
~~15. Now the scratchpad should never unfocus. Ensure that this is the case by waiting past the default timeout length. This should happen if both keyboard input is entered, or not.~~
16. Set the timeout to a custom duration (i.e. 5 seconds). Ensure that the scratchpad will lose focus after this time. This should happen if both keyboard input is entered, or not. The valid range should be 5-120 seconds.
17. Go into the options. Disable MCDU keyboard input (NONE). Reset the MCDU if prompted. Check that previous behavior is retained.

Please note these testing instructions are just here to *inform* your testing procedure, don't just stick to this. Do try your best to break things.  


2hwk:
Needs a lot of work to exit draft at this stage.

I can't figure out how to disable keybinds in the sim (for camera movement, instrument and switch adjustment shortcuts/keybinds, etc) when the MCDU is in focus, so you will probably need to unbind all controls on the alphanumeric cluster on your keyboard (or move them into modifier + key binds) for this to be actually useful. I don't expect to solve this one unless somebody has some ideas.

IMO this is especially bad for the inexperienced if you have the default keybinds set up (I can already see people happily using the MCDU with this feature with the default keybinds, hitting Y, going into slew mode and then being "stuck".)

Secondly from a UX perspective it would be good to have some kind of visual display that the MCDU is currently in focus (so you don't input things when you don't mean to)

And finally it should be a toggleable option in the options menu (off by default). This helps to mitigate the first issue a bit.

MisterChocker:
@2hwk some ideas that came to my mind:
maybe make it, that you'll need to "open" the mcdu in a new (separate) window. Iirc my joystick only works when focused in the main window of the sim. 
Basically have the mcdu "pop out" to allow for keyboard inputs.

2hwk:
> @2hwk some ideas that came to my mind:
> maybe make it, that you'll need to "open" the mcdu in a new (separate) window. Iirc my joystick only works when focused in the main window of the sim.
> Basically have the mcdu "pop out" to allow for keyboard inputs.

Yeah I believe the EFB has something similar, this may be a good enough solution.

Handling the logic of the +/- is quite interesting. Being able to use arrow keys with keyboard arrow keys as well. There is more to this than meets the eye. 

I also was thinking that pressing the keyboard should probably MOVE the physical buttons in the cockpit but I mean XML so...

2hwk:
Well I have not found a perfect solution, but there's potential here. 

Unfortunately arrow keys do not seem to pass events to JS, so, I can't get the MCDU arrow keys working with keyboard input. Everything else more or less functions as it should. Now if only there was a way to disable sim keybind events...

ssewell:
This would be killer feature for cockpit builders with one minor change: Allow for a key bind that would directly feed into the MCDU (both alphanumeric and even the soft keys without the need to set focus beforehand).

This would enable any external keyboard HID to directly control the MCDU, enabling quick development of external hardware to control the unit. I've looked over the official MSFS 2020 SDK documentation, and I can't find any obvious way to feed keyboard into the MCDU via SimConnect (is this even possible?).

2hwk:
> This would be killer feature for cockpit builders with one minor change: Allow for a key bind that would directly feed into the MCDU (both alphanumeric and even the soft keys without the need to set focus beforehand).
> 
> This would enable any external keyboard HID to directly control the MCDU, enabling quick development of external hardware to control the unit. I've looked over the official MSFS 2020 SDK documentation, and I can't find any obvious way to feed keyboard into the MCDU via SimConnect (is this even possible?).
> 
> Ninja edit: Would it be possible to use an approach as described in https://web.dev/hid/ to obtain input values from specific HID hardware in the event that multiple HID devices of the same class are connected to the computer?

This is probably a bit beyond the scope of this PR, but, I can point you in a helpful direction, since this should already be possible with various tools the community has already built. 

Currently what you are describing would not be possible using _SimConnect_ no, as the mcdu runs off HTML events (H: Events) for the scratchpad input from the push buttons. I don't believe at this moment that you can interact with HTML events without WASM. The LVars I have added only affect the animation of the MCDU keyboard. 

I believe MobiFlight has produced a WASM module that can interact with H:Events. 

https://youtu.be/PKBjEl9E5A4?t=218
https://www.mobiflight.com/forum/topic/6528.html

The relevant H:Events for the MCDU would be in the format H:A320_Neo_CDU_{L/R}_BTN_{X} 
i.e. H:A320_Neo_CDU_L_BTN_G would be the event for button G on the Left MCDU.

As for web HID, maybe, but, bearing in mind how ancient the webkit engine that the game uses is... I wouldn't be that confident on that right now. I don't think that would be helpful.

ssewell:
> > This would be killer feature for cockpit builders with one minor change: Allow for a key bind that would directly feed into the MCDU (both alphanumeric and even the soft keys without the need to set focus beforehand).
> > This would enable any external keyboard HID to directly control the MCDU, enabling quick development of external hardware to control the unit. I've looked over the official MSFS 2020 SDK documentation, and I can't find any obvious way to feed keyboard into the MCDU via SimConnect (is this even possible?).
> > Ninja edit: Would it be possible to use an approach as described in https://web.dev/hid/ to obtain input values from specific HID hardware in the event that multiple HID devices of the same class are connected to the computer?
> 
> This is probably a bit beyond the scope of this PR, but, I can point you in a helpful direction, since this should already be possible with various tools the community has already built.
> 
> Currently what you are describing would not be possible using _SimConnect_ no, as the mcdu runs off HTML events (H: Events) for the scratchpad input from the push buttons. I don't believe at this moment that you can interact with HTML events without WASM. The LVars I have added only affect the animation of the MCDU keyboard, I may rename these so as to not mislead.
> 
> I believe MobiFlight has produced a WASM module that can interact with H:Events.
> 
> https://youtu.be/PKBjEl9E5A4?t=218
> https://www.mobiflight.com/forum/topic/6528.html
> 
> The relevant H:Events for the MCDU would be in the format H:A320_Neo_CDU_{L/R}_BTN_{X}
> i.e. H:A320_Neo_CDU_L_BTN_G would be the event for button G on the Left MCDU.

Thank you! This should give me enough thread to pull so I can work on my own implementation.

2hwk:
> > > This would be killer feature for cockpit builders with one minor change: Allow for a key bind that would directly feed into the MCDU (both alphanumeric and even the soft keys without the need to set focus beforehand).
> > > This would enable any external keyboard HID to directly control the MCDU, enabling quick development of external hardware to control the unit. I've looked over the official MSFS 2020 SDK documentation, and I can't find any obvious way to feed keyboard into the MCDU via SimConnect (is this even possible?).
> > > Ninja edit: Would it be possible to use an approach as described in https://web.dev/hid/ to obtain input values from specific HID hardware in the event that multiple HID devices of the same class are connected to the computer?
> > 
> > 
> > This is probably a bit beyond the scope of this PR, but, I can point you in a helpful direction, since this should already be possible with various tools the community has already built.
> > Currently what you are describing would not be possible using _SimConnect_ no, as the mcdu runs off HTML events (H: Events) for the scratchpad input from the push buttons. I don't believe at this moment that you can interact with HTML events without WASM. The LVars I have added only affect the animation of the MCDU keyboard, I may rename these so as to not mislead.
> > I believe MobiFlight has produced a WASM module that can interact with H:Events.
> > https://youtu.be/PKBjEl9E5A4?t=218
> > https://www.mobiflight.com/forum/topic/6528.html
> > The relevant H:Events for the MCDU would be in the format H:A320_Neo_CDU_{L/R}_BTN_{X}
> > i.e. H:A320_Neo_CDU_L_BTN_G would be the event for button G on the Left MCDU.
> 
> Thank you! This should give me enough thread to pull so I can work on my own implementation.

Sorry, one small correction, it should be H:A320_Neo_CDU_{1/2}BTN{X}. Please look at our WASM modules (in /src/) as well as Mobiflight's WASM bridge (https://bitbucket.org/mobiflight/mobiflightwasm/src/master/) to get an idea of how you can implement this. 

On that note I realise I could simplify my code a lot more and I was overthinking my solution a bit. Will just quickly duck back in and fix that.

tyler58546:
I think I may have just found a solution by reverse engineering the text fields for the sim's UI.
`Coherent.trigger('FOCUS_INPUT_FIELD')` - Disables all keybinds and locks the camera
`Coherent.trigger('UNFOCUS_INPUT_FIELD')` - Enables keybinds and unlocks the camera

tracernz:
> 
> 
> I think I may have just found a solution by reverse engineering the text fields for the sim's UI.
> `Coherent.trigger('FOCUS_INPUT_FIELD')` - Disables all keybinds and locks the camera
> `Coherent.trigger('UNFOCUS_INPUT_FIELD')` - Enables keybinds and unlocks the camera

Just tried on the EFB input component:
```ts
    useEffect(() => {
        if (focusActive) {
            Coherent.trigger('FOCUS_INPUT_FIELD');
        } else {
            Coherent.trigger('UNFOCUS_INPUT_FIELD');
        }
    }, [focusActive]);
```

IT WORKS!! 👏

ZigTag:
#4026 needs this

sidnov:
> #4026 needs this

cc @hiaaryan

2hwk:
Tyler has been the hero as usual, although locking the camera is not 100% ideal but this is a good solution

2hwk:
> > #4026 needs this
> 
> cc @hiaaryan

Please note:  Be careful with handling FOCUS_INPUT_FIELD as if for any reason you are able to escape without calling UNFOCUS_INPUT_FIELD, you will be locked out and unable to do anything without the user having to restart the entire sim.

2hwk:
I don't see arrow key functionality being possible for now, so, I'm scrapping it. On reflection, I'm removing pop-up functionality as well, I think it works best as intended, being able to mask keyboard events makes it unnecessary.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
```diff
@@ -73,5 +73,5 @@ const NXFictionalMessages = {
     noWptInfos:             new McduMessage("NO WAYPOINT INFOS", false, false),
     emptyMessage:           new McduMessage(""),
     weightUnitChanged:      new McduMessage("UNIT CHANGED RELOAD A/C", true, true),
-    resetMcdu:              new McduMessage("RESET MCDU TO APPLY", true, false)
+    reloadPlane:             new McduMessage("RELOAD PLANE", true, false)
```
tracernz, line 76: Nitpick: can we make it RELOAD A/C, or RELOAD AIRCRAFT, to match the message immediately above?

tracernz (commented):
Looks good apart from one small comment.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
2hwk: Sure, should be an easy change

aguther:
One observation that happened to me is that the options page the value does not update if you change it. So say if you've selected "NO INPUT" and click on "ALLOW INPUT" the screen doesn't change. Functional wise it works though. If you exit the page and reenter it, it's correctly displayed. This is the case in both directions.

2hwk:
> One observation that happened to me is that the options page the value does not update if you change it. So say if you've selected "NO INPUT" and click on "ALLOW INPUT" the screen doesn't change. Functional wise it works though. If you exit the page and reenter it, it's correctly displayed. This is the case in both directions.
> 
> Other than that a really great feature!
> 
> I tested also enable keyboard input and then reload the flight. Wasn't causing any issues.

This was a bug that was introduced when I got rid of the MCDU refresh, I'll make some changes.

giridat:


https://user-images.githubusercontent.com/61822722/120941522-c9a80c00-c740-11eb-9c60-5648f0ae36fe.mp4

https://user-images.githubusercontent.com/61822722/120941525-cf055680-c740-11eb-9447-c2c8ce5085cb.mp4




https://user-images.githubusercontent.com/61822722/120941518-c4e35800-c740-11eb-9376-2b5439986916.mp4



https://user-images.githubusercontent.com/61822722/120941507-b72dd280-c740-11eb-9516-85b8510cb831.mp4


Discord: Giridhar Raut (Giridhar [z+5:30])
Object of testing: (#4902 )
Tier of Testing: 1
Date: (07/06/2021)

Testing Process: Conducted a flight from EDDF - EGCC
Route: EDDF SID ULKIG Y180 BITBU Y181 DEMUL M624 BUB L608 DENUT L610 GILDA M14 STOAT L613 MOGLI UP6 LESTA STAR EGCC

Steps:

Followed the test instructions:

1. Ensure MCDU functions normally as a baseline (working as intended)
2. 
3. Attempt keyboard input without clicking on the MCDU. Nothing should happen.  (working as intended)
4. 
5. Attempt keyboard input after clicking on MCDU. Nothing should happen (still).  (working as intended)
6. 
7. Go into MCDU Options > Options > Realism > Keyboard Input. Turn this to ALLOW INPUT. You will have to exit to the menu and reload the flight for this change to occur (consistent with units change from kg/lb).  (working as intended)
8. 
9. Click on the MCDU screen, and attempt keyboard input.. The scratchpad should be highlighted. Letters, numbers, forward  slashes and periods (.) should appear on the scratchpad. Backspace should act with the same behavior as the CLR button.  (working as intended) 
10. 
11. NOTE: Focusing the MCDU for keyboard input will stop all sim keybinds from working. Most notably, you will not be able to move the camera. Unfocusing should return control (i.e. camera)  (working as intended)
12. 
13. Attempt to flood the MCDU and break it with keyboard input somehow. Slam the keyboard if you have to. Check that there are no inputs that would be impossible using the manual method (clicking on buttons).  (working as intended) ( Didnt slam the keyboard though) 
14. 
15. Check that the normal input method still works (using cockpit mcdu buttons to input into scratchpad).  (working as intended)
16. 
17. Check that each button (on both MCDUs) moves and reacts to the keyboard.  (working as intended)
18. Check that all the pages are still functioning as intended. Check for visual UI glitches.  (working as intended) 
19. 
20. Check that inputting data from the scratchpad still works as you would expect.  (working as intended)
21. Double check that no inappropriate inputs that would not be possible with the usual method (clicking on the buttons) are possible. (working as intended)
22. 
23. Clicking again on the MCDU should disable keyboard input. The scratchpad should no longer be highlighted. Note: By default, there is a 60 second timeout on keyboard input. If no inputs are received in this time, the scratchpad will unfocus itself. This is intentional behaviour.  (working as intended)
24. Now test the aforementioned timeout function. Make an input, then wait for a minute. The scratchpad should unfocus itself. Check that keyboard input is no longer possible. (working as intended)
25. 
26. Go back to MCDU options > Options > Realism > MCDU Timeout (Seconds) and set the timeout to 0s. NONE* should appear
27. Now the scratchpad should never unfocus. Ensure that this is the case by waiting past the default timeout length. This should happen if both keyboard input is entered, or not.  (working as intended)
28. Set the timeout to a custom duration (i.e. 4 seconds). Ensure that the scratchpad will lose focus after this time. This should happen if both keyboard input is entered, or not. (working as intended)
29.
30. Go into the options. Disable MCDU keyboard input (NONE). Reset the MCDU if prompted. Check that previous behavior is retained.  (working as intended)


Negatives: Found a cyan color band or light bluish band on the header covering the fond. 
Expected Result: The color band on the header should not be present throughout scratchpad usage.

Testing Results: Passed

Conclusions: Big relief thanks for implementing good to go.
Media: 

https://user-images.githubusercontent.com/61822722/120941501-b2691e80-c740-11eb-9ecb-ee666125b6ae.mp4

2hwk:
The cyan band is intentional, since disabling keybinds now disables the camera, I need to provide a very obvious visual indicator that the MCDU input is active so the user does not get confused and they know what is causing their camera to be locked.

tracernz (approved):

Cryptic56:

Discord: Cryptic#4731
Object of testing: #4902 
Tier of Testing: 1
Date: 07/05/2021

Testing Process: Loaded in to gate at YSSY
Attempted to type on MCDU (should not work)
Change settings to allow keyboard inputs. Reload A/C
Attempt to type (should not work)
Click on MCDU and type (should work)
Check that MCDU buttons still work.
Wait 60s for timeout
Check all MCDU functions and pages.
Test letters, numbers, backspace, forward/back slash, period, minus/plus, space.

Negatives: None

Testing Results: Passed

Conclusions: Everything is working as expected including the intentional features that have been discussed already (cyan bar etc). This is a great feature and worked flawlessly, great work!

Media: 
Microsoft Flight Simulator Screenshot 2021 06 07 - 14 16 36 75

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
```diff
@@ -72,5 +72,6 @@ const NXFictionalMessages = {
     noRefWpt:               new McduMessage("NO REF WAYPOINT", false, false),
     noWptInfos:             new McduMessage("NO WAYPOINT INFOS", false, false),
     emptyMessage:           new McduMessage(""),
-    weightUnitChanged:      new McduMessage("UNIT CHANGED RELOAD A/C", true, true)
+    weightUnitChanged:      new McduMessage("UNIT CHANGED RELOAD A/C", true, true),
```
MisterChocker, line 75: I suggest removing the `weightUnitChanged` and using the new `reloadPlaneApply` because they both have the same intention.

MisterChocker (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
2hwk: This isn't used in this, out of scope for this PR, that's for weights lb/kg.

tracernz: MisterChocker is saying share the same reload message for both changing units, and changing the keyboard setting... i.e. delete the unit changed one and replace it with the new more general one.

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #4902
Tier of Testing: 1
Date: 7.6.2021

Testing Process: (Protocol)

- Attempt keyboard input without clicking on the MCDU. Nothing should happen. ==> OK
- Attempt keyboard input after clicking on MCDU. Nothing should happen (still). ==> OK
- Go into MCDU Options > Options > Realism > Keyboard Input. Turn this to ALLOW INPUT. You will have to exit to the menu and reload the flight for this change to occur (consistent with units change from kg/lb). ==> OK
- Click on the MCDU screen, and attempt keyboard input.. The scratchpad should be highlighted, and the title should be highlighted in cyan as well. ==> OK
- Letters, numbers, forward slashes and periods (.) should appear on the scratchpad. Backspace should act with the same behavior as the CLR button. ==> OK
- NOTE: Focusing the MCDU for keyboard input will stop all sim keybinds from working. Most notably, you will not be able to move the camera. Unfocusing should return control (i.e. camera) ==> confirmed
- Attempt to flood the MCDU and break it with keyboard input somehow. Slam the keyboard if you have to. Check that there are no inputs that would be impossible using the manual method (clicking on buttons). ==> OK
- Check that the normal input method still works (using cockpit mcdu buttons to input into scratchpad) ==> OK
- Check that each button (on both MCDUs) moves and reacts to the keyboard. ==>
- Check that all the pages are still functioning as intended. Check for visual UI glitches. ==> OK
- Check that inputting data from the scratchpad still works as you would expect. ==> OK
- Double check that no inappropriate inputs that would not be possible with the usual method (clicking on the buttons) are possible. ==> OK
- Clicking again on the MCDU should disable keyboard input. The scratchpad should no longer be highlighted. ==> OK
- By default, there is a 60 second timeout on keyboard input. If no inputs are received in this time, the scratchpad will unfocus itself. This is intentional behaviour. Now test the aforementioned timeout function. Make an input, then wait for a minute. The scratchpad should unfocus itself. Check that keyboard input is no longer possible. ==> OK
- Go back to MCDU options > Options > Realism > MCDU Timeout (Seconds) and set the timeout to 0s. NONE* should appear ==> OK
- Now the scratchpad should never unfocus. Ensure that this is the case by waiting past the default timeout length. This should happen if both keyboard input is entered, or not. ==> OK
- Set the timeout to a custom duration (i.e. 4 seconds). Ensure that the scratchpad will lose focus after this time. This should happen if both keyboard input is entered, or not. ==> OK
- Go into the options. Disable MCDU keyboard input (NONE). Reset the MCDU if prompted. Check that previous behavior is retained. ==> OK
- Ensure MCDU functions normally as a baseline: ==> OK
- Various attempts to try to break stuff: ==> OK 

Negatives: None (German keyboard not having a "/" is known - workarounds with other keys OK imho)

Testing Results: (Passed/Not Passed)

Conclusions: LGTM

Media: (Pictures/videos if applicable - delete line if not applicable)
image

(Extra: German keyboard see screenshot)
Observations: 
Key "^" (left of the number 1) works as "/"
Key "+" (left of enter) toggles "+" and "-"
Key "-" (left of right shift) toggles "+" and "-"
Key "<" (right of left shift) works as "/"
Num Key work as well but not / (":-")
Suggestion: use the num key "/" as "/" as well - this would work on German keyboards
F1-F12 works nice for input buttons

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
2hwk: yes but this PR is the one that adds the reloadPlaneApply NXSystemMessage, weightUnitChanged is used in the ECAM so if I simply deleted it it would cause an exception there.

2hwk: This has been changed in the latest commit

tracernz (approved):
Code looks fine now.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
```diff
@@ -72,5 +72,5 @@ const NXFictionalMessages = {
     noRefWpt:               new McduMessage("NO REF WAYPOINT", false, false),
     noWptInfos:             new McduMessage("NO WAYPOINT INFOS", false, false),
     emptyMessage:           new McduMessage(""),
-    weightUnitChanged:      new McduMessage("UNIT CHANGED RELOAD A/C", true, true)
+    reloadPlaneApply:       new McduMessage("RELOAD A/C TO APPLY", true, false)
```
MisterChocker, line 75: Why did you changed it to a type I message?
I'd assume this message has be actively processed by the pilot.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -19,10 +19,23 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
+        const mcduInput = NXDataStore.get("MCDU_KB_INPUT", "NONE");
+        switch (mcduInput) {
+            case "ALL":
+                return true;
+            case "RALT":
+                return false;
+            default:
+                return false;
+        }
+    }
```
MisterChocker, line 38: Cant this be simplified?

```suggestion
    get isInteractive() {
        return NXDataStore.get("MCDU_KB_INPUT", "NONE") === "ALL";
    }
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -831,6 +896,96 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
+    initKeyboardScratchpad() {
+        window.document.addEventListener('click', () => {
+
+            const mcduInput = NXDataStore.get("MCDU_KB_INPUT", "NONE");
+            const mcduTimeout = parseInt(NXDataStore.get("CONFIG_MCDU_KB_TIMEOUT", "60"));
+            const isPoweredL = SimVar.GetSimVarValue("L:A32NX_ELEC_AC_ESS_SHED_BUS_IS_POWERED", "Number");
+            const isPoweredR = SimVar.GetSimVarValue("L:A32NX_ELEC_AC_2_BUS_IS_POWERED", "Number");
+
+            // TODO: L/R MCDU
+            if (mcduInput != "NONE") {
```
MisterChocker, line 918: Use strict compare instead (if possible)

e.g.

```suggestion
            if (mcduInput !== "NONE") {
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_MCDU_KB.js
```diff
@@ -0,0 +1,84 @@
...
+        let all = "*ALLOW INPUT[color]cyan";
+        let none = "*NO INPUT[color]cyan";
+
+        switch (mcduInput) {
+            case "ALL":
+                all = "ALLOW INPUT[color]green";
+                break;
+            default:
+                none = "NO INPUT[color]green";
+        }
```
MisterChocker, line 21: I don't see the need for a switch here...

```suggestion
        let all = "*ALLOW INPUT[color]cyan";
        let none = "*NO INPUT[color]cyan";
        
        if (mcduInput === "ALL") {
            all = "ALLOW INPUT[color]green";
        } else {
            none = "NO INPUT[color]green";
        }
```

one could even go ahead and make it const

```suggestion
        const [all, none] = mcduInput === "ALL" ? ["ALLOW INPUT[color]green", "*NO INPUT[color]cyan"] : ["*ALLOW INPUT[color]cyan", "NO INPUT[color]green"];
```

And while I am at it, I prefer using `{green}mytext{end}` over `mytext[color]green`, but in this case it makes no difference because there aren't multiple colors in one line/string.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_MCDU_KB.js
```diff
@@ -0,0 +1,84 @@
+class CDU_OPTIONS_MCDU_KB {
+    static ShowPage(mcdu) {
+
+        mcdu.clearDisplay();
+
+        const mcduInput = NXDataStore.get("MCDU_KB_INPUT", "NONE");
```
MisterChocker, line 6: I don't understand why you are using NONE and ALL instead of 0 and 1 to mimic Boolean behavior which you are strictly following.

MisterChocker (changes requested):

MisterChocker:
Looking at the isInteractive get you have created, why don't you make it a utility function and use it everywhere?

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
tracernz: Good catch. I missed that. 😞

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
2hwk: Leftover fragment from before the change, good catch

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_MCDU_KB.js
2hwk: Was leftover for when there were 3 modes earlier on, reduced down to 2. I agree that it's not needed now.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
2hwk: Noted, also changing to === true rather than !== false

2hwk:
Removed everything that was related to the old right alt pop-out function that was removed, some leftover code wasn't cleaned up.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -19,10 +19,15 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
         this.activeSystem = 'FMGC';
         this.messageQueue = [];
+        this.inFocus = false;
+        this.lastInput = 0;
     }
     get templateID() {
         return "A320_Neo_CDU";
     }
+    get isInteractive() {
+        return NXDataStore.get("MCDU_KB_INPUT", "DISABLED") === "ENABLED";
```
MisterChocker, line 29: What I originally meant was something like this when I talked about Boolean.
Actually have a number 0 or 1 (Boolean) stored as string ofc and when reading converting it back to a Boolean.

```js
Boolean(Number("0")));
Boolean(Number("1")));
```

Also why is it that I don't see any usage of this method??? This method should be used instead or scrapped.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -831,6 +888,96 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
+        this.getChildById("header").style = null;
+        if (this.check_focus) {
+            clearInterval(this.check_focus);
+        }
+    }
+
+    initKeyboardScratchpad() {
+        window.document.addEventListener('click', () => {
+
+            const mcduInput = NXDataStore.get("MCDU_KB_INPUT", "DISABLED");
```
MisterChocker, line 904: use `this.isInteractive()` instead OR scrap the method.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_MCDU_KB.js
```diff
@@ -0,0 +1,75 @@
+class CDU_OPTIONS_MCDU_KB {
+    static ShowPage(mcdu) {
+
+        mcdu.clearDisplay();
+
+        const mcduInput = NXDataStore.get("MCDU_KB_INPUT", "DISABLED");
```
MisterChocker, line 6: same here. `mcdu.isInteractive()`

And when doing this, I'd rename (for simplicity) `mcduInput` into `mcduIsInteractive` so the variable name actually represents it's functionality/usage.

MisterChocker (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
2hwk: get isInteractive() is overriding an inherited function from A32NX_BaseAirliners (Line 381).

Long story short this function is called when the MCDU is first initialised (even before main loop is created) to determine if the screen should capture touch input or not.

2hwk: get isInteractive() is overriding an inherited function from A32NX_BaseAirliners (Line 381).

Long story short this function is called when the MCDU is first initialised (even before main loop is created) to determine if the screen should capture touch input or not.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
2hwk: See above comment on get isInteractive()

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/OPTIONS/A320_Neo_CDU_OPTIONS_MCDU_KB.js
2hwk: See above comment on get isInteractive()

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
2hwk: Also as for the syntax around boolean NXDataStorage items, the existing convention seems to be to use "ENABLED"/"DISABLED" as flags, also I think this is a lot more readable than trying to force boolean types, even if this would save on some string processing I don't think it's worth the loss of readability.

MisterChocker (approved):
LGTM

marcman86:
Discord:marcman86#4907
Object of testing: #4902
Tier of Testing: 1
Date: 11/06/2021

Testing Process: 
- Spawned C&D at EDDB
- Start up with classic MCDU interface
- Respawned C&D at EDDB
- MCDU Options > Options > Realism > Keyboard Input
- Reload to apply
- Start up with MCDU keyboard inputs


Negatives: DEL dont clears messages like GPS PRIMARY or values for V1, VR, V2 and ZFW/ZFWCG. Maybe something for a future PR.

Testing Results: Passed
Conclusions: Works. Nice feature. Would be nice to have this for the EFB.

Hope this helps :-)

FlightSimmer7700:
Not able to complete a full qa review because my sim's navdata is not playing ball, noticed 2 things, 
- when trying to enter a - or + you have to double click on the - OR + to get a + and simply clicking the + gives you a -
- when holding down backspace CLR flashes a lot, I think this should be changed to match what happens when CLR is held if possible

2hwk:
> Not able to complete a full qa review because my sim's navdata is not playing ball, noticed 2 things:
> 
> * when trying to enter a - or + you have to double click on the - OR + to get a + and simply clicking the + gives you a -
> * when holding down backspace CLR flashes a lot, I think this should be changed to match what happens when CLR is held if possible

The first is to emulate pressing the +/- button as there is no dedicated + and - buttons on the MCDU, intentional behaviour.
Onclr held behaviour will be implemented in a future PR along with arrow keys.

2hwk:
I've modified the +/- behaviour to be more intuitive. Now it should enter + when you press + and - when you press -, but it will still prevent you from entering invalid scratchpad input.

CLR functionality has also been modified to prevent the CLR blinking issue.

Samueleonard:

Discord: 1samueleonard1
Object of testing: (#4902)
Tier of Testing: (1)
Date: (16/06/2021)

Testing Process: 
.loaded in on runway, switched on input option, reloaded
. used keyboard to enter flightplan - normal operations worked fine
. spammed/tried to overload - worked fine

TESTING LATEST ADDITIONS
. used +/- keys - worked intuitively
. pressed DEL - functions as CLR

Negatives: sim limitation, im sure, but when the keyboard is toggled, camera cannot be moved, not really a negative, just a point if it can be overcome

Testing Results: (Passed)

Conclusions: great addition, works fine for me, maybe get 1 more test to make sure nothing was missed with the new additions
Media: n/a

wpine215:
> personally, i think it would be better to have this option enabled by default, i dont see any reasons why it shouldnt be

Looking at the commit history, it seems the option that is disabled by default was a remnant of the PR before Tyler found the correct coherent events to mask all input. Given your positive review of the PR without any issues found, I agree that it should be enabled by default.

Also, just my opinion, but the highlighting on the title (top line) of the MCDU looks a bit out of place. Perhaps the scratchpad highlighting is sufficient, or a slight border around the entire display (1-2 px, possibly white or yellow) could work?

FlightSimmer7700:
Quality Assurance Tester

Object of testing: 4902
Tier of Testing: 1
Date: 20/06/21

Testing Process: 
Loaded in while C&D
Tried all keyboard buttons
Tried on different pages of the mcdu
Negatives: 
- When clicking on the + on the numpad it acts like + used to do next to the backspace key
- was continuously clicking on the mcdu and basically hitting the keyboard and now I am stuck in the pilot's view and can't move as the keyboard input is on and the view is looking away from the mcdu, I think 0 seconds/never should not be allowed otherwise people may get into this situation or there should be a keybind unlocking the view. Or maybe a "cooldown" between the two as a buffer like 30ms so that this doesn't happen? Not sure how much of this is possible in code lol
- if stuck as above, restarting a flight doesn't help so you have to go back to the main menu (most of the time)
- ("feature")1 to 5 seconds should probably say "ENTRY OUT OF RANGE" when put as the timeout

Testing Results: Not Passed

Conclusions: Very good work overall but has a few small bugs
Media:
Clicking the mcdu screen continuously and spamming keyboard in the S, D, F, G above and below area
https://user-images.githubusercontent.com/70103461/122686605-d67d3300-d209-11eb-83b7-a954d83184e2.mp4

FlightSimmer7700:
> Also, just my opinion, but the highlighting on the title (top line) of the MCDU looks a bit out of place. Perhaps the scratchpad highlighting is sufficient, or a slight border around the entire display (1-2 px, possibly white or yellow) could work?

I agree

2hwk:
> personally, i think it would be better to have this option enabled by default, i dont see any reasons why it shouldnt be
> 
> Looking at the commit history, it seems the option that is disabled by default was a remnant of the PR before Tyler found the correct coherent events to mask all input. Given your positive review of the PR without any issues found, I agree that it should be enabled by default.
> 
> Also, just my opinion, but the highlighting on the title (top line) of the MCDU looks a bit out of place. Perhaps the scratchpad highlighting is sufficient, or a slight border around the entire display (1-2 px, possibly white or yellow) could work?

It needs to be obvious since it would be quite easy for new NX users to click on the mcdu and not realize they've activated the text input. As it locks out the camera this would lead to more things that support have to deal with. I tried a border and just having the scratchpad highlighting but it was not obvious enough. 

By default we have all convenience features such as fast ADIRS alignment and self-test time set to realistic settings, so this is just keeping in line with that.

2hwk:
Improvements have been made:

- Shift+Backspace functions as a CLR HELD keybind, hold both keys down for 1 second to clear the scratchpad
- +/- functionality has been improved
- Focus timeout - You can no longer select no timeout option, valid range is 5-120 seconds.

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #
Tier of Testing: 1
Date: 21.6.2021

Testing Process: Re-test of changes

- Shift+Backspace functions as a CLR HELD keybind, hold both keys down for 2 seconds to clear the scratchpad ==> OK
- +/- functionality has been improved - Invalid input like -+-+-+-+ should not be possible still, but + and - functions in a more intuitive way. ==> OK 
- Focus timeout - You can no longer select no timeout option, valid range is 5-120 seconds. ==> OK
- You can now unfocus the MCDU by pressing CTRL+Z or ALT. (ESC cannot be used as it is used by the pause menu) ==> OK

Negatives: None 

Testing Results: Passed

FlightSimmer7700:
Quality Assurance Tester

Object of testing: #4902 
Tier of Testing: 1
Date: 22/06/21

Testing Process: 
Loaded in and tried all the new changes made:

- "Shift+Backspace functions as a CLR HELD keybind, hold both keys down for 2 seconds to clear the scratchpad

- +/- functionality has been improved - Invalid input like -+-+-+-+ should not be possible still, but + and - functions in a more intuitive way.

- Focus timeout - You can no longer select no timeout option, valid range is 5-120 seconds.

- You can now unfocus the MCDU by pressing CTRL+Z or ALT. (ESC cannot be used as it is used by the pause menu)"

Negatives: N/A
Testing Results: Passed

Conclusions: LGTM, no issues found great work on fixing the issues promptly

PR #4996, opened 2021-06-04
feat(mcdu): clear whole scratchpad when CLR held

tracernz:
Fixes #[issue_no]

## Summary of Changes

When the CLR button on the MCDU is held down for a period of time, the whole scratchpad is cleared. ~~Awaiting pilot feedback on what that period is before this goes out of draft...~~ The scratchpad clears when the button has been held for approx. 1 second, thanks to Capt. Chaos.

## Additional context

Unfortunately had to use Lvars rather than Ovars to hold the state because the ASOBO_GT_Push_Button_Airliner template creates a sub-component, putting the LEFT_SINGLE_CODE in a different scope to the update.

## Testing instructions

Try the CLR button with the scratchpad in various states (messages up, empty, with one character, with multiple characters, empty, etc...).

### DIFF
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
@@ -100,6 +100,12 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
             }
             this.tryShowMessage();
         };
+        this.onClrHeld = () => {
+            if (this.inOut !== "" && this.inOut !== FMCMainDisplay.clrValue) {
+                this.inOut = "";
+                this.tryShowMessage();
+            }
+        };
 
         this.PageTimeout = {
             Prog: 2000,
@@ -996,6 +1002,8 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
                 setTimeout(() => {
                     this.onClr();
                 }, this.getDelaySwitchPage());
+            } else if (input === "CLR_Held") {
+                this.onClrHeld();
             } else if (input === "DIV") {
                 setTimeout(() => {
                     this.onDiv();
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -91,6 +91,7 @@
 1. [OVHD] Correct EVAC button behaviours - @tracernz (Mike)
 1. [OVHD] Fix invisible fault emissive decal on centre fuel pump 2 - @tracernz (Mike)
 1. [VFX] Add APU exhaust heat blur effect - @wpine215 (Iceman)
+1. [MCDU] Clear scratchpad when CLR held down - @tracernz (Mike)
 
 ## 0.6.0
 1. [CDU] Added WIND page - @tyler58546 (tyler58546)
--- a/src/behavior/src/A32NX_Interior_MCDU.xml
+++ b/src/behavior/src/A32NX_Interior_MCDU.xml
@@ -312,8 +312,9 @@
             <BASE_NAME>OVFY</BASE_NAME>
         </UseTemplate>
 
-        <UseTemplate Name="FBW_MCDU_BUTTON_Template">
+        <UseTemplate Name="FBW_MCDU_Holdable_BUTTON_Template">
             <BASE_NAME>CLR</BASE_NAME>
+            <MIN_HELD_DURATION>1</MIN_HELD_DURATION>
         </UseTemplate>
     </Template>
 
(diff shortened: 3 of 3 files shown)


davidwalschots (approved):

Samueleonard:

Discord: 1samueleonard1
Object of testing: #4996 
Tier of Testing: 1
Date: 04/06/2021

Testing Process: 
. entered chars in scratchpad
. cleared single digits as normal - works
. entered chars in scratchpad
. held button - all chars disapeared

Negatives: n/a 
Testing Results: (Passed)

Conclusions: works

Media: n/a

marcman86:
Category: QA Tester
Name: marcman86#4907
Date of testing: 4.06.2021
Version of the sim: 1.16.2.0
PR Tested: #4996
Tier of testing: 1
Changes to observe: MCDU CLR button hold function
Testing technique: Tryout

1.) Spawned C&D at EDDP
2.) Turned on bat. and ext. power
3.) Typed a few digits into MCDU
4.) Press CLR to delete a single digit
5.) Press and hold CLR to delete a all digits

Negatives: -

Result: -

Conclusion: When pressed and hold first a single digit (the last one) gets cleared before the rest clears. Dont know if this is the wanted behaviour. If yes then this is GTG.

Hope this helps :-)

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -100,6 +100,12 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
             }
             this.tryShowMessage();
         };
+        this.onClrHeld = () => {
```
MisterChocker, line 103: Please follow the same logic as the onClr event...

Also when holding down the clear button, does it clear the first character and then the entire scratchpad?

How does it behave when an type I or type II message is displayed?

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -100,6 +100,12 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
...
+            if (this.inOut !== "" && this.inOut !== FMCMainDisplay.clrValue) {
+                this.inOut = "";
+                this.tryShowMessage();
+            }
+        };
```
MisterChocker, line 108:
```suggestion
            if (this.inOut === "") {
                this.inOut = FMCMainDisplay.clrValue;
            } else if (this.inOut === FMCMainDisplay.clrValue) {
                this.inOut = "";
            } else if (this.isDisplayingErrorMessage || this.isDisplayingTypeTwoMessage) {
                this.tryRemoveMessage();
                this.lastUserInputToScratchpad();
                this._inOutElement.className = "white";
                this.isDisplayingErrorMessage = false;
                this.isDisplayingTypeTwoMessage = false;
            } else {
                this.inOut = "";
            }
            this.tryShowMessage();
        };
```

MisterChocker (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
tracernz: > EDIT: The follow needs to be true
> 
> When holding clear button, our onClr event should not be triggered.

That _cannot_ be true, because:
- we cannot tell in advance whether the user will hold the button down for longer or not,
- the real MCDU does not do this for the same reason

The real MCDU clears the last character on the rising edge, as we do by triggering onClr, and then the rest later if it is held down. Additionally, if the scratchpad is empty, the rising trigger (onClr) adds a CLR to the pad, and holding down has no further affect in this PR. I don't have a video reference for this behaviour, so can't confirm it 100%, but it seems to match the FCOM.

> Also now logic for type I and type II messages need to be added to this function.

This part I now see is a problem. The user input buffer is shared with the FMC messages, which is a design mistake we should probably rectify to simplify the logic, or we can hax around it with additional logic in onClrHeld.

I am busy with work on the new ND and FMGC thus I don't have more time to work on the scratchpad logic for now, so I'll close this.

MisterChocker: True,

when pressing the button the event onClr fires and when then holding the button, the event onClrHold gets executed next.

IMO it still needs to account for a message being displayed when there is only one char on the scratchpad beforehand.

tracernz: Yes, you are correct and I don't think it does that correctly at the moment. I might get back to this one day, or somebody else might.

MisterChocker: Maybe I'll pick up on this one

tracernz: > Maybe I'll pick up on this one

Yep, if you can that'd be good. 👍

PR #5009, opened 2021-06-04
chore: add pre-commit hook

### LINKED ISSUE #4982: Add pre-commit hook to ensure code is linted before allowing push to branch
### Aircraft Version

Development

### Description

Add a pre-commit hook to lint and format code before code can be committed.

This will help reduce the need for resolving linting errors as it will already have been done prior to pushing to branch.

Also, suggest that prettier or another code formatter be executed at pre-commit also to ensure code looks... pretty.

### References (optional)

_No response_

As voiced by several, it's not ideal to have pre-commit hooks for linting as this repo is not just used by JS Devs.

The alternative approach here is to make this an optional feature so it's not executed by those who don't need it, nor by CI. There is no reason it can't be included, just needs to be updated so that it can be opted into by the contributor - perhaps with a command like lint:init - just needs scripts written to handle install and config. For the guys doing a lot of JS work, this is super handy because it catches issues before it even reaches the pipeline lint step which obviously helps maintain the quality of code committed and reduces the need for rework.

antiso

AlexChesters:
Fixes #4982 

## Summary of Changes

Adds a pre-commit hook (using [`husky`](https://www.npmjs.com/package/husky)) that will run `npm run lint` on every commit

## Screenshots (if necessary)

N/A

## References

N/A, only reference used was #4982 itself

## Additional context

* I considered adding `npm run prettier` to this PR as well, but that command seems to result in unintended changes for me (it changes compiled files in `.target/` as well as files in `.github`; if we decide that this PR should also include prettier I can address that in this PR and update .`prettierignore`
* I don't have access to a Linux machine (only Windows) so I haven't been able to test on Linux

Discord username (if different from GitHub): ShermheadRyder#0052

## Testing instructions

Make a change on any branch and run `git commit`


davidwalschots (commented):
I don't think we should include a pre-commit hook.

While trying it I ran into the following things:

1. One has to manually run `npm install` and `husky install` to make it work. This goes against the containerised development environment philosophy.
2. After running `husky install` a `.gitignore` file is created which one would need to commit.
3. Committing takes a long time as the linting takes a long time, this is a big workflow killer for someone who frequently commits (me), especially for those working mostly on Rust code (me) or C++ code.

aguther:
I'm against it, last time we had husky in it was running minutes for every small commit. Sorry no...

AlexChesters:
not adding it is fine by me, can a contributor close #4982 with a summary of those points raised above so contributors know it's not something desired by the project?

antonyisoardi:
The alternative approach here is to make this an optional feature. There is no reason it can't be included, just needs to be updated so that it can be opted into but the contributor - perhaps with a command like `link:init` - just needs scripts written to handle install and config. For the guys doing a lot of JS work, this is super handy because it catches issues before it even reaches the pipeline lint step which obviously helps maintain the quality of code commited and reduces the need for rework.

foestauf:
> I'm against it, last time we had husky in it was running minutes for every small commit. Sorry no...it was a nightmare for devs. Why not fail the build on a PR? That's completely enough and when devs are too lazy to check themselves then they learn it when their PR is not building. Simple as that.

The best approach for that is to combined husky with lint-staged so it's only looking at files on the commit not the whole repo.

foestauf:
> I don't think we should include a pre-commit hook.
> 
> While trying it I ran into the following things:
> 
> 1. One has to manually run `npm install` and `husky install` to make it work. This goes against the containerised development environment philosophy.
> 2. After running `husky install` a `.gitignore` file is created which one would need to commit.
> 3. Committing takes a long time as the linting takes a long time, this is a big workflow killer for someone who frequently commits (me), especially for those working mostly on Rust code (me) or C++ code.

It is true that npm install must be run for the git hook to be installed however husky install is not necessary by every developer.
On point 3 yes if husky is used without lint-staged it will need to parse the entire repo which is very time consuming. lint-staged will only parse through the files on the commit.

tracernz:
husky hijacks the git hooks dir entirely which makes it a) badly behaved, and b) not even trivial to get rid of for people who are not confident with git. Our previous experience with it was very poor for these reasons, and also it's (lack of) speed.

antonyisoardi:
The latest Husky (6.0.0) requires to you explicitly define your hooks, so it doesn't happen at `npm install` anymore.

Even if Husky is installed, there are no hooks defined so it's not going to run anything anyway.

Hijacking of git hooks directory probably isn't going to be an issue either if there are no hooks for anyone other than contributors who decide to use them

As suggested, create a script to init the pre-commit hook, include the `.husky` directory in `.gitignore`.

Also agree with using lint-staged. Limit what needs to be linted. I've been using this combo for a long time and it's never been an issue - and speed comes down to eslint, not Husky.

tracernz:
The issue is, when something goes wrong, and all of a sudden our local builds take ages, it slows our work down, so we look in .git/hooks, as we’ve done for the past 10 years, but there’s nothing in there… what’s going on? Nobody remembers who or what husky is.

PR #5017, opened 2021-06-05
build: quick fix for component errors 🎉

hiaaryan:
## Summary of Changes

Fixes build errors with broken v0.2.9 `react-components`.

Discord username (if different from GitHub): blurpyx#2092

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   },
   "dependencies": {
     "@flybywiresim/api-client": "~0.7.0",
-    "@flybywiresim/react-components": "^0.2.7",
+    "@flybywiresim/react-components": "^0.3.0",
     "@tabler/icons": "^1.41.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
   },
   "dependencies": {
     "@flybywiresim/api-client": "~0.7.0",
-    "@flybywiresim/react-components": "^0.2.7",
+    "@flybywiresim/react-components": "^0.3.0",
```
Saschl, line 65: Should we rather keep the caret or set a specific version?

Benjozork: This version was supposed to be `0.2.10`, but was named `0.3.0` by mistake. This will not be a problem if future versions are named correctly.

Saschl (approved):

sidnov:
Tested the artifact. LGTM

PR #5018, opened 2021-06-05
feat: chrono auto brightness

sidnov:
Fixes #[issue_no]

## Summary of Changes

- A basic CSS-based implementation which modifies the brightness of the Chrono text depending on the time of the day. 
- This feature considers the ~~`LOCAL TIME`~~ `TIME OF DAY` simvar available by default in the sim.
- `TIME OF DAY`, type "enum": Returns 0,1,2,3 which corresponds to dawn, daylight, dusk and night respectively marked as phases of the day. Brightness change should be observed from phase 0->1 and phase 2->3.
- Works in polar regions and similar locations where days with no sunrise/sunsets exist
- This method was considered because of 2 main reasons: there is currently no way to implement the IRL chrono brightness sensor in sim (at least not known to me :P), and the text on the chrono display is not controlled by any brightness knobs either.
- The auto brightness is a fair trade-off between poor text visibility in daylight and an overly bright chrono in the night.

## Screenshots (if necessary)

**BEFORE**
image

**AFTER**
image

## Additional context

## Testing instructions

1. Spawn either on runway or C/D
2. Observe the chrono on the front panel
3. Use the in-sim weather panel to transition through different times of the day.
4. Observe brighter display in daylight and a dimmer display during night.


src/instruments/src/Clock/Components/DayNight.tsx
```diff
@@ -0,0 +1,15 @@
...
+    const displayTime = [0, 0, 0];
+    displayTime[0] = Math.floor(seconds / 3600);
+    return displayTime;
+};
+
+const DayNight = () => {
+    const [localTime] = useSimVar('E:LOCAL TIME', 'seconds', 200);
+    const displayTime = secondsToDisplay(localTime);
+    return ((displayTime[0] >= 6 && displayTime[0] <= 18) ? 'day' : 'night');
+};
```
beheh, line 13: You can't do this. Either DayNight is a React component, or it's not. If it is, you can use hooks, but you need to actually render and return JSX, not a string. If it is not (and it's just a normal function), you cannot use hooks. [These are the rules of hooks](https://reactjs.org/docs/hooks-rules.html).

To fix this you could make DayNight an actual function like `getDayNightClassName` and pass it localTime as a parameter.

beheh (dismissed):

src/instruments/src/Clock/Components/DayNight.tsx
sidnov: @beheh Thanks for the feedback! As you can see with this goof up above, I'm fairly new to React. 😄 I'll look into it and make the suggested changes

tracernz: >  [These are the rules of hooks](https://reactjs.org/docs/hooks-rules.html).

I'm in the same boat as sidnov. This page helped explain some things I saw, so thank you. 👍

krangerich:
A suggestion if you want to make it more sophisticated: There are sunset/sunrise algorithms, which depend on date and location (lat/lon).
https://edwilliams.org/sunrise_sunset_algorithm.htm

(I've no clue, how this works, I just googled it)

src/instruments/src/Clock/Components/DayNight.tsx
```diff
@@ -0,0 +1,7 @@
+function getDayNightState(seconds: number) {
+    const localHour = Math.floor(seconds / 3600);
+    const timeOfDay = (localHour >= 6 && localHour <= 18) ? 'day' : 'night';
+    return timeOfDay;
+}
+
+export default getDayNightState;
```
beheh, line 7: While this is better, with it being in a Components folder this could still be mistaken for a component. You could move it to a separate file in the Clock directory and give it some kind of more generic filename, like `utils` or `helpers`.

sidnov:
> A suggestion if you want to make it more sophisticated: There are sunset/sunrise algorithms, which depend on date and location (lat/lon).
> https://edwilliams.org/sunrise_sunset_algorithm.htm
> 
> (I've no clue, how this works, I just googled it)

I like the idea, couldn't work on it though for the past few days. Having a look rn

sidnov:
Update:
TIME OF DAY, type "enum": Returns 0,1,2 and 3 which corresponds to dawn, daylight, dusk and night respectively marked as phases of the day. Brightness change should be observed from phase 3->0 and phase 1->2.
Works in polar regions and similar locations where some days with no sunrise/sunset exist.

MisterChocker (commented):
Wondering if it makes sense to implement a smooth transition, to kind of fade out/in the brightness change.

sidnov:
> Wondering if it makes sense to implement a smooth transition, to kind of fade out/in the brightness change.

YES, it's possible and it should give a better effect overall. Trying it out now :)

sidnov:
@MisterChocker 7 second transition
https://user-images.githubusercontent.com/29005005/122742049-3542bb00-d2a3-11eb-88e2-8d0268182c1c.mp4

TheSledge-SLEJ:
"This method was considered because of 2 main reasons: there is currently no way to implement the IRL chrono brightness sensor in sim (at least not known to me :P), and the text on the chrono display is not controlled by any brightness knobs either"

Prob with this is ignoring light changes for sun and cloud affecting cockpit during daylight hours. Consider measuring the changes in RGB or HSL (Luminance) values of the background adjacent the digital display itself? This must be captured somewhere to display the shadow fall or (no) direct sunlight effect?

sidnov:
> "This method was considered because of 2 main reasons: there is currently no way to implement the IRL chrono brightness sensor in sim (at least not known to me :P), and the text on the chrono display is not controlled by any brightness knobs either"
> 
> Prob with this is ignoring light changes for sun and cloud affecting cockpit during daylight hours. Consider measuring the changes in RGB or HSL (Luminance) values of the background adjacent the digital display itself? This must be captured somewhere to display the shadow fall or (no) direct sunlight effect?

We could only wish there are such advanced techniques in sim that could help us capture those parameters, but that's not possible with the current state of the sim. The current brightness mode on the other hand is somewhat a median of different daylight conditions, be it sunny or cloudy. For reference,

TheSledge-SLEJ:
> > <snip>there is currently no way to implement the IRL chrono brightness sensor in sim (at least not known to me :P), <snip>
> 
> We could only wish there are such advanced techniques in sim that could help us capture those parameters <snip>

Yh. I figured you may well be following the best possible current course. As you were and thanks for all the effort.

davidwalschots (approved):
Cool stuff!

AviationSpotter:

Object of testing: #5018
Tier of Testing: 1
Date: 27/06/2021

Testing Process:

1. Spawned C&D
2. Switched "Time of Day" to daytime and night time
3. After switching the time of day, observed chrono auto brightness

Negatives: N/A
Testing Results: Passed

Conclusions: Auto brightness works like a charm! (Too bad auto brightness can't adjust when the sun is blocked by clouds)

PR #3625, opened 2021-02-19
feat(elec): power consumption and batteries

davidwalschots:
## Summary of Changes
The electrical system developed in #2582 does not yet include power consumption. This means the load on origins of electric potential is always 0%. This PR aims to provide a semi-realistic power consumption simulation. It does so by:

- Approximately simulating power consumption during different phases of flight, based on data provided by @komp1821.
- Providing programmatic types to simulate power usage of individual components, such that the approximate simulation above can slowly be replaced by an exact variant.

## Review recommendations
This PR is rather large. For code reviewing I have the following recommendations in terms of files to review:
- [src/systems/a320_systems/src/power_consumption.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-7cc551d958928ec423f659ddb4a65de17e83f520ba101d467cf99be51ed4c62b)
- [src/systems/systems/src/electrical/battery.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-6f6ac61925c52a35b92ca7ca8199e65df85885e5074eb99fcbeeeadc8de8ca16)
- [src/systems/systems/src/electrical/battery_charge_limiter.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-6f6ac61925c52a35b92ca7ca8199e65df85885e5074eb99fcbeeeadc8de8ca16)
- [src/systems/systems/src/electrical/consumption.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-6f6ac61925c52a35b92ca7ca8199e65df85885e5074eb99fcbeeeadc8de8ca16)
- [src/systems/systems/src/electrical/mod.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-c36ef03a9c19212e4229162e88bd2c7e9a8a46a615b662bd4b74b7e238f51265) regarding the `Potential` type.
- [src/systems/systems/src/simulation/test.rs](https://github.com/flybywiresim/a32nx/pull/3625/files#diff-6f6ac61925c52a35b92ca7ca8199e65df85885e5074eb99fcbeeeadc8de8ca16)

## Technical changes

- [x] Testing of types implementing `SimulationElement` is easier with a `SimulationTestBed` which invokes all the functions as they would be invoked when running in the simulator itself.

## Functionality
- [x] Power consumption during different phases of flight, with randomisation of consumption over time (i.e. base usage lookup table for phase of flight where randomisation sets actual consumption to 90-110% of base usage every 5 to 30 seconds).

[...]

PR #2582, opened 2020-12-21
feat: electrical systems

davidwalschots:
## Summary of Changes
This PR implements the electrical systems using Rust and WASM. Initial design thoughts concerning systems in Rust can be found [here](https://github.com/davidwalschots/rfcs/blob/systems-design/text/000-systems-design.md).

The electrical system contains many features which shouldn't all be implemented as part of a single PR. Therefore, this PR intents to provide a foundation to build upon. **The single most important thing to note is that the inner workings of the MSFS electrical system are _not_ yet replaced.**

### Note for code review

This is the second big system related PR. As I'm learning more about Rust every day some of my learnings have also become part of this PR. These learnings do not only affect the ELEC system, but also the APU created earlier. I have attempted to keep things "small" by not including features such as power consumption, generator load, and the effect on V and Hz as electrical demand fluctuates. Despite all this, the PR is still huge. Thanks for bearing with me in all this and thanks for taking the effort of reviewing my work. I know it's a long one once again...

### Thanks to

Big thanks to @komp1821 for providing information on all the little details.

#### Engine generator

The engine generator is not yet fully implemented. Some code that existed as part of an earlier C++ implementation has been translated into Rust. The engine generator will have to be revisited as part of a separate PR. For now:

- [x] When engine N2 is at or above 58%, the integrated drive generator (IDG) will supply electricity after 0,5 seconds.
- [x] When engine N2 decreases to below 56%, the IDG stops supplying electricity.
- [x] When supplying electricity, the output is: 115V, 400Hz, and 0% load. _Fluctuations in V, Hz and load calculation will be added later_.
- [x] When not supplying electricty, the output is: 0V, 0 Hz, and 0% load.
- [x] A running engine increases IDG oil temperature.
- [x] IDG oil temperature decreases to ambient when the IDG is disconnected.
- [x] IDG oil temperature decreases to ambient when the engine is not running.

#### APU generator

The APU generator is unchanged.

#### External power

- [x] When external power is available, the output is: 115V, and 400Hz.
- [x] When external power is not available, the output is: 0V, and 0Hz.

#### Emergency generator

There is some code for the emergency generator. However, as it requires a hydraulics implementation it hasn't been fully implemented yet.

#### Static inverter

- [x] When the static inverter is powered by battery 1 , the output is: 115V, and 400Hz. _Fluctuations in V and Hz will be added later_.
- [x] When the static inverter is not powered, the output is: 0V, and 0Hz.

#### Transformer Rectifiers

[...]

PR #5043, opened 2021-06-07
refactor(elec): prepare for advanced scenarios

davidwalschots:
## Summary of Changes

In #2582 I implemented the electrical system. #3625 extended this work with power consumption. While working on #3625 I noticed that my earliest design had a flaw related to the `Potential` struct. 

### Current implementation
Currently, instances of this struct are copied while passed between electrical elements such as contactors, buses, etc. This causes (at least) the following problems:
- To enable a scenario in which batteries charge each other I currently bounce the `Potential` back and forth: DC<sub>1</sub> → DC<sub>BAT</sub> → BAT<sub>1</sub> → DC<sub>BAT</sub> → BAT<sub>2</sub> → DC<sub>BAT</sub> → BAT<sub>1</sub>.
- Simulating correct APU start power draw is hard. The TR is limited to a maximum of 300A with the batteries picking up the remaining power drawn.

Thus, it hinders the implementation of various advanced scenarios, some of which @tracernz would be very happy to add at some point, but are too difficult to add right now.

### New implementation

In this PR I will fix this design mistake by having a single `Potential` instance per actual potential in the system.

Consider a situation where both engine generators and both batteries are connected, four `Potential` instances (P<sub>1</sub> to P<sub>4</sub>) would then be found in the system. These would have the following origins:

- Origin(P<sub>1</sub>) = GEN<sub>1</sub>
- Origin(P<sub>2</sub>) = GEN<sub>2</sub>
- Origin(P<sub>3</sub>) = TR<sub>1</sub> + BAT<sub>1</sub> + BAT<sub>2</sub>
- Origin(P<sub>4</sub>) = TR<sub>2</sub>

With the following elements referencing each (contactors omitted):

- Ref(P<sub>1</sub>) = { GEN<sub>1</sub>, AC<sub>1</sub>, AC<sub>ESS</sub>, AC<sub>ESS SHED</sub>, TR<sub>1 (input)</sub> }
- Ref(P<sub>2</sub>) = { GEN<sub>2</sub>, AC<sub>2</sub>, TR<sub>2 (input)</sub> }
- Ref(P<sub>3</sub>) = { TR<sub>1 (output)</sub>, DC<sub>1</sub>, DC<sub>BAT</sub>, DC<sub>ESS</sub>, DC<sub>ESS SHED</sub>, BAT<sub>1</sub>, HOT<sub>1</sub>, BAT<sub>2</sub>, HOT<sub>2</sub> }
- Ref(P<sub>4</sub>) = { TR<sub>2 (output)</sub>, DC<sub>2</sub> }

## Code reviewing

This PR again is very large, as there are some necessary improvements that must be done in order to make testing with the newly introduced `Electricity` type less convoluted.

Primary files to review are:
- src/systems/systems/src/electrical/mod.rs
- src/systems/systems/src/simulation/test.rs

### Electricity

The newly introduced `Electricity` type itself handles the references to `Potential`. With it one can `flow` `Potential` between `ElectricalElement`s and `transform` it into another `Potential`. `ElectricalElement`s all have one or two identifiers which are handed out to them when the `Aircraft` is constructed. Using `String` for this would have removed the need to pass `&mut Electricity` when constructing `Aircraft`, however I've chosen to not use it, as to not add the burden of creating additional data on the heap during every request for an `ElectricalElement`'s identifier(s).

With `Electricity` created and a requirement for constructing an `Aircraft` and running a `Simulation.tick` on it, I ran into the problem that every single test and/or custom `TestBed` implementation needed to instantiate `Electricity` and pass it to the `Aircraft`. This wasn't great, also because it , and thus I decided to move `Electricity` into the `Simulation` type, thereby circumventing the need to instantiating it everywhere.

### Default `TestAircraft`

Accessing the default `TestAircraft`'s element (T) in `SimulationTestBed<TestAircraft<T>>` then became a pain, with one having to write `test_bed.aircraft().element()` or `test_bed.aircraft_mut().element_mut()`. Through one of the glorious features of Rust, that could be shortened to `test_bed.element()` and `test_bed.element_mut()`:

```rust
impl<T: SimulationElement> SimulationTestBed<TestAircraft<T>> {
    pub fn element_mut(&mut self) -> &mut T {
        self.aircraft_mut().element_mut()
    }

    pub fn element(&self) -> &T {
        self.aircraft().element()
    }
}
```

### Instantiating default `TestAircraft`

Instantiating a new `SimulationTestBed<TestAircraft<T>>` when one doesn't care about `Electricity` was also a pain, which is why I added a `From<SimulationElement>` implementation, resulting in the following improvement:

**Before**
```rust
SimulationTestBed::new(|_| TestAircraft::new(LandingGear::new()));
```
**After**
```rust
SimulationTestBed::from(LandingGear::new());
```

Similarly, I provided a version for `SimulationTestBed<TestAircraft<T>>` where one _does_ care about `Electricity`:

**Before**
```rust
SimulationTestBed::new(|electricity| TestAircraft::new(contactor(electricity)));
```
**After**
```rust
SimulationTestBed::from(ElementCtorFn(contactor));
```

### `SimulationTestBed` includes `Simulation` instance

As `Electricity` and `Aircraft` became inseparable within the `Simulation` instance, as a result the `SimulationTestBed` had to include `Simulation` and thus change from `SimulationTestBed` to `SimulationTestBed<T: Aircraft>`. This in turn enabled the creation of a common testing interface. 

There was a lack of testing interface commonality between custom test bed implementations. I.e. the `A320ElectricalTestBed` had functions which differed from `A320HydraulicsTestBed` or the default `SimulationTestBed`. The `TestBed` trait, which should be implemented by all custom `TestBed`s solves this problem. As a result, each `TestBed` now exposes the following functions (default implementation not shown):

```rust
trait TestBed {
    fn run(&mut self);
    fn run_with_delta(&mut self, delta: Duration);

    fn command<V: FnOnce(&mut Self::Aircraft)>(&mut self, func: V);
    fn query<V: Fn(&Self::Aircraft) -> W, W>(&self, func: V) -> W;
    fn query_elec<V: Fn(&Self::Aircraft, &Electricity) -> W, W>(&self, func: V) -> W;

    fn set_indicated_airspeed(&mut self, indicated_airspeed: Velocity);
    fn indicated_airspeed(&mut self) -> Velocity;
    fn set_indicated_altitude(&mut self, indicated_altitude: Length);
    fn set_ambient_temperature(&mut self, ambient_temperature: ThermodynamicTemperature);
    fn set_on_ground(&mut self, on_ground: bool);

    fn contains_key(&self, name: &str) -> bool;
}
```

Of biggest interest are `command`, which lets one perform mutations on the `Aircraft` contained within the `TestBed`:
```rust
self.command(|a| a.set_battery_push_button_off());
```

`query`, which lets one request information from the `Aircraft` contained within the `TestBed`:
```rust
self.query(|a| a.battery_contactor_is_closed())
```
and `query_elec` which is similar to `query`, but includes the `Electricity` type:
```rust
self.query_elec(|a, elec| a.generator_is_unpowered(elec))
```

### Common `read` and `write` interface for simulator variables

The `Read` and `Write` traits which had implementations previously defined on the `SimulatorReader` and `SimulatorWriter` have now been abstracted further, such that those implementations can not only be used on the `SimulatorReader` and `SimulatorWriter`, but also on any `TestBed`. This should help any new developer get into the code base more easily, as there are less differences between the testing and actual implementation portion of the code, for example:
**Before**
```rust
fn green_pressure(&mut self) -> Pressure {
    Pressure::new::<psi>(self.read_f64("HYD_GREEN_PRESSURE"))
}
```
**After**
```rust
fn green_pressure(&mut self) -> Pressure {
    self.read("HYD_GREEN_PRESSURE")
}
```

## Additional context


tracernz (approved):
Absolutely brilliant, both in terms of implementation, and the branch nicely broken into reviewable commits. 🥇

davidwalschots:
Thanks for reviewing once again @tracernz. I appreciate the compliments. 
Merged it, as I like livin' on the edge with all the automated tests 🤪.

PR #5062, opened 2021-06-08
fix: synchronising park brake lever value from sim

crocket63:
## Summary of Changes

WASM mechanism for park brakes wasn't allowing to read from .FLT files variable changes when starting a session.
As a consequence, starting in flight was with parking brake on by default.

## Additional context

## Testing instructions

1. Starting on ground the variable A32NX_PARK_BRAKE_LEVER_POS should be 1, and park brake lever in cockpit should be ON
2. Starting in flight the variable A32NX_PARK_BRAKE_LEVER_POS should be 0, and park brake lever in cockpit should be OFF
3. Parking brake function normally using clickable lever or keyboard bindings


davidwalschots (approved):
I wonder if the variables as set in the `.flt` file are always available when the WASM is loaded, or if the variables might be set afterwards. I do recommend testing this. It would be great if we could add an `Init` trait to the `SimulatorAspect`, and then only execute that code once.

Irregardless of the outcome, the current code LGTM.

crocket63:
> I wonder if the variables as set in the `.flt` file are always available when the WASM is loaded, or if the variables might be set afterwards. I do recommend testing this. It would be great if we could add an `Init` trait to the `SimulatorAspect`, and then only execute that code once.
> 
> Irregardless of the outcome, the current code LGTM.

Not sure FLT process is really repeatable indeed....  But in that particular case we might want to do this always so that if any external software tries to write the var of park lever it would react accordingly.

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #5062
Tier of Testing: 1
Date: 8.6.2021

Testing Process: 
Protocol:
Starting on ground the variable A32NX_PARK_BRAKE_LEVER_POS should be 1, and park brake lever in cockpit should be ON
==> OK (normal function in UI, key binding and LVAR (SPAD)
Starting in flight the variable A32NX_PARK_BRAKE_LEVER_POS should be 0, and park brake lever in cockpit should be OFF
==> OK (normal function in UI, key binding and LVAR (SPAD - ECAM warning msg comes up as expected)
Parking brake function normally using clickable lever or keyboard bindings
==> OK

Negatives: none
Testing Results: Passed
Conclusions: LGTM

Samueleonard:

Discord: 1samueleonard1
Object of testing: (#5062)
Tier of Testing: (1)
Date: 08/06/2021

Testing Process:

loaded on stand - parking brake set
loaded on runway - parking brake set
loaded in air - parking brake released

Negatives: n/a
Testing Results: (Passed)

Conclusions: works as intended
Media: n/a

tracernz (approved):

PR #5044, opened 2021-06-07
fix(mcdu): init-a tropo entry validation

MisterChocker:
Fixes #5032

## Summary of Changes

- Fixed tropo field input handling as per reference

## References

See linked issue

## Additional context

Discord username (if different from GitHub): DerL30N#3751

## Testing instructions

Manipulate tropo field/value on init-a
_allowed format is NNNN or NNNNN_

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #5067, opened 2021-06-08
feat(eicas): Show ISA DEV on EICAS

### LINKED ISSUE #5029: DELTA ISA <*| Optional feature
### Aircraft Version

Development

### Description

When the CM1 altimeter knob is pulled (STD MODE) and there is a valid SAT value, the ISA DEV is shown like this:

DSC-31-20x

### References (optional)

Refer to DSC-31-20 for further info:

DSC-31-20

_No response_

Illuminati

BlueberryKing:
Fixes #5029 

## Summary of Changes
The ISA DEV is now shown on the EICAS if the ADIRS are aligned and the altimeter knob is set to STD. This is computed by:
```
ISA = 15 - (AltStd * 1.98 / 1000)
delta ISA = SAT - ISA
```
where AltStd is the altimeter reading while barometer is set to Std. The international standard atmosphere (ISA) model is expected to have a temperature of 15 °C at sea level and decrease by about 1.98 °C / 1000 ft up to the tropopause. At which point it stays at -56.5 °C. The EICAS shows the deviation of the current actual air temperature to the temperature predicted by ISA. Thank you to @lukecologne and @donstim for explaining!

Furthermore, I right-aligned all temperature values.

## Screenshots (if necessary)
Before: (still what you see if altimeter is not set to std)
aligned-not-std
After:
right-aligned-std

## References
- FCOM
- Discord user Illuminati#0055 (type rated on the A320) opened the PR which should validate this logic

## Additional context

## Testing instructions
- Spawn in cold and dark, power up the airplane without aligning the ADIRS. SAT and TAT should be shown in amber XX. ISA is not visible.
- Align ADIRS and make sure ISA is only shown if altimeter setting is STD.
- Ensure all temperature values are right-aligned, meaning they should all have the same spacing to the °C symbol.
- Change ISA in the ingame weather settings. Changes should be reflected on the EICAS.
- Take off and climb up to about FL 380 (EXP CLB is helpful for this). ISA should stay about the same during the whole climb while SAT and TAT decrease.
- Ensure in particular that ISA does not change once passing tropopause at 36089 feet. SAT and TAT should stop decreasing too.


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.html
```diff
@@ -22,6 +22,12 @@
             <text class="Unit" x="150" y="87" text-anchor="start">&#176;C</text>
         </g>
 
+		<g id="ISA">
```
tracernz, line 25: Looks like an alignment issue there?

lukecologne:
ALTStd does not mean the standard altimeter setting, but rather the barometric altitude with the standard reference (i.e. altimeter setting in STD). What this is supposed to show is the difference between the current temperature and the International standard atmosphere  temperature (ISA), so `ΔISA = TAT - ISA`, with `ISA = 15°C - h * 6.5°C/1000m ≈ 15°C - h * 2/1000 °C/ft`

ExampleWasTaken:
What about implementing `getISA()` in [A32NX_Util](https://github.com/flybywiresim/a32nx/blob/master/flybywire-aircraft-a320-neo/html_ui/JS/A32NX_Util.js) to enable other features to access it in the future?

donstim:
> ALTStd does not mean the standard altimeter setting, but rather the barometric altitude with the standard reference (i.e. altimeter setting in STD). What this is supposed to show is the difference between the current temperature and the International standard atmosphere temperature (ISA), so `ΔISA = TAT - ISA`, with `ISA = 15°C - h * 6.5°C/1000m ≈ 15°C - h * 2/1000 °C/ft`

I think you mean 'ΔISA = **S**AT - ISA`. **Below the tropopause (36089 ft),** ISA, in degrees Celsius is 15 - 0.0019812 * h, where h is the true height above sea level in feet. At the tropopause and above (for altitudes within the A320 operating envelope), ISA is a constant -56.5° C.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -135,19 +163,27 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
                 this.tatText.classList.add("Value");
                 this.tatText.classList.remove("Warning");
+                this.isaText.classList.add("Value");
+                this.isaText.classList.remove("Warning");
             }
         }
     }
+    getISA() {
+        // EIS2 S13-2 ref
+        return this.currentSAT + (1013 * 2 / 1000) - 15;
```
MisterChocker, line 186: (1013 * 2 / 1000) - 15 could be simplified to a const 

```suggestion
        return this.currentSAT -12,974;
```

wondering why the Isa calculation is that different from the fmgc...

image

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -13,6 +13,9 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
     init() {
         this.tatText = this.querySelector("#TATValue");
         this.satText = this.querySelector("#SATValue");
+        this.isaText = this.querySelector("#ISAValue");
+        this.isaContainer = this.querySelector("#ISA");
+        this.areAdirsAligned = SimVar.GetSimVarValue("L:A320_Neo_ADIRS_STATE", "Enum") == 2;
```
MisterChocker, line 18: also what is it with the areAdirsAligned?

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -135,19 +163,27 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
     refreshADIRS() {
         if (this.tatText != null && this.satText != null) {
             if (SimVar.GetSimVarValue("L:A320_Neo_ADIRS_STATE", "Enum") != 2) {
+                this.areAdirsAligned = false;
```
MisterChocker, line 166: same here

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -135,19 +163,27 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
                 this.tatText.textContent = "XX";
                 this.tatText.classList.add("Warning");
                 this.tatText.classList.remove("Value");
                 this.satText.textContent = "XX";
                 this.satText.classList.add("Warning");
                 this.satText.classList.remove("Value");
             } else {
+                this.areAdirsAligned = true;
```
MisterChocker, line 174: and here

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
BlueberryKing: From the FCOM, I gathered that ISA is only shown when there is a valid SAT, which is when ADIRS are aligned (or I assume that's what `SimVar.GetSimVarValue("L:A320_Neo_ADIRS_STATE", "Enum") != 2` means). But since I need to check for aligned ADIRS and whether altimeter is STD, I just save the SimVar value of `refreshADIRS()` in an extra variable, so I don't need to access it again in `refreshISA()`.

BlueberryKing:
Thank you for everyone's comments. I will fix the ISA calculation to match the FMGC's. However, I will extract the calculation into `A32NX_Util.js`, so the code can be used by the FMGC, EICAS and anything else that might be using it in the future.

> Just make sure that at 36089 ft and above, ISA temp is a constant -56.5° C.

Regarding this, should I use this constant or actually access the tropopause which can be entered in the FMGC?

MisterChocker:
> > Just make sure that at 36089 ft and above, ISA temp is a constant -56.5° C.
> 
> Regarding this, should I use this constant or actually access the tropopause which can be entered in the FMGC?

Yeah that would be a nice idea!

BlueberryKing:
I have implemented the requested changes and updated the PR to include testing instructions to fly up to FL 380 (over the tropopause) and observe behavior of ISA.

For now, I just used 36089 ft as tropopause until the correct behavior is confirmed. Also, I am not sure how I would read the FMC entered value from the EICAS. Is this what SimVars are for?

donstim:
> > > Just make sure that at 36089 ft and above, ISA temp is a constant -56.5° C.
> > 
> > 
> > Regarding this, should I use this constant or actually access the tropopause which can be entered in the FMGC?
> 
> Yeah that would be a nice idea!
> 
> @donstim would this actually be correct tho?

Yes, I forgot that you can enter a different value for trop in the MCDU.

donstim:
Whatever value is used for the temperature lapse rate (2/1000, 1.98/10000, or 1.9812/1000), why not just go ahead and do the division and enter only the constant (0.002, 0.00198, or 0.0019812)? It may only save a picosecond of calculation time, but why make the code do that each time?

lukecologne:
The formula that you got was correct, the EIS doesn't care at all about the tropopause, probably they decided it wasn't really relevant.

> I think you mean 'ΔISA = SAT - ISA`

Yep, was a typo, corrected it

donstim:
Don't know what you mean by the tropopause not being relevant. The tropopause is inherently part of the definition of the international standard atmosphere (ISA). Surely, the displayed ISA deviation value in the airplane doesn't leave it out, does it? That would be a major error xD!

MisterChocker:
@BlueberryKing to use the tropo value entered in the mcdu, we'll need to create a new local variable.

The fmgc rounds the tropo to the last 10s, so in this case it will round up.
This is what it would look like when integrated into the tropo updater (FMCMainDisplay):

`SimVar.SetSimVarValue("L:A32NX_TROPO", "number", 36090);` also needs to be added to the init.

image

flybywire-aircraft-a320-neo/html_ui/JS/A32NX_Util.js
```diff
@@ -125,6 +125,25 @@ A32NX_Util.greatCircleIntersection = (latlon1, brg1, latlon2, brg2) => {
...
 };
 
+/**
+ * Returns the ISA temperature for a given altitude
+ * @param alt {number} altitude in ft
+ * @returns {number} ISA temp in C°
+ */
+A32NX_Util.getIsaTemp = (alt = Simplane.getAltitude()) => {
+    const altOrTropopause = Math.min(alt, 36089);
+    return altOrTropopause / 1000 * (-1.98) + 15;
```
MisterChocker, line 135:
```suggestion
    return Math.min(alt, 36089) * 0.0019812 + 15;
```

BlueberryKing:
Thank you for the example, @MisterChocker. :) I can implement it this way.

I think it's worth pointing out that Wikipedia has a section on the ICAO Standard Atmosphere https://en.wikipedia.org/wiki/International_Standard_Atmosphere#ICAO_Standard_Atmosphere.

This simply assumes the tropopause to be at 11000 m or 36000 ft with no mention of taking into account where the tropopause actually is on a given day in a given region which would speak against using the tropopause field in the first place. Of course, we don't know if that's what Airbus actually does.

donstim:
> This simply assumes the tropopause to be at 11000 m or 36000 ft with no mention of taking into account where the tropopause actually is on a given day in a given region which would speak against using the tropopause field in the first place. Of course, we don't know if that's what Airbus actually does.

11000 m is 36089 ft.  (For some reason, the MCDU rounds this to 36090.) You are correct that we don't know for sure whether the TROP value entered in the MCDU affects this particular display. That value may only be used in MCDU performance predictions. May be prudent not to use this value for the ISA deviation display unless you can get confirmation that it is based on the MCDU TROP entry.

BlueberryKing:
Ok, I'm trying to get feedback from pilots, but it's quite the edge case. Apart from that, the PR should be ready.

MisterChocker (approved):
LGTM

donstim:
> Ok, I'm trying to get feedback from pilots, but it's quite the edge case. Apart from that, the PR should be ready.

Thinking about it some more, as you noted above ISA has a defined profile including the troposphere. If the tropopause changes, it no longer conforms to ISA, just like the temperature deviations. Based on that, I don't think it would be correct to adjust the tropopause level when calculating the ISA deviation. ISA is ISA and does not change. (Well, unless ICAO amends it or adopts a different standard.)

BlueberryKing:
> Thinking about it some more, as you noted above ISA has a defined profile including the troposphere. If the tropopause changes, it no longer conforms to ISA, just like the temperature deviations. Based on that, I don't think it would be correct to adjust the tropopause level when calculating the ISA deviation. ISA is ISA and does not change. (Well, unless ICAO amends it or adopts a different standard.)

You are correct. I just confirmed this on Discord. If we adjusted ISA based on conditions and tropo entry, it wouldn't make sense to display a deviation from it in the first place

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.html
```diff
@@ -22,6 +22,12 @@
             <text class="Unit" x="150" y="87" text-anchor="start">&#176;C</text>
         </g>
 
+        <g id="ISA">
+            <text class="Title" x="74" y="111" text-anchor="end">ISA</text>
+            <text id="ISAValue" x="112" y="111" text-anchor="middle" class="Value">XX</text>
+            <text class="Unit" x="150" y="111" text-anchor="start">&#176;C</text>
+        </g>
```
bouveng, line 29: Hi! When referring to your images, the aligning is a bit wrong (also for the other ones).

These should be right aligned and not centered as your single digit shows.

Try to set the `text-anchor` to `end`, and then adjust the `x`-axis accordingly, for **all** these 3 svg groups.

Reference: 
isa

Mock/p.o.c in actual svg:
isa-svg

Br.

/J

BlueberryKing:
Ah, good catch! Temperature values should now be properly aligned. I don't have a good eye for this kind of stuff, so let me know if you want me to nudge them to the left or right a little more. I updated the "After" picture in the PR to reflect the change.

bouveng:
@BlueberryKing Alignment looks perfect now! I have a good eye and are picky with minor details. =))

MisterChocker:
_I'll add the needs changelog just so we don't forget to add it later on_

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -67,6 +73,27 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
...
+
+        this.currentISA = _value;
+        if (this.isaText != null) {
+            if (this.currentISA > 0) {
+                this.isaText.textContent = "+" + this.currentISA.toString();
+            } else {
+                this.isaText.textContent = this.currentISA.toString();
+            }
+        }
+        //}
```
beheh, line 95:
```suggestion
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
```diff
@@ -67,6 +73,27 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
         }
         //}
     }
+    refreshISA(_value, _force = false) {
+        //if ((_value != this.currentISA) || _force) {
```
beheh, line 77:
```suggestion
```
I'm not sure why we still have this in the other functions. I would suggest to not add it back.

BlueberryKing: Works for me. In my opinion, we should also remove it from the other functions then, including the `_force` parameter. It looks like the commented out lines have been there since implementation. It's unlikely someone is gonna want to uncomment them again. If it turns out this logic is required again for some reason, it's easy enough to reimplement. Would you agree?

BlueberryKing: Actually, thinking about it now. I think we should put the check for `_value != this.currentISA` back in for everything (TAT, SAT, ISA) to avoid unnecessary DOM updates, but leave out the `_force` parameter until there's a need for it.

TheSquirtingPantin:
Quality Assurance Tester

Discord: Remi#8871
Object of testing: #5044
Tier of Testing: 1
Date: 10/06/2021

Testing Process:
1) CD at KMIA
2) Full flight from KMIA to KDEN. Observed the variation of the DELTA ISA value throughout flight

Negatives: None. 
Testing Results: Passed. 

Conclusions: LGTM. DELTA ISA is always calculated correctly. Logic for display on DELTA ISA is correct (ADIRS and STD Altimeter setting). ISA temp locked at -56 reaching FL360. 
Media: N/A

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
Benjozork: @BlueberryKing I already did that in a branch, and nothing broke. So you can re-add it safely.

BlueberryKing: @Benjozork Are you talking about putting the if statement back in? Because I ran into an issue where, once the ADIRS aligned, I would still see "XX" but in green for SAT and TAT because the `this.currentSAT` had not changed under the hood, so with this code, there was no need to rewrite the actual temperature values to the DOM until the temperature actually changes. We would at least need to force an update once the ADIRS aligned, if I identified correctly.

In my view, there are two options: We leave the `if` commented out, updating the DOM every time. This is not optimized, but we don't need to change any code and the QA is still valid. Or I could implement some optimizations to only update the DOM when needed, which would probably need to be retested. I don't mind which option we choose.

Benjozork: I'd go with the second one. Actually I just remembered this but and I had to switch to tracking the entire text and not just the temperature value.

BlueberryKing: Sounds like a good solution. At this point, I will also make sure that `refreshADIRS()` doesn't trigger an update every frame.

marcman86:
Discord:marcman86#4907
Object of testing: #5067
Tier of Testing: 1
Date: 11/06/2021

Testing Process: 
- Spawned C&D at EDDB
- Turned on bat. and ext. power > TAT XX, SAT XX, ISA invisible
- Align ADIRS > TAT +26°C, SAT +26°C, ISA invisible
- Altimeter to STD > TAT +26°C, SAT +26°C, ISA +11°C
- Took off to FL380


Positives: 
- Digit alignment looks fine
- On Climb ISA stays almost the same
- ISA value stays at -56°C above FL360 (Fig1, Fig2)

Negatives: -
Testing Results: Passed
Conclusions: Works
Media: 


Fig1


Fig2

Hope this helps :-)

BlueberryKing:
Alright, I implemented some caching to only update the DOM when needed. I don't think another QA round is needed, but I've been notoriously wrong about this in the past.

BlueberryKing:
@Illuminati270 I am not sure I am understanding you correctly. Please correct me. Is the point of ISA DEV not to show how different the current conditions are from the standard atmosphere, which does not change? 

So, if the tropopause in Ushuaia happens to be at FL300 today, then we expect the SAT to stop decreasing at FL300. This is not modeled in the simulator apparently, but if it were, we would expect for the ISA DEV to _increase_ between FL300 and ISA tropopause FL360, since ISA temperature decreases and SAT stays the same, so the ISA DEV = SAT - ISA will increase. At FL 360 and above, SAT, ISA, and ISA DEV would then stay the same.

Illuminati270:
> @Illuminati270 I am not sure I am understanding you correctly. Please correct me. Is the point of ISA DEV not to show how different the current conditions are from the standard atmosphere, which does not change?
> 
> So, if the tropopause in Ushuaia happens to be at FL300 today, then we expect the SAT to stop decreasing at FL300. This is not modeled in the simulator apparently, but if it were, we would expect for the ISA DEV to _increase_ between FL300 and ISA tropopause FL360, since ISA temperature decreases and SAT stays the same, so the ISA DEV = SAT - ISA will increase. At FL 360 and above, SAT, ISA, and ISA DEV would then stay the same.

You got me correctly.

The whole thing here is: ISA DEV is related to **ISA TEMP at TROPO** and not related to ISA TROPO ALT.

So, given the following:

TROPO at FL340
Aircraft flying at FL340
SAT -56
ISA DEV should read 0

TROPO at FL340
Aircraft flying at FL380
SAT -56
ISA DEV should read 0

TROPO at FL340
Aircraft flying at FL340
SAT -50
ISA DEV should read +6

TROPO at FL340
Aircraft flying at FL380
SAT -50
ISA DEV should read +6
 

Also:

I downloaded the new artifact, and I can see that the ISA DEV works (No longer XX), but now the SAT doesn't dected the real-world temperature anymore. It will always show ISA temps.

Not sure if that happened to this specific build, but the Thrust Levers scale turned to 0% throttle being full rev and 100% throttle being TOGA instead of 0% throttle being Idle.

BlueberryKing:
But looking at this, ISA temperature seems to always be -56 °C, so the ISA DEV shown on ECAM is always just SAT - (-56) regardless of where you are flying and where the tropopause is?

Illuminati270:
Not quite.

The aircraft will recognize the lack of temperature change and understand it has reached the tropopause. At that stage, it will base calculations on -56 rather than constant thermal gradient.

BlueberryKing:
Ah, let's say the tropopause was at FL 300 and we fly at FL 290 with SAT -50 °C. Since we have not reached tropopause yet, the plane assumes the tropopause is at FL 360 as defined by ISA. At FL 290 the ISA temperature would then be about -42.5 °C. So the ECAM shows ISA DEV -7.5 °C. However, as soon as we cross the tropopause at FL 300 and the plane detects that SAT has stopped decreasing (say it was still at -50 °C), the plane assumes ISA at FL 300 is -56 °C, so ISA DEV on the ECAM will jump to +6 °C?

Illuminati270:
> Ah, let's say the tropopause was at FL 300 and we fly at FL 290 with SAT -50 °C. Since we have not reached tropopause yet, the plane assumes the tropopause is at FL 360 as defined by ISA. At FL 290 the ISA temperature would then be about -42.5 °C. So the ECAM shows ISA DEV -7.5 °C. However, as soon as we cross the tropopause at FL 300 and the plane detects that SAT has stopped decreasing (say it was still at -50 °C), the plane assumes ISA at FL 300 is -56 °C, so ISA DEV on the ECAM will jump to +6 °C?

That's it! You got it.

You'll hardly such a temperature jump in only 1,000ft. But that's essentially as you now described.

donstim:
> 
> The whole thing here is: ISA DEV is related to **ISA TEMP at TROPO** and not related to ISA TROPO ALT.
No, the whole thing is that ISA DEV is related to measured SAT compared to the ISA temperature. This does not change for a tropo alt that is different from the ISA tropo alt.
> So, given the following:
> 
> TROPO at FL340
> Aircraft flying at FL340
> SAT -56
> ISA DEV should read 0
No, ISA temp at FL340 is -52.4. ISA DEV should read -4
> 
> TROPO at FL340
> Aircraft flying at FL380
> SAT -56
> ISA DEV should read 0
This one is correct.
> 
> TROPO at FL340
> Aircraft flying at FL340
> SAT -50
> ISA DEV should read +6
No, ISA DEV should read +2

> TROPO at FL340
> Aircraft flying at FL380
> SAT -50
> ISA DEV should read +6
This one is also correct.

Are you sure about this? The ISA is a defined standard. It doesn't vary, even though it does not represent the real world atmosphere at any particular time or place.

donstim:
> > Ah, let's say the tropopause was at FL 300 and we fly at FL 290 with SAT -50 °C. Since we have not reached tropopause yet, the plane assumes the tropopause is at FL 360 as defined by ISA. At FL 290 the ISA temperature would then be about -42.5 °C. So the ECAM shows ISA DEV -7.5 °C. However, as soon as we cross the tropopause at FL 300 and the plane detects that SAT has stopped decreasing (say it was still at -50 °C), the plane assumes ISA at FL 300 is -56 °C, so ISA DEV on the ECAM will jump to +6 °C?
> 
> That's it! You got it.
> 
> That's essentially as you now described.
> 
> Because we actually provide the expected Tropopause level at INIT-A, the temperature profile won't be surprised by actual flight conditions. So that happens really smoothly.

If that's what it does, then the ECAM display is not showing the actual ISA DEV. Do you have any video/screenshots of the real airplane to support this claim? I have one in the #private references channel on Discord that does not support this (unless the IRL Tropo is actually at 36089 for that flight, which would be a real coincidence).

Illuminati270:
> > > Ah, let's say the tropopause was at FL 300 and we fly at FL 290 with SAT -50 °C. Since we have not reached tropopause yet, the plane assumes the tropopause is at FL 360 as defined by ISA. At FL 290 the ISA temperature would then be about -42.5 °C. So the ECAM shows ISA DEV -7.5 °C. However, as soon as we cross the tropopause at FL 300 and the plane detects that SAT has stopped decreasing (say it was still at -50 °C), the plane assumes ISA at FL 300 is -56 °C, so ISA DEV on the ECAM will jump to +6 °C?
> > 
> > 
> > That's it! You got it.
> > That's essentially as you now described.
> > Because we actually provide the expected Tropopause level at INIT-A, the temperature profile won't be surprised by actual flight conditions. So that happens really smoothly.
> 
> If that's what it does, then the ECAM display is not showing the actual ISA DEV. Do you have any video/screenshots from the airplane to support this claim? See the one I have in the #private references channel on Discord.

Did a bit of research through all Airbus aircraft, and to my surprise, you're actually right.

Funny though, is that the aircraft's ceiling capability would be improved along the climb (Taking QRH as a reference).

donstim:
> Did a bit of research through all Airbus aircraft, and to my surprise, you're actually right.
> 
LOL. It's happened a few times in my life. I try not to make it a habit, though!

The trop input (and sensing) is used to make the vertical flight path prediction more accurate, where, for example, the true airspeed when flying a constant Mach changes with temperature.

Benjozork:
@BlueberryKing Please add a changelog

BlueberryKing:
@Benjozork Done.

PR #4799, opened 2021-05-13
feat(hyd): hydraulics base refactoring

crocket63:
Fixes #5077

## Summary of Changes

This is a core hydraulics pressure regulation overhaul.
Currently we are computing max pumping capacity and using what is needed for pump regulation deep inside.
This refactoring aims at making the pump regulation easier to understand, better, and will expose clearly pump states like pump actual displacement (currently already existing but hidden deep inside).

- [x] Each hydraulic circuit is separated in two sections: main pump section (EDP or blue elec pump), and secondary section with electric pumps and various consumers. 
This will expose two different hydraulic pressures in each section like in the real plane.
- [x] Pressure regulation is split into clearer steps for each section
- [x] Pump characteristics are now the actual pumps max pump characteristic of real pumps. The regulation will lookup those characteristics and position the actual displacement of each pump accordingly so it is easy to expose this in external variables (pump sound need displacement to be accurate).
- [x] Reservoirs are now actual objects. Making this refactoring will make it easy to model pneumatic pressurising and pump cavitation for too low pressurisation levels in a later PR
- [x] Build the actual logic of the HYD PTU ecam memo
- [x] NW STRG DISC ecam memo computed from hydraulics
- [x] Core hydraulics code clean up
- [x] Replace current A320 hydraulic with the new core structure

## Additional context

Discord username (if different from GitHub):

## Testing instructions

Known issues:
- Blue elec pump has to fight valve restriction of its pump section and this is causing a blue system slower to reach 3000psi. Valve system will receive further improvements in later dedicated PR.

Testing:

1. This is a complete overhaul of already available hydraulic system. In that sense, the main testing work should be focused on the fact we don't have regressions from the previous system.
Attention should be focused on: 

- Pressure values doesn't show unrealistic instabilities in nominal conditions
- Playing with the overheads buttons doesn't cause strange pressure instabilities
- Faults on buttons / Amber on ECAM hyd screen are still showing when pressure is lower than expected
- Reservoir volumes are changing accordingly while hydraulics are in use, and are not higher / lower after a period of time where we get back to same plane state.

Going deeper, the behaviour of pumps vs system sections can be checked by looking at pressures:
A32NX_HYD_{loop_name}_PUMP1_SECTION_PRESSURE
A32NX_HYD_{loop_name}_SYSTEM1_SECTION_PRESSURE
For example, when engine pump is off, pressure of pump section should get back to 0. Meanwhile, system section can still stay at 3000psi if an electric pump is available, or ptu, or RAT.
[...]

PR #5097, opened 2021-06-12
fix(ecam): Fix HYD PTU message on ECAM

### LINKED ISSUE #5077: HYD PTU memo on ECAM incorrect behaviour
### Aircraft Version

Development

### Build info

```json
{
  "built": "2021-06-09T07:52:45+00:00",
  "ref": "master",
  "sha": "e3cda27aa50d6a8173363ccb7ae256888dc67842",
  "actor": "davidwalschots",
  "event_name": "manual"
}
```

### Describe the bug

On a single engine taxi procedure, after you start engine 1 and turn on the hyd yellow eletric pump, you can check on the HYD sys page that the PTU is correctally inactive, since pressure is beeing provided by engine 1 on the blue sys and the eletric pump on the green sys

Even tho the PTU is OFF as confirmed by the ECAM HYD page, the HYD PTU memo is present on the ECAM.

On a related matter HYD PTU memo also shows on the ground even if the PTU and ELEC pumps are off and pressure on the Yellow sys is 0 on the ground (screenshot included)

Also, HYD PTU memo does not show in flight even if it's in fact active due to ENG 2 (or 1) PUMP beeing set off or if the engine 2 (or 1) itself is off

### Expected behavior

HYD PTU ecam memo should only appear if the PTU is in fact active (HYD pressure difference between yellow and green systems > 500 PSI

### Steps to reproduce

1. Go throught normal pre-flight
2. Perform a single engine taxi out procedure:
2b.Start engine 1
2c Set ignition normal
2d. Set APU bleed off
2e. Set Yellow Hyd eletric pump ON
3. Verify PTU ON memo on ECAM, very that the PTU is actually of in the HYD sys page

HYD PTU memo will also show if Eng1 is on, Yellow electric pump is off and there is not pressure on the yellow line (PTU inactive on the ground)

### References (optional)

image

image

image

image

image

image

image

image

_No response_

raelias #5157

BlueberryKing:
Fixes #5077 

## Summary of Changes
The logic for when "HYD PTU" showed on the ECAM was incorrect. More correctly, the memo should now show whenever the PTU is on according to the new hydraulics system implementation.

This is only a temporary solution until the FWC/FWS is implemented properly. This is not an optimal solution. If the performance hit is too hard. Feel free to just reject the PR, so the logic can be implemented properly later on.

## Screenshots (if necessary)
HYD PTU message is no longer shown after engine 1 start:
eng-start-no-ptu
HYD PTU message is shown with 2 engines, but yellow hydraulics pump turned off and PTU on HYD ECAM page (also works for other side):
yellow-hyd-pump-off-ptu

## References

- @raeliass who filed the issue is a type rated A320 pilot linked references in issue #5077.

## Additional context

## Testing instructions
- Spawn in C/D, go through normal preflight.
- Perform a single engine taxi out procedure:
  - Start engine 1
  - Set ignition normal
  - Set APU bleed off
  - Set Yellow Hyd eletric pump ON
- At no point during this should the HYD PTU memo show on the ECAM, neither should the PTU come active on the lower ECAM HYD page.
- Turn off yellow electric pump
- Start engine 2
- Turn off engine 2 hydraulics pump (yellow) on the overhead panel, verify PTU comes on on lower ECAM and HYD PTU message shows.
- Explicitly turn off PTU valve on the overhead panel, verify PTU turns off on lower ECAM and HYD PTU message is not shown.
- Turn PTU valve back on
- Turn on yellow electric pump on the overhead panel, verify PTU turns off on lower ECAM and HYD PTU message is not shown. 
- Turn off yellow electric pump
- Turn off engine 1 hydraulics pump (green) on the overhead panel, verify PTU comes on on lower ECAM and HYD PTU message shows.

I am not very familiar with the hydraulics system, so I would be thankful if anyone testing the PR could also try other known scenarios where they would expect the PTU to come on/stay off.


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -105,6 +105,9 @@ var A320_Neo_UpperECAM;
             this.iceNotDetTimer1 = new NXLogic_ConfirmNode(60);
             this.iceNotDetTimer2 = new NXLogic_ConfirmNode(130);
             this.predWsMemo = new NXLogic_MemoryNode(true);
+            this.showHydPtuMemoTimer = new NXLogic_ConfirmNode(0);
```
MisterChocker, line 108: the `NXLogic_ConfirmNode` parameter is a the amount of time in ms it takes until the state becomes true.

So I am actually not sure whether a confirmation node is actually needed, but if so, it probably should not be 0ms and 2ms.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -1262,7 +1265,7 @@ var A320_Neo_UpperECAM;
                         isActive: () => {
                             // Rough approximation until hydraulic system fully implemented
                             // Needs the last 2 conditions because PTU_ON is (incorrectly) permanently set to true during first engine start
-                            return (this.getCachedSimVar("L:XMLVAR_PTU_ON", "Bool") == 1) && (SimVar.GetSimVarValue("ENG N1 RPM:1", "Percent") < 1 || SimVar.GetSimVarValue("ENG N1 RPM:2", "Percent") < 1);
+                            return this.hydPtuMemoVisible.read();
```
MisterChocker, line 1268: Would it be possible to just replace the condition, because I don't see the need and understand the reason behind the usage of confirmation and memory nodes. But please do explain your thoughts on this so I may be able to understand you doing so.

`this.getCachedSimVar("L:A32NX_HYD_PTU_ACTIVE_L2R", "bool") || this.getCachedSimVar("L:A32NX_HYD_PTU_ACTIVE_R2L", "bool")`

BlueberryKing: I think the argument in `NXLogic_ConfirmNode` is the time in seconds, not milliseconds, right? I was running into a problem where the HYD PTU message would flicker on and off because the PTU pressurizes in pulses, essentially it pressurizes the system to 3000 psi, then lets off for about a second after which it pressurizes again. To prevent this, I wanted to have the HYD PTU message show immediately when PTU flow was detected but only disappear if it hadn't done anything in 2s. 

Let me know if this can be done maybe with only one ConfirmNode or maybe with none at all?

MisterChocker: You are totally right, it's in seconds.. my bad.

raeliass: > I think the argument in `NXLogic_ConfirmNode` is the time in seconds, not milliseconds, right? I was running into a problem where the HYD PTU message would flicker on and off because the PTU pressurizes in pulses, essentially it pressurizes the system to 3000 psi, then lets off for about a second after which it pressurizes again. To prevent this, I wanted to have the HYD PTU message show immediately when PTU flow was detected but only disappear if it hadn't done anything in 2s.
> 
> Let me know if this can be done maybe with only one ConfirmNode or maybe with none at all?

Yeah the PTU simbology stays on in the lower ecam HYD page but it pressurizes in pulses like you said, maybe you could use the condition for the lower ecam simbology and copy it?

BlueberryKing: I did actually try and do that, but still ran into the flickering issue. I don't know why this doesn't happen on the lower ECAM to be honest.

raeliass (approved):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
MisterChocker: The pr is not yet ready for testing and it's not labeled as such.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
raeliass: There is a side effect on this fix: During the second engine start, the PTU self-tests itself (the famous barking sound in the cabin) during this PTU activation the HYD PTU ecam memo should be inhibited but it briefly apears as the system build pressure and deactivates it, it's worth noting that in the sim that PTU is not actually doing it's self test, but activating as soon as you set the second engine master to on since it detecs that the engine is on but there is no hyd pressure on that side, might be a deeper hyd system issue not yet fesible to implement a fix, but if possible, inhibit the HYD PTU message until the second engine finishes starting

crocket63: I think there is no actual "PTU self test" per say as far as I understand. When starting the second engine, it's just supposed to activate PTU valve, and computer just waits for pressure to rise to 3000 in less than X seconds to validate PTU test (as the engine driven pump is not supposed to provide enough flow on the very begining of engine spool up). I don't know if there is a specific condition to inhibit the message on upper ecam though.

BlueberryKing: I think their point is simply that, while it is correct for the PTU to come on at this stage, the HYD PTU memo should not appear on the ECAM. I am not sure how I should inhibit this. Checking if both engines are available is not correct since the PTU should come on in the event of an engine failure.

crocket63: Just checked the logic and there's actually no inhibit condition. If edp pressure is low on one side, and system pressure is >1450 on both sides it should show basically

raeliass: > Just checked the logic and there's actually no inhibit condition. If edp pressure is low on one side, and system pressure is >1450 on both sides it should show basically

Like I explained, the PTU self-test itself on second engine start, but it does not show the HYD PTU memo during the test or the engines start as a whole

https://www.youtube.com/watch?v=L94f0ad9x3Q

MisterChocker: you could use the FWC flight phase to determine whether to actually show the message or not.

BlueberryKing:
Thank you for the input everybody. I am now checking `L:A32NX_FWC_FLIGHT_PHASE` to determine whether to show it or not. I could not find documentation on this variable, but from looking at the ModelBehavior variables, `L:A32NX_FWC_FLIGHT_PHASE >= 2` seems appropriate. One issue I've noticed is that it takes about 30 sec for the flight phase to update from 1 to 2 when spawning on the runway, so the memo will not be shown if you turn off a hydraulic pump before that.

raeliass:
I think thats ok, I mean when you dont spawn cold and dark it is to be expected some system logic error, for instance the rat seem to be active on the hyd page if you start hot, even tho its not deployed

Great work!

davidwalschots (commented):
I'd like to note that according to the documentation, the HYD PTU message isn't inhibited in any flight phase. Thus, if you decide to keep it in the code, then it is best to add a comment indicating why you added it, and stating that this shouldn't really be inhibited.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -105,6 +105,9 @@ var A320_Neo_UpperECAM;
             this.iceNotDetTimer1 = new NXLogic_ConfirmNode(60);
             this.iceNotDetTimer2 = new NXLogic_ConfirmNode(130);
             this.predWsMemo = new NXLogic_MemoryNode(true);
+            this.ptuOnTimer = new NXLogic_ConfirmNode(0);
```
beheh, line 108: You should never use 0 for this. Either you need a confirmation timer, or you don't.

beheh (changes requested):
As has been mentioned elsewhere, this implementation is fairly inaccurate and uses the flight phase inhibition to fake an inhibition where there is none. In reality there are additional requirements for a pressure difference in addition to the PTU Valve being open that trigger this memo. I'm not sure if this (fairly small) PR is even worth merging, as the accuracy is only marginally improved.

crocket63:
> As has been mentioned elsewhere, this implementation is fairly inaccurate and uses the flight phase inhibition to fake an inhibition where there is none. In reality there are additional requirements for a pressure difference in addition to the PTU Valve being open that trigger this memo. I'm not sure if this (fairly small) PR is even worth merging, as the accuracy is only marginally improved.

Problem is we have the needed inputs already for real logic to be implemented, but I can't see how it would not show on engine startup.  So either I just did not see a simple thing, which is totally possible, or this is down to a set of things to do right like PTU model, engine N2 speed profile and oil pressure rising time....
Sadly didn't have time to test the actual logic on what we currently have to see what is missing.

beheh:
I think what could solve this open question about the memo is a video of a second engine start with the HYD page open. Then we can look at the precise pressure timings and figure out why it's not showing.

raeliass:
> I think what could solve this open question about the memo is a video of a second engine start with the HYD page open. Then we can look at the precise pressure timings and figure out why it's not showing.
> 
> As it stands, I'd rather not merge this PR as-is.

I think that's irrelevant really, regardless of pressures during the second engine start the HYD PTU memo NEVER shows, you can observe this in any number of videos, but it will be very difficult for you to find an engine start with the HYD page open since the ENG page is used to monitor the engine start

The Airbus manual are knowingly lacking in some "useless" (and some useful) info, something as trivial as this ecam memo inhibition logic is something that wouldn't be mentioned anywhere

Benjozork:
> The Airbus manual are knowingly lacking in some "useless" (and some useful) info, something as trivial as this ecam memo inhibition logic is something that wouldn't be mentioned anywhere

The logic we have for the FWC is exact. The condition probably lies in one of the inputs to the FWC.

beheh:
@BlueberryKing With #4799 being where it is right now, we can probably close this?

BlueberryKing:
Absolutely! Closing since proper logic is going to be in place soon.

PR #5115, opened 2021-06-14
feat(isis): ISIS auto brightness

### LINKED ISSUE #5101: ISIS auto brightness
### Aircraft Version

Development

### Description

Hey,

There is a light sensor on the A320NEO on the ISIS (Integrated System Instrument Standby, here PFD shown on this screen). The brightness is therefore adjusted according to the ambient light.

It is a small black square.

Capture d’écran (99)_LI

### References (optional)

_No response_

_No response_

AXWB#8487

BlueberryKing:
Fixes #5101

## Summary of Changes
Automatically set ISIS (Integrated Standby Instrument System) to appropriate brightness based on time of day. The real aircraft does this using an ambient light sensor. To my knowledge, there is no way to extract this value in the simulator though, so we do what a simulator does and simulate it by checking whether it is daytime where the airplane is which typically means that sunlight falls into the cockpit causing the ISIS to brighten up. We check for daytime using the `E:TIME OF DAY` SimVar. This is the same logic the automatic brightness control the chrono instrument uses (#5018). When a change in the time of day, the brightness setting transitions from to opposite setting.

No change in brightness will be observed if ambient light changes in the cockpit by other means that time of day (dome lights, clouds, etc.)

The manual brightness buttons still work by adjusting the increasing/decreasing the current brightness by some granularity. The manually set brightness is then taken over as the new day/night preset value. If the brightness is adjusted during a transition, the same happens and the transition is halted.

It is quite simple for me to adjust levels of brightness for the default day/night presets if desired.

## Screenshots (if necessary)

Before:
isis-before
After:
isis-after

## References

- AntoineXWB who opened issue 5101 is an Airbus engineer and provided information on how the system works in the real aircraft. We try to simulate this.

## Additional context

## Testing instructions

- Spawn in (C/D or on the runway). 
- The ISIS should have a reasonable brightness level.
- Use the in-game menu to change the time of day either from day to night or night to day.
- Observe how the brightness transitions to the new brightness setting (the transition should take about 7 seconds, but it can take up to 10 seconds to start). Again, the ISIS should be at a reasonable setting for this time of day.
- Manually change the brightness of the instrument using the buttons to the left. This should behave as it previously has; adjusting brightness between a minimum and maximum. Holding down a button should also behave as if the button were repeatedly pressed.
- For maximum effect, set an extreme brightness setting manually (either full brightness or full darkness).
- Change time of day and observe that instrument is in extreme setting it had previously been set to manually.


MisterChocker (commented):
Wondering if it makes sense to implement a smooth transition, to kind of fade out/in the brightness change.

BlueberryKing:
@MisterChocker I have now implemented a smoother transition. Unfortunately, I couldn't really use CSS for it, because I need to handle the manual brightness buttons still.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/SAI/A320_Neo_SAI.js
```diff
@@ -1125,13 +1125,30 @@ class A320_Neo_SAI_Brightness extends NavSystemElement {
...
+        if (!this.isInTransition && this.updateThrottler.canUpdate(_deltaTime) === -1) {
+            return;
+        }
+
+        if (this.isDaytime !== this.getIsDaytime()) {
+            this.isDaytime = this.getIsDaytime();
+            this.isInTransition = true;
+        }
+
+        this.updateTransition(_deltaTime);
```
MisterChocker, line 1151: You should never ever overrule the updateThrottler.

this.updateTransitions seems to only have one use case so I don't see the need to create a new method for it.

Generally there is one key issue, the "day" brightness can be reduced below the "night" brightness, so the brightness may increase on a day->night transition.

I am still looking further into it..

MisterChocker (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/SAI/A320_Neo_SAI.js
MisterChocker: I've taken a look at how to solve these issues, I've made a pr to your fork with a suggestion on how to solve the issue.
https://github.com/BlueberryKing/a32nx/pull/1

BlueberryKing: Thank you for your effort! I will keep the correct usage for `updateThrottler` in mind. Your changes look good. The only thing I've noticed as a side effect of not having the "reverse" brightness transitions is that the brightness does not go back to your set presets, so if you increase the brightness at night from 0.5 to 0.7, then it turns daytime, so brightness is 0.85, but after it turns night again, it will go back to 0.5, not 0.7

To be honest, this is quite a rare scenario and I'm not even sure it's necessarily a bad thing. 

I think we should have some people test it with the goal of the system working without them ever having to touch it, possibly using the buttons in case of clouds or something.

MisterChocker: Well if this actually is a thing, we still can use the diff to calculate the offset and adopt it to the new target brightness anyways...

So basically apply the brightness changes to the new "preset" like this?
Well, I kind of like that idea, ngl.

```js
this.targetBrightness = isDaytime ?
    Math.max(Math.min(this.maximum, this.targetBrightness * 2 - this.nightBrightness), this.dayBrightness) : 
    Math.min(Math.min(this.maximum, this.targetBrightness * 2 - this.dayBrightness), this.nightBrightness);
```

BlueberryKing: Some of those `Math.max`, `Math.min` and `this.maximum` look wrong. Now, I'm not sure what you're proposing 😄. To implement what I'm saying, we would probably need to modify `this.dayBrightness` and `this.nightBrightness` at some point just to save the "preset" value somewhere. Right now, these values are (purposely) never mutated.

We just need to make sure that, if, for example, `this.dayBrightness` were to drop below `this.nightBrightness` through manual adjustment, we also "push" `this.nightBrightness` down.

We might be overengineering this a bit, but I'll think about it some more.

MisterChocker: That's exactly what my proposal does 😅 it pushes the offset to the next "phase" and maintains the difference while also ensuring we are not exceeding the maximum brightness,that's what the "inner" Math.Max is for.

MisterChocker: Alright this is the switcher, it will ensure the corrected value is within min max range and will adjust for the difference

The switch happens by multiplying the difference of day and night normal values (.45) with 1 or -1 depending on the transition.

```js
this.targetBrightness = Math.max(this.minimum, Math.min(this.maximum, this.targetBrightness - (this.nightBrightness - this.dayBrightness) * (isDaytime ? 1 : -1)));
```

BlueberryKing: Ah, now this looks like what I had in mind with the offset to the presets. I think there is still a small inconsistency in this scenario:
At night, you manually adjust from 0.5 to 0.8, so offset is 0.3. During the automatic transition to daytime, the value gets clamped from (0.85 + 0.3 = 1.15) to 0.99. Now the offset is only 0.99 - 0.85 = 0.14, so on the next nighttime transition, brightness will be 0.5 + 0.14 = 0.69, not 0.8 as you originally set it.

Admittedly, this is a very constructed example and not necessarily unexpected behavior.

MisterChocker: Keep in mind that 0.3 means that you've increased the brightens by 30% (and that's in the night) and end up with a brightness 5% less than daylight brightness. (with this setting the isis basically is a flashbang)
I don't really see this scenario being realistic.

I'd say lets add the dynamic change and let people play around with it during QA and gather feedback.
We can still adjust the behavior afterwards with a new pr if people dislike the behavior.

BlueberryKing: Fair enough, sounds good 👍.

MisterChocker (approved):
LGTM

BlueberryKing:
Hey, thanks for those changes. Will add the changelog tonight. Sorry if my PR caused you more work than it saved lol.

MisterChocker:
No worries, helping were I can ;)

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5115 
Tier of Testing: 1
Date: 23/06/2021

**Testing Process:**
1. Loaded in on RW in daylight
2. Set nighttime
3. Set manual low brightness
4. Set daytime

**Negatives:** N/A
**Testing Results:** Passed

AviationSpotter:

Object of testing: #5115
Tier of Testing: 1
Date: 23/06/2021

Testing Process: 

1. Spawned C&D
2. Changed time of day
3. Monitored ISIS

Negatives: N/A
Testing Results: Passed

Conclusions: The auto brightness is pretty smooth. If were not looking for a change of brightness, you probably wouldn't notice the fade out/fade in! :)

dragold:
Changes to observe:
-Automatically set ISIS (Integrated Standby Instrument System) to appropriate brightness based on time of day.
-The manually set brightness is then taken over as the new day/night preset value. If the brightness is adjusted during a transition
-the same happens and the transition is halted.


Testing technique: Tryout

Methodology:
-Quickstart
-- Random Time input
-- Random button input
-- Random Time input post button input (brightness)
-- Random button input(brightness) pos time input
-- Random button input(brightness) Time set (Live) 

-- Time set (Live) post button input (brightness)(Failed)
The process to  reproduce the issue
1- go to the airport with live time is day
2- rapid random selection(make sure you select) on the time bar then set it to nighttime
3- rapid  button input (brightness "-")
4- set time to LIVE
 
Media-   https://cdn.discordapp.com/attachments/754130199804772372/857385518882750475/2021-06-24_01-22-10.mkv

Observed Observation- "manually set brightness is overtaken by the auto as the new day/night preset value" it did not after said takeover time that is 10 seconds

Expected Observation- manually set brightness is to be overtaken by the auto as the new day/night preset value by or at 10seconds.

references- https://ptb.discord.com/channels/738864299392630914/754130199804772372/857385425630265365

Negatives: refer above

Testing Results: Fail 

Conclusions: I suggest this for a code review and reference with an IRL A320 pilot... Good luck best wishes

TheSquirtingPantin:
Quality Assurance Tester

Discord: Remi#8871
Object of testing: #5115
Tier of Testing: 1
Date: 23/06/2021

Testing Process:
1) C/D at KDEN
2) Changed time, multiple adjustment to manual brightness
3) Reproduced scenario from previous negative points

Negatives: Same negative points than above. But an overall positive feeling with the improvement. Auto Brightness logic following the time of day is correct. 
Testing Results: Passed. 

Conclusions: Passed due to the negative point only regarding a single scenario not affecting the user experience (rapid change of lighting condition alongside manual input). Change of brightness smooth and barely noticeable. 
Media: N/A

AntoineXWB:
Hey I have a video in #private references to help you on Discord. Maybe a bit late :/

PR #5086, opened 2021-06-10
refactor: (ecam) Convert Cruise page to React

RichardPilbery:
Fixes #[issue_no]

## Summary of Changes

Conversion of Cruise page to React

## Screenshots (if necessary)

ECAM-CRUISE-2
ECAM-CRUISE-1

## Additional context

Discord username (if different from GitHub):

## Testing instructions

1. Programme a flight with a destination
2. Fly it
3. Once airbourne check the following on the  Cruise page:
     a. Check LDG ELEV is auto and the elevation shown matches your destination elevation to the nearest 50ft
     b. Adjust the landing elevation dial on the overhead panel and check elevation changes
     c. Press the cabin pressure mod select button to show MAN and check that arrow is replaced by gauge
     d. Check correct arrow (up or down) is shown when cabin pressurisation is increasing/decreasing 
     e. Adjust the overhead cockpint, forward and aft cabin temperature dials and the temperatures in the bottom left of the screen should start to change (may take 10-30 seconds)

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #5128, opened 2021-06-16
docs: adds pages (/docs) for table formatted simvars and events

frankkopp:
## Summary of Changes
Add pages to /docs for simvars and events in table format.  
Removes old style pages (content is current as of 18.6., including changes from PR# 5135 and PR#4970

## References
Example: http://aviation.allanville.com/misc/fsx-pmdg-737-ngx

Discord username (if different from GitHub): Cdr_Maverick

## Testing instructions
Compare existing pages with content and check for copy & paste mistakes :)
General mistakes - typos, format, etc. 
Please check or give feedback regarding categorization into the correct ATA chapters - pilot or engineer feedback appreciated.


docs/a320-simvars-table.md
```diff
@@ -0,0 +1,284 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
davidwalschots, line 153: **issue**: Items which have some placeholder (e.g. `{name}`) have that placeholder name and the appropriate values listed in odd ways. For example, in:
> Bool: True when the push button is ON {name}

Having `{name}` at the end doesn't make sense.

davidwalschots (commented):
A few thoughts:
- Consider using a column for the type, instead of it being part of the description.
- Pick one of the options for the `Name/ID` column. I'd go with `Name`.
- Consider adding a vscode extension recommendation to the project to format tables in the editor, such as `takumii.markdowntable`.

docs/a320-simvars-table.md
frankkopp: Thanks for the feedback - was already thinking along the same lines. Tried to do it 1:1 first.
Will invest more time to improve - adding column and looking through the placeholders

For Markdown I use "Markdown Navigator Enhanced" for Clion (payware). Not sure this is a good recommendation as it is not free. Will do some research for a free alternative (for vscode). Where would you put that recommendation - in the PR or on the page itself?

davidwalschots: You could try `takumii.markdowntable` in vscode, it might be good enough.

Extension recommendations can be added to `.vscode/extensions.json`. That way they'll automatically be recommended when a developer opens the project's folder in vscode.

frankkopp: Is it helpful to also separate data type from unit? 
Bool, Integer, Float vs. Number, Percent, Celcius, Hz, etc.?
(most do not specify data type currently but unit - e.g. Volts for A32NX_ELEC_{name}_POTENTIAL - I think that is a float)

davidwalschots: It's probably best to have one column with the unit, as that is what one uses in code:

```javascript
SimVar.GetSimVarValue("L:A32NX_ELEC_APU_GEN_1_POTENTIAL", "Volts")
```

frankkopp: @davidwalschots do you think sorting for column name would be helpful? I would say yes as I think there is no real coordinated sorting otherwise.

davidwalschots: I'm not sure what sorting means? As far as I know GitHub doesn't support sorting columns.

frankkopp: Would be a one time sort with the table editor plugin. Right now ot does not seem to have an order at all.

davidwalschots: I have ordered the simvars I added such that the ELEC simvars are (mostly) together. We should probably categorise those (and also HYD).

frankkopp: Alright - will do and within the groups I will sort alphabetically - the names support sorting and grouping anyway nearly automatically

frankkopp: I have updated the table:

- added more categories (might have overdone it - feedback welcome)
- categories should be checked by someone with more A320 system knowledge - would be best to stick to the correct naming and system-groups
- added some missing LVARS and documentation

Also added a list of all LVARS (discovered through SPAD.next) for additional documentation in future PRs. The list might already be somewhat helpful for users looking for a specific LVAR.

I would like to recommend the following:

- quickly get this rolled out so that devs can start using this table instead to avoid that the two pages drifting apart. 
- get additional feedback for the categories - there are several ways to structure this - e.g. by system or by place in cockpit
- I'll be happy to continue to work on this to build an excellent documentation for everybody who wants to use hardware to control the plane
- Maybe the team should create a general naming convention and also do a cleanup round to make naming of existing LVARs more consistent and logical (better now then wait even longer as this will break stuff for users who use the LVARs)

davidwalschots: The old format should definitely be removed as part of this PR. We won't be maintaining two versions of the same thing.

Benjozork:
I don't know how I feel about this. I preferred the old format. I see little advantages to this.

frankkopp:
> I don't know how I feel about this. I preferred the old format. I see little advantages to this.

I always thought it should be a table for better readability. I  aleready started this some weeks ago but did not continue as I got used to the old format. A few days ago I got feedback on the SPAD Discord that this list is really hard to read and I picked up the idea again. Especially if this list continues to grow even more which it will as still several LVARS are missing. 

For someone who wants to configure hardware mapping it is currently still very much reverse engineering. E.g. I did all the lighting with the SPAD monitor. It would have helped to have a little more documentation there. 

Do you think the table would have disadvantages?

frankkopp:
Compare these:

Old format
image

New format
image

I really think it makes a huge difference and it still can be improved further in the future (within the limits of github MD pages)

Benjozork:
What exactly is this improving?

The example you gave is a list of boolean simvars. It looks better with those. But look at this:

image

image

And compare to this:

image

This is impossible to write. The old format didn't necessitate HTML tags, or to deal with markdown tables.

I don't necessarily dislike the tables themselves, but it actually is more difficult to read on small screens.

frankkopp:
I agree that writing the table in MD is a bit more difficult but I have the reader in mind as I think developers know how to use the right plugins.

The section you chose is indeed a more complex one but on my screen it looks like this:
image
(use the plugin and do not wrap lines)

I find this even easier to manage for devs than the old style. But having long list within a cell is indeed a bit less readable as MD forces us to pack everything into one line with <br\/>. 

Still this is how I work and everybody how uses the plugin could work:
image

I also think looking ahead bringing this doc to the website a more dynamic HTML/JS table could improve the user experience even further. Sortable, Collabbse/Expand, etc.

frankkopp:
Maybe we should NOT use "variables" but use a line for each variation instead:
(Especially when there are many variations as in this example)

image

Not much more work - just some additional copy&paste for the benefit of readability.

docs/a320-simvar-list.md
```diff
@@ -0,0 +1,692 @@
+# List all LVARs for the A32NX
```
davidwalschots, line 1: **issue**: Duplicate information, which won't be maintained by developers.
**suggestion**: Remove this file and only keep `docs/a320-simvars-table.md`

docs/a320-simvars-table.md
```diff
@@ -0,0 +1,550 @@
...
+| A32NX_AUTOTHRUST_TLA:{index}                                               | Number (degrees) | Indicates the TLA of the throttle lever {index}, first throttle lever has index 1 <br/>Position = Value <br/>REVERSE = -20 <br/>REV_IDLE = -6 <br/>IDLE = 0 <br/>CLB = 25 <br/>FLX/MCT = 35 <br/>TOGA = 45 |
+| A32NX_THROTTLE_MAPPING_A32NX_<br/>THROTTLE_MAPPING_{name1}_{name2}:{index} | Number           | Indicates the low or high value to latch into the given detent<br/>Range is from -1 to 1<br/>{name1}<br/>REVERSE<br/>REVERSE_IDLE<br/>IDLE<br/>CLIMB<br/>FLEXMCT<br/>TOGA<br/>{name2}<br/>LOW<br/>HIGH     |
+| A32NX_THROTTLE_MAPPING_INPUT:{index}                                       | Number           | Indicates the raw input values for throttle axis {index}, first axis has index 1<br/>Range is from -1 to 1                                                                                                 |
```
davidwalschots, line 548: **question**: Should these also be split?

docs/a320-simvar-list.md
frankkopp: Ok, will do. Can be recreated any time anyway.

docs/a320-simvars-table.md
frankkopp: With split you mean expanding without {index} or moving the A32NX_AUTOTHRUST_TLA to Autothrust?
I will expand {name} into lines. I think the {index} is fine, ok?

davidwalschots (approved):

MisterChocker:
Can we remove the A32NX from each entry and just say at the beginning A32NX local variables

like this

| Name | Type | Description |
|:-----------------------------------|:---------------|:---------------------------------------------------------------------------------------|
| _AVIONICS_COMPLT_ON           | Boolean        | True if avionics comp lt is on.                                                        |
| _CREW_HEAD_SET | Boolean | True if CVR crew head set is being pressed. |
...

Also can we please name the file to what's inside? Such as a320-local-variables

And rename `# A320neo Local SimVars` to `# A320neo Local Variables`

And for the events I'd like to see the event type specified.

frankkopp:
> Can we remove the A32NX from each entry and just say at the beginning A32NX local variables
> 
> like this
> 
> Name	Type	Description
> _AVIONICS_COMPLT_ON	Boolean	True if avionics comp lt is on.
> _CREW_HEAD_SET	Boolean	True if CVR crew head set is being pressed.
> ...		

Good idea - makes it more compact.
I would even leave out the _ and put a paragraph at the beginning explaining the naming.

> Also can we please name the file to what's inside? Such as a320-local-variables

Also good - I just used the previous name plus -table. I there are not vetos I will do this as well. 

> And rename `# A320neo Local SimVars` to `# A320neo Local Variables`
OK

> And for the events I'd like to see the event type specified.
Where can I see the event type? Can't find anything in the code.

frankkopp:
I will leave the "_" as some few variables don't have the A32NX
E.g.: PUSH_DOORPANEL_VIDEO, XMLVAR_MOMENTARY_PUSH_OVHD_ANTIICE_WING_PRESSED, etc.

MisterChocker:
> Where can I see the event type? Can't find anything in the code.

For instance `A32NX.ATHR_RESET_DISABLE` is a K-Event.

I am not sure if all the others are also K-Events. I am asking andreas for more info.

Also would it make sense to group certain such as _ELEC_CONTACTOR_...
So maybe say like _ELEC_CONTACTORS: and then the same as with the A32NX_name

Obviously only if it makes sense!

MisterChocker:
So after talking to andreas, these are all K-Events.

frankkopp:
> Also would it make sense to group certain such as _ELEC_CONTACTOR_...
> So maybe say like _ELEC_CONTACTORS: and then the same as with the A32NX_name

As a developer I agree - repetition is evil.
But for documentation it is probably more readable by just repeating stuff instead

davidwalschots:
Removing the `A32NX` prefix for variables that should have it doesn't make sense to me. As a developer I sometimes search for the full variable name to figure out what it means, or to rename the variable throughout the project. It isn't great if I can't find this particular usage.

frankkopp:
> Removing the `A32NX` prefix for variables that should have it doesn't make sense to me. As a developer I sometimes search for the full variable name to figure out what it means, or to rename the variable throughout the project. It isn't great if I can't find this particular usage.

That's a good point. Haven't thought of this. Was too focussed on making it compact. 
@MisterChocker agreed?

Benjozork (changes requested):
I'll be straight up and upfront about this, I really dislike the table format.

I think it works well for scalar value simvars with no parameters. But for complex ones, it turns to hell very quickly, and with your latest change, it solves the width problem but introduces a LOT of repetition and noise:

See this (new):

image

Compared to this (old):

image

I see the second (old) format as MUCH easier to read and comprehend, and carries the information with MUCH less visual noise and clutter.

I do want to ask, what problem are we trying to solve here? There is going to be a lot of data in those files regardless, how exactly is the list format hard to read?

While naming verbosity does help reduce guesswork and confusion in a codebase, this is just too much IMO. There is no immediate indication that those simvars are all really the same, while that was dead obvious in the old format.

If the problem was file length, then we can split it up in area-specific definition files.

Benjozork:
Besides, I find it a bit too much to have to install a plugin for something which didn't require one before.

Benjozork:
An alternative solution could be to define `{name}` somewhere else, like some kind of enum. But this could require a lot of work and necessitate jumping between sections.

frankkopp:
I understand that as a developer you really dislike repetition. As do I in code. 

But in documentation I try to think from the user's/reader's side. Your list above is as long vertically as the table. But to find the variable you had to jump over hundreds of lines of irrelevant text (other variables AND their details) until you found the right variable. In a table you would be much quicker (less lines).

Also it is much harder to make sure that the variables are sorted in some way without a table. Currently they are all over the place and sorting is very important to actually quickly find the variable you are looking for even if you don't know the variable's name (btw. I'm not saying alphabetically is the best sorting - grouping them to related functions might make sense**)

Obviously it would be much better if MD would allow for more complex tables which would be easier to write and read. But I still think it is an improvement for people who are looking for things and are not as familiar with the plane as you. 

Also, if we would at some point move/copy this to the website we would have the chance to add a little dynamic to it as well (sorting, collapsing, etc.).

I have one more idea how to visualize that extreme example with the contactors:
image

Disadvantage would be that it is not possible to sort automatically any more - but sorting manually is still easier than with the previous format.

Could you agree to this format? 

** I would even go a step further and say we should sort it from a cockpit process point of view along the checklist and also add non local varibales (SIMCONNECT, MSFS). That is how some people, me included, setup their hardware.

Benjozork:
> But in documentation I try to think from the user's/reader's side. Your list above is as long vertically as the table. But to find the variable you had to jump over hundreds of lines of irrelevant text (other variables AND their details) until you found the right variable. In a table you would be much quicker (less lines).

So my suggestion is to separate all those variables into separate files, related to the area the variable relates to:

```
docs/
    simvars/
        electrical.md
        hydraulics.md
        autofllight.md
```

They could even be sorted by ATA chapter:

```
docs/
    simvars/
        24-electrical.md
        29-hydraulics.md
        22-autofllight.md
```

This solution would also address this point:

> Also it is much harder to make sure that the variables are sorted in some way without a table. Currently they are all over the place and sorting is very important to actually quickly find the variable you are looking for even if you don't know the variable's name (btw. I'm not saying alphabetically is the best sorting - grouping them to related functions might make sense**)

With this, I can reasonably see the list format taking up much less space.

---

> Also, if we would at some point move/copy this to the website we would have the chance to add a little dynamic to it as well (sorting, collapsing, etc.).

I have been thinking of ways to achieve this while still having the usual plain-md format available. Perhaps something needs to be generated.

frankkopp:
> So my suggestion is to separate all those variables into separate files, related to the area the variable relates to:
> 
> ```
> docs/
>     simvars/
>         electrical.md
>         hydraulics.md
>         autofllight.md
> ```
> 
> They could even be sorted by ATA chapter:
> 
> ```
> docs/
>     simvars/
>         24-electrical.md
>         29-hydraulics.md
>         22-autofllight.md
> ```

I think this could be a second step. Although my personal opinion would be to have them on one page to make browser based search easier. 
 
> > I have one more idea how to visualize that extreme example with the contactors:
> 
> This could work. Although I would not use `{}`, as it's not immediately obvious what that is without being aware of Rust format strings.

That was a quick-shot for this discussion - we can go back to {name} or something similar. 

How about I work on this and you can have a look at the result?

PS: I really wish we could do spans in MD

frankkopp:
BTW - Is this what you refer to with ATA:  https://en.wikipedia.org/wiki/ATA_100

frankkopp:
@Benjozork I have reorganized everything according to ATA chapters. Also used some grouping to make it more readable.

The only two "difficult" parts are A32NX_ELEC_CONTACTOR_{name}__IS_CLOSED and A32NX_ELEC_{name}_BUS_IS_POWERED.
Everything else is good to read and understand. Especially if grouped by function/button and not how it is coded. 

E.g. all BAT_1 are together, all BAT_2, etc.:
```
- A32NX_ELEC_BAT_1_CURRENT          
- A32NX_ELEC_BAT_1_CURRENT_NORMAL   
- A32NX_ELEC_BAT_1_POTENTIAL        
- A32NX_ELEC_BAT_1_POTENTIAL_NORMAL 

- A32NX_ELEC_BAT_2_CURRENT          
- A32NX_ELEC_BAT_2_CURRENT_NORMAL   
- A32NX_ELEC_BAT_2_POTENTIAL        
- A32NX_ELEC_BAT_2_POTENTIAL_NORMAL 
```

This is how someone would actually look for values. 

I really hope you like it this time. I put in a lot of time and I really think this is huge improvement.
Also we can improve it even further in the future with feedback from developers and users. 
It lays the foundation for additional improvements - splitting into pages, dynamic tables, etc. 

Help needed: I'm not a pilot or plane engineer so my categorization might be wrong sometimes. I tried to follow this [ATA100](http://www.s-techent.com/ATA100.htm#[26]) as best as I could. 

Looking forward to your feedback.

Benjozork:
I'm on mobile at the moment, but this is so far approaching a point I'd consider ideal. Will take a look to confirm later.

frankkopp:
@Benjozork did you have a chance to look at it? If it is acceptable it would be great to change soon before more PRs with changes to events and vars come along. Apart from #5086 I think I caught all of them so far. But it would be good if people could use the new format asap to avoid missing things. Thanks

Benjozork (approved):

Benjozork:
I'm fine with it being merged, I'll see what others think of it though.

frankkopp:
@Benjozork , @davidwalschots can you guys look at this again and maybe we can come to a decision?
I have also opened a PR in docs - it really looks good there and we can add sorting etc. later (there is already a feature for sorting in mkdocs afaik).
https://docs-git-fork-frankkopp-varsevents-flybywire.vercel.app/a32nx-dev/a32nx-local-vars/

davidwalschots:
The last pending action is this one:

> I'm fine with it being merged, I'll see what others think of it though.

I was and still am am fine with this being merged 🙂.

Benjozork:
We will merge this for now. We are working on a better system which would auto-generate documentation and present data through a React component.

davidwalschots (approved):

davidwalschots:
I've observed many discussions on this PR and the format change. From what I've been able to tell the vast majority of people do not like the change. Therefore we'll not merge this one.

PR #5183, opened 2021-06-22
feat(lights): Add stepless potentiometer to INTEG LT knobs

bouveng:
## Summary of Changes

Adds stepless, silent, potentiometers to
* INTEG LT MAIN PNL & PED
* OVHD INTEG LT

This allows for a finer control of the integrated light settings.
No other changes to integrated lights in this PR.

## Screenshots (if necessary)

ovhd-stepless

## Video
https://user-images.githubusercontent.com/433429/122993073-29d8b800-d3a7-11eb-96fb-09e00c225efe.mp4

Added video as gif above was not smooth enough

## Testing instructions

1, Spawn anywhere
2, Power on aircraft
3, Turn the INTEG LT MAIN PNL & PED and OVHD INTEG LT knobs and ensure they work as intended


davidwalschots (approved):
You might consider creating a custom template to remove the repetition. With or without it, this LGTM.

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5183
Tier of Testing: 1
Date: 22/06/2021

**Testing Process:**
1. Loaded C&D in the night
2. Tested different knob positions in different lighting conditions

**Negatives:** 
- Again I dont like the fact, that the mouse wheel turn is so slow
- It seems like after the 11 o clock position the brightness does not change much, so turning the knob to the full stop is pretty useless. I think this needs adjustments

**Testing Results:** Not Passed

bouveng:
> * Again I dont like the fact, that the mouse wheel turn is so slow. I would prefer faster mouse wheel and clicking for precision

This has to do with that the knob now haver 100 "steps" (0-100%) instead of 10 (0%-10%-20% etc).
I do not know how to make the knob behave differently when using the mouse wheel.
Is there any "stepless" knob in any plane on the sim that works better you can ref. me to?
What about the 3 temperature knobs on the OVHD, do they behave better?

> * It seems like after the 11 o clock position the brightness does not change much, so turning the knob to the full stop is pretty useless. I think this needs adjustments

This is related to the emissive itself and how those are lit. I have not changed anything regarding the the emissives and/or how bright the are. This exact problem was in the previous knob as well, but using 10 steps at a time.

The knob output is 100% linear, but the perceived light from the emissives are not.
This PR justs implements the knobs themselves, I could try to do something about the "range" of light so we can "fake" an non-linear output, perhaps by dividing the output below 30%, and multiply it above 60%.

Suggestions and desired behaviour is welcome. =)

bouveng:
**UPDATE!**

Hi testers!

I have been digging around in the codebase and I think I will be able to turn the knobs differently when using the mouse-wheel.

There is `WheelUp` and `WheelDown` events!

I will have to do my own template for this. That is a good idea anyway since this will be used in a few places.
Give me a couple of days to build this.
Expect updated PR towards/in the weekend.

bouveng:
This PR has been **closed** and **replaced** by #5244

PR #5186, opened 2021-06-22
feat: EFB - List of ATC - decrease refresh rate

ZeroKaa:
## Summary of Changes
Reduce the interval between refresh to 1 second


davidwalschots:
21 commits for a 1 line change? This PR increases the refresh rate. It doesn't decrease it.

ZeroKaa:
> 21 commits for a 1 line change? This PR increases the refresh rate. It doesn't decrease it.



> 21 commits for a 1 line change? This PR increases the refresh rate. It doesn't decrease it.

Hello, indeed ... I update the title ;)

for the commits, I don't know why git add the already merged commits. I did a pull from the branch first

src/instruments/src/EFB/ATC/ATC.tsx
```diff
@@ -18,11 +18,16 @@ export const ATC = () => {
...
-        loadAtc();
-        setInterval(() => loadAtc(), 2 * 60 * 1000);
-    }, []);
+        const interval = setInterval(() => {
+            loadAtc();
+            setLastUpdate(new Date().toString());
+        }, 60 * 1000);
+
+        return () => clearInterval(interval);
+    }, [lastUpdate]);
```
beheh, line 30: Why have you added `lastUpdate` as a dependency to this hook? You're not using it.

ZeroKaa: the hook executes everytime lastUpdate change.
I had to use this 'hack' otherwise the "currentLatitude" and "currentLongitude" wasn't updated within the setInterval()

hiaaryan (approved):

beheh (changes requested):

ZeroKaa:
@beheh Can you show me an example with a set interval ?
I did my change based on
 https://stackoverflow.com/questions/59621744/how-to-use-setinterval-in-react

ZeroKaa:
@beheh : I updated the code based on your remark. Could you validate this is better now ?
Thanks in advance

beheh (approved):

src/instruments/src/EFB/ATC/ATC.tsx
```diff
@@ -34,26 +34,26 @@ export const ATC = () => {
...
+                    a.visualRange = 100;
                 }
             }
             allAtc.sort((a1, a2) => (a1.distance > a2.distance ? 1 : -1));
             allAtc = allAtc.slice(0, 26);
             allAtc.push({ callsign: 'UNICOM', frequency: '122.800', type: apiClient.AtcType.RADAR, visualRange: 999999, distance: 0, latitude: 0, longitude: 0, textAtis: [] });
             setControllers(allAtc.filter((a) => a.distance <= a.visualRange));
         });
-    };
+    }, [currentLatitude, currentLongitude]);
```
beheh, line 56:
```suggestion
    }, [currentLatitude, currentLongitude, atisSource]);
```

beheh (changes requested):
Sorry for the delay, one more required change: atisSource is a dependency, so it must be in the hooks array.

There is another general cleanup you could do (but I don't see as a strict requirement): Rather than repeatedly setting apiClient.ATC.getAtc with the same atisSource every 5 seconds when the lat/long changes, you could split that part out into a separate memo which you only redo when atisSource alone changes. Then you simply redo the filtering whenever lat/long changes. In that case you could even completely drop useInterval.

ZeroKaa:
@beheh 

I remove atisSource from dependencies array because the only way to change source is via the settings page and going to settings page and coming back to atc page will reload the component

I call getAtc every 5 minutes because the list of controllers can change in case a controller join or left the Network. So refresh is needed for location changes but also on regular interval.

beheh:
> I remove atisSource from dependencies array because the only way to change source is via the settings page and going to settings page and coming back to atc page will reload the component

If it doesn't change, it's a no-op, so it doesn't hurt having it. Imagine we refactor this component in future and there's another way for the ATIS source to change, maybe a second EFB or an API for a third-party tool or something - at that point we would want this! In all other cases, it does not hurt. Until you're very familiar with hooks I would highly suggest sticking exactly to the [rules of hooks](https://reactjs.org/docs/hooks-rules.html), because unfortunately there's a lot of hidden footguns otherwise. Religiously following the rules of hooks protects you from these.

> I call getAtc every 5 minutes because the list of controllers can change in case a controller join or left the Network. So refresh is needed for location changes but also on regular interval.

That seems fair

ZeroKaa:
> > I remove atisSource from dependencies array because the only way to change source is via the settings page and going to settings page and coming back to atc page will reload the component
> 
> If it doesn't change, it's a no-op, so it doesn't hurt having it. Imagine we refactor this component in future and there's another way for the ATIS source to change, maybe a second EFB or an API for a third-party tool or something - at that point we would want this! In all other cases, it does not hurt. Until you're very familiar with hooks I would highly suggest sticking exactly to the [rules of hooks](https://reactjs.org/docs/hooks-rules.html), because unfortunately there's a lot of hidden footguns otherwise. Religiously following the rules of hooks protects you from these.
> 
> > I call getAtc every 5 minutes because the list of controllers can change in case a controller join or left the Network. So refresh is needed for location changes but also on regular interval.
> 
> That seems fair. Another way you could do this is following my suggestion, but having a separate counter state variable that you increment every few minutes (and you can reset after a few iterations), so it always triggers an invalidation of the callback.

Make sense indeed. I will add atisSource.

ZeroKaa:
@beheh  All requested changes implemented. thanks

beheh (approved):

PR #5200, opened 2021-06-23
refactor(adirs): move logic to rust

davidwalschots:
## Summary of Changes
I'm moving the ADIRS logic to Rust to prepare for a more advanced ADIRS implementation.

## Additional context

## Testing instructions
It should work as it did before! 😀


tracernz (approved):
Code LGTM. I haven't checked accuracy, and given the goal of this PR is to move the existing implementation to wasm for future accuracy and improvement work I don't think it's necessary at this stage.

davidwalschots:
Can be tested already, but is awaiting #5241.

FlightSimmer7700:
Quality Assurance Tester

Object of testing: #5200 
Tier of Testing: 1
Date: 30/6/21

Testing Process:
Loaded C&D and tried with all different speeds: instant, fast and real
Loaded on the runway and tested
^Did the same for all spawn conditions

Negatives: Loaded in C&D, then changed the speed of the adirs to instant and restarted the flight. Then as soon as it loaded I pressed fly now (known to cause bugs with some prs that's why it was tested this way) and turned on the adirs into nav and turned on the batteries/plane and once it had loaded it was aligning normally as if the setting was on real

Also happens when the batteries are turned on first but as long as the plane has just been loaded and the batteries say 0.0v it will start doing a normal align instead of instant

Testing Results: Not passed (passed if not fixable)

Conclusions: Works as expected in normal use but can be broken
Media: 
image

davidwalschots:
We discussed the problem mentioned by FlightSimmer7700 on Discord. The alignment process was started before the aircraft was fully loaded (i.e. the batteries were also indicating 0.0V). As the code to copy the align time setting didn't run yet at that point, it will assume the default value which is REAL. Sadly there's nothing I can do about that situation. Just like I can't display the battery voltage before the aircraft is fully loaded. Once the aircraft is fully loaded, moving to NAV will instantly align the ADIRU. So the code works, it simply didn't run yet.

PR #5203, opened 2021-06-24
refactor(aoc/apt): housekeeping

MisterChocker:
## Summary of Changes

- I've removed a constant which Imho is unnecessary.
- I've also converted the function `assembleAirportInfo` to a method and reworded to `getAssembledAirportInfo` to better represent it's purpose.
- Reworded `canUpdate` to `tryUpdate` because `can` indicates a Boolean return whereas `try` is more intuitive.
- Changed `getAssembledAirportInfo()` parameter `_isManaged` default to `true` since there are 5 use cases of true and only 2 for false.

---

```js
if (this.rows[_index].isManaged || (!this.rows[_index].isManaged && !this.rows[_index].icao)) {
```
does the same as the one below because it's either managed and we can override or it's selected and therefore must be empty.

```js
if (this.rows[_index].isManaged || !this.rows[_index].icao) {
```

---

```js
this.getAssembledAirportInfo(_dep, true)
```
In this case, `_managed` can always be true, due to the fact that `getAssembledAirportInfo()` checks whether `_icao` is defined or not and only sets managed to true if the `_icao` parameter is also defined.

This way the responsibility is directly handed over to the airport info assembler instead of some outside logic or function.

---

_This pr doesn't alter the logic or workflow of the aoc apt list_

## Additional context

## Testing instructions

No changes to observe -> observe the aoc weather request airport list to behave normally.


Benjozork (approved):

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5203 
Tier of Testing: 1
Date: 05/07/2021

**Testing Process:**
1. Quick checked behavior of AOC weather request list (No FPLN, with FPLN, ALT, deleting, overwriting, requesting)

**Negatives:** N/A
**Testing Results:** Passed

Issue #2123, opened 2020-11-29
Cold & Dark + Slew Mode


CapStar362:
starting a flight with cold and dark status,  going to Slew Mode,  getting airborne and returning to normal flight causes the Engine Selectors to remain in the "OFF" Position.   this causes permanent erroneous EICAS and MASTER WARNING alerts throughout the flight, and cannot be cleared even if moving the Engine Selectors to the "ON" Positions.  the only way to clear them is to shutdown the entire electrical system and reboot the entire cockpit. 

shutting down the aircraft by moving the selectors back to the "OFF" position is disabled,  shutting down the fuel pumps is also non-effective in this sequence of events. 

just a report for anyone who slews around at first and then jumps into the air and returns to normal flight from a cold and dark status.

Benjozork:
Please follow the issue template

PR #5219, opened 2021-06-26
fix(efb): Upgrade @flybywiresim/react-components to version 0.3.1

### LINKED ISSUE #5011: EFB Map units not updating when display units changed
### Aircraft Version

Stable

### Build info

```json
{
  "built": "2021-05-05T23:03:06+00:00",
  "ref": "v0.6.1",
  "sha": "4a1cccc90ab3997109e394402ab9b41a80fbf325",
  "actor": "Benjozork",
  "event_name": "manual"
}
```

### Describe the bug

When the measurement units are changed on the EFB, the unit distance indicator in the lower left corner of the map doesn't change with it.

Subsequently, when using the track tool, it will only display the unit value in the measures on each point of what was on the display measurement at the time the waypoint was added.

The same is also true for the website map (https://www.flybywiresim.com/map) as it also shares the same component.

### Expected behavior

The unit distance indicator should display the same unit of measurement that is toggled to in the right hand side control panel.

All waypoint data should also display its measurement units in the same unit of measurement that is toggled to in the right hand side control panel.

### Steps to reproduce

1. Start a new flight
2. Turn on the EFB
3. View dashboard
4. Observe the behavior of the unit distance indicator when clicking on the measurement unit button in the control panel on the lower right of the map
5. Observe the behavior when using the tracking tool to track while changing the unit of measurement.

### References (optional)

**Sim**
image

**Website**
image

_No response_

antiso

BlueberryKing:
Fixes #5011

## Summary of Changes
The bug is described in issue #5011. Previously, when using the distance measuring tool. The units of distance would not update on tooltips of a path that was not finished yet. Changing units would only affect existing paths. Now, all tooltips should update even if the path is not finished yet.

I have made a change to the plugin we are using for the map and published the change via the react-component (https://github.com/flybywiresim/react-components/pull/11) library, which I am now updating here (and on the website).

## Screenshots (if necessary)

Before (notice all different units):

After:
Screenshot 2021-06-26 124900

## Additional context

## Testing instructions

1. Load up anywhere
2. Open EFB.
3. Use the distance measuring tool to draw a path. Finish the path by clicking on the endpoint.
4. Draw a new path, do not finish it. Instead, just use the mouse (which should still be moving the endpoint of the path) to change the units on the control panel.
5. Observe how all tooltips have the units selected on the control panel.


davidwalschots (approved):

marcman86:
Discord:marcman86#4907
Object of testing: #5219
Tier of Testing: 1
Date: 1/07/2021

Testing Process:
- Spawned C&D at EDDP
- Turned on EFB
- Marked a route on the map and changed unit
- Marked a route, while still marking changed unit

Negatives: 
- Map not in dark mode

Negatives (not of this PR): 
- Plane icon flickering

Testing Results: Passed, if light mode is on purpose
Conclusions: Works
Media:-

Hope this helps :-)

Benjozork:
@BlueberryKing Can you check why light mode is on by default ?

BlueberryKing:
As opposed to the map, I'm still a bit in the dark on this. The map is in light mode for me even on the master branch. I made sure to downgrade to react-components@0.3.0 which should not be using any of my changes. I'm gonna need some more time to investigate. Sorry about that!

Benjozork:
The map is supposed to be in dark mode, I think someone made the component light by default recently.

Might need to be explicitly specified now.

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5219 
Tier of Testing: 1
Date: 05/07/2021

**Testing Process:**
1. Loaded onto RW and turned on EFB
2. Tested distance measuring
3. New line not ending, changing units, end line

**Negatives:** 
- I think the "m" unit should be called "km" as thats the unit the measurement is done and also every other unit is in the big nm or mi unit
image

**Testing Results:** Maybe fix the above typo, then its good to go :) Didnt even know this feature existed before :D

BlueberryKing:
@St54Kevin Thank you. I will have to make this change in the react-components repo, which will have to be published to npm, but I can do it.

@Benjozork I can do this in the A32NX repo.

2supersanjay:
### Quality Assurance Trainee Report

Object of testing: PR #5219 
Tier of Testing: 1
Date: 6/7/2021

### Testing Process: 

1. Started up the sim in a cold/dark state
2. Powered up the aircraft and turned on the EFB
3. Used the distance measuring tool 
4. Clicked on the dot to end the path
5. Created another path
6. Instead of ending the path, I changed the units of the paths

### Negatives: 
One doubt is that when I did "shift + click" in order to delete the point, instead it just zoomed all the way where my aircraft was. Another doubt is that the unit was called 'M' instead of 'KM' since 'KM' is what is measured for the paths instead of 'M'. 

Testing Results: Passed

### Conclusions: Everything worked as suppose too. I am able to change the units of all paths even the ones that I have not completed. The distance measuring tool works as it is suppose too. Just be sure to fix the "shift + click" unless it is suppose to zoom in on the aircraft and change the measurement 'M' to 'KM'.

Media: https://youtu.be/m9TMsqE-TqE

BlueberryKing:
I looked into the km/m issue and it looks like the logic is rooted in the https://github.com/ppete2/Leaflet.PolylineMeasure library we're using for the measurement tool control panel, so not an easy fix. In my opinion, it's not a big issue and if we decide to fix it in the future, we can do this with a separate PR.

marcman86:
The unit "m" is fine, as this is the "unit of lenght". See: https://en.wikipedia.org/wiki/Metre
The "k" is only a prefix (factor 1000). So i would stay to "m".

St54Kevin:
> The unit "m" is fine, as this is the "unit of lenght". See: https://en.wikipedia.org/wiki/Metre
> The "k" is only a prefix (factor 1000). So i would stay to "m".

I do not agree with this. m is not the same compared to mi or nm. But km is. and the lines are actually measured in km not m.

marcman86:
>  and the lines are actually measured in km not m.

They do, if you have lines less then 1km.

St54Kevin:
> > and the lines are actually measured in km not m.
> 
> They do, if you have lines less then 1km.
> 
> 

okay you won :D

marcman86:
Retested: Not rebased. Still the FlyOS v1.

BlueberryKing:
Ah, sorry about that. I could've seen that coming! Should be fixed now.

marcman86:
Retested: Works fine. I think this is good to go. :-)

TheSquirtingPantin:
Quality Assurance Tester

Discord: Remi#8871
Object of testing: #5219
Tier of Testing: 1
Date: 26/07/2021

Testing Process:
1) C/D at KMIA
2) Powered EFB up
3) Used map to check units coherence, path design with change of units.
4) Multiple resets of map and redraw to ensure EFB is robust.

Negatives: Some characters are showing as square for IN and OUT values. Seems similar to the ATSU ATIS issue we had previously. 

Conclusions: PARTIAL PASS for squared characters for IN and OUT values. Not affecting user experience.
Media: N/A

2hwk:
-- SU5 after this comment --

NUTTTTTTTT:

Discord: █▀█ █▄█ ▀█▀#2123
Object of testing: #PR 5219
Tier of Testing: 1
Date: 03/08/2021

Testing Process: 
- Spawned in on runway at EGKK and opened EFB. 
- Drew a path with the distance measuring tool.
- Finished the path and drew a new one without finalising, instead changed the units using the controls on screen between KM, NM and Mi.

Observed:
- Units for points 2 and 3 drawn out showed distance of leg, and total distance from start to end of the second point, this gave correct values on all three measurements.
- When Mi or Nm selected and distance is less than 1nm/mi, value is shown in feet.
- When Km is selected and distance is less than 1km, value is shown in meters.
- No issues with tracking larger numbers of points at once.

Negatives: N/A
Testing Results: Passed

Conclusions: EFB map is now displayed in the darker theme, so all good there. Everything otherwise seems to be working exactly as intended for the PR and no new issues observed with the EFB or aircraft.

Media:
image

davidwalschots:
lol. I'm a bit lost as to why this PR exists. We don't have a `package-lock.json` file, and the version in the `package.json` is `^0.3.0` so this will update to anything `< 0.4` automatically. Therefore, the moment `@flybywiresim/react-components`'s `v0.3.1` was published all new builds used that newer version already. 😀

BlueberryKing:
> We don't have a package-lock.json file

Wait, so there's no `package-lock.json` anywhere? So not even on whatever machine generates the builds for the installer? I didn't even know that was an option 😀. I feel like I tried it on master after my @flybywiresim/react-components PR got merged and it hadn't updated, but I was probably doing something wrong. Anyways, I still suggest merging this just to make sure it's updated on the new builds. What do you think?

davidwalschots:
Sure. There's still a conflict to be solved though!

PR #5228, opened 2021-06-28
fix(hyd): update hydraulics to use new engine n2 uncorrected value

crocket63:
Fixes #[issue_no]

## Summary of Changes

-New uncorrected N2 parameter is now used as engine driven pump speed source
-New uncorrected N2 parameter is now used as blue epump auto running condition
-Blue epump auto running condition on N2 threshold adjusted to start sooner at 55% uncorrected N2.

## Additional context

## Testing instructions

-Engine driven pumps should behave normally on startup / idle / full power / engine spool down
-Blue electric pump should start automatically around 55% N2 value of the first engine to start.
-Blue electric pump should stop automatically around 55% N2 value of the last engine to spool down.

### DIFF
--- a/src/systems/systems/src/electrical/engine_generator.rs
+++ b/src/systems/systems/src/electrical/engine_generator.rs
@@ -5,7 +5,7 @@ use super::{
 };
 use crate::{
     shared::{
-        calculate_towards_target_temperature, EngineFirePushButtons, EngineN2,
+        calculate_towards_target_temperature, EngineCorrectedN2, EngineFirePushButtons,
         PowerConsumptionReport,
     },
     simulation::{
@@ -48,7 +48,7 @@ impl EngineGenerator {
     pub fn update(
         &mut self,
         context: &UpdateContext,
-        engine: &impl EngineN2,
+        engine: &impl EngineCorrectedN2,
         generator_buttons: &impl EngineGeneratorPushButtons,
         fire_buttons: &impl EngineFirePushButtons,
     ) {
@@ -170,7 +170,7 @@ impl IntegratedDriveGenerator {
     pub fn update(
         &mut self,
         context: &UpdateContext,
-        engine: &impl EngineN2,
+        engine: &impl EngineCorrectedN2,
         generator_buttons: &impl EngineGeneratorPushButtons,
         fire_buttons: &impl EngineFirePushButtons,
     ) {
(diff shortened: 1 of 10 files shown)


src/systems/systems/src/electrical/engine_generator.rs
```diff
@@ -300,10 +300,13 @@ mod tests {
...
         }
     }
-    impl EngineCorrectedN2 for TestEngine {
+    impl EngineN2 for TestEngine {
         fn corrected_n2(&self) -> Ratio {
             self.corrected_n2
         }
+        fn uncorrected_n2(&self) -> Ratio {
+            self.corrected_n2
+        }
```
davidwalschots, line 309: **issue**: An `uncorrected_n2` fn returning `corrected_n2` doesn't make sense. Even though it is a test it still breaks the trait's contract.
**suggestion**: Instead of having one trait, keep the `EngineCorrectedN2` and add `EngineUncorrectedN2`.

davidwalschots (changes requested):

src/systems/systems/src/electrical/engine_generator.rs
crocket63: I was indeed trying not to have multiple traits for related N2 parameters. Can either update the test engine to have a fake uncorrected value even unused as of now, or separate in multiple traits.  Preference for multiple traits then?

davidwalschots: Definitely. The usages are separated (i.e. engine gen only cares about corrected, the other parts only care about uncorrected) and therefore it's a good idea to let them rely on separate traits as well. If in the future we have usages that require both, we can pass `&(impl EngineCorrectedN2 + EngineUncorrectedN2)`.

davidwalschots (approved):

TheSquirtingPantin:
Quality Assurance Tester

Discord: Remi#8871
Object of testing: #5228
Tier of Testing: 1
Date: 28/06/2021

Testing Process:
1) C/D at LFBO
2) ENG 2 start. Monitoring of HYD BLUE EPUMP behavior through the Model Behavior page (HYD EPUMP ACTIVE variable + Reservoir + Pressure)
3) ENG 2 shutdown. Monitoring of HYD BLUE EPUMP behavior through the Model Behavior page (HYD EPUMP ACTIVE variable + Reservoir + Pressure)
4) Full Flight Test (LFBO-LFBO)

Negatives: None

Conclusions: PASS. No issue during ground and flight phase.
Media: N/A

PR #5254, opened 2021-07-02
feat(hyd): weight on wheels sensors to replace SIM ON GROUND

crocket63:
## Summary of Changes

As various systems use different informations from gear position/compression, this is a first basic implementation of the LANDING GEAR CONTROL INTERFACE UNIT. Two redundant computers are monitoring landing gear sensor inputs, and provide those informations to any system requiring it.

## Additional context

## Testing instructions

1)Autobrake armed on approach engages in 2seconds after both main gear on ground even if nose gear up

2)Blue epump should auto run even on engine off as long as center gear is in the air
Test case idea: fly approach -> all engines off -> N2 values < 30% (BLUE PUMP STAYS ON) -> touch down main gears (BLUE PUMP STAYS ON) -> Nose gear touches ground BLUE PUMP IS OFF

2)Ptu should stop inhibiting when center gear is in the air
Test case idea: fly approach -> Engine 1 Master switch off -> Park brake ON (PTU STAYS ON) -> touch down main gears (PTU STAYS ON) -> Nose gear touches ground (PTU IS INHIBITED)

3)Autobrake MAX should disarm after 10s in the air, starting from the last main gear leaving the ground (was nose gear previously).
Test case idea: Start on runway -> Manage to rotate but all main gears stay on ground as long as possible(more than 10s if possible): Autobrake still armed. Start a timer from when last main gear leaves ground, from that point autobrake should disarm after 10s.


beheh:
I'll definitely be using this in the FWS as well!

From an FWC perspective, I believe the real aircraft has multiple systems to determine this:
1) LGCIUs (redundantly)
2) A discrete signal, I believe powered by one of the ESS busses?

I'm not sure which of those is being implemented here.

crocket63:
> I'll definitely be using this in the FWS as well!
> 
> From an FWC perspective, I believe the real aircraft has multiple systems to determine this:
> 
> 1. LGCIUs (redundantly)
> 2. A discrete signal, I believe powered by one of the ESS busses?
> 
> I'm not sure which of those is being implemented here.

So here it's the sensor abstraction only getting a boolean from gear extension.  So philosophically it's the LGCIU input.
Until we decide we need a complete LGCIU implementation, I'm using it directly as if it was a LGCIU output

beheh (approved):
Overall, this is a great improvement! This will greatly accuracy of ground detection and should improve future timings around gear compressions.

I had a hand in designing the LGCIU trait and I think it pretty accurately matches the LGCIU discretes that are used by a lot of other systems. While we don't precisely model the electric potentials for the signals, we have a logical output that supports the special ground power override used by some systems. The ARINC outputs are also very simplified for now, but in future those could be extended, if needed.

This PR also lays the groundwork for somebody in future building a much more in-depth implementation of the LGCIUs with self-tests, cross-checks, standby/active stuff, but it's perfectly functional for now. What we have now a good basis to accurately hook up other systems that will then benefit from any future improvements to the LGCIUs.

src/systems/systems/src/shared/mod.rs
```diff
@@ -43,11 +43,31 @@ pub trait RamAirTurbineHydraulicLoopPressurised {
...
+    fn right_gear_extended(&self, treat_ext_pwr_as_ground: bool) -> bool;
+
+    fn left_gear_compressed(&self, treat_ext_pwr_as_ground: bool) -> bool;
+    fn left_gear_extended(&self, treat_ext_pwr_as_ground: bool) -> bool;
+
+    fn left_and_right_gear_compressed(&self, treat_ext_pwr_as_ground: bool) -> bool;
+    fn left_and_right_gear_extended(&self, treat_ext_pwr_as_ground: bool) -> bool;
+
+    fn nose_gear_compressed(&self, treat_ext_pwr_as_ground: bool) -> bool;
+    fn nose_gear_extended(&self, treat_ext_pwr_as_ground: bool) -> bool;
```
beheh, line 62: A note for other reviewers: If you think of these functions as discrete connections going from the specific LGCIU to another system, then in reality there is various ways to encode this. A signal might be truthy when 28V DCC is applied, it might be truthy when GND is applied, and some connections have a special transistor to explicitly treat external power as ground.

While we don't explicitly model the outgoing voltage here (we describe the outgoing logical meaning of a signal), every consumer can individually decide whether it wants the logically positive or negative case (compressed vs. extended), and whether it wants the LGCIU to treat it's specific connection with the special external power ground case. The positive/negative cases are important for system to decide which way to fail in a fault case (default to extended vs. compressed, or flight vs. ground), and each consumer can also decide individually whether external power is good enough to force the ground case.

This is why all these combinations are supplied, and why the external ground power override is modeled as a parameter interpreted inside the LGCIU (rather than the target system).

crocket63: We'd probably want to make in the future a list of the system users and what signal they should use. And mark this as "reality checked" or "Rule of thumb only, awaiting real data confirmation"

St54Kevin:
Cant test, looks like something is broken with the checks

beheh:
@crocket63 Can you push another commit so GitHub rebuilds?

crocket63:
> @crocket63 Can you push another commit so GitHub rebuilds?

Will do  later today !

beheh:
Checks look good now! @St54Kevin

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5254 
Tier of Testing: 1
Date: 12/07/2021

**Testing Process:**
1. Loaded onto RW
2. Takeoff to check Autobrake MAX off after 10s (tested with early rotation to get nose wheel off the ground)
3. Turned on APU, turned off ENGs and checked B ELEC Pump running

**Negatives:** N/A
**Testing Results:** Partial Passed

@crocket63 Please provide more extensive testing instruction especially with deep systems like this. I had to dig into my IRL documents to find out that there is no information about any system inhibitions :D Im not able to test Point 1 and 3, because I dont know what to do.

beheh:
So to clarify from what I know (@crocket63 can surely add more):

- This PR introduces the first concept of an LGCIU with shock absorber sensors. That means we no longer look at a magical SimVar that tells us whether we're on the ground, instead it actually looks at the gear compression.
- The LGCIUs are far from complete and aren't even hooked up to a power source (so they can't fail). For now they're just a helper to get a more accurate signal that we can directly use in the right places, and then in future we can come in and add all the other things.
- This new gear compression signal is hooked up only to the hydraulic system for now. So nothing should have changed outside the hydraulic system.
- In addition, we have new downlock and uplock signals
- In the hydraulic system, these new LGCIU signals are currently used in the following places, as far as I can tell:
  - EDP fault if on ground and engine oil pressure is low
  - Blue Pump running automatically in the air
  - PTU running automatically in the air with only one engine running
  - Autobrake disable after uplock on takeoff
  
So a QA test should verify that these still work as expected, and that they are mostly controlled by main gear compression.
Some of these may be hard to test, and in addition some of the behaviours are not fully obvious until we have actual failures. Nonetheless I think a few touch and gos with this PR while playing with the hydraulic system should give us confidence that we can merge this.

crocket63:
Yup, it's just a starting point so we have the ability to check any gear on ground for example. We will need to build a exhaustive list of all system users of those lgciu outputs so we have a defined requirement for all of them.
For now it's just done so it uses what we "think" is right, until we can build a list of actual requirements.

For autobrake there was the goal to start the 10s timer for it to disarm automatically after main gears are off the ground, whereas before it was after nose wheel in flight only.

Current actual implementation is already checking for power available to computers though, but that's probably hard to check in sim without failures?

ShadowFrost3:
Following these testing instructions as the basis for pass/fail. 

1)Autobrake armed on approach engages in 2seconds after both main gear on ground even if nose gear up

2)Blue epump should auto run even on engine off as long as center gear is in the air
Test case idea: fly approach -> all engines off -> N2 values < 30% (BLUE PUMP STAYS ON) -> touch down main gears (BLUE PUMP STAYS ON) -> Nose gear touches ground BLUE PUMP IS OFF

2)Ptu should stop inhibiting when center gear is in the air
Test case idea: fly approach -> Engine 1 Master switch off -> Park brake ON (PTU STAYS ON) -> touch down main gears (PTU STAYS ON) -> Nose gear touches ground (PTU IS INHIBITED)

3)Autobrake MAX should disarm after 10s in the air, starting from the last main gear leaving the ground (was nose gear previously).
Test case idea: Start on runway -> Manage to rotate but all main gears stay on ground as long as possible(more than 10s if possible): Autobrake still armed. Start a timer from when last main gear leaves ground, from that point autobrake should disarm after 10s.

Object of testing: #5254 
Tier of Testing: 1
Date: (28/07/2021)

Testing Process: Test 1, hotstart (on runway), fly approach, touch down hold nosewheel off, check if autobrake engages while nosewheel off the ground. Autobrake engages approximately 2-3 seconds after main gear touch down with nosewheel held off the ground. 
Test 2.1, hotstart (on runway NASA airfield), APU on, begin takeoff roll and kill both engines, rotate then land, hold nosewheel off keep mains on ground, green and yellow slowly lose pressure towards 0, blue stays on until nosewheel contacts ground then loses pressure. 
Test 2.2, hotstart, takeoff, N1 engine shutdown, parking break on, hydraulic page shows green PTU bar, main gears on ground, hydraulic page shows green PTU, front gear touches down, orange PTU bar on hydraulic page. 
Test 3, hotstart (long runway useful again), set high flex temp, deploy speed brakes for extra help avoiding early main gear liftoff, hold nose off at approximately 70 knots, balance to keep mains on ground for as long as reasonably possible but do not drop nose on the ground, approximately 10 seconds from main gear rotation max autobrake disarms. Disarm appears to be tied to main gear ground contact and not nose wheel contact. 
Then perform the opposite, to confirm 10 seconds accuracy, hold the nose wheel down until about 140-160kts and then perform a quick rotation. With the results of the nosewheel hold tests and the quick rotation, it indeed appears that the autobrake disarms at 10 seconds 
Negatives: N/A
Testing Results: Passed, for the four tests applicable, this appears to be working as intended. 
See notes on test 2.2 to ensure my testing accuracy to the results. 

Conclusions: My knowledge on the systems being tested is limited so I can only test for the system functions as the tests are described. The note I would add is for test 2.2. The transition from PTU green to orange (amber?) is the noticed change and I assume what this PR is attempting to do at least from a pilot's perspective. If test 2.2 is incorrect or deviating from expected then I need to know and results should change from passed to partial. In regards to test 1, 2.1, and 3, I do not have any known uncertainties.

crocket63:
> 
> Conclusions: My knowledge on the systems being tested is limited so I can only test for the system functions as the tests are described. The note I would add is for test 2.2. The transition from PTU green to orange (amber?) is the noticed change and I assume what this PR is attempting to do at least from a pilot's perspective. If test 2.2 is incorrect or deviating from expected then I need to know and results should change from passed to partial. In regards to test 1, 2.1, and 3, I do not have any known uncertainties.

This is exactly what was expectec for 2.2, amber means PTU valve is closed, so LGTM!  Thanks!

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5254 
Tier of Testing: 1/2
Date: 30/07/2021

**Testing Process:**
1. Takeoff with long rotation -> Autobrake MAX disengages 10s after main gear lift off
2. Normal approach -> Autobrake engaging 2s after main gear touchdown
3. Higher approach turning both ENG off -> BLUE Pump ON until nose gear touchdown
4. One ENG off approach and Parkbrake ON -> PTU ON until nose gear touchdown

**Negatives:** N/A
**Testing Results:** Passed

PR #4750, opened 2021-05-09
refactor: instruments build and framework cleanup

Benjozork:
## Summary of Changes

* [x] Use parallelized build for instruments
* [x] Enable support for TypeScript path mappings (depends on https://github.com/simonhaenisch/rollup-plugin-typescript-paths/pull/9)
* [x] Enable use of `rollup.config.js` regardless of cwd
* [x] Separate simulator and browser builds
* [ ] Add `devServer`
  * [x] Base 
  * [ ] `SimVar` shim
* [x] Adapt A32NX build scripts

Discord username (if different from GitHub): someperson#4953

## Testing instructions

* Ensure all instruments and ECAM pages work

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #5278, opened 2021-07-06
feat(build): multi-platform file path support

davidwalschots:
## Summary of Changes
Couldn't build instruments in watch mode on Windows using the changes introduced in  #4750. This PR enables multi-platform file path support.

## Additional context

## Testing instructions
No need to test by QA. Maybe a developer can test on their non-Windows environment if it still work as expected 😀.

### DIFF
--- a/src/instruments/buildSrc/directories.mjs
+++ b/src/instruments/buildSrc/directories.mjs
@@ -6,8 +6,8 @@ import { fileURLToPath } from 'url';
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export const Directories = {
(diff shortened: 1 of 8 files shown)


src/instruments/buildSrc/directories.mjs
```diff
@@ -1,11 +1,13 @@
 import os from 'os';
+import path from 'path';
```
Benjozork, line 2: These imports should be consistent, we use `import { join } from 'path'` in all other files.

Benjozork (changes requested):

Benjozork (approved):

PR #5341, opened 2021-07-17
feat(hyd) : rigid body simulation for cargo door

### LINKED ISSUE #5175: Cargo door opens faster than IRL
### Aircraft Version

Development

### Build info

```json
{
  "built": "2021-06-20T23:50:13+00:00",
  "ref": "master",
  "sha": "85470d645814125ff79304bfd19a438e721c736d",
  "actor": "MisterChocker",
  "event_name": "manual"
}
```

### Describe the bug

The new sound of the hydraulic cargo door opening and closing is very realistic, but the speed of the cargo door opening and closing is very fast

### Expected behavior

Cargo Door open/close animation should be a bit slower

### Steps to reproduce

1. Load A32NX at any airport parking/gate
2. Look at the cargo door using external/wing/drone(any) view
3. Call for luggage, or open cargo door using PushbackHelper(3rd party, so not recommended)
4. Door opens within 2 - 3 seconds, whereas sound lasts accurately.

### References (optional)

https://www.youtube.com/watch?v=a2k4i_Su__o

Cargo door opens slower IRL

This could qualify as an enhancement, so if required, you may change the tag please.

tbone0793

crocket63:
Fixes #5175

## Summary of Changes

This PR is the first step for a complete hydraulic actuator simulation. Starting with cargo door.

The goal is to simulate the cargo door using a physical rigid body simulation. Then linear actuators connected to the door control arm will be able to move it using hydraulic power.
The whole principle will then apply for all moving parts of the plane.

- [x] First rigid body simulation of the door
- [x] Retrieve plane and world acceleration from FS2020 and apply it to the door
- [ ]  Replace cargo door animation with the position of the simulated rigid body
- [ ] Linear actuator simulation
- [ ] Connect actuator to hydraulic system
- [ ] Wind gusts can make the door oscillate
- [ ] Add a cargo door "controller" that will act as a ground operator
- [ ] Add a manual pump to move the door when no ground power

## Additional context


crocket63:
I currently fail at writing the position of the door to FS2020.
My attempts to write to INTERACTIVE POINT OPEN:5 or EXIT OPEN:5 all failed sadly.
I'm open to ideas to make this work :)

tracernz:
> 
> 
> I currently fail at writing the position of the door to FS2020.
> My attempts to write to INTERACTIVE POINT OPEN:5 or EXIT OPEN:5 all failed sadly.
> I'm open to ideas to make this work :)
> 
> If it's definitly not possible maybe last chance is to replace the interactive point animation with a "classic" animation driven by a position simvar? No idea if this is possible.

Can't we just drive the animation off an Lvar that you write?

crocket63:
> 
> Can't we just drive the animation off an Lvar that you write?

Maybe, but I'm not sure how to do this... I can only find this and "door_cargo" in the lod files. 
 <Component ID="INTERACTIVE_POINTS">
            <UseTemplate Name="ASOBO_DOOR_InteractivePoint_Template">
                <ID>5</ID>
                <ANIM_NAME>door_cargo</ANIM_NAME>

Risk is then to have discrepency between physical door and interactive point driving the luggage treadmil. But that's another story :)

I'll ask in modeling section maybe :)

src/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -83,13 +88,15 @@ fn create_aircraft_variable_reader(
...
-    reader.add("EXIT OPEN", "Percent", 3)?;
     reader.add("PUSHBACK ANGLE", "Radian", 0)?;
     reader.add("PUSHBACK STATE", "Enum", 0)?;
     reader.add("ANTISKID BRAKES ACTIVE", "Bool", 0)?;
     reader.add("ACCELERATION BODY Z", "feet per second squared", 0)?;
+    reader.add("ACCELERATION BODY X", "feet per second squared", 0)?;
+    reader.add("ACCELERATION BODY Y", "feet per second squared", 0)?;
+
+    reader.add("PLANE PITCH DEGREES", "Degrees", 0)?;
+    reader.add("PLANE BANK DEGREES", "Degrees", 0)?;
```
davidwalschots, line 99: These definitions are duplicate.

src/systems/systems/src/hydraulic/brake_circuit.rs
```diff
@@ -110,6 +100,10 @@ impl Actuator for BrakeActuator {
     fn reservoir_return(&self) -> Volume {
         self.volume_to_res_accumulator
     }
+    fn reset_accumulators(&mut self) {
+        self.volume_to_res_accumulator = Volume::new::<gallon>(0.);
+        self.volume_to_actuator_accumulator = Volume::new::<gallon>(0.);
+    }
```
davidwalschots, line 106: Formatting (missing blank line before fn).

src/systems/systems/src/hydraulic/brake_circuit.rs
```diff
@@ -310,6 +299,10 @@ impl Actuator for BrakeCircuit {
...
```
davidwalschots, line 305: Formatting (missing blank line before fn).

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    fn time_step(&self) -> Duration {
+        // println!("fixed::time_step: time{:.3}", self.time_step.as_secs_f64(),);
+        self.time_step
+    }
+
+    fn _loop_number(&self) -> u32 {
+        self.number_of_loops_this_iteration
+    }
```
davidwalschots, line 27: With `_` because it is unused? If so, then please remove, otherwise rename to not have `_`.

davidwalschots (commented):
A few comments but by no means a full review.

BlueberryKing (commented):
This is brilliant! The amount of physical detail put into this is truly inspiring! 🚀

I left some comments regarding tight coupling between components. As always, feel free to correct me on any assumptions I made and let me know if you don't agree with my suggestions. These are just my personal opinions after all 😄

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,707 @@
...
+    fluid_compression_damping_constant: f64,
+    active_damping_constant: f64,
+}
+impl LinearActuator {
+    const DEFAULT_I_GAIN: f64 = 0.2;
+    const DEFAULT_P_GAIN: f64 = 0.05;
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        connected_body: &RigidBodyOnHingeAxis,
```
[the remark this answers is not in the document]
crocket63: I had the same concern indeed, and it's difficult for me to truly separate those two. In fact if you need a min and max linear distance... that would only be for a linear actuator. Other way for an actuator to be connected would be maybe by a direct torque application, where in that case you wouldn't probably need that min/max distance.  Don't know what to do :D :D

The fact the actuator "asks" for min and max is that this way, it will always have the correct dimentions depending on the actuated body dimensions.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,707 @@
...
+
+pub trait HydraulicAssemblyController {
+    fn mode_requested(&self) -> LinearActuatorMode;
+    fn position_request(&self) -> Ratio;
+    fn should_lock(&self) -> bool;
+    fn lock_position_request(&self) -> Ratio;
+}
+pub struct HydraulicActuatorAssembly {
+    linear_actuator: LinearActuator,
+    rigid_body: RigidBodyOnHingeAxis,
```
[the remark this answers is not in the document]
crocket63: Fact is I think everything on the plane is door like :D 
I agree it's not super satisfying to have a particular actuator type in that structure. But after thinking about this there's pretty much no other type of actuator I would add there.  Well there is one indeed: the hydraulic motor actuators for flap and slat systems.  But the structure of flap/slat system is so different that I don't think I would use the same HydraulicActuatorAssembly container.  Maybe correct naming would be to have a HydraulicLinearActuatorAssembly and a HydraulicMotorActuatorAssembly ?   
Will have a second look at how this could be merged in a one only structure, but sounds difficult to me looking at how motors interact with flaps and slats. I'm still not too much at ease with the traits concept :p

src/systems/systems/src/hydraulic/linear_actuator.rs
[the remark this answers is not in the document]
BlueberryKing: > In fact if you need a min and max linear distance... that would only be for a linear actuator.

Ok, I see that. But if I understand it correctly, we would only be modifying the `LinearActuator` to accept a more generic rigid body, so it's fine if this concept of max and min only applies to it, no?

In this case, my second solution of accepting `max_pos` and `min_pos` as parameters instead of `RigidBodyOnHingeAxis` would still make it a bit easier to test this code as well, since we don't have to instantiate a rigid body every time we create a `LinearActuator`.

Anyways, I don't think this too important, since you do elaborate on the reason for the tight coupling between the actuator and the body later on, so separating them more here wouldn't really matter too much. 😄

src/systems/systems/src/hydraulic/linear_actuator.rs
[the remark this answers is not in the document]
BlueberryKing: Ok, that's perfectly reasonable! Yeah, I agree with you that there's no point in trying to optimize prematurely here. We can think about making this more generic if we implement the flaps/slats with the hydraulic system and notice a lot of duplication. Only then is it worth considering merging some of the concepts. 👍

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,292 @@
...
+        Self {
+            lag_time_accumulator: Duration::from_millis(0),
+            time_step,
+            number_of_loops_this_iteration: 0,
+            number_of_loops_remaining: 0,
+        }
+    }
+
+    fn time_step(&self) -> Duration {
+        // println!("fixed::time_step: time{:.3}", self.time_step.as_secs_f64(),);
```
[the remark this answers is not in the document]
crocket63: Those damn printf never die :D  Thanks!

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,707 @@
...
+        let mut volume_to_actuator = Volume::new::<cubic_meter>(0.);
+        let mut volume_to_reservoir = Volume::new::<cubic_meter>(0.);
+
+        if self.delta_displacement > Length::new::<meter>(0.) {
+            volume_to_actuator = self.delta_displacement * self.bore_side_area;
+            volume_to_reservoir = volume_to_actuator / self.volume_extension_ratio;
+        } else if self.delta_displacement < Length::new::<meter>(0.) {
+            volume_to_actuator = -self.delta_displacement * self.rod_side_area;
+            volume_to_reservoir = volume_to_actuator * self.volume_extension_ratio;
+        }
```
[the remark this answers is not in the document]
crocket63: There's the sign but the areachosen for the calculation too :)

src/systems/systems/src/hydraulic/rigid_body.rs
```diff
@@ -0,0 +1,491 @@
...
+use crate::simulation::UpdateContext;
+
+// RigidBodyOnHinge represent any physical object able to rotate on a hinge axis.
+// It can be a gear, elevator, cargo door...... Only one rotation degree of freedom is handled.
+// An actuator or multiple actuators can apply forces to its control arm
+//
+// Coordinates as follows:
+// on x (left->right looking at the plane from the back)
+// on y (down->up)
+// on z (aft->fwd)
```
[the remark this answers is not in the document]
crocket63: Felt more natural this way but yeah just a convention

BlueberryKing (approved):

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -503,137 +556,146 @@ impl A320Hydraulic {
...
         engine_fire_push_buttons: &U,
         lgciu1: &impl LgciuInterface,
         lgciu2: &impl LgciuInterface,
     ) {
-        self.brake_computer.send_brake_demands(
-            &mut self.braking_circuit_norm,
-            &mut self.braking_circuit_altn,
-        );
+        // Then run fixed update loop for main hydraulics
+        while let Some(cur_time_step) = self.core_hydraulic_updater.next() {
```
davidwalschots, line 568: **praise**: I like using an `Iterator` for this. Good idea!
**suggestion**: The idea is great, but this doesn't yet really use the `Iterator` semantics in the sense that you could also add a `next` function to the `FixedStepLoop` without implementing `Iterator` and it would still work. I would suggest really implementing this throughout such that you end up with simply having a for loop over items coming from `self.core_hydraulic_updater`.

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -430,6 +455,27 @@ impl A320Hydraulic {
         self.yellow_loop.is_pressurised()
     }
 
+    #[allow(clippy::too_many_arguments)]
```
davidwalschots, line 458:
```suggestion
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -251,19 +322,12 @@ impl A320Hydraulic {
...
-            Duration::from_millis(Self::HYDRAULIC_SIM_TIME_STEP_MILLISECONDS);
-
-        self.total_sim_time_elapsed += context.delta();
+        self.core_hydraulic_updater.update(context);
+        self.physics_updater.update(context);
 
-        // Time to catch up in our simulation = new delta + time not updated last iteration
-        let time_to_catch = context.delta() + self.lag_time_accumulator;
+        // Here we update at the same rate as the sim or max time step if sim rate is too slow
+        self.update_max_fixed_step(&context);
```
davidwalschots, line 329: **thought**: The comment here and on the function itself are different. Consider removing this comment.

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -430,6 +455,27 @@ impl A320Hydraulic {
         self.yellow_loop.is_pressurised()
     }
 
+    #[allow(clippy::too_many_arguments)]
+    // Update with same refresh rate as the sim
+    fn update_max_fixed_step(&mut self, context: &UpdateContext) {
+        while let Some(cur_time_step) = self.physics_updater.next() {
```
davidwalschots, line 461: **thought**: The `MaxStepLoop`'s name differs from the `max_fixed_step` name of this function. I wonder if those names should be the same?

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -430,6 +455,27 @@ impl A320Hydraulic {
...
+                self.yellow_loop.pressure(),
+            );
+
+            self.ram_air_turbine
+                .update_physics(&cur_time_step, context.indicated_airspeed());
+        }
+    }
+
     #[allow(clippy::too_many_arguments)]
     // Update with same refresh rate as the sim
```
davidwalschots, line 480: **thought**: The comment on `update_every_frame` is equal to the comment on `update_max_fixed_step`, yet they do seem to update at different rates. Something feels wrong with this.

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -274,55 +338,16 @@ impl A320Hydraulic {
...
+        // Here we update at a fixed rate only. Mainly core hydraulics
+        self.update_fixed_step(
+            &context,
+            engine1,
+            engine2,
+            overhead_panel,
+            engine_fire_push_buttons,
+            lgciu1,
+            lgciu2,
+        );
```
davidwalschots, line 350: **suggestion**: Consider changing the structure of the `update` function, such that the amount of "steps" taken is only known at this level and doesn't leak further down. For example:

```rust
while let Some(step_delta) = self.physics_updater.next() {
    self.update_physics(&context.with_delta(step_delta))
}

while let Some(step_delta) = self.some_new_updater_which_does_it_per_frame.next() {
    self.update_something(
            &context.with_delta(step_delta),
            &overhead_panel,
            &autobrake_panel,
            rat_and_emer_gen_man_on,
            emergency_elec_state,
            lgciu1,
            lgciu2,
    );
}

while let Some(step_delta) = self.core_hydraulic_updater.next() {
    self.update_core_hydraulics(
            &context.with_delta(step_delta)
            engine1,
            engine2,
            overhead_panel,
            engine_fire_push_buttons,
            lgciu1,
            lgciu2,
    );
}
```

Note that this example hasn't yet taken into account the `Iterator` comment I made elsewhere.

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+
+            should_close_valves: true,
+            control_position_request: Ratio::new::<ratio>(0.),
+            should_unlock: false,
         }
     }
 
-    fn has_moved(&self) -> bool {
-        (self.position - self.previous_position).abs() > f64::EPSILON
+    fn update(&mut self, context: &UpdateContext, door: &CargoDoor, pressure_available: Pressure) {
```
davidwalschots, line 1594: **suggestion**: `pressure_available` reads like it would be a `bool` argument, but it is not. Rename to `available_pressure` to make clear you're talking about the available amount of pressure.
**note**: I would like to note that in other situations this is named `current_pressure`. I'd pick either `available_pressure` or `current_pressure` and use that throughout the code.
```suggestion
    fn update(&mut self, context: &UpdateContext, door: &CargoDoor, available_pressure: Pressure) {
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+            DoorControlState::UpLocked => {
+                self.should_close_valves = true;
+            }
+        }
+    }
+
+    fn determine_mode(
+        &mut self,
+        door: &CargoDoor,
+        pressure_available: Pressure,
```
davidwalschots, line 1639:
```suggestion
        available_pressure: Pressure,
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -503,137 +556,146 @@ impl A320Hydraulic {
...
-            context,
-            &self.yellow_loop,
-            &self.yellow_electric_pump_controller,
-        );
+            self.engine_driven_pump_2.update(
+                context,
+                &self.yellow_loop,
+                engine2
+                    .hydraulic_pump_output_speed()
+                    .get::<revolution_per_minute>(),
```
davidwalschots, line 630: **suggestion**: The actual measurement unit used is an internal concern of the `EngineDrivePump`. Thus I would suggest passing the `AngularVelocity` itself and only inside the `EngineDrivenPump` and `Pump` converting it to the quantity you need (basically at the point of calculation).

```suggestion
                engine2.hydraulic_pump_output_speed()
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -487,14 +539,15 @@ impl A320Hydraulic {
...
     }
 
     fn update_yellow_actuators_volume(&mut self) {
         self.yellow_loop
-            .update_actuator_volumes(&self.braking_circuit_altn);
-        self.braking_circuit_altn.reset_accumulators();
+            .update_actuator_volumes(&mut self.braking_circuit_altn);
+
+        self.yellow_loop
+            .update_actuator_volumes(self.forward_cargo_door.actuator());
```
davidwalschots, line 550: **thought**: Is the `aft_cargo_door` missing here?

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
     }
 }
 
-struct Door {
-    exit_id: String,
-    position: f64,
-    previous_position: f64,
+#[derive(PartialEq, Clone, Copy)]
+pub enum DoorControlState {
```
davidwalschots, line 1551: **suggestion**: As this enum is currently unused outside of the module, it can remain private.

```suggestion
enum DoorControlState {
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+                    self.control_position_request = Ratio::new::<ratio>(0.);
+                }
+            }
+            DoorControlState::UpLocked => {
+                self.should_close_valves = true;
+            }
+        }
+    }
+
+    fn determine_mode(
```
davidwalschots, line 1636: **thought**: `determine_mode` returns `DoorControlState`. Should `determine_mode` be named `determine_state`?
**thought**: The function also contains writes to `self.should_unlock`, thus `determine_state` might not fully capture the intent either.

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+            }
+            DoorControlState::UpLocked => {
+                if self.position_requested < Ratio::new::<ratio>(1.)
+                    && pressure_available > Pressure::new::<psi>(1000.)
+                {
+                    DoorControlState::HydControl
+                } else {
+                    DoorControlState::UpLocked
+                }
+            }
```
davidwalschots, line 1677: **suggestion**: Consider using match guards to reduce the duplication requiring an else branch returning the same state as before:

```suggestion
        match self.state {
            DoorControlState::DownLocked if self.position_requested > Ratio::new::<ratio>(0.) => {
                self.should_unlock = true;
                DoorControlState::NoControl
            }
            DoorControlState::NoControl
                if self.duration_in_no_control > Self::DELAY_UNLOCK_TO_HYDRAULIC_CONTROL =>
            {
                DoorControlState::HydControl
            }
            DoorControlState::HydControl if door.is_locked() => {
                self.should_unlock = false;
                DoorControlState::DownLocked
            }
            DoorControlState::HydControl
                if door.position() > Ratio::new::<ratio>(0.9)
                    && self.position_requested > Ratio::new::<ratio>(0.5) =>
            {
                self.should_unlock = false;
                DoorControlState::UpLocked
            }
            DoorControlState::UpLocked
                if self.position_requested < Ratio::new::<ratio>(1.)
                    && pressure_available > Pressure::new::<psi>(1000.) =>
            {
                DoorControlState::HydControl
            }
            _ => self.state,
        }
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+            DoorControlState::HydControl => {
+                self.should_close_valves = false;
+                if self.position_requested > Ratio::new::<ratio>(0.)
+                    || self.duration_in_hyd_control < Self::UP_CONTROL_TIME_BEFORE_DOWN_CONTROL
+                {
+                    self.control_position_request = Ratio::new::<ratio>(1.);
+                } else {
+                    self.control_position_request = Ratio::new::<ratio>(0.);
+                }
```
davidwalschots, line 1628:
```suggestion
                self.control_position_request = if self.position_requested > Ratio::new::<ratio>(0.)
                    || self.duration_in_hyd_control < Self::UP_CONTROL_TIME_BEFORE_DOWN_CONTROL
                {
                    Ratio::new::<ratio>(1.)
                } else {
                    Ratio::new::<ratio>(0.)
                }
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+    }
+    fn position_request(&self) -> Ratio {
+        self.control_position_request
+    }
+    fn should_lock(&self) -> bool {
+        !self.should_unlock
+    }
+    fn lock_position_request(&self) -> Ratio {
+        Ratio::new::<ratio>(0.)
+    }
```
davidwalschots, line 1703: **nitpick**: These functions need some breathing space (i.e. whitespace).

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1479,28 +1547,217 @@ impl SimulationElement for A320BrakingForce {
...
+}
+impl SimulationElement for A320DoorController {
     fn read(&mut self, state: &mut SimulatorReader) {
-        self.previous_position = self.position;
-        self.position = state.read(&self.exit_id);
+        self.position_requested = Ratio::new::<ratio>(state.read(&self.requested_position_id));
```
davidwalschots, line 1707:
```suggestion
        self.position_requested = Ratio::new::<ratio>(reader.read(&self.requested_position_id));
```

src/systems/a320_systems/src/hydraulic.rs
crocket63: Damn I thought that's what I did by implementing the iterator next() trait
"impl Iterator for FixedStepLoop "

You mean I should override more iterator default methods?

src/systems/a320_systems/src/hydraulic.rs
crocket63: Well, yes but it's missing in the plane too :D

src/systems/a320_systems/src/hydraulic.rs
crocket63: Thanks, was obvious to do it this way, I totally brain farted!

src/systems/a320_systems/src/hydraulic.rs
crocket63: That's more a MaxFixedStepLoop maybe :)

src/systems/a320_systems/src/hydraulic.rs
crocket63: So. I made it work using #[derive(Clone, Copy)] for iterators. 

It works but I'm not sure of what this implies under the hood there.

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.number_of_loops_remaining > 0 {
+            self.number_of_loops_remaining -= 1;
+            Some(self.time_step)
+        } else {
+            None
+        }
+    }
+}
```
davidwalschots, line 69:
```suggestion
impl IntoIterator for &mut FixedStepLoop {
    type Item = Duration;

    type IntoIter = std::vec::IntoIter<Self::Item>;

    fn into_iter(self) -> Self::IntoIter {
        let mut v = vec![];
        while self.number_of_loops_remaining > 0 {
            self.number_of_loops_remaining -= 1;
            v.push(self.time_step);
        }

        IntoIterator::into_iter(v)
    }
}
```

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,284 @@
+use crate::simulation::UpdateContext;
+use std::{iter::Iterator, time::Duration};
+
+#[derive(Clone, Copy)]
```
davidwalschots, line 4:
```suggestion
```

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -251,19 +320,13 @@ impl A320Hydraulic {
...
-
-        // Time to catch up in our simulation = new delta + time not updated last iteration
-        let time_to_catch = context.delta() + self.lag_time_accumulator;
+        self.core_hydraulic_updater.update(context);
+        self.physics_updater.update(context);
 
-        // Number of time steps (with floating part) to do according to required time step
-        let number_of_steps_floating_point =
-            time_to_catch.as_secs_f64() / min_hyd_loop_timestep.as_secs_f64();
+        for cur_time_step in self.physics_updater {
```
davidwalschots, line 326:
```suggestion
        for cur_time_step in &mut self.physics_updater {
```

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: This means it will copy the `FixedStepLoop` type every time you iterate over it, which isn't really what you want. I'll add a separate review with suggestions.

davidwalschots (commented):
RE: Iterator. I'll leave implementing it on the other types as an exercise to the author. Be sure to read [this page](https://doc.rust-lang.org/std/iter/index.html) in the documentation for a good overview.

Note that a (potentially nicer) alternative would be creating your own Iterator type which would take a reference to the type that calculates the duration (so `FixedStepLoop` and the other types), and would implement `Iterator`.

src/systems/systems/src/hydraulic/update_iterator.rs
crocket63: Ok makes more sense! So it's called once to actually build the vector at the start of for loop.

lolisec (approved):

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -36,7 +48,61 @@ use systems::{
     },
 };
 
+struct A320CargoDoorFactory {}
+impl A320CargoDoorFactory {
+    fn a320_cargo_door_actuator(rigid_body: &LinearActuatedRigidBodyOnHingeAxis) -> LinearActuator {
+        LinearActuator::new(
+            rigid_body.max_linear_distance_to_anchor(),
+            rigid_body.min_linear_distance_to_anchor(),
```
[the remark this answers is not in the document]
davidwalschots: @crocket63 reopening this one. The intent is to replace the argument of `LinearActuator`, otherwise this whole exercise wasn't really useful 😀.

```rust
    fn a320_cargo_door_actuator(bounded_linear_length: &impl BoundedLinearLength) -> LinearActuator {
        LinearActuator::new(
            rigid_body
            rigid_body
            2,
            Length::new::<meter>(0.04422),
            Length::new::<meter>(0.03366),
            VolumeRate::new::<gallon_per_second>(0.01),
            600000.,
            15000.,
            500.,
        )
    }

impl LinearActuator {
    const DEFAULT_I_GAIN: f64 = 0.2;
    const DEFAULT_P_GAIN: f64 = 0.05;

    #[allow(clippy::too_many_arguments)]
    pub fn new(
        bounded_linear_length: &impl BoundedLinearLength,
        number_of_actuators: u8,
        bore_side_diameter: Length,
        rod_diameter: Length,
        max_flow: VolumeRate,
        fluid_compression_spring_constant: f64,
        fluid_compression_damping_constant: f64,
        active_hydraulic_damping_constant: f64,
    ) -> Self {
        let total_travel = bounded_linear_length.max() - bounded_linear_length.min();
        // ...
    }
}
```

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: @crocket63 Unresolved this one for now. The implementation still leaks, in the sense that the function names at the lower level are still called `update_max_fixed_step`, `update_every_frame`, and `update_fixed_step`, thereby removing the point of the exercise of moving the "execution control" to a higher level.

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: @crocket63 Unresolving, as `Pump` still receives an `f64`.

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: @crocket63 This confuses me, as an `aft_cargo_door` field is found in `A320Hydraulic`. Do you mean the current 3D model doesn't contain the door? If so, then I think you need to pick and choose: either add the code for that door even if it's not there at the moment, or remove all references to it. I would go for the former 😀.

crocket63: Indeed we don't have it on the model sadly, but for consistency I added it there because that will be the really easy thing to forget to add when adding the aft door...  I think code is complete for it to work on hydraulic part, we only need to wait for a animation node now

src/systems/a320_systems/src/hydraulic.rs
crocket63: Do you mean in hydraulic/mod.rs ?  I did not go further down as this should all be updated in the next overhaul PR

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: So the aft cargo door when actually available doesn't affect the yellow loop?

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: Yes, as a general rule always try to pass `uom` types until the moment you need to get the actual measurement unit in a calculation.

src/systems/a320_systems/src/hydraulic.rs
crocket63: It is definitly and if you're not seeing the line I added for the aft door I don't know where the lines I wrote went :(

crocket63:
Need to investigate some broken tests plus some code I wrote that disapeared somehow :(

BritishAvgeek:

Object of testing: #5341 
Tier of Testing: 1
Date: 20/09/2021

Testing Process: Spawned C&D and began a turnaround - tested Cargo door open and close functions multiple times

Negatives: N/A
Testing Results: Passed - though I note crocket mentioning above there may be something broke somewhere, I couldn't find anything but will test again later should something some up and I have time

Media:
Door was cracked when I selected it, then began to open as the cart began to move towards the aircraft. Beautifully timed. Door bounced slightly at the top as it got to the fully open position.

Wow!
image

crocket63:
> Quality Assurance Tester Report
> 
> Object of testing: #5341
> Tier of Testing: 1
> Date: 20/09/2021
> 
> Testing Process: Spawned C&D and began a turnaround - tested Cargo door open and close functions multiple times
> 
> Negatives: N/A
> Testing Results: Passed - though I note crocket mentioning above there may be something broke somewhere, I couldn't find anything but will test again later should something some up and I have time
> 
> Media:
> Door was cracked when I selected it, then began to open as the cart began to move towards the aircraft. Beautifully timed. Door bounced slightly at the top as it got to the fully open position.
> 
> Wow!
> image

Might be that you can only access a build where tests and code was still good. 
Hope you'll be happy to open it once again when I finally have some time to investigate those issues :)
Thanks!

src/systems/a320_systems/src/hydraulic.rs
davidwalschots: Overlooked it!

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
...
+            self.number_of_loops_this_iteration = 0;
+        } else {
+            // Int part is the actual number of loops to do
+            // rest of floating part goes into accumulator
+            self.number_of_loops_this_iteration = number_of_steps_floating_point.floor() as u32;
+
+            self.lag_time_accumulator = Duration::from_secs_f64(
+                (number_of_steps_floating_point - (self.number_of_loops_this_iteration as f64))
+                    * self.time_step().as_secs_f64(),
+            ); // Keep track of time left after all fixed loop are done
```
davidwalschots, line 51: **suggestion**: Be sure to test this, I think it should work, but not entirely sure. See the [docs](https://doc.rust-lang.org/std/primitive.f64.html#method.fract).

```suggestion
            // Keep track of time left after all fixed loop are done
            self.lag_time_accumulator =
                Duration::from_secs_f64(number_of_steps.fract() * self.time_step().as_secs_f64());
```

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
...
+    time_step: Duration,
+    number_of_loops_this_iteration: u32,
+    number_of_loops_remaining: u32,
+}
+impl FixedStepLoop {
+    pub fn new(time_step: Duration) -> Self {
+        Self {
+            lag_time_accumulator: Duration::from_millis(0),
+            time_step,
+            number_of_loops_this_iteration: 0,
```
davidwalschots, line 15: **suggestion**: `number_of_loops_this_iteration` can be removed and all writes to it can write to `number_of_loops_remaining`.
```suggestion
```

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
...
+            self.lag_time_accumulator = Duration::from_secs_f64(
+                number_of_steps_floating_point * self.time_step().as_secs_f64(),
+            );
+            self.number_of_loops_this_iteration = 0;
+        } else {
+            // Int part is the actual number of loops to do
+            // rest of floating part goes into accumulator
+            self.number_of_loops_this_iteration = number_of_steps_floating_point.floor() as u32;
+
+            self.lag_time_accumulator = Duration::from_secs_f64(
```
davidwalschots, line 48: **suggestion**: Consider ordering the setting of variables in the same way in both the `if` and `else` block.

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
+use crate::simulation::UpdateContext;
+use std::time::Duration;
+
+pub struct FixedStepLoop {
```
davidwalschots, line 4: **thought**: I think this type and `MaxFixedStepLoop` can really use a docs comment describing what they are. Right now to use them one has to read the code to figure out what it is they do. As these structs might be useful to other problem spaces outside of hydraulics, having better comments would be good.

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
...
+        while self.number_of_loops_remaining > 0 {
+            self.number_of_loops_remaining -= 1;
+            v.push(self.time_step);
+        }
+
+        IntoIterator::into_iter(v)
+    }
+}
+
+pub struct MaxFixedStepLoop {
```
davidwalschots, line 73: **question**: Reading the code in this type I've been unable to decipher what exactly it tries to do. Could you elaborate in functional terms (either here or as part of a docs comment) on what this type is trying to achieve?

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,280 @@
...
+    max_time_step: Duration,
+    num_of_max_step_loop: u32,
+    remaining_frame_duration: Option<Duration>,
+    number_of_loops_this_iteration: u32,
```
davidwalschots, line 77: **issue**: Field is never read.
```suggestion
```

src/systems/systems/src/simulation/update_context.rs
```diff
@@ -108,6 +128,22 @@ impl UpdateContext {
         self.longitudinal_acceleration
     }
 
+    pub fn lat_accel(&self) -> Acceleration {
+        self.lateral_acceleration
+    }
+
+    pub fn vert_accel(&self) -> Acceleration {
+        self.vertical_acceleration
+    }
```
davidwalschots, line 137: **suggestion**: As these three types of acceleration are related, consider wraping `longitudinal_acceleration`, `lateral_acceleration`, and `vertical_acceleration` into a struct of its own which is contained within `UpdateContext`. This struct could then also contain functions such as returning a m/s^2 `Vector3`.

```rust
context.acceleration().to_ms2_vector();
```
```

File: src/systems/systems/src/simulation/update_context.rs [hunk 37]
```diff
@@ -108,6 +128,22 @@ impl UpdateContext {
...
+
+    pub fn pitch(&self) -> Angle {
+        self.pitch
+    }
+
+    pub fn bank(&self) -> Angle {
+        self.bank
```
davidwalschots, line 144: **thought**: Maybe you also want to do something funky like I suggested above here?

src/systems/systems/src/hydraulic/update_iterator.rs
crocket63: Pretty sure I had something but I might have put it in the update part not here indeed, will write something!

src/systems/systems/src/hydraulic/update_iterator.rs
crocket63: So basically you just force a max time step allowed. It can go faster, not slower.
If max step is 20ms, and sim refreshes in 10ms, it will just do one 10ms step.
If sim refreshes in 30ms, it will first do a 20ms step, then complete with the remaining 10ms step.

jcalor:

Discord          : Jorge#6816
Object of testing: #5341 
Tier of Testing  : 1
Date             : 20/09/2021

Testing Process: 
- Spawned C&D, opened Cargo Door a few times to see its operation.
- Checked ECAM EWD DOOR page; cargo door was opened
- Once door is closed, checked again ECAM EWD DOOR page; door was closed.
- Checked ECAM EWD HYD page; cargo door starts yellow pump
- Once cargo door was opening and closing, I´ve been alt-tabbing windows - no issues at all

Negatives: N/A

Testing Results: Passed

Conclusions: No issues were found, PR works as expected - cargo door pressurizes yellow pump, once it is opened, yellow pump depressurizes; when door is closed, yellow pump pressurizes again and when door is completely closed, pump depressurizes again. PS: Phisical movement is amazing!

Media:       
image
image
HYD Page once cargo door is fully opened:
image

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,712 @@
...
+    f64::*,
+    force::newton,
+    length::meter,
+    ratio::ratio,
+    velocity::meter_per_second,
+    volume::{cubic_meter, gallon},
+    volume_rate::gallon_per_second,
+};
+
+use super::rigid_body::LinearActuatedRigidBodyOnHingeAxis;
```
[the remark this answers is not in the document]
crocket63: Had a lot of thoughts and hesitations on this one. The rigid body implemented is already a really particular one that is hinge constrained and can be moved by a linear actuator connected to its control arm.

Generally speaking, we would have a standard non constrained rigid body accepting a force applied anywhere or a torque applied anywhere. Then a constrained rigid body would be the same but with a different update. Then a linear actuated one would be the constrained one but having a control arm to connect a linear actuator.

BUT... for now I just can't think of a different one we'd have to simulate actually as all moving parts on that plane are actually sumed up in a linear actuated rigid body.

So I'd keep it simple like this until proved otherwise by an actuator/body couple I didn't think of.  Ideas?  Can only think of the flap slat mechanism, but in all my researches I ended up not simulating the rigid body at all for flaps and going for a different approach.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,712 @@
...
+            fluid_compression_damping_constant,
+            active_hydraulic_damping_constant,
+        }
+    }
+
+    fn update_before_rigid_body(
+        &mut self,
+        connected_body: &mut LinearActuatedRigidBodyOnHingeAxis,
+        requested_mode: LinearActuatorMode,
+        available_pressure: Pressure,
```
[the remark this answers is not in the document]
crocket63: Let's standardise!

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,712 @@
...
+        let position_error = self.opened_valves_target_position - self.position_normalized;
+
+        let mut open_loop_flow_target = 0.;
+        if position_error >= Ratio::new::<ratio>(0.001) {
+            open_loop_flow_target = self.max_flow.get::<gallon_per_second>();
+        } else if position_error <= Ratio::new::<ratio>(-0.001) {
+            open_loop_flow_target = self.min_flow.get::<gallon_per_second>();
+        }
+
+        let flow_error = open_loop_flow_target - self.signed_flow.get::<gallon_per_second>();
```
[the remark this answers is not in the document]
crocket63: Must admit I really dislike that rust sugar syntax :p  Will try but... beuaaaargh :p

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,712 @@
...
+
+    fn open_control_valves_for_position_control(&mut self) {
+        self.mode = LinearActuatorMode::PositionControl;
+    }
+
+    fn set_position_target(&mut self, target_position: Ratio) {
+        self.opened_valves_target_position = target_position;
+    }
+
+    fn compute_control_force(&mut self, hydraulic_pressure: Pressure) {
```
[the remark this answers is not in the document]
crocket63: Need to have a closer look on that implementation, not sure how it would look with that Enum/structs !?

For the modes, those are actually functional modes of the actuator implying different way of computing forces. This is basically an abstraction of what would happen with internal solenoids and super low level hydraulic sorcery we can't even imagine simulate.
There is a state machine but it is at the charge of the actuator user. The actuator just receives current mode through trait HydraulicAssemblyController::requested_mode().

For example an aileron controller would implement HydraulicAssemblyController::requested_mode(), and will send position control if there is pressure and computer is alive, else damping mode.

lolisec (approved):

src/systems/systems/src/hydraulic/linear_actuator.rs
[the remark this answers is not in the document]
davidwalschots: As discussed on Discord, it might be that parts of both modules are so (logically) interconnected that they belong in the same module.

sidnov:
@lolisec kindly refrain from approving changes which are not authorised

src/systems/systems/src/hydraulic/update_iterator.rs
```diff
@@ -0,0 +1,288 @@
...
+use std::time::Duration;
+
+/// This is a structure that will handle a fixed step like update.
+/// When update() is called, it will build an array of frame updates to do, all having the same fixed duration
+/// If all those fixed steps added do not sum up as the whole update time, a time lag accumulator is incremented so we catch up
+/// with real time at next updates.
+///
+/// Usage:
+/// Call update() at each sim updtate
+/// Then a for loop over this as an iterator will return one frame per loop with corresponding fixed step duration
```
davidwalschots, line 11: **thought**: I find this description to be very technical. It describes the implementation and after reading it I still have a hard time understanding what this type is supposed to do.

```suggestion
/// Provides fixed time interval looping.
///
/// ## Example scenario
/// With a fixed time interval of 10 ms and a frame delta of 35 ms, this type will provide three
/// iterations of 10 ms. The remaining 5 ms will carry over to the next frame. When the next frame
/// takes 15 ms, this will result in 2 iterations of 10 ms (15 ms + 5 ms).
```

src/systems/systems/src/simulation/update_context.rs
davidwalschots: I definitely like the `LocalAcceleration` change! Putting things in an array in `LocalAcceleration` isn't really necessary. By just having all three values as fields you could get rid of some complexity, so I'd suggest maybe still doing that.

src/systems/systems/src/simulation/update_context.rs
davidwalschots: Indeed just putting it in an array isn't very useful. I'd suggest a type:
```rust
struct Attitude {
    pitch: Angle;
    bank: Angle;
}
impl Attitude {
    fn pitch_rotation(&self) -> Rotation3<f64> { // Not quite sure about this function name, you probably have a better one!
        Rotation3::from_axis_angle(&Vector3::x_axis(), context.pitch().get::<radian>())
    }
}

File: src/systems/systems/src/hydraulic/linear_actuator.rs [hunk 43]
```diff
@@ -0,0 +1,1302 @@
...
+/// functional modes
+///
+/// ClosedValves -> Turns actuator in a high constant spring/damper system simulating a closed actuator
+/// only constrained by its own fluid compressibility
+///
+/// ActiveDamping -> Actuator use internal valves to provide a force resisting to its own movements, dampening
+/// the piece movements it's connected to
+///
+/// PositionControl -> Actuator will try to use hydraulic pressure to move to a requested position, while
+/// maintaining flow limitations
```
davidwalschots, line 56: **suggestion**: Rust docs support markdown, using it makes things clearer.
**suggestion**: Consistent usage of `.`.

```suggestion
/// Represents an abstraction of the low level hydraulic actuator control system that would in real life consist of a lot of
/// solenoid control valves, spring loaded valves, and a differential pressure mechanism.
///
/// We don't want to simulate all of those little bits, so the functions of the actuator are split into
/// the following functional modes:
///
/// - [LinearActuatorMode.ClosedValves]: Turns actuator in a high constant spring/damper system simulating a closed actuator
/// only constrained by its own fluid compressibility.
/// - [ActiveDamping.LinearActuatorMode]: Actuator use internal valves to provide a force resisting to its own movements, dampening
/// the piece movements it's connected to.
/// - [LinearActuatorMode.PositionControl]: -> Actuator will try to use hydraulic pressure to move to a requested position, while
/// maintaining flow limitations.
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+            fluid_compression_damping_constant,
+            max_flow,
+            min_flow,
+            flow_error_prev: VolumeRate::new::<gallon_per_second>(0.),
+            bore_side_area,
+            rod_side_area,
+            last_control_force: Force::new::<newton>(0.),
+        }
+    }
+    fn update_force(
```
davidwalschots, line 105:
```suggestion
    }
    
    fn update_force(
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+    fn actions_from_current_to_closed_valves(&mut self, position_normalized: Ratio) {
+        match self.current_mode {
+            LinearActuatorMode::ClosedValves => {}
+            LinearActuatorMode::PositionControl => {
+                self.go_to_close_control_valves(position_normalized);
+            }
+            LinearActuatorMode::ActiveDamping => {
+                self.go_to_close_control_valves(position_normalized);
+            }
+        }
```
davidwalschots, line 144: **suggestion**: Can be shortened to:
**note**: After this change, you could consider inlining `go_to_close_control_valves` and the other functions, as there is only one usage.

```suggestion
        match self.current_mode {
            LinearActuatorMode::ClosedValves => {}
            LinearActuatorMode::PositionControl | LinearActuatorMode::ActiveDamping => {
                self.go_to_close_control_valves(position_normalized);
            }
        }
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+    fn actions_from_current_to_position_control(&mut self) {
+        match self.current_mode {
+            LinearActuatorMode::ClosedValves => {
+                self.go_to_position_control();
+            }
+            LinearActuatorMode::PositionControl => {}
+            LinearActuatorMode::ActiveDamping => {
+                self.go_to_position_control();
+            }
+        }
```
davidwalschots, line 156:
```suggestion
        match self.current_mode {
            LinearActuatorMode::PositionControl => {}
            LinearActuatorMode::ClosedValves | LinearActuatorMode::ActiveDamping => {
                self.go_to_position_control();
            }
        }
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+    fn actions_from_current_to_damping(&mut self) {
+        match self.current_mode {
+            LinearActuatorMode::ClosedValves => {
+                self.go_to_damping();
+            }
+            LinearActuatorMode::PositionControl => {
+                self.go_to_damping();
+            }
+            LinearActuatorMode::ActiveDamping => {}
+        }
```
davidwalschots, line 168:
```suggestion
        match self.current_mode {
            LinearActuatorMode::ActiveDamping => {}
            LinearActuatorMode::ClosedValves | LinearActuatorMode::PositionControl => {
                self.go_to_damping();
            }
        }
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+                required_position,
+                position_normalized,
+                signed_flow,
+                current_pressure,
+                speed,
+            ),
+        }
+    }
+
+    fn force_damping(&mut self, speed: Velocity) -> Force {
```
davidwalschots, line 207:
```suggestion
    fn force_damping(&self, speed: Velocity) -> Force {
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+        Force::new::<newton>(
+            -speed.get::<meter_per_second>() * self.active_hydraulic_damping_constant,
+        )
+    }
+
+    fn force_closed_valves(&mut self, position_normalized: Ratio, speed: Velocity) -> Force {
```
davidwalschots, line 213:
```suggestion
    fn force_closed_valves(&self, position_normalized: Ratio, speed: Velocity) -> Force {
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -0,0 +1,1302 @@
...
+            LinearActuatorMode::ActiveDamping => self.force_damping(speed),
+            LinearActuatorMode::PositionControl => self.force_position_control(
+                required_position,
+                position_normalized,
+                signed_flow,
+                current_pressure,
+                speed,
+            ),
+        }
+    }
```
davidwalschots, line 205: **thought**: The name makes this function feel like a getter or something that calculates a value and returns it. This is true for `ClosedValves` and `ActiveDamping`. `PositionControl` however does adapt `self`. Should this function also be called `update_force`? Should an `update` function return a value, or should the updating and getting be separated?

davidwalschots (commented):
Small stuff! After this we're fully ready!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Clippy wouldn't catch those normally?

davidwalschots: Nope!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Sounded easy to use down into the LinearActuator but in fact just updating then getting is perfectly nice, and I removed the useless force from LinearActuator

davidwalschots (approved):
🥳

crocket63:
> 🥳

Thanks for the big help!   More fun to come :D

BritishAvgeek:

Object of testing: #5341
Tier of Testing: 1
Date: 23/09/2021

Testing Process: Spawned C&D and began a turnaround - tested Cargo door open and close functions multiple times. A repeat of my previous test to find any differences and similarities

Negatives: N/A
Testing Results: Passed

Such a small feature but when you see it/watch it, it's wonderful to observe! All the little accurate animations etc make this so immersive in addition to all the other work done so thank you :) I tried all "normal" type operating use and couldn't break it - not gone farfetched with it in testing in the time I had

crocket63:
Thanks! Little detail but all control surfaces and gear will benefit from the same thing :)

FlightSimmer7700:
Quality Assurance Tester

Object of testing: #5341
Tier of Testing: 1
Date: 23/9/21

Testing Process: 
Loaded in C&D at multiple airfileds and tried opening the cargo door and observed the animation
Loaded on approach, runway and mid-air to test if the door was able to be openened 
Used Slew mode to get airborne as soon as the door opened and observed it close automatically
Negatives: N/A
Testing Results: Passed

Conclusions: Looks amazing great work!
Media: https://streamable.com/pgn9ue

FlightSimmer7700:
> 
> 
> Quality Assurance Tester
> 
> Discord: FlightSimmer7700#5438
> Object of testing: #5341
> Tier of Testing: 1
> Date: 23/9/21
> 
> Testing Process:
> Loaded in C&D at multiple airfileds and tried opening the cargo door and observed the animation
> Loaded on approach, runway and mid-air to test if the door was able to be openened
> Used Slew mode to get airborne as soon as the door opened and observed it close automatically
> Negatives: N/A
> Testing Results: Passed
> 
> Conclusions: Looks amazing great work!
> Media: https://streamable.com/pgn9ue

Was doing some further testing after this test and found an issue when "cancelling" a cargo door opening where the elec yellow pump stays on and doesn't turn off as expected
https://streamable.com/ysbbs9

ShadowFrost3:


Object of testing: #5341
Tier of Testing:1
Date: 23/09/2021

Testing Process: 
C&D 
Open door - works
Close door - works
Open door but then decide to close - works but the animation waits on the people loading cargo and Yellow pump stays on - fail
Turn yellow pump elec off - stays on - fail
Turn yellow pump off on new aircraft - door opens still yellow pump turns on - fail 
-notes- considering I'm hearing quite the whine from the pumps and lower ECAM indication I am assuming they are not using a handpump system which is available 
Negatives: 
Testing Results: Failed

Conclusions: Please educate me on the ways of hydraulics if I am misinformed on the lower points

crocket63:
> 
> Object of testing: #5341
> Tier of Testing:1
> Date: 23/09/2021
> 
> Testing Process:
> C&D
> Open door - works
> Close door - works
> Open door but then decide to close - works but the animation waits on the people loading cargo and Yellow pump stays on - fail
> Turn yellow pump elec off - stays on - fail
> Turn yellow pump off on new aircraft - door opens still yellow pump turns on - fail
> -notes- considering I'm hearing quite the whine from the pumps and lower ECAM indication I am assuming they are not using a handpump system which is available
> Negatives:
> Testing Results: Failed
> 
> Conclusions: Please educate me on the ways of hydraulics if I am misinformed on the lower points
> Media:
> image
> (https://www.aviationmatters.co/airbus-a320-hydraulic-system/)
> 
> Edit- After looking at your PR I realize the last point may or may not be applicable? I defer to your discussion and will make a decision on pass/fail once that has been discussed and the pump staying on has been remedied.

Expected behaviour is that you can't turn off the yellow pump manually if it's used for cargo door operations, which overrides it (cargo door moving + 20s). Only way to turn it off "manually" is shutting its power off, which means unpluging ground electric power.

So nothing unexpected here, I'll still investigate the always on issue!

crocket63:
> Door opens with no ground power or battery (pump off) using handpump? - Correct? (Do you actually have a handpump modeled? Residual Pressure? It should be noted I never pressurized the hyd system in this test, no elec power was every available)

Let me clear some details :)

So first thing, hand pump is not modeled.

Then, for ground power:  it's available for the yellow pump as soon as it is plugged to the plane. Wether you click or not on the external power button in the plane has NO INFLUENCE.  To test without ground power means that plane is not connected, so it's in a place where the ext power "avail" on the button would not show up.

> Door does not close when using ground power/battery off conditions (handpump) - Incorrect
Can you elaborate on what you are doing there? What conditions are you expecting to close the door there?

> Door close once Ground power is restored but does not operate on battery power alone - unknown
Can you also exactly tell what are the steps of this test case? Door won't open if plane is in a country airport without available ground power maybe?

ShadowFrost3:


Object of testing: #5341
Tier of Testing:1
Date: 24/09/2021

Testing Process: C&D
Open cargo door with ground power = checked
Close door with ground power mid-process = door close pump remains on ~20 seconds = checked and correct?
Open cargo with no power (grnd switch) (grnd availible) = opens and pump on (audible) = correct
Close cargo door with no power (grnd switch) (grnd availible) = closes and pump on (audible) = correct
Open cargo door with no power (grnd not availible) = opens without pump on (not audible) = handpump? = incorrect
Close door with no power (grnd not availible) = does not close = incorrect

Negatives: Unpower cargo door still moves and does not close
Testing Results: Partial pass - Normal operations work as expected
Conclusions:

ShadowFrost3:


Object of testing: #5341
Tier of Testing:1
Date: 24/09/2021

Testing Process: 
Open and close with power = checked
Open without power = gravity position = checked
Close without power = gravity position = checked
Remove power while opening = gravity position = checked

Negatives: If the door is open all the way the ground crew will phase through the door to load packages
Testing Results: Overall, 95% passed 

Conclusions: A small issue in a extremely unlikely range and designed scenario to test the limits of the system. There are several options to avoid the above that can be pursued at a future date. (IE locking the door when no power is applied to the aircraft SOP correct or modeling a hand pump SOP available but not correct for cargo operations AFAIK)  
Media: 
image

crocket63:
> 
> Object of testing: #5341
> Tier of Testing:1
> Date: 24/09/2021
> 
> Testing Process:
> Open and close with power = checked
> Door can be open/closed mid process (You can change your mind, IE pressing the button mid action) = checked
> Open without power = gravity position = checked
> Close without power = gravity position = checked
> Remove power while opening = gravity position = checked
> 
> Negatives: If the door is open all the way the ground crew will phase through the door to load packages.
> (EFB) Baggage will indicate blue if ground crew completes their job but the cargo door is not locked (due to lack of power).
> Testing Results: Overall, 95% passed
> 
> Conclusions: A small issue in a extremely unlikely range and designed scenario to test the limits of the system. There are several options to avoid the above that can be pursued at a future date. (IE locking the door when no power is applied to the aircraft SOP correct or modeling a hand pump SOP available but not correct for cargo operations AFAIK)
> Media: Gravity positon
> image

Good to me, we'll try to handle edge case with MSFS2020 animation and EFB interactions in a next step if we feel it's needed.

Thanks for the help testing these edge cases!

jcalor:

Discord          : Jorge#6816
Object of testing: #5341 
Tier of Testing  : 1
Date             : 27/09/2021

Testing Process: 

GPU connected, APU OFF:

- First opening and closing door done with Yellow Pump ON, no issues - system were pressurizing and cargo door page showing opened when open was open and closed when it was closed.
- Second opening and closing door done with Yellow Pump OFF, no issues - system were pressurizing and cargo door page showing opened when open was open and closed when it was closed.

GPU disconnected, APU ON:
- First opening and closing door done with Yellow Pump ON, no issues - system were pressurizing and cargo door showing opened when open was open.
- Second opening and closing door done with Yellow Pump OFF, no issues - system were pressurizing and cargo door showing opened when open was open.

Additional tests: 

Opened and closed door several times while it was opening/closing (on its middle action) - no issues.
Opened and closed door without letting HYD system to depressurize - no issues

Negatives: N/A

Testing Results: Passed

Conclusions: No issues were found - PR works as expected.

Media:
Both tests were succesfull both GPU connected + APU OFF and GPU disconnected and APU ON = same screenshots

1 - Door opened and system presurized with yellow pump ON (system pressure working well)
image
image
2 - Yellow pump OFF. System presurized with the opening and closing door
image
image
image

PR #5344, opened 2021-07-18
fix: FlyPad Unit Selection Breaks Weights, Duplicates MCDU Settings

tracernz:
Fixes #[issue_no]

## Summary of Changes

The FlyPad set the `CONFIG_USING_METRIC_UNITS` setting to 1 for metric/kg and 0 for imperial/lb, which seems reasonable given the name of the setting. Unfortunately it was a setup, and that setting should have been called CONFIG_WEIGHT_CONVERSION_FACTOR or something. This caused all weights to be zero when imperial units were selected on the FlyPad.

The FlyPad duplicates many of the settings on the "fake" MCDU options pages. I have removed most of them (partly to avoid fixing things that should be dead).

I have implemented syncing of NXDataStore settings across instruments, and added a new unit conversion class for the MCDU/ECAM code that takes care of this in one place. The aircraft no longer needs reloaded to change the unit selection.

TODO in future (I will open an issue, these are notes to self until then):
- Fix SimBrief setting on FlyPad so it uses the proper setting, and stores pilot ID rather than username, then remove that setting from the MCDU.
- Add Telex setting + warning to FlyPad and remove from MCDU
- Add keyboard input setting to FlyPad and remove from MCDU
- 

## Additional context


flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXUnits.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Unit conversion utilities
+ */
+
+class NXUnits {
+    static get metricWeight() {
+        if (NXDataStore._metricWeight === undefined) {
```
tracernz, line 7: derp. this is why it's WiP

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Refuel.js
```diff
@@ -68,7 +67,7 @@ class A32NX_Refuel {
         const LOutTarget = LOutTargetSimVar;
         const RInnTarget = RInnTargetSimVar;
         const ROutTarget = ROutTargetSimVar;
-        if (refuelRate == 2) {
+        if (refuelRate == '2') {
```
Benjozork, line 70: **suggestion:** Since we are gonna rewrite this, and don't wanna refactor too much, can we add a comment saying what that magic value is ?

Benjozork (changes requested):

ErickSharp:
> Unfortunately it was a setup,

LOL. Instead of using `CONFIG_USING_METRIC_UNITS`, kgToUser is preferred, correct?

tracernz:
> > Unfortunately it was a setup,
> 
> LOL. Instead of using `CONFIG_USING_METRIC_UNITS`, kgToUser is preferred, correct?

That’s the idea once this is merged. Never have to worry about doing the conversion, or syncing the setting across instruments then as it’s all taken care of.

Benjozork:
Will review soon

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -333,13 +336,31 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
                     console.log("TELEX MSG FETCH FAILED");
                 });
         }, NXApi.updateRate);
+
+        SimVar.SetSimVarValue("L:A32NX_GPS_PRIMARY_LOST_MSG", "Bool", 0).then();
```
Benjozork, line 340: Is this meant to be included here ?

src/.eslintrc.js
```diff
@@ -99,5 +99,7 @@ module.exports = {
         Utils: 'readonly',
         JSX: 'readonly',
         Coherent: 'readonly',
+        ViewListener: 'readonly',
+        RegisterViewListener: 'readonly',
```
Benjozork, line 103: Is this still needed after the TS changes in #5359 ?

src/instruments/src/Common/persistence.tsx
```diff
@@ -59,18 +15,20 @@ export const usePersistentProperty = (propertyName: string, defaultValue?: strin
...
+        NXDataStore.set<string>(propertyName, value);
         rawPropertySetter(value);
     };
 
     return [propertyValue, propertySetter];
 };
 
 export const usePersistentPropertyWithDefault = (propertyName: string, defaultValue: string): [string, (string) => void] => {
-    const [propertyValue, rawPropertySetter] = useState(() => NXDataStore.get(propertyName, defaultValue));
+    const [propertyValue, rawPropertySetter] = useState(() => NXDataStore.get<string>(propertyName, defaultValue));
```
Benjozork, line 26: Shouldn't the `<string>` generic also be added in `usePersistentProperty` ?

src/instruments/src/Common/persistence.tsx
```diff
@@ -59,18 +15,20 @@ export const usePersistentProperty = (propertyName: string, defaultValue?: strin
...
+
+    NXDataStore.subscribe<string>(propertyName, (key, value) => rawPropertySetter(value));
```
Benjozork, line 28: **issue:** If we are going to add add a subscription to this hook, shouldn't we also do it for `usePersistentProperty` ?

Benjozork, line 28: **issue:** This callback is never cancelled. A `useEffect` hook should be used to remove the subscription.

See: https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

src/instruments/src/EFB/react-app-env.d.ts
```diff
@@ -20,6 +20,8 @@ declare global {
          * @param args any extra arguments to be passed to the event handlers.
          */
         function trigger(name: string, ...args: any[]): void;
+
+        function on(name: string, callback: (event, ...args) => void): { clear: () => void };
```
Benjozork, line 24: Is this still needed after the TS changes in #5359 ?

Benjozork (changes requested):

PR #5349, opened 2021-07-18
feat(sound): custom cockpit sounds part 3

ImenesFBW:
## Summary of Changes
 
Added customs sound events from @hotshotp to the following items:

- [x] PFD/ND XFR switch (also added dummy button functionality, it can be pressed but reports "Inoperant" via the tooltip)
- [x] RADAR PANEL: all clickable items should now have custom sound
- [x] ISIS: all clickable items should now have custom sound
- [x] ACP: volume knobs should now be silent when turned with custom sound when pushed in or out
- [x] ANTISKID switch
- [x] GUARDED switches: all guarded switches besides APU/ENG FIRE should now be silent
- [x] FCP: guarded switches cover and fire push button 
- [x]  Changed the ECAM pages button sound events with feedback from @St54Kevin 

Also cleaned up redundant templates and lines in Airbus.xml and AirlinerCommon.xml they should now be referencing our custom sounds only.

- [x] Check compatibility with cFPM and ISI branches 

## Additional context
Sounds were advised by Boris.

Discord username: Imenes#1

## Testing instructions
Test all the listed items plus the ND range selector knob, and if you have time as many clickable items in the cockpit as you can to make sure no functionality is lost, and also to maybe help identify and of the default sounds that might still be hiding in the cockpit. 


hotshotp (approved):
Contacted over discord to review the sounds and all seem correct. LGTM!

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5349 
Tier of Testing: 1
Date: 19/07/2021

**Testing Process:**
1. Loaded C&D
2. Tested the PRs buttons
3. As you asked for this, i checked all other buttons and writing down what I think below

**Negatives:** 
- These push type button doesnt sound like the real one. I sim its a clack, while IRL its a very heavy click (hard to describe sounds with text :D), Same with the fire TEST buttons BTW, those are also clicky
image
- same with this button type. They sound weird to me, should be just a simple click
image
- The MECH button seems to make a sound in the cockpit, while it shouldnt. Its an outside horn and should be barely hearable in the cockpit, if at all

**Testing Results:** Passed

src/behavior/src/A32NX_Interior_Handling.xml
```diff
@@ -623,4 +623,22 @@
...
+            <WWISE_EVENT_2>lswitch</WWISE_EVENT_2>
+        </DefaultTemplateParameters>
+        <Component ID="#NODE_ID#" Node="#NODE_ID#">
+            <UseTemplate Name="ASOBO_GT_Switch_Code">
+                <LEFT_SINGLE_CODE>
+                    (&gt;K:ANTISKID_BRAKES_TOGGLE)
+                    (A:ANTISKID BRAKES ACTIVE, Bool) ! if{
+                        0 (&gt;L:XMLVAR_Autobrakes_Level)
+                    }
+                </LEFT_SINGLE_CODE>
```
davidwalschots, line 638: Has this code been discussed with @crocket63?

crocket63: Nope :)  XMLVAR_Autobrakes_Level is not used any more and removed from code. What is the functional need here?

davidwalschots (changes requested):
Based on the discussion above. I'll mark this as "request changes" for now to ensure nobody else starts testing this.

src/behavior/src/A32NX_Interior_Handling.xml
crocket63: If the aim of this is just to disable autobrake once button is switched off, it can be simply removed as new autobrake will process this already in wasm part by looking at ANTISKID BRAKES ACTIVE simvar

tracernz:
This will have quite some conflicts with cFPM branch and maybe also ISIS.

src/behavior/src/A32NX_Interior_Handling.xml
ImenesFBW: I just imported the template we were using for the switch, the same code that was used before but instead of calling on asobos template I moved it to interior_handeling, changed the name of the template and sound event, if it's not used I can just remove that, but it was calling for that template out of our code before as well.

ImenesFBW:
> This will have quite some conflicts with cFPM branch and maybe also ISIS.

Does cFMP use the template for ND range knob I removed?

src/behavior/src/A32NX_Interior_Handling.xml
crocket63: Mmmm maybe I'm doing something wrong but I can't find the XMLVAR anywhere doing a search in last dev.

ImenesFBW: It was not in the mod it was looking for a template in official adobo packages, look for : ASOBO_HANDLING_Switch_AntiSkid_Template

ImenesFBW: Pushed a commit that reduces that to `<LEFT_SINGLE_CODE>(&gt;K:ANTISKID_BRAKES_TOGGLE)</LEFT_SINGLE_CODE>` since that part is still needed for the switch to function.

crocket63: Just need to make sure this still makes ANTISKID BRAKES ACTIVE change its state, then it's good :)

ImenesFBW: Just checked, it does toggle ANTISKID BRAKES ACTIVE it also triggers the ecam and everything so we should be fine with only that one line of LEFT_SINGLE_CODE

ImenesFBW:
> Based on the discussion above. I'll mark this as "request changes" for now to ensure nobody else starts testing this.

The convo above should resolve this part.  With initial implementation, nothing was changed except the template that was adopted from official packages to A32NX, with new sound events. 
After the feedback above, I removed the unneeded bits of LEFT_SINGLE_CODE.  So that part should now be more streamlined than it is in the current dev. 

I am still awaiting feedback on the conflicts mentioned by @tracernz  for those I can like I mentioned above omit the template clean-up part of this PR for later.  That should minimize any conflicts with the mentioned branches.

FlightSimmer7700:
Quality Assurance Tester

Object of testing: #5349 
Tier of Testing: 1
Date: 19/7/21

Testing Process: 
Loaded in and pressed every single button and checked the sound
Negatives: 
- NAV button on the RMP has a cover that has sound when it shouldn't
- One of the ACP volume knobs doesn't make a sound when pressed
Testing Results: Not Passed

Conclusions: Sounds like all buttons are changed apart from the 2 mentioned above
Media:
https://user-images.githubusercontent.com/70103461/126229306-ab3d5ec3-aa22-470a-8d79-d3a3f8728336.mp4
https://user-images.githubusercontent.com/70103461/126229428-938ad90e-2b9b-4ace-99f2-7a9fc16f8ec8.mp4

FlightSimmer7700:
Issue 1 tested and fixed, ECAM sound events also tested and fixed. Issue 2 talked about on discord and currently not fixable

davidwalschots (approved):

Benjozork:
Do we need more testing on this?

tracernz:
> > Do we need more testing on this?
> 
> Not from my end, but @tracernz mentioned some conflicts with cFPM. But those can probably get cleared up by reverting the cleanup commits if necessary.

Yeah, it's fine. We'll take care of them. :)

PR #5359, opened 2021-07-20
feat: failures

### LINKED ISSUE #2217: [REQUEST] Failures system
No problem. 

 A app to b e downloaded with the aircraft that allows you to fail even very small systems, and the aircraft displays ECAM messages due to said system failing, like when hydraulics fail it display's a warning..

**References**
 Not possible.. 

**Additional context**
 Nope.. 

davidwalschots:
Fixes #2217

## Summary of Changes

This PR provides the foundational work for triggering failures.

Functionally, a user can now fail the following equipment through a failures page within the EFB:

- [x] Captain PFD display
- [x] F/O PFD display
- [x] TR 1
- [x] TR 2
- [x] ESS TR

It should be noted that simulation of these failures isn't fully finished as the primary goal of this PR isn't implementing these individual failures. I've added a warning text in the EFB to make this clear.

## Code review notes

Please read through the **Software Design** section below.

This PR adds Jest for JS unit testing. It also fixes some JS environment related issues.

The majority of JS for failures is put in the `@flybywiresim/failures` module. As this is an internal only module for now, I'm simply using path based module resolution for it.

### Communication

The software design below discusses two directions of communication:

- Orchestrator to failable modules
- Failable modules to orchestrator

The first is fully implemented within both JavaScript and Rust. The second is implemented on the JavaScript side, but not within Rust, as I haven't found an actual failure thus far that requires it. This means that `TransactionReader` and `TransactionWriter` are unused for now. I decided to keep them in the PR as they are fully covered by unit tests and will be useful the moment we add a failure that requires the second form of communication.

## Software Design

Let's start with some definitions we use within the software design section.

- _Equipment_ is some piece of the aircraft, whether hard- or software, which is necessary for its operation.
- A _failure_ is the abnormal functioning of a piece of equipment.
- A _module_ is any separated piece of software at runtime, e.g. `systems.wasm`, the browser running to display the PFD, etc.
- Zero to many modules that have equipment which can fail are _failable_.
- One module that orchestrates failures is the _orchestrator_.
- Each failure has a unique _failure identifier_.

### Communication protocol

Failure configuration and (de)activating failures is handled by the orchestrator. The orchestrator most likely will run in the EFB module. When the orchestrator (de)activates a failure, the failure's identifier needs to be communicated to the failable module which can handle the failure. Likewise, when the failable module (de)activates a failure it will need to communicate this information back to the orchestrator.

Sadly, the potential solutions for the communication protocol are highly limited, due to some types of modules only being able to communicate floating point numbers through SimVars.

We describe the communication protocol below.

#### Orchestrator to failable modules

We define SimVars for communication. These SimVars can be set to contain a failure's identifier. Two SimVars are used for communication from the orchestrator to the failable modules:

`A32NX_FAILURES_TO_FAILABLE_ACTIVATE`
`A32NX_FAILURES_TO_FAILABLE_DEACTIVATE`

Each failable module is aware of the failure identifiers it handles. Each failure identifier can only be handled by a single failable module. Due to this, when the orchestrator (de)activates a failure, there will be only one failable module picking it up.

The SimVars are only written by the orchestrator. The orchestrator queues their writing, ensuring no written value is overwritten before it is read by a failable module. To notify the orchestrator the value was read, the failable module which handles the failure overwrites the variable with 0.

#### Failable modules to orchestrator

Four SimVars are used for communication from the failable modules to the orchestrator, and for confirming the receival
of the message by the orchestrator:

`A32NX_FAILURES_TO_ORCHESTRATOR_ACTIVATE`
`A32NX_FAILURES_TO_ORCHESTRATOR_DEACTIVATE`
`A32NX_FAILURES_ORCHESTRATOR_ACTIVATE_RECEIVED`
`A32NX_FAILURES_ORCHESTRATOR_DEACTIVATE_RECEIVED`

The first two are written by failable modules. Failable modules are unaware of each other's writing which may cause concurrency issues. As it is likely a rare event for two failable modules to write a failure (de)activation at the same time, we apply an optimistic concurrency control solution.

After receiving a (de)activation, the orchestrator writes the failure identifier to the respective RECEIVED SimVar. These SimVars are observed by the failable modules and the one failable module that handles the given failure identifier will overwrite the value with 0 to confirm. At that point it knows the orchestrator received the failure (de)activation. In the event where the orchestrator does not confirm the receival within a certain time frame, the failable module retries writing the failure identifier to the appropriate (de)activation SimVar.

The failable module does not have to check the RECEIVED SimVars when it is not expecting any more confirmations.

Failable modules should take care to synchronise their internal activate and deactivate writing, making sure that if those events happen in rapid succession, they are written and confirmed as received in the correct order.

#### Failure identifier ranges

From a software point of view, it does not matter which numeric identifier _F<sub>i</sub>_ is used for which failure. For ease of understanding however, using ATA ranges per area is helpful (for example ELEC, HYD). For example:

- _F<sub>elec</sub> = 24000, ⋯, 24999_
- _F<sub>hyd</sub> = 29000, ⋯, 29999_

Should an area ever exceed the range allotted, another range can be assigned to it, or we can then decide to increase all ranges from 1000 to 10000.

## Additional context

## Testing instructions

Try triggering some failures through the EFB 😀.


bouveng:
This is just what we need for the circuit breakers! :shipit:

JJimmyYoung:
Just to add something useful to the list of triggers. When we are training in the simulators it's normal to use a speed reference to trigger failures so that they occur before / at / after V1 on the take off roll. Would be great if this PR allows for some failures (usually engine related!) to be triggered based on speed.

src/failures/src/communication/writers/transaction-writer.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Requires that the written variable is read by a transactional reader.
+ */
+export class TransactionWriter implements Writer {
+    private retryAfterNumberOfUpdates: number;
+
+    private valueSimVar: Writer & Updatable;
+
+    private transactionSimVar: Reader & Writer;
+
+    private openTransactions: Record<number, Transaction> = {};
```
beheh, line 15: This has hurt me in the past, because all Javascript object keys are strings internally. I would suggest either using a `Map`, or explicitly casting the keys to strings in your code to avoid future confusion about Javascript automatically casting the type:

```
> x[1] = "bar"
'bar'
> x
{ '1': 'bar' }
> Object.keys(x)
[ '1' ]
```

src/failures/src/communication/writers/transaction-writer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     *
+     * @param valueSimVar The writer to use for writing values.
+     * @param transactionSimVar The reader/writer to use for managing the transaction.
+     * @param retryAfterNumberOfUpdates The number of calls to `update` to wait before making another write attempt. Defaults to 30.
+     */
+    constructor(valueSimVar: Writer & Updatable, transactionSimVar: Reader & Writer, retryAfterNumberOfUpdates?: number) {
+        this.retryAfterNumberOfUpdates = retryAfterNumberOfUpdates || 30;
```
beheh, line 24:
```suggestion
    constructor(valueSimVar: Writer & Updatable, transactionSimVar: Reader & Writer, retryAfterNumberOfUpdates: number = 30) {
        this.retryAfterNumberOfUpdates = retryAfterNumberOfUpdates;
```

This is better because || is a falsy-check - so passing 0 would be automatically replaced with 30.

An alternative would be to explicitly check for `=== undefined` (or `typeof x ==="undefined"` if you're really paranoid).

src/.eslintrc.js
```diff
@@ -24,6 +24,20 @@ module.exports = {
...
+            env: { jest: true },
+            plugins: ['jest'],
+            rules: {
+                'jest/no-disabled-tests': 'warn',
+                'jest/no-focused-tests': 'error',
+                'jest/no-identical-title': 'error',
+                'jest/prefer-to-have-length': 'warn',
+                'jest/valid-expect': 'error',
+            },
+        },
```
beheh, line 40: So what can be better here is to defer to the list of "recommended" rules from ESLint:

```json
{
  "extends": ["plugin:jest/recommended"]
}
```

Personally I would also suggest the auto-fixable style rules to enforce consistency in our tests: https://www.npmjs.com/package/eslint-plugin-jest#rules

...or we also just additionally include those:

```json
{
  "extends": ["plugin:jest/style"]
}
```

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
```diff
@@ -0,0 +1,53 @@
...
+});
+
+test('reads a value that is in the collection', async () => {
+    await expectToBeCalledTimes(1, async (r) => {
+        await SimVar.SetSimVarValue(simVarName, 'number', inCollectionIdentifier);
+        r.update();
+    });
+});
+
+test('doesn\'t read the same value multiple times', async () => {
```
beheh, line 18: Personally I fall back from single quotes to double quotes for string containing a single quote, but that's a preference:

```suggestion
test("doesn't read the same value multiple times", async () => {
```

I believe prettier would do this as well, if we ever used it.

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
```diff
@@ -0,0 +1,53 @@
+import { QueuedSimVarReader } from '.';
+import { SimVarReaderWriter } from '..';
+
+test('does not read a value that is not in the collection', async () => {
```
beheh, line 4: In jest it's common to put tests in a file or belonging to the same testable thing in a [describe block](https://jestjs.io/docs/api#describename-fn). Personally I find it makes the console output much more readable.

So assuming these tests are for the QueuedSimVarReader:

```suggestion
describe('QueuedSimVarReader ', () => {
    test('does not read a value that is not in the collection', async () => {
```

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    await expectToBeCalledTimes(0, async (r) => {
+        await SimVar.SetSimVarValue(simVarName, 'number', notInCollectionIdentifier);
+        r.update();
+    });
```
beheh, line 8: I'm not a huge fan of being to clever in tests with helper methods, as they tend to obscure the actual call. I find explicit code in tests, even if they are repetitive, more helpful in the case of a test failure (because you're not tracing assertions in sub-functions).

beheh (changes requested):
A few comments, there's lots of personal preference in here though.

For example, I'm personally used to tests being organized in a `__tests__` folder, so that you don't have 20 files with `src` and tests alternating per directory. Especially helpful when you just want to consume code.

Additionally I'm used to a more TDD-style naming (with `.spec.ts` files), where you write tests like this:

```javascript
describe('MyQueue', () => {
  describe("when initialized with an empty queue", () => {
    it("emits the next item in the queue", () => {
      // ...
    });

    it("throws an error when the queue is empty", () => {
      // ...
    });
  });

  describe("when initialized with a prefilled queue", () => {
    it("emits the next item in the queue", () => {
      // ...
    });
  });
});
```

you get a console output looking like this:

image

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
davidwalschots: Agreed. I believe somehow our eslint rules conditioned me into thinking that was not an option, but it is also my preferred method for handling such strings. 😀

Benjozork: We can enable the `avoidEscape` option to allow this

davidwalschots: Nothing needs to be changed to allow this. It's already allowed.

Benjozork: Ah okay, I thought you were implying that eslint was complaining.

davidwalschots: What I was trying to say is that eslint complains so much that by now I even believe eslint complains when it doesn't! 😀

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
davidwalschots: I have moved the `expect` calls into the test function body.

src/failures/src/communication/readers/queued-sim-var-reader.test.ts
davidwalschots: Funny how I've [done this before](https://github.com/davidwalschots/angular-reactive-validation/blob/master/angular-reactive-validation/src/validation-messages/validation-messages.component.spec.ts) but somehow didn't in this project. 😅

davidwalschots:
Thanks for reviewing @beheh. I've applied many of your suggestions. With regards to the test file locations I've kept most of them next to the unit under test. I quite like being able to find the tests right there, instead of having to walk another directory tree. It might be something I'm used to from back in my Angular days. The integration tests moved into `__tests__` though, as they're clearly not related to a single unit.

src/instruments/src/Common/displayUnit.tsx
```diff
@@ -42,13 +43,15 @@ export const DisplayUnit: React.FC<DisplayUnitProps> = (props) => {
...
+        if (state !== DisplayUnitState.Off && props.failed) {
+            setState(DisplayUnitState.Off);
+        } else if (state === DisplayUnitState.On && (potentiometer === 0 || electricityState === 0)) {
             setState(DisplayUnitState.Standby);
             setTimer(10);
         } else if (state === DisplayUnitState.Standby && (potentiometer !== 0 && electricityState !== 0)) {
             setState(DisplayUnitState.On);
             setTimer(null);
-        } else if (state === DisplayUnitState.Off && (potentiometer !== 0 && electricityState !== 0)) {
+        } else if (state === DisplayUnitState.Off && (potentiometer !== 0 && electricityState !== 0 && !props.failed)) {
```
2hwk, line 54: I almost wonder if it's better to just declare a new state ("Failed") that is entered (and exited) upon failure and recovery and avoiding tagging on this condition for state OFF. It kind of breaks the flow of this hook which makes sense as controlling display standby/self-test/on timers, which already is a bit over-complex as is.

I think this would be a better foundation for more advanced failures involving displays, especially once we have to consider the display swapping.

src/failures/src/a320.ts
```diff
@@ -0,0 +1,9 @@
+// One can rightfully argue that this constant shouldn't be located in @flybywiresim/failures.
+// Once we create an A320 specific package, such as @flybywiresim/a320, we can move it there.
+export const A320Failure = Object.freeze({
```
2hwk, line 3: This might be pedantic but  A320Failure? Shouldn't it be prefixed as A32NX or FBW? (or just Failure)

2hwk (changes requested):

src/failures/src/a320.ts
davidwalschots: I always avoid prefixing variables with information that is already apparent from the context. The A320 in this variable name isn't intended as a prefix.

src/instruments/src/Common/displayUnit.tsx
davidwalschots: As discussed, this comment is more of a functional nature. I've added a few failures to show how the failure system works, but the actual individual failures will have to be built and/or improved in separate PRs.

2hwk (approved):

beheh (approved):
I've reviewed this PR a few times now, and there's been a bunch of improvements that make this great on the JS side now. I'm by no means a Rust expert, but the changes seem appropriate.

src/failures/src/a320.ts
```diff
@@ -0,0 +1,9 @@
...
+    TransformerRectifier1: 24000,
+    TransformerRectifier2: 24001,
+    TransformerRectifierEssential: 24002,
+    LeftPfdDisplay: 31000,
+    RightPfdDisplay: 31001,
+});
```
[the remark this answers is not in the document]
davidwalschots: I could do that, but it does increase the size of the generated code quite a bit. here's currently no need for the string, as a nicer user-friendly string is used. I tend to prefer `const enum`, but sadly you can't do that combined with babel, unless you use another plugin which I attempted but just created more problems 😀.

tracernz (approved):

ewold:

Discord: E-wold#8195
Object of testing: #5359 
Tier of Testing: 1
Date: 10/08/2021

Testing Process:
Started EFB and tested these failures. 
 Captain PFD display OK ,display went back. Failure off and screen restarted. 
 F/O PFD display OK  ,display went back. Failure off and screen restarted. 
 TR 1 OK  Failure on ELEC Page
 TR 2 OK  Failure on ELEC Page
 ESS TR OK Failure on ELEC Page

Negatives: I would if more failures come to the plaine put these in functions like displays/elec/hyd failures. It took me some time to figure out where to find the TR1/TR2/ESS TR failure. 

Testing Results: Passed

Conclusions: Failures worked from EFB / overall status page didn;t guve the failures yet .

sidnov:
QA Report

Object of testing: #5359 
Tier of Testing: 1
Date: 11/8/2021

Testing Process:
Spawn on the runway, trigger failures on the EFB, observe changes around the flightdeck. Restore functionality, depart for a circuit pattern, trigger failures again mid-air, observe changes, land back safely.

Negatives: None
Testing Results: Passed (Aircraft systems successfully failed ⚡ )

Conclusions: The existing parameters on the EFB Failures page work well when triggered, and restoring functionality occurs with ease. Can't wait to see the scope of this module expand 🚀

NUTTTTTTTT:

Discord: █▀█ █▄█ ▀█▀#2123
Object of testing: #PR 5359
Tier of Testing: 1
Date: 11/08/2021

Testing Process:
- Loaded in on runway, started EFB and tested failures independently and together, on the ground and then again in the air.
- Captain PFD - Cap PFD disabled - OK.
- F/O PFD - F/S PFD disabled - OK.
- TR1 only - all systems functional, some lights flicker.
- TR2 only - all systems functional, some lights flicker.
- ESS TR only - all systems functional.

- TR1+TR2 - DCDU, rudder trim strip, F/O side comms disabled - all other systems functional
- ESS TR + TR1 OR TR2 - ISIS, Clock, Cap side comms disabled - all other systems functional
- ESS TR + TR1 AND TR2 - FCU displays, DCDU, both comms, ISIS and Cloock disabled - Primary functions still functional

Noted:
- ISIS and DCDU perform full duration self-tests, PFD's briefly flicker with self test screen before coming back fully functional.
- When all systems disabled, aircraft still functions as expected.
- TR 1 and 2 both have different sounds when disabled, none when disabling ESS TR.

Negatives: N/A
Testing Results: Passed

Conclusions: Would obviously like to see more failures in due course, and options to randomly fail systems throughout a given flight or at random points, but as proof of concept and a first base on failures, this is great stuff, and my other conclusions are more suggestions for the future. 

Media:
image
image
image
image

davidwalschots:
@NUTTTTTTTT See [here](https://github.com/flybywiresim/rfcs/blob/1ed0dc339c3a43e95e6cd1636cf2619f007ee841/text/000-failures.md) under requirements for a listing of some of the things that are considered.

ErikNL80:
One thing that comes into mind now, I am used to turn the F/O displays off manually everytime to save some FPS. Now we have the failure framework in place, would it be possible to disable the F/O screens via the EFB (not as a failure, but to save FPS)?

NUTTTTTTTT:
> @NUTTTTTTTT See [here](https://github.com/flybywiresim/rfcs/blob/1ed0dc339c3a43e95e6cd1636cf2619f007ee841/text/000-failures.md) under requirements for a listing of some of the things that are considered.

These look brill! 
"Real-world failure rate: executes based on per component real-world or MTBF statistics.
Stress/misuse: executes when exceeding the limits of the component."

--------------------------------------------------------------


> One thing that comes into mind now, I am used to turn the F/O displays off manually everytime to save some FPS. Now we have the failure framework in place, would it be possible to disable the F/O screens via the EFB (not as a failure, but to save FPS)?

Turning off those displays doesn't really have any impact on FPS since these displays were converted.

ErikNL80:
Ok, good to know, please disregard my previous message then!

PR #5365, opened 2021-07-21
build: bump @flybywiresim/eslint-config to 0.2.2

Benjozork:
## Summary of Changes

* Waiting for #5541
* Bump `@flybywiresim/eslint-config` to 0.2.2.

## Testing Instructions

- Make sure **all** flyPad features still work
- Make sure DCDU still has same very basic functionality
- Make sure SD pages still work


Benjozork:
cc @Lucky38i

davidwalschots:
Created 19 days ago in a failed lint state. What are we doing with this PR?

Benjozork:
@davidwalschots I meant to make this a draft, I am waiting on #5541 since Erick wanted to do the refactoring part and in the mean time conform to the new ESLint rules.

bouveng (changes requested):
Hi! Is this PR supposed to contain work on DCDU, PFD, RMP, EICAS and EFB?

Benjozork:
Yes, it fixes lint errors and warnings.

Benjozork:
Converting to draft since not done with warnings yet

Benjozork:
Only few errors/warnings to go...

Benjozork:
DONE!!!!!

tracernz:
> 
> 
> DONE!!!!!

CI doesn't agree. RIP. 🥲

package.json
```diff
@@ -23,7 +23,7 @@
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
     "@babel/preset-typescript": "~7.12.7",
-    "@flybywiresim/eslint-config": "~0.1.0",
+    "@flybywiresim/eslint-config": "^0.2.3",
```
beheh, line 26: I just realized we don't have a lock file 🤔

In any case, I would suggest sticking with the `~`-style dependency, then it wil only upgrade to 0.2.4, 0.2.5... but not to 0.3.0. That seems safer.

src/instruments/src/EFB/Dashboard/Dashboard.tsx
```diff
@@ -52,7 +52,8 @@ const Dashboard: React.FC<DashboardProps> = (props) => {
                     </div>
 
                     <div className="h-3/5 ml-3 mt-3 rounded-2xl shadow-lg">
-                        <Map currentFlight={getCurrentFlightData} forceTileset="carto-dark" disableMenu hideOthers />
+                        {/* eslint-disable-next-line react/jsx-handler-names */}
```
beheh, line 55: I think either we should enforce this rule and apply it everywhere, or we should remove the rule from the config. Or is there anything special about the usage here that warrants an exception?

src/instruments/src/EFB/TODCalculator/GroundSpeed/GroundSpeedAuto/GroundSpeedAuto.tsx
```diff
@@ -16,21 +16,17 @@ const GroundSpeedAuto = ({ groundSpeedData, currentAltitude, setTodGroundSpeed,
     let [simGroundSpeed] = useSimVar('GPS GROUND SPEED', 'knots', 1_000);
     simGroundSpeed = round(simGroundSpeed);
 
-    const setCurrentGroundSpeed = () => {
+    const { groundSpeed } = last(groundSpeedData);
```
beheh, line 19: While this PR doesn't really change this, this behaviour seems dodgy for two reasons:
- If you're storing data in a React component, that is state, and should be treated as such (useState over last)
- Storing past prop values can be dodgy for the same reasons as prevTime in the clock components above. I'm confused why this is even necessary in this component.

src/instruments/src/RMP/Components/StandbyFrequency.tsx
```diff
@@ -99,22 +99,23 @@ const offsetFrequencyChannel = (spacing: ChannelSpacing, channel: number, offset
  * Hooks to outer and inner rotary encoder knobs.
  * Renders standby frequency RadioPanelDisplay sub-component.
  */
-export const StandbyFrequency = (props: Props) => {
+export const StandbyFrequency:React.FC<Props> = ({ side, value, setValue }) => {
```
beheh, line 102:
```suggestion
export const StandbyFrequency: React.FC<Props> = ({ side, value, setValue }) => {
```

beheh (changes requested):
Mostly typos, but a few . Some of the hook dependency changes could definitely uncover or change component behaviours, especially in the Clock and EFB from what I could see.

One more tip: The changes are correct in that technically if you call a setter e.g. `setMyStateVariabe` in useEffect, its a dependency. However, it's a (not very well documented) React convention that setter coming from another hook should not change and are thus usually excluded from dependency arrays (unless e.g. you pass a prop called setX, that might have to be passed in!). Usually I would expect eslint to pick that up itself and to not add variables starting with `set` coming from a hook to the dependency array. However, I'm seeing that it's inconsistentally applied in this PR - sometimes they're there, sometimes not.

src/instruments/src/EFB/Dashboard/Dashboard.tsx
Benjozork: This is because:

1. a aprop with prefix `on` has to take a function prefixed with `on|handle`;
2. a prop being passed a function prefixed with `on|handle` must be prefixed with `on`.

The mistake here was re-naming `getCurrentFlightData` to `handleGettingCurrentFlightData,` which is incorrect since it is not an event handler.

I will fix this.

package.json
Benjozork: From npm docs:

image

TL;DR: `^` does not allow minor version updates when major version is 0.

src/instruments/src/Common/hooks.tsx
```diff
@@ -57,7 +57,7 @@ export const useInteractionEvents = (events: string[], handler: (any?) => void):
         return () => {
             events.forEach((event) => getRootElement().removeEventListener(event, wrappedHandler));
         };
-    }, [
-        ...events,
+    }, [ /* eslint-disable-line react-hooks/exhaustive-deps */
+        ...events, /* eslint-disable-line react-hooks/exhaustive-deps */
     ]);
```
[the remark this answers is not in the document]
Benjozork: This is caused by ESLint not being able to verify the dependency array, since we are using a spread operator.

image

I can't really think of a good way to fix this, other than create a custom hook to memoize the dependencies that would have an ESLint exception inside it to circumvent the warning in the call site.

src/instruments/src/Common/simVars.tsx
```diff
@@ -296,8 +296,10 @@ export const useSimVar = (
         return [0, () => {}];
     }
 
+    /* eslint-disable react-hooks/rules-of-hooks */
     const value = useSimVarValue(name, unit, maxStaleness);
     const setter = useSimVarSetter(name, unit);
+    /* eslint-enable react-hooks/rules-of-hooks */
```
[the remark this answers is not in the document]
Benjozork: Removed `SIMVAR_DISABLE`. ACE does not use it.

src/instruments/src/Common/simVars.tsx
```diff
@@ -486,8 +488,10 @@ export const useSplitSimVar = (
         return [0, () => {}];
     }
 
+    /* eslint-disable react-hooks/rules-of-hooks */
     const value = useSimVarValue(readName, readUnit, maxStaleness);
     const setter = useSimVarSetter(readName, writeUnit || readUnit, writeName);
+    /* eslint-enable react-hooks/rules-of-hooks */
```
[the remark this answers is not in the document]
Benjozork: See above.

src/instruments/src/EFB/ATC/ATC.tsx
```diff
@@ -77,9 +72,13 @@ export const ATC = () => {
         loadAtc();
     }, [loadAtc]);
 
+    /* eslint-disable react-hooks/exhaustive-deps */
```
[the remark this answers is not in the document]
Benjozork: Fixed.

src/instruments/src/RMP/Components/StandbyFrequency.tsx
Benjozork: Fixed.

src/instruments/src/Common/hooks.tsx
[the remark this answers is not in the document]
beheh: To be specific I really meant the double comment. The spread operator seems acceptable to me.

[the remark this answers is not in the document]
Benjozork: Fixed

beheh (approved):
All the critical feedback has been addressed and this would seem ready for testing.

xZreai:
@Benjozork Any specific testing instructions?

Benjozork:
Added.

ShadowFrost3:
I had (haven't retested since then) some serious FPS issues with this build. (Asked in discord) Any word on whether this issue is limited to me?

Benjozork:
No, I had the issue too, but I cannot find the cause. Waiting on help from some other people.

Benjozork:
EFB infinite rendering issues, will wait until flypadosv3 because I was not able to figure them out for now.

Benjozork:
Replaced by #7344

PR #5384, opened 2021-07-23
feat(bleed): replace default bleed system

BlueberryKing:
## Summary of Changes

Implementation of the Bleed system written in Rust.

As a complete compressible fluid flow simulation is not feasibly implementable, we settle for a basic simulation assuming isentropic flow.

In code, this is implemented using basic components: containers and valves. A container contains some amount of fluid (in this case air). The container is responsible for storing thermodynamic quantities of the fluid stored in it, i.e pressure and temperature. 
A valve on the other hand is responsible for transferring fluids between containers. The valve removes some amount of air from one container and adds it to the other container. Consequently, both containers update their macroscopic quantities using equations for an isentropic process. There are more smaller components to support the rest of the system.

The implementation of the system can be roughly divided into the subtasks listed below. Note that tasks marked as complete may still require tuning.
- [x] Basic fluid simulation elements (valves, containers, etc.)
- [x] Engine related systems
  - [x] Basic engine model (i.e pressure in the compression chambers at different altitudes/speeds)
  - [x] HP / IP valve switching logic
  - [x] PRV regulation logic
  - [x] Engine starter valve
  - [x] Fan air valve
  - [x] Precooler
- [x] APU
  - [x] APU bleed valve
- [x] Overhead panel integration
- [x] Cross bleed valve
- [x] Other bleed air consumers
  - [x] A very basic implementation of the packs.
  - [x] Hydraulic reservoirs
- [x] Connect everything to proper electrical busses

Not addressed in this PR:
- Cabin pressurization
- Packs

Known issues & limitations:
- Nothing of the new bleed system is shown on the ECAM yet. All the variables are visible in the ModelBehaviours screen. I will most likely do this in a separate PR.
- There are a lot of parameters that still need to be tuned to make the pressure and temperature values shown more accurate. This is quite hard to do due to lack of references.
- While the bleed system has the logic for engine starting in place, this is not actually connected to the engines yet. This also means that crossbleed engine starts are not yet possible
- All pressure containers are initialized at 1 standard atmosphere (14.7 psi). This is ok for now since the system reaches the proper steady-state pretty quickly.
- Heat convection is not yet modelled, only heat conduction.

## References

Tons of them

## Additional context

A slightly outdated overview of the bleed system in the A320.
image

## Testing instructions

To test this PR, you need to have the ModelBehaviours window open in MSFS. That's how you can check the values of the pneumatic SimVars.

- Start cold & dark, do not turn anything on for now.
- Open the debugger, type in "PNEU" to see all the bleed system-related variables.
- Make sure all pressure values are around atmospheric pressure (14.7 psi). These do not change with the ambient pressure for now.
- Make sure all of the temperatures are at around 15 °C.
- Since the aircraft does not have any power, doing the following should not have any effect on valves.
  - Try clicking the PACK 1 and PACK 2 pushbuttons.
  - Try clicking the ENG 1 BLEED and ENG 2 BLEED pushbuttons
  - Turn the crossbleed valve control knob
- Start up APU
- Make sure APU bleed valve is closed, turn on the APU bleed. Check:
  - The APU bleed valve should now be open. (Look for 'apu' where you currently have 'pneu' in the variables window and find `A32NX_APU_BLEED_AIR_VALVE_OPEN`)
  - The Crossbleed valve should also be open
  - Check that the pressure regulating valves are closed.
  - Observe both `PRECOOLER_OUTLET_TEMPERATURE` and `PRECOOLER_INLET_PRESSURE` rising. It is expected that pressures also rise in other sections of the system. Make sure that there is no big increase in the `HP_PRESSURE` or `IP_PRESSURE`.
- Turn the crossbleed valve control knob to the SHUT position. Make sure that pressure on the number 2 system drops by a bit.
- Turn the knob back to AUTO and see how the pressure comes back.
- For both sides, turn off the PACK pushbutton. Observe:
  - The pack flow dropping to zero on that side.
- Start the engines.
  - You should see HP pressure and temperature increase with N2. 
  - IP pressure and temperature should also increase once N1 is spinning.
- Turn off the APU bleed. Observe:
  - The Pressure regulating valve opening (`A32NX_PNEU_ENG_1_PR_VALVE_OPEN` and `A32NX_PNEU_ENG_2_PR_VALVE_OPEN`)
  - Precooler outlet pressures should settle at around 46 psi.
- Do a runup. On one or both engines
  - Slowly increase thrust and observe pressures and temperatures increasing in HP and IP.
  - If HP valve was open previously, it should close at some point during the runup and the IP valve should open instead.
  - The precooler inlet and outlet pressure and temperatures should only increase slightly.
- Taxi and takeoff
  - Observe how the precooler outlet pressures stay around 46 psi.
  - Play around with the system. Make sure no pressure or temperature ever has a value of `nan`, `inf` or `-inf`.


crocket63:
Did you plan for the hydraulic reservoir supply?  I think it bypasses crossbleed valve.

BlueberryKing:
> Did you plan for the hydraulic reservoir supply? I think it bypasses crossbleed valve.

Yeah, it's on my radar. It's gonna be put in eventually, but I'm not sure how complex the interaction with the hydraulics system is, so might be addressed during the second iteration of this PR.

crocket63:
> > Did you plan for the hydraulic reservoir supply? I think it bypasses crossbleed valve.
> 
> Yeah, it's on my radar. It's gonna be put in eventually, but I'm not sure how complex the interaction with the hydraulics system is, so might be addressed during the second iteration of this PR.

I think there's nothing special to implement for hydraulics, as long as you have a pressure output in crossbleed duct I think we can just use this.  First hyd implementation can be as simple as having reservoir_air_press = max(reservoir_air_press,bleed press).

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,26 +1,1711 @@
...
+        self.regulated_pressure = sensors.regulated_pressure();
+
+        self.hpv_output = self
+            .hpv_pid
+            .next_control_output(self.transfer_pressure.get::<psi>())
+            .output;
+        self.prv_output = self
+            .prv_pid
+            .next_control_output(self.regulated_pressure.get::<psi>())
+            .output;
```
crocket63, line 466: I just noticed those are not using a time dt to compute the control, so it would probably do strange stuff when dt is changing from frame to frame. This could be enough but take care :D

BlueberryKing: Oh, this is a good point. The PID library I'm using even explicitly states this as an assumption :D I will have to look into it more. Thanks for pointing this out!

crocket63: I was thinking of using this lib to do some quick tests and noticed it requires fixed steps indeed. Could be worth implementing your own pid, or maybe I can try to share the one I use for autobrake, which is in a recursive form and doesn't require a dt value

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3055 @@
...
+        }
+    }
+
+    pub(crate) fn update<T: EngineCorrectedN1 + EngineCorrectedN2>(
+        &mut self,
+        context: &UpdateContext,
+        engines: [&T; 2],
+        overhead_panel: &mut A320PneumaticOverheadPanel,
+        engine_fire_push_buttons: &impl EngineFirePushButtons,
+        hydraulics: &A320Hydraulic,
```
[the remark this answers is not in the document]
crocket63: The newest hydraulic Traits in latest PR might come handy for this :)

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3055 @@
...
+struct PackFlowValveController {
+    pack_toggle_pb_id: VariableIdentifier,
+    pack_pb_is_auto: bool,
+    pid: PidController,
+}
+impl PackFlowValveController {
+    fn new(context: &mut InitContext, engine_number: usize) -> Self {
+        Self {
+            pack_toggle_pb_id: context
+                .get_identifier(format!("AIRCOND_PACK{}_TOGGLE", engine_number)),
```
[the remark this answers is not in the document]
BlueberryKing: I agree, but I did not make this SimVar. Since it's all over the code base, I did not feel like it would make sense to rename this in this PR.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+
+        self.green_hydraulic_reservoir_with_valve
+            .update_flow_through_valve(context, &mut left_system[0]);
+        self.blue_hydraulic_reservoir_with_valve
+            .update_flow_through_valve(context, &mut left_system[0]);
+        self.yellow_hydraulic_reservoir_with_valve
+            .update_flow_through_valve(context, &mut left_system[0]);
+
+        self.packs[0].update(context, &mut left_system[0]);
+        self.packs[1].update(context, &mut right_system[0]);
```
davidwalschots, line 256: **suggestion**: The code below is longer, but can be used to generalise to any number of packs and engines.
```suggestion
        self.packs
            .iter_mut()
            .zip(self.engine_systems.iter_mut())
            .for_each(|(pack, engine_system)| {
                pack.update(context, engine_system);
            });
```

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+
+        let (left_system, right_system) = self.engine_systems.split_at_mut(1);
+        self.apu_bleed_air_valve
+            .update_move_fluid(context, &mut self.apu, &mut left_system[0]);
+
+        self.cross_bleed_valve.update_move_fluid(
+            context,
+            &mut left_system[0],
+            &mut right_system[0],
+        );
```
davidwalschots, line 246: I have read this comment on the PR:
> While the bleed system has the logic for engine starting in place, this is not actually connected to the engines yet. This also means that crossbleed engine starts are not yet possible

**question**: I assume this is the place where I can see this isn't implemented yet. What are you thoughts on the implementation? Will it be the same lines of code, but just left and right reversed?

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+
+    fn write(&self, writer: &mut SimulatorWriter) {
+        writer.write(
+            &self.cross_bleed_valve_open_id,
+            self.cross_bleed_valve.is_open(),
+        );
+        writer.write(
+            &self.apu_bleed_air_valve_open_id,
+            self.apu_bleed_air_valve.is_open(),
+        );
```
davidwalschots, line 306: **thought**: Would it maybe make sense to move these writes into the types themselves?

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+            // I am not exactly sure if both BMCs should actually control this valve all the time.
+            self.cross_bleed_valve
+                .update_open_amount(&bleed_monitoring_computer.main_channel);
+        }
+
+        let (bleed_monitoring_computer_one, bleed_monitoring_computer_two) =
+            self.bleed_monitoring_computers.split_at_mut(1);
+
+        bleed_monitoring_computer_one[0].check_for_failure(&mut bleed_monitoring_computer_two[0]);
+        bleed_monitoring_computer_two[0].check_for_failure(&mut bleed_monitoring_computer_one[0]);
```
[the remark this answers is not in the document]
BlueberryKing: Oh lovely! I knew there had to be a better way of doing this :D I think this is not the only occurrence of the weird `split_at_mut` behemoth.

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: Ah, the comment does not make this clear, but I was talking about the fact that the bleed system does pump air into the engines at the correct time, as can be seen here:

https://github.com/BlueberryKing/a32nx/blob/0ba61c788d1fe0d64548dc620b20f36257123337/src/systems/a320_systems/src/pneumatic.rs#L920-L926

But there is no logic in the Engine/C++ code to check this pressure apart from it directly checking whether the APU bleed is on.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3055 @@
...
+        self.open_amount = Ratio::new::<ratio>(
+            2. / PI
+                * (pressure_difference.get::<psi>() * self.spring_characteristic)
+                    .atan()
+                    .max(0.),
+        );
+    }
+
+    fn update_open_amount(&mut self, controller: &impl ControllerSignal<CrossBleedValveSignal>) {
+        if let Some(signal) = controller.signal() {
```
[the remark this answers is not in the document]
BlueberryKing: I am not sure what happens in the real aircraft when the valve does not receive a signal from either BMC. I assume it will go into a pneumatic backup mode, where it is only controlled by physical forces due to the pressure differences between either side of the valve.

On a high level, when both BMCs fail in my code, this is also what will happen. But it's not because both send a `None` signal, it's because the only way for both BMCs to fail right now is the failure of the DC 2 and DC ESS SHED bus. In this case, the valve will also be unpowered and it will revert to pneumatic backup mode as soon as it realizes it's not powered, i.e when the electricity visitor comes for a visit.

I haven't given this too much thought to be honest because they are mostly all linked to failure scenarios. This is also a limitation of the PR (I will add it to the description).

I think it's a good concept that a valve can be connected to multiple controllers, where some controllers simply give a `None` signal which are ignored in favor of the other signals containing information. So I am leaning towards leaning this in, but I will have to think more concretely about the use cases.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+                self.apu_bleed_air_valve.is_open(),
+                overhead_panel,
+                engine_fire_push_buttons,
+                self.cross_bleed_valve.is_open(),
+                &self.fadec,
+            );
+
+            // I am not exactly sure if both BMCs should actually control this valve all the time.
+            self.cross_bleed_valve
+                .update_open_amount(&bleed_monitoring_computer.main_channel);
```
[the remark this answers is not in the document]
BlueberryKing: Oh, that's actually a good suggestion for the other signals as well. I'll have to look into it again, but I think this will allow me to get rid of the whole `channel_for_engine` stuff.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3041 @@
...
+            return Some(HighPressureValveSignal::new_closed());
+        }
+
+        Some(HighPressureValveSignal::new(Ratio::new::<ratio>(
+            self.high_pressure_valve_pid.output(),
+        )))
+    }
+}
+impl ControllerSignal<PressureRegulatingValveSignal> for BleedMonitoringComputerChannel {
+    fn signal(&self) -> Option<PressureRegulatingValveSignal> {
```
davidwalschots, line 608: **thought**: The same idiomatic Rust comment regarding usage of `return` applies here. 
**suggestion**: Do consider creating functions that wrap longer expressions, such as:
```rust
self.is_apu_bleed_on
            && self.is_apu_bleed_valve_open
            && (self.engine_number == 1 || self.cross_bleed_valve_is_open)
```

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3041 @@
...
+            .update(context, engine);
+        self.high_pressure_compression_chamber_controller
+            .update(context, engine);
+
+        self.fan_compression_chamber
+            .update(&self.fan_compression_chamber_controller);
+        self.intermediate_pressure_compression_chamber
+            .update(&self.intermediate_pressure_compression_chamber_controller);
+        self.high_pressure_compression_chamber
+            .update(&self.high_pressure_compression_chamber_controller);
```
davidwalschots, line 897: **thought**: This is an interesting piece of code. I'm wondering if we should design the software such that the `Engine` itself owns these pieces, but maybe these pieces are provided into the `Engine`'s constructor function by the pneumatic system? Is this a perspective you have considered?

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3041 @@
...
+        self.high_pressure_valve
+            .update_open_amount(high_pressure_valve_controller);
+        self.pressure_regulating_valve
+            .update_open_amount(pressure_regulating_valve_controller);
+        self.engine_starter_valve
+            .update_open_amount(engine_starter_valve_controller);
+        self.fan_air_valve
+            .update_open_amount(fan_air_valve_controller);
+
+        // Update valves (fluid movement)
```
davidwalschots, line 909: **thought**: The function names are good enough for this and some of the other comments in this function to not really be necessary.
```suggestion
```

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3041 @@
...
+            &self.pack_flow_valve_flow_rate_id,
+            self.pack_flow_valve.fluid_flow(),
+        );
+    }
+}
+
+// In the future, this will be done by the ACSC, hence why I have used an external controller and not the BMC
+struct PackFlowValveController {
+    pack_toggle_pb_id: VariableIdentifier,
+    pack_pb_is_auto: bool,
```
davidwalschots, line 1246: **thought**: Should this be represented by an overhead panel type?

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3041 @@
...
+/// This is a unique valve (and specific to the A320 probably) because it is controlled by two motors. One for manual control and one for automatic control
+pub struct CrossBleedValve {
+    open_amount: Ratio,
+    connector: PneumaticContainerConnector,
+    spring_characteristic: f64,
+    is_powered_for_manual_control: bool,
+    is_powered_for_automatic_control: bool,
+}
+impl CrossBleedValve {
+    pub fn new(spring_characteristic: f64) -> Self {
```
davidwalschots, line 1290: **question**: Are you expecting differences in `spring_characteristic`? Currently the only usage initialises this with `1.`

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+            / (container_two
+                .pressure()
+                .get::<pascal>()
+                .powf(1. / Self::HEAT_CAPACITY_RATIO)
+                * container_one.volume()
+                + container_one
+                    .pressure()
+                    .get::<pascal>()
+                    .powf(1. / Self::HEAT_CAPACITY_RATIO)
+                    * container_two.volume());
```
davidwalschots, line 260: **suggestion**: There's some repetition of calculations here. Maybe you can extract this into a few variables?

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+        container_one.update_temperature(
+            -coefficient
+                * temperature_gradient
+                * (1. - (-Self::HEAT_TRANSFER_SPEED * context.delta_as_secs_f64()).exp()),
+        );
+        container_two.update_temperature(
+            coefficient
+                * temperature_gradient
+                * (1. - (-Self::HEAT_TRANSFER_SPEED * context.delta_as_secs_f64()).exp()),
+        );
```
davidwalschots, line 296: Note that `x` in my suggestion is not a very good name 😉.
```suggestion
        let x = coefficient
            * temperature_gradient
            * (1. - (-Self::HEAT_TRANSFER_SPEED * context.delta_as_secs_f64()).exp());
        container_one.update_temperature(-x);
        container_two.update_temperature(x);
```

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    /// Transfer heat between two containers depending on the temperature difference
+    fn heat_conduction(
+        &self,
+        context: &UpdateContext,
+        container_one: &mut impl PneumaticContainer,
+        container_two: &mut impl PneumaticContainer,
+    ) {
+        let coefficient = 1e-1;
```
davidwalschots, line 280: **suggestion**: Move this closer to usage.

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+        .max(-from.volume());
+
+        from.change_fluid_amount(fluid_to_move);
+
+        self.fluid_flow = -fluid_to_move / context.delta_as_time();
+    }
+
+    fn calculate_required_volume_for_target_pressure(
+        &self,
+        from: &mut impl PneumaticContainer,
```
davidwalschots, line 360:
```suggestion
        from: &impl PneumaticContainer,
```

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+            true,
+            Acceleration::new::<foot_per_second_squared>(0.),
+            Acceleration::new::<foot_per_second_squared>(0.),
+            Acceleration::new::<foot_per_second_squared>(0.),
+            Angle::new::<radian>(0.),
+            Angle::new::<radian>(0.),
+        )
+    }
+
+    // It's a bit of a pain to initialize all the units manually
```
davidwalschots, line 447: 😀

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: I think it's a good idea to move the controllers' logic to the engine code. The idea will be to have the engines provide the target pressure and keep the `CompressionChamber` in the bleed system. That way, I can also get rid of some of the inappropriate `ControllerSignal` usage.

src/systems/systems/src/pneumatic/valve.rs
```diff
@@ -0,0 +1,690 @@
...
+        self.open_amount
+    }
+}
+impl PneumaticValve for PurelyPneumaticValve {
+    fn is_open(&self) -> bool {
+        self.open_amount.get::<percent>() > 0.
+    }
+}
+
+pub struct ElectroPneumaticValve {
```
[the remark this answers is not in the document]
BlueberryKing: Yeah, I've thought about this, but the solutions I've considered seem a bit like forced abstractions where modifications in the future would be harder because they're tied together too strictly.

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: It turns out there is quite a bit of work to this and I cannot really figure out a good solution, so I'll leave this as it is for now.

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: Related to an earlier comment, the issue is that the local variables `AIRCOND_PACK{}_TOGGLE` do not follow the `...PB_IS_ON` and `...PB_HAS_FAULT` convention that's required to use the `AutoOffFaultPushButton`. I don't really feel comfortable renaming them in the whole project for now.

src/systems/systems/src/shared/mod.rs
```diff
@@ -373,6 +388,38 @@ pub fn to_bool(value: f64) -> bool {
     (value - 1.).abs() < f64::EPSILON
 }
 
+pub struct InternationalStandardAtmosphere;
```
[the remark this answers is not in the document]
BlueberryKing: I like having it abstracted into this for now. I don't really like the idea of having functions floating around the module.

src/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -123,6 +123,18 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
...
+                "Bool",
+                1,
+                vec!["OVHD_PNEU_ENG_1_BLEED_PB_IS_AUTO".to_owned()],
+            )?
+            .provides_aircraft_variable_with_additional_names(
+                "BLEED AIR ENGINE",
+                "Bool",
+                2,
+                vec!["OVHD_PNEU_ENG_2_BLEED_PB_IS_AUTO".to_owned()],
+            )?
```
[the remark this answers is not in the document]
BlueberryKing: Hmm, not sure how to make this more clear. I think I might just be blind it was perfectly obvious to other people. Another option would be not to sort alphabetically but more thematically, maybe? As in, variables belonging to the same system are grouped?

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: I use a `DefaultValve` for the APU bleed air valve and I don't really want to inject a variable name into that struct, just so it can write its state. For consistency, I will also refrain from doing that for the cross bleed valve.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1,27 +1,3039 @@
...
+                        context,
+                        channel,
+                        channel,
+                        &self.engine_starter_valve_controllers[index],
+                        channel,
+                        engines[index],
+                    );
+                }
+            }
+        }
```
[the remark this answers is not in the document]
BlueberryKing: `.zip` doesn't work in my case. For both engines, I want to query both BMC's

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -1,40 +1,910 @@
...
+            .update_move_fluid(context, connected_container, &mut self.container);
+    }
+
+    pub fn container(&mut self) -> &mut T {
+        &mut self.container
+    }
+
+    pub fn pressure(&self) -> Pressure {
+        self.container.pressure()
+    }
```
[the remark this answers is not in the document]
BlueberryKing: This doesn't seem to be working for me. It makes it so the method is not found anymore in the test module. Not sure where I'm going wrong.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -1,40 +1,910 @@
...
+}
+impl TargetPressureSignal {
+    pub fn new(target_pressure: Pressure) -> Self {
+        Self { target_pressure }
+    }
+
+    pub fn target_pressure(&self) -> Pressure {
+        self.target_pressure
+    }
+}
```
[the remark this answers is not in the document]
BlueberryKing: This is used in APU code, so I don't think I can make it non-pub.

src/systems/systems/src/apu/mod.rs
```diff
@@ -126,23 +121,24 @@ impl<T: ApuGenerator, U: ApuStartMotor> AuxiliaryPowerUnit<T, U> {
         fire_overhead: &AuxiliaryPowerUnitFireOverheadPanel,
         apu_bleed_is_on: bool,
         apu_gen_is_used: bool,
+        bleed_air_valve: &mut impl ControllablePneumaticValve,
```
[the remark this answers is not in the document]
BlueberryKing: Yeah, so I've played around with this a bit, but I couldn't really get anything working well. I tried to explain this in the comment here: 

> // TODO: Returning a mutable reference here is not great. I was running into an issue with the update order:
    // - The APU turbine must know about the bleed valve being open as soon as possible to update EGT properly
    // - To open the bleed valve, we need a signal from the ecb
    // - To get a signal from the ECB to open the bleed valve, we have to update the APU.
    // For now, we just pass over control of the bleed valve to the APU, so it can be updated after the ECB update but before the turbine update.

src/systems/a320_systems_wasm/src/lib.rs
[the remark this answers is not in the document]
davidwalschots: I'd stick to alphabetical, because thematically leaves things open to interpretation, whereas alphabetical is a standardised order.

src/systems/systems/src/pneumatic/mod.rs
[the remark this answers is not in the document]
davidwalschots: I overlooked the fact that usages are in different crates!

davidwalschots (approved):

mehoygg:

Object of testing: #5384
Tier of Testing  : 1
Date             : 09/12/2021

Testing Process: 
I spawned in C/D at KLAX. I brought up the Dev tools to be able to look at the values as stated in the Testing instructions.  I did have help from BlueBerry in discord.
When I opened up the PNEU in the debug tool I looked at the values to make sure that the temps were around 15C and Pressure was around 14.7  The Values I saw, temps ranged from 15-18C and pressure values were from 14.7-14.9.
Turning off PACK1 and PACK2 did not affect the values. 
Turning off ENG 1/2 BLEED(s) did not affect values.
Turning off cross bleed valve control knob did not affect values. 
APU was then started.  APU Bleed valve was open as shown in the value (1.00).
Pressure regulating valves were closed.
No big increase noticed in HP/IP_PRESSURE.
Turning the CROSS BLEED VALVE to SHUT position NUM2 Systems did begin to drop and would come back when Knob would go back Open.
PACKS to OFF would drop to 0.00.
I started Engine 1 and the HP Press/Temp increased as the N2 started to rise. IP started to rise once N1 started to rise. 
Once both engines were on, I shut off the APU and the Press Regulating valves opened as stated.
Precooler Outlet Pressure was settling around 46PSI.
With both engines on, I started to push the throttle levers forward to about 30-40% and the Pressure/Temp for HP/IP would rise with it and if left, would stabilize and slightly lower.
I did also observe HP closing while IP Valve would open on "run up" and then IP would close and HP would open when I lowered towards idle. 
On take off, Precooler Outlet Pressure did stay pretty stable around 46PSI.  I departed 25R and climbed to about 5000FT and flew around in HDG Mode. 
I started playing with the systems, and was flying with an engine on idle for a 30-60 seconds at a time. The values for each side, 1/2, would increase and decrease respectively. 
I did NOT observe any Temp/Pressure value display "nan," "inf," or "-inf."


Negatives:       
N/A

Testing Results: 
PASSED

Conclusions: 
All worked as intended and some was seen via discord. 

Media:       
N/A

ewold:

Discord: E-wold#8195
Object of testing: #5384 
Tier of Testing: 1
Date: 12/12/2021

Testing Process: 
Start cold & dark, do not turn anything on for now. ✅ 
Open the debugger, type in "PNEU" to see all the bleed system-related variables.✅ 
Make sure all pressure values are around atmospheric pressure (14.7 psi). These do not change with the ambient pressure for now.✅ 
Make sure all of the temperatures are at around 15 °C.✅ 
Since the aircraft does not have any power, doing the following should not have any effect on valves.
Try clicking the PACK 1 and PACK 2 pushbuttons. ✅ 
Try clicking the ENG 1 BLEED and ENG 2 BLEED pushbuttons✅ 
Turn the crossbleed valve control knob ✅ 
Start up APU✅ 
Make sure APU bleed valve is closed, turn on the APU bleed. Check:  
The APU bleed valve should now be open. (Look for 'apu' where you currently have 'pneu' in the variables window and find A32NX_APU_BLEED_AIR_VALVE_OPEN) ✅ 
The Crossbleed valve should also be open  ✅ 
Check that the pressure regulating valves are closed.  ✅ 
Observe both PRECOOLER_OUTLET_TEMPERATURE and PRECOOLER_INLET_PRESSURE rising. It is expected that pressures also rise in other sections of the system. Make sure that there is no big increase in the HP_PRESSURE or IP_PRESSURE.  ✅ 
Turn the crossbleed valve control knob to the SHUT position. Make sure that pressure on the number 2 system drops by a bit.  ✅ 
Turn the knob back to AUTO and see how the pressure comes back.  ✅ 
For both sides, turn off the PACK pushbutton. Observe:
The pack flow dropping to zero on that side.  ✅ 
Start the engines.  ✅ 
You should see HP pressure and temperature increase with N2.  ✅ 
IP pressure and temperature should also increase once N1 is spinning.  ✅ 
Turn off the APU bleed. Observe:
The Pressure regulating valve opening (A32NX_PNEU_ENG_1_PR_VALVE_OPEN and A32NX_PNEU_ENG_2_PR_VALVE_OPEN)  ✅ 
Precooler outlet pressures should settle at around 46 psi.   ✅ 
Do a runup. On one or both engines
Slowly increase thrust and observe pressures and temperatures increasing in HP and IP.  ✅ 
If HP valve was open previously, it should close at some point during the runup and the IP valve should open instead.  ✅ 
The precooler inlet and outlet pressure and temperatures should only increase slightly.  ✅ 
Taxi and takeoff ✅ 
Observe how the precooler outlet pressures stay around 46 psi.✅ 
Play around with the system. Make sure no pressure or temperature ever has a value of nan, inf or -inf. ✅ 
Negatives:  Pneumatics worked as expected
Testing Results: Passed

Conclusions: No issues seen with the Behaviours


Media: 

pneumatics

PR #5421, opened 2021-07-27
fix(chrono): pause chrono when sim is paused using escape key

ManyPandas:
Fixes #[issue_no]

## Summary of Changes

When the sim is paused with the escape key, the chronometer now pauses as well. 

## Additional context

ManyPandas#0839

### DIFF
--- a/docs/a320-simvars.md
+++ b/docs/a320-simvars.md
@@ -126,6 +126,11 @@
 - A32NX_OVHD_HYD_PTU_PB_HAS_FAULT
     - Bool
     - True if PTU fault
+    
+- A32NX_GAME_PAUSED
+    - Bool
+    - True if game is paused using esc
+    - Used by chrono to determine whether to pause 
 
 - A32NX_OVHD_HYD_PTU_PB_IS_AUTO
     - Bool
--- a/docs/a320-simvars.md
+++ b/docs/a320-simvars.md
@@ -130,7 +130,6 @@
 - A32NX_GAME_PAUSED
     - Bool
     - True if game is paused using esc
-    - Used by chrono to determine whether to pause 
 
 - A32NX_OVHD_HYD_PTU_PB_IS_AUTO
     - Bool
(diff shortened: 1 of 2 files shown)


ManyPandas:
@davidwalschots This PR needs some work from you. Currently, the chronometer is reading a simvar which requires an event listener for when the game is paused with esc. This cannot be done directly inside the chrono itself. Would you mind taking a look and possibly implement a listener which updates the simvar I'm looking for?

ErickSharp:
There are different variations of 'Pause' but I am sure we will get it ironed out!

davidwalschots:
There are various forms of pausing. Currently we don't support any of them in the aircraft. I'm not planning on working on supporting pausing any time soon. If anyone else feels like looking at it they should feel free to do so.

MisterChocker:
Hopping onto the discussion after sharing my thoughts in discord.. _(just to have this documented idea somewhere related)_

I have had the idea to add a check for pauses to the `updateThrottler`, so that the systems won't be updated while paused.

ErickSharp:
> There are various forms of pausing. Currently we don't support any of them in the aircraft. I'm not planning on working on supporting pausing any time soon. If anyone else feels like looking at it they should feel free to do so. However, it should be noted that pausing is in fact a concern not just affecting the chrono, but all parts of the aircraft. As such, I do not recommend moving forward with this PR, and instead recommend solving the problem throughout the whole aircraft.

ManyPandas opened this draft PR to wait for someone to eventually implement listening for a Pause event. He himself won't actually be able to proceed in doing that himself as he just doesn't have the knowledge of how to do so yet.

davidwalschots:
If that's the case then I'd recommend closing this PR at the moment. Draft PRs aren't really intended for waiting on another PR which will come at an undetermined point in the future 😀.

ManyPandas:
Closed at the recommendation of @davidwalschots

Issue #5424, opened 2021-07-27
Option to disable Flight Display Pixelization


Croniac:
### Aircraft Version

Development

### Description

While the new pixelization looks great on screen, in VR it creates a bad moire effect that makes the screens hard to read unless zoomed in close.

Please consider an option to disable this feature as it renders the cockpit very uncomfortable to use.

### Discord Username (optional)

Croniac

SierraBravo1:
This issue is not only related to VR. I have the moiré effect also on my screen (1080p).
Please add an option to disable it.
It looks poor:
![grafik](https://user-images.githubusercontent.com/76132388/127289032-c066d71f-ca54-417e-826c-a453e0199668.png)

bouveng:
Hi! This is being looked at, in various forms.
I think the impl. will work so that the pixels are only visible when watching/zooming really close to the screens.

TLSY2:
same issue here, not only on VR!

daddyali:
![image](https://user-images.githubusercontent.com/6809626/127452556-64566b84-6245-4756-9a9b-7f41f6f502df.png)
 Ditto on 1080p - showing the 'popped out' version too which looks even worse. Definitely would appreciate a 'turn it off' option please.

TLSY2:
maybe not completely off, it isnt even realistic as it looks now.

cloudynetwork:
It looks great for me. I only notice any form of pixels when zoomed right in and certainly don't get the kind of things being posted here. What's the difference? Are you running non-native resolutions or something?

TLSY2:
> It looks great for me. I only notice any form of pixels when zoomed right in and certainly don't get the kind of things being posted here. What's the difference? Are you running non-native resolutions or something? 

If youve got a nasa pc then you dont see, but us who cant read it by not even zooming out more than deafault is a pain!

daddyali:
The work done on https://github.com/flybywiresim/a32nx/pull/5483 has massively improved this for me - in fact, it's no longer an issue at all. I now only see the pixels when I effectively have my virtual nose up against the screen in question. In normal use, pop-up use, even head forward the screens are back to being crisp with no artifacts. Great work from @2hwk

tracernz:
> It looks great for me. I only notice any form of pixels when zoomed right in and certainly don't get the kind of things being posted here. What's the difference? Are you running non-native resolutions or something?

The issue was fixed the day before.

tracernz:
[...]

PR #5483, opened 2021-07-30
feat(displays): Improve pixelated screens

2hwk:
Improves #5424

## Summary of Changes

- Pixels will only appear when the cockpit camera view is close to the screens (taking into account zoom).
- Backlight bleed is detached from pixel mask and opacity increased to previous values
- Added useGameVar to simvar hooks.

This has been tested to have a negligible effect on performance, as the sim variables do not trigger re-renders unless the camera or zoom is moved, and the update/read rate of the 2 accessed simvars is throttled back, with a css transition used to mask the intervals.  

Note that moving the camera in general (in any aircraft) as of SU5 has an issue where only the terrain in the current FOV will be loaded, which will cause frame rate to drop. This is not related to this PR. 

Exact details on implementation to follow.

## How it works

The effect for the pixels is driven by overlaying a background image over the displays, like layering a transparency over the top of the displays. The opacity value of this effect is what we will be targeting.

In order to fade the effect in and out to prevent the moire effect at a distance, we will have to track the position of the user relative to the displays, taking into account zoom levels.

Your view in MSFS is determined by a camera position, along 3 axis X, Y and Z. From the context of the cockpit, X+ is sideways to the left, Y+ is upwards and Z+ is forwards.

There is a GameVar for this camera position. It can be accessed by calling ``SimVar.GetGameVarValue('CAMERA POS IN PLANE', 'xyz')``. This produces a XYZ object, the relevant attributes we will look at in this example is ``z`` which reflects the z-coordinate of the camera position.

Now, image that there is an infinite plane (surface) perpendicular to the Z-axis, right in front of the display. Let's discard the need to account for FOV and zoom for now.

fencing

We will use this as a baseline to create a "virtual fence" or a limit to where the transition between the effect being prominently displayed and not. In front of this plane, the effect will display at maximum opacity, before this point, it will first fade before disappearing entirely.

This takes advantage of the fact that all the MFDs are in more or less the same plane vertically, As such, for the MFDs, we only have to track the position of the camera along one axis and not on the others. 

However, there is also a significant zoom in feature available in MSFS, and it has the visual illusion of appearing to move the camera forward at times. It would make sense to the user that they have physically moved closer to the displays in the virtual space.

We account for this by creating a model to move this "virtual fence" dynamically to counter the effect of the zoom in/reduced FOV. 

In order to do this, we will have to track the zoom level of the camera. There is another SimVar we can watch for this, ``SimVar.GetSimVarValue('COCKPIT CAMERA ZOOM', 'percent')``.

A measurement is taken in the virtual cockpit (using the GameVar for coordinates) to determine comfortable zoom levels at 100% and at a reduced zoom level. This creates two points which we can use to create a linear formula for at what Z-position the infinite plane/fence should be set at.

image

i.e. at 100%, target Z position will be 11.62, at 75% zoom, target Z position will be 11.7. This makes sense, at a lower zoom level, you will have to be "closer" to the displays for the fade-in to appear.

Now I have a good framework to work on that is highly performant and does not require complex calculations to work. The problem is this is still a bit simplistic, as if the effect simply "pops in" and out as we cross the virtual fence, this can be jarring for users. We should try to modulate this.

First, we can track the position of the camera position from the virtual fence, we will call this z Difference (zDelta).

To do this I have created yet another linear model for pixel image mask opacity vs this zDifference value. Aka, the closer we get to the "fence" the stronger the opacity should be. In effect we have constructed a 3 dimensional model using 2 linear models, that takes into account the distance of the displays from the user, the camera's FOV or zoom level and this "fade-in" effect which we wish to achieve vs pop-in.

For performance reasons, this driving logic only runs when the camera is moved or zoomed, as well as only every 150 ms. To help to interpolate the opacity jumps that this would otherwise create, a css transition ``transition:opacity 0.5s`` is used to help to smooth out this effect.

Note that this logic has to be done again for the MCDU specifically, as it does not lie in the same plane as the others. The only complication here is that the Y position coordinate increases as you move away from the MCDU vertically, so the formulas will be -ve signed, but otherwise similar.

Note that a more accurate model would probably require more complex functions, however as we want to minimise the performance impact of this change strictly, this has been left out deliberately. 

A similar approach may also be taken for more accurate backlight bleed.

You will note that currently, the model does not account for camera tilt or angles. However, extension is possible, by creating yet another model of Opacity vs Camera tilt to account for this and turn this into a 4-dimensional model.

## Screenshots (if necessary)

Note: Quite hard to capture this in screenshots without losing quality, you will have to see it in sim for the full effect.
image
image
FlightSimulator_Lk7IKp4cEG
FlightSimulator_hB9EbUnlzw

## Additional context

## Testing instructions

- Check that the pixel fade in and fade out effect does not break. Zoom in and out of screens repeatedly, move the camera around the cockpit. (In the SDK > Model Behaviours > Local Variables - filter by MASK_OPACITY
- Check that 


src/instruments/src/Common/displayUnit.tsx
```diff
@@ -39,6 +42,27 @@ export const DisplayUnit: React.FC<DisplayUnitProps> = (props) => {
...
+            // opacity: 4zΔ + 0.5 < 0.5
+            setOpacity(Math.min(0.5, 4 * (zDelta) + 0.5));
+            setSimVar('L:A32NX_LCD_MASK_OPACITY', opacity, 'number');
+        }, [camXyz, zoomLevel]);
+    } else {
+        const [maskOpacity] = useSimVar('L:A32NX_LCD_MASK_OPACITY', 'number');
+        useEffect(() => {
+            setOpacity(maskOpacity);
+        }, [maskOpacity]);
+    }
```
davidwalschots, line 64: I know this PR is still in draft, but I'm afraid I'll forget otherwise.

**issue**: Encapsulation breaks down.
**suggestion**: Move the things that do not directly relate to this component into their own component. This means `DisplayUnit` will only contain the code reading the SimVar and setting the opacity. Not the code calculating the opacity and writing it to a SimVar.
**suggestion**: Decide whether or not to render the newly created component within the parent, such that you don't need to check for equality to potentiometer 88 (which is only true for this aircraft/for now); but instead can check `isCaptainSide()`.
**suggestion**: Avoid using `setSimVar` in a React context. Use `useSimVar`.

ErickSharp: Hopefully I can learn from this question and I hope it doesn’t sound stupid, but what is the point of separation of logic out of displayUnit if the logic immediately ties into its functionality? Wouldn’t further abstraction just result in more clutter?

davidwalschots: Feel free to ask questions. I'll happily explain my thoughts.

## Potentiometer 88
When thinking of components, think of them as a tree. Each item in the tree is allowed to know (1) about itself, and (2) about its children. A parent can provide a child with information so that it can make decisions (as is done by passing the `potentiometerIndex`), but the child shouldn't have knowledge about the implementation of the parent.

The fact that potentiometer 88 is the captain side PFD is a fact only known to the `PFD` component. Imagine we change the logic in the `PFD`. We will then also have to change the logic here. If we forget, then things will break (and we might not notice). We can prevent this problem by not depending on knowing 88 = captain PFD at all.

## Remaining implementation
It seems we only want to execute the calculation once. Instead of using the potentiometer we could pass a boolean indicating if the calculation should take place in this particular `DisplayUnit`. However, we shouldn't, as we can't reasonably argue that a single `DisplayUnit` should have the responsibility of calculating the opacity for all display units, as we have to remember that the `DisplayUnit` should be unaware of any other `DisplayUnit`, given that those other `DisplayUnit`s are not below it in the tree.

In fact, moving this up the tree might not be good enough either, as we're still calculating this within a single instance of a PFD.

So there are two choices:
1. Calculate it in every DisplayUnit (which would be my preferred option), or
2. Calculate and write the opacity in `A32NX_Core` and then each `DisplayUnit` would merely read the SimVar.

2hwk: Yes this was just a bit of mess I left in as I was drafting. I haven't decided which approach I want to take on this yet.

davidwalschots: @2hwk 1 feels more encapsulated than 2. Though both are good options and 2 might be better in the current situation as we're having multiple ways of coding displays (React, old style, etc).

2hwk: > @2hwk 1 feels more encapsulated than 2. Though both are good options and 2 might be better in the current situation as we're having multiple ways of coding displays (React, old style, etc).

Ok, I see your point, I'll probably be moving this logic into A32NX_Core then, I would probably adopt the first stance if we still didn't have to worry about the remaining original javascript, but as it is I think I will leave that work to future contributors.

ErickSharp: I have a question, why is the logic tied to the potentiometerIndex?

2hwk: That was just a fast and dirty way of restricting the useEffect hook that controls the logic to only ran on the captain's side PFD.

Benjozork:
simvar hook changes should also be ported to https://github.com/flybywiresim/react-msfs/

2hwk:
> simvar hook changes should also be ported to https://github.com/flybywiresim/react-msfs/

https://github.com/flybywiresim/react-msfs/pull/4

davidwalschots:
@Benjozork Why do we still have a copy of `react-msfs` in this project?

davidwalschots:
@Benjozork will replace usages in another PR.

flybywire-aircraft-a320-neo/html_ui/CSS/A32NX_Display_Common.css
```diff
@@ -56,18 +57,32 @@ text {
   /* custom svg */
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 10'%3E%3Cpath d='M0 0 V10 Z H10 Z' stroke='%23000' stroke-width='3' fill='none'/%3E%3C/svg%3E");
   background-size: 2px;
-  opacity: 0.35;
+  opacity: 0;
+  transition:opacity 0.5s;
   /* subtle tint and edge bleed */
   background-color: rgba(0, 0, 255, 0.025);
   box-shadow: inset 0px 0px 30px 10px rgba(0, 75, 255, 0.04);
   border-radius: 30px;
```
bouveng, line 65: Does `#LcdOverlay` need the tint+bleed now when this is moved to its own layer?

bouveng (approved):
LGTM!

flybywire-aircraft-a320-neo/html_ui/CSS/A32NX_Display_Common.css
```diff
@@ -46,8 +46,9 @@ text {
   fill: none;
 }
 
+
 /* LCD pixel effect for screens */
-#panel:after {
+#LcdOverlay{
```
hiaaryan, line 51: `#LcdOverlay {`

flybywire-aircraft-a320-neo/html_ui/CSS/A32NX_Display_Common.css
```diff
@@ -56,18 +57,32 @@ text {
   /* custom svg */
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 10'%3E%3Cpath d='M0 0 V10 Z H10 Z' stroke='%23000' stroke-width='3' fill='none'/%3E%3C/svg%3E");
   background-size: 2px;
-  opacity: 0.35;
+  opacity: 0;
+  transition:opacity 0.5s;
```
hiaaryan, line 61: `transition: opacity 0.5s;`

hiaaryan (changes requested):

Pushok154:
pls make a switch (in flypad or in mcdu) to get an option to manage with this feature!

2hwk:
> pls make a switch (in flypad or in mcdu) to get an option to manage on/off with this feature!

Should not be necessary after this change, but we will look into a setting once this PR is merged.

2hwk:
Jumped the gun a bit, forgot to fix the ISIS backlight.

ErickSharp:
> pls make a switch (in flypad or in mcdu) to get an option to manage on/off  with this feature!

We are still waiting for a couple pre requisites before I can get started on creating a better settings page so we can add new things to it as it's currently overcrowded!

2hwk:
@hiaaryan Changes made, awaiting re-review.

hiaaryan (approved):

FinalLightNL:

**Discord:** FinalLight#2113
**Object of testing:** #5483 
**Tier of Testing:** 1
**Date:** 02/08/2021

**Testing Process:** Spawned in, turned the aircraft on and checked the displays from afar and from up close.
Moved position manually with the arrows to check it out, and also used zooming in and out.
**Negatives:** Maybe the pixel density is too much, they are a bit hard to notice unless you fully zoom in which no1 does.
**Testing Results:** Passed

**Conclusions:** The functionality with zooming in and out works perfectly, in normal sitting distance they don't appear as intended.
They are properly disabled from a certain distance and appear when moving past the "virtual fence". great work!
**Media:** 
Normal View

MCDU

Moved forward   zoomed in fully

MFD

**-it becomes noticeable when really zooming in far.**
MFD max zoom

davidnorfolk96:

Discord: !davidn123#0001
Object of testing: #5483
Tier of Testing: 1
Date: 02/08/2021

Testing Process: Spawned into runway at LOWI with all systems on & engines running. Check displays via zooming in & out & moving around the flightdeck for any unusual behaviour. Then repeated the steps as mentioned with MASK_OPACITY. Nothing to report on this & everything seemed to be working well in 1080P. There's no breaks or anything unusual regarding the zooming in & out aswell as moving around. I did however, find that turning up the brightness somewhat "drowns" out & becomes hard to see the pixals however this could be a 1080p problem. lowering the brightness on the displays does however show them up close & zoomed in. 
Negatives: As mentioned, i did find that on a higher brightness level. The pixals become harder to see zoomed in. (This could related to the resolutoion of my monitor) 

Testing Results: Passed

Conclusions: The steps that were asked showed no problems..this includes breaks or anything unusual. From a pilots view or jumpseat all looks normal. 

Media:
Screenshot_4
Screenshot_5
Screenshot_6
Screenshot_7
Screenshot_8
Screenshot_9
Screenshot_10

NUTTTTTTTT:

Discord: █▀█ █▄█ ▀█▀#2123
Object of testing: #PR 5483
Tier of Testing: 1
Date: 02/08/2021

Testing Process:
Tested in 1440p
-Spawned on runway + opened model behaviour variables for mask opacity.
-Checked each screen from various perspectives for any sign of the previous side effects, none can be observed and MASK_OPACITY values remain 0.000 until moving the camera around the cockpit close to a display.

Pixels are only shown when moving the camera closer to the displays, not possible to zoom in from the seat and view the pixels. No visible Moiré effect from any angle.


Negatives: At times you need to get really close to the displays to see anything, even with full zooom, and it can be difficult to actually get them properly into view for the intended effect.

Testing Results: Passed

Conclusions: I was a big fan of the pixels in the first place and having to get so close somewhat feels more like a complete reversion of the initial feature rather than a fix, however without any visible Moiré effect in this PR, I feel it's still a positive change. 

Media:
image
image

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_LocalVarUpdater.js
```diff
@@ -215,5 +232,41 @@ class A32NX_LocalVarUpdater {
...
+            // y-axis - away from screen >> +ve
+            const yTarget = (zoomLevel + 423.33) / 333.33;
+            const yDelta = yTarget - camXyz.y;
+            // opacity: 0 < [4zΔ + 0.5] < 1
+            SimVar.SetSimVarValue('L:A32NX_MCDU_MASK_OPACITY', 'number', Math.max(0, Math.min(1, 4 * (yDelta) + 0.5)));
+
+            this.camY = camXyz.y;
+            this.zoomLevel = zoomLevel;
+        }
+    }
```
davidwalschots, line 269: These functions should return the value that should be set to the SimVar declared on lines 93 and 99. Due to this not returning the value it tries to set null values, which gives the following console output:

image

2hwk: Resolved in new PR https://github.com/flybywiresim/a32nx/pull/5557

PR #5499, opened 2021-08-01
refactor: switch to react-msfs

Benjozork:
## Summary of Changes

* switch usage of simVar/persistence hooks to `react-msfs`


tracernz (commented):
That changes https://github.com/flybywiresim/a32nx/pull/5344 quite a bit, as react-msfs does not have the required functionality, and now it's a multi-step process to add it there, then publish a release, then finally actually make the changes to a32nx.

What about [src/shared/persistence](https://github.com/flybywiresim/a32nx/blob/master/src/shared/src/persistence.ts)? react-msfs' FsDataStore looks like an inferior impl.

Benjozork:
I can port those changes to `react-msfs` if needed.

src/instruments/src/util.js
```diff
@@ -78,14 +78,6 @@ export function setSimVar(name, value, type = SIMVAR_TYPES[name]) {
...
 
-export function useSimVar(name, type) {
-    const [value, updater] = useState(0);
-
-    useUpdate(() => updater(getSimVar(name, type)));
-
-    return value;
-}
-
 export function useGlobalVar(name, type) {
```
davidwalschots, line 81: **question**: `react-msfs` contains `useGlobalVar` now. Can this be removed?

davidwalschots (commented):
Thanks for your work.

In reply to @tracernz: in some ways `react-msfs` was extracted too soon. It would have been better to keep it as a separate project within the same repo for now and use TS paths to link `@flybywiresim/react-msfs` to that repo. (And don't forget to add a module resolver configuration, see example [here](https://github.com/flybywiresim/a32nx/pull/5359/files#diff-8054277b56da998d1c093401bfdc6dff087be6b1cb22c10e93903a185aa62304)).

I'd like to note that the README on `react-msfs` itself doesn't make any mention of the hooks. So having them in there might be questionable.

Lastly I'd like to mention that many of the SimVar hook stuff was created by theomessin in collaboration with beheh under GPLv3 license. Although the repo later did briefly change to MIT, I do hope we have asked both for permission to extract it into a library under the MIT license.

src/instruments/src/util.js
Benjozork: > 
> 
> Thanks for your work.
> 
> In reply to @tracernz: in some ways `react-msfs` was extracted too soon. It would have been better to keep it as a separate project within the same repo for now and use TS paths to link `@flybywiresim/react-msfs` to that repo. (And don't forget to add a module resolver configuration, see example [here](https://github.com/flybywiresim/a32nx/pull/5359/files#diff-8054277b56da998d1c093401bfdc6dff087be6b1cb22c10e93903a185aa62304)).
> 
> I'd like to note that the README on `react-msfs` itself doesn't make any mention of the hooks. So having them in there might be questionable.
> 
> Lastly I'd like to mention that many of the SimVar hook stuff was created by theomessin in collaboration with beheh under GPLv3 license. Although the repo later did briefly change to MIT, I do hope we have asked both for permission to extract it into a library under the MIT license.

`react-msfs` was supposed to be changed to GPLv3. Forgot that (I already have permission from the other authors of the library). Will mark as draft until that is done.

Benjozork: As for the hooks in the README, I was going to draft up one but noticed that the bit of doc there was actually supposed to be in the map component folder. I will move the existing README and write a new one mentioning the hooks.

Benjozork: > **question:** react-msfs contains useGlobalVar now. Can this be removed?

Yes. Will fix.

Benjozork:
> In reply to @tracernz: in some ways react-msfs was extracted too soon. It would have been better to keep it as a separate project within the same repo for now and use TS paths to link @flybywiresim/react-msfs to that repo. (And don't forget to add a module resolver configuration, see example here).

Thank you for the advice - seems like a better solution going forward.

Benjozork:
I have discussed with another author of the library and we have decided to remove persistent property prefixes from `FsDataStore` - that will need some changes inside the `a32nx` codebase.

Benjozork:
Waiting on #5344 to integrate new DataStore into library

Benjozork:
Will do this at a later point since there is some movement in the simvar hooks

PR #5504, opened 2021-08-01
feat(press): manual cabin pressurization

### LINKED ISSUE #5087: Landing elevation (Auto & Manual)
### Aircraft Version

Development

### Build info

```json
{
  "built": "2021-06-09T07:52:45+00:00",
  "ref": "master",
  "sha": "e3cda27aa50d6a8173363ccb7ae256888dc67842",
  "actor": "davidwalschots",
  "event_name": "manual"
}
```

### Describe the bug

- Landing elevation is not showing up correctly when in AUTO mode.

- Manual landing elevation not working properly.

### Expected behavior

- Auto landing elevation should be based on arrival airport on the MCDU.

- Manual landing elevation should become selectable once the rotary knob is pulled and turned on the overhead.

### Steps to reproduce

1. Select an airport destination with a different landing elevation (not sea level)
2. Landing elevation won't be correct.
3. Turn (should be pulled first to allow the turn) the manual landing elevation knob: nothing happens.

### References (optional)

FCOM DSC 21-20-40
Press

Selecting Manual System should change from Sys 1 to Sys 2 as well. Maybe the fix could also include that behavior.

Illuminati

mjuhe:
Fixes #5087 

## Summary of Changes

Second iteration of the cabin pressurization system. This one implements most of the fun bits, but it also changes some of the underlying logic of the first one, so it's important to ensure nothing has been broken on the automatic behaviour of the system. 

A few of the new modelled items and expected responses:
- When ditching overhead pushbutton is pressed, the outflow valve closes completely and packs turn off.
- When manual landing elev overhead knob is rotated so it's not in default position, the aircraft targets the manual elevation for landing (note this will create a pressure differential at touchdown!).
- When MODE SEL overhead pb is set to MAN, the outflow valve stops moving and responds to only manual inputs.
- MODE SEL MAN has precedence over Ditching (i.e. if the outflow valve is in manual, pressing the ditching button will not close it).
- Moving the MAN V/S CTL switch opens and closes the outflow valve and affects the vertical speed of the cabin.
- If cabin differential pressure goes above 8.6 PSI or below -1 PSI, the safety valves open
- The Residual Pressure Control Unit (RPCU) automatically depressurizes the aircraft in case of abnormal residual pressure on ground.
- New ECAM memos and warnings as per FCOM.

Items changed from PR #5159 :
- The system now commands a position of the outflow valve and then calculates the vertical speed of the cabin from there, much more realistic than before.
- Effects of packs on or off now should reflect on the pressurization schedule and the outflow valve opening ratio.
- The new LGCIU has been integrated.

Things not yet modelled:
- Effect of RAM AIR pushbutton on to the outflow valve.
- Electrical system integration.

## Screenshots (if necessary)

I will add a few screenshots and videos when I have this more advanced.

## References

Again, I've mainly used the FCOM and the AMM 21-30 - PRESSURIZATION CONTROL as reference. I've also had some help from a couple of active A320 pilots, but this doesn't mean everything in here is 100% accurate to real life. Any further reference or help is always welcome!!

Also a good video with good info: https://www.youtube.com/watch?v=iORXooxhYN8

## Additional context

Again I would like to thank @davidwalschots for all the help and the reviews on the first PR. I hope you can't find any of the same issues on this one 😄 .

Also a big thank you to @ParzivalSan for early testing, feedback and real life references!

## Testing instructions

This should hopefully be a fun one to test. In addition to ensuring all the above items behave as advertised, here are a few suggestions to try:
1. Spawn on runway, C&D or in the air.
2. Manually close the outflow valve. 
    - If packs off, pressure should remain constant in the cabin. If in the air, a tiny amount of leakage should occur (cabin altitude should increase very slowly).
    - If packs on, pressure should build up in the cabin.
3. Manually fully open the outflow valve. 
    - If on the ground, nothing should happen.
    - If in the air, there should be rapid decompression of the cabin and a red master warning.
    - Cabin altitude should rapidly increase until it is equal to ambient.
    - Maybe follow the Excessive Cabin Altitude abnormal procedure 😉 
4. Verify cabin altitude doesn't go _above_ ambient, also that as the aircraft climbs or descends, it stays same as ambient.
5. Pressurize the cabin again by closing the outflow valve.
6. Select a manual landing elevation on the overhead different to your actual landing elevation.
7. Land with MODE MAN and validate the RPCU opens the outflow valve to depressurize the cabin.
8. Throughout the test, validate nothing "jumps", all transitions are smooth and nothing unrealistic happens.


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -766,7 +770,57 @@ var A320_Neo_UpperECAM;
                                         action: "MAN ON"
                                     },
                                 ]
-                            }
+                            },
+                            {
+                                message: "EXCESS RESIDUAL PR",
```
beheh, line 775: As per reference: only one S
```suggestion
                                message: "EXCES RESIDUAL PR",
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -717,9 +717,13 @@ var A320_Neo_UpperECAM;
...
                                 level: 3,
                                 flightPhasesInhib: [1, 2, 3, 4, 5, 7, 8, 9, 10],
+                                page: "PRESS",
                                 isActive: () => (
                                     !Simplane.getIsGrounded() &&
-                                    this.getCachedSimVar("PRESSURIZATION CABIN ALTITUDE", "feet") > 10000
+                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > 9550 &&
+                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > (this.getCachedSimVar("L:A32NX_DEPARTURE_ELEVATION", "feet") + 1000) &&
+                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > (this.getCachedSimVar("L:A32NX_OVHD_PRESS_LDG_ELEV_KNOB", "feet") + 1000
+                                        || this.getCachedSimVar("L:A32NX_PRESS_AUTO_LANDING_ELEVATION", "feet") + 1000)
```
beheh, line 726: From an FWC perspective this is actually just a single signal that is received through the SDACs from the CPCUs.

As such, in order to be ready for the FWS rework, I would suggest simply moving this logic to your Rust CPC and exposing an individual boolean LVar (e.g. `A32NX_PRESS_EXCESSIVE CAB ALT`). At that point I can then just hook up the FWS to that, rather than having to then move this logic into the CPC myself.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -766,7 +770,57 @@ var A320_Neo_UpperECAM;
...
+                            {
+                                message: "LO DIFF PR",
+                                level: 2,
+                                flightPhasesInhib: [2, 3, 4, 5, 7, 8, 9, 10],
+                                page: "PRESS",
+                                isActive: () => (
+                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_DELTA_PRESSURE", "PSI") < 1.45 &&
+                                    this.getCachedSimVar("VERTICAL SPEED", "feet per minute") < -500 &&
+                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > (this.getCachedSimVar("L:A32NX_OVHD_PRESS_LDG_ELEV_KNOB", "feet") + 1500
+                                        || this.getCachedSimVar("L:A32NX_PRESS_AUTO_LANDING_ELEVATION", "feet") + 1500)
```
beheh, line 810: Similar to the other issue, I would suggest moving this to a single boolean LVar (e.g. `A32NX_PRESS_LOW_DIFF_PRESS`) so that the logic lives in Rust, and the FWS can just consume it (here for now or later through Rust from the individual CPCUs).

beheh (changes requested):
I reviewed the FWS-relevant aspects for now: One typo, and two suggestions to move logic from Javascript into Rust for more realistic CPCU outputs and a more accurate FWS in future.

mjuhe:
Thanks for the review @beheh, all changes incorporated!

beheh (changes requested):
Thanks for incorporating my suggestions! However just to be clear, I didn't mean "move all the FWC logic into the CPC", but rather make the CPCs do these internal number calculations themselves, so that the FWC can take the boolean and look at e.g. the LGCIUs and flight phase itself.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -719,11 +719,7 @@ var A320_Neo_UpperECAM;
                                 flightPhasesInhib: [1, 2, 3, 4, 5, 7, 8, 9, 10],
                                 page: "PRESS",
                                 isActive: () => (
-                                    !Simplane.getIsGrounded() &&
-                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > 9550 &&
-                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > (this.getCachedSimVar("L:A32NX_DEPARTURE_ELEVATION", "feet") + 1000) &&
-                                    this.getCachedSimVar("L:A32NX_PRESS_CABIN_ALTITUDE", "feet") > (this.getCachedSimVar("L:A32NX_OVHD_PRESS_LDG_ELEV_KNOB", "feet") + 1000
-                                        || this.getCachedSimVar("L:A32NX_PRESS_AUTO_LANDING_ELEVATION", "feet") + 1000)
+                                    this.getCachedSimVar("L:A32NX_PRESS_EXCESS_CAB_ALT", "bool")
```
beheh, line 722: Add the grounded check back here.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -772,13 +768,11 @@ var A320_Neo_UpperECAM;
...
                             {
-                                message: "EXCESS RESIDUAL PR",
+                                message: "EXCES RESIDUAL PR",
                                 level: 2,
                                 flightPhasesInhib: [1, 2, 3, 4, 5, 6, 7, 8, 9],
                                 isActive: () => (
-                                    Simplane.getIsGrounded() &&
-                                    !this.isEngineRunning(1) && !this.isEngineRunning(2) &&
-                                    Math.abs(this.getCachedSimVar("L:A32NX_PRESS_CABIN_DELTA_PRESSURE", "PSI")) > 0.01
+                                    this.getCachedSimVar("L:A32NX_PRESS_EXCESS_RESIDUAL_PR", "bool")
```
beheh, line 775: Add the grounded and engine running checks back here (but keep the SimVar).

mjuhe:
Got you now 😄 moved some of the logic back to the FWC as suggested. Happy to discuss this further offline if you want, specially for when the new FWC comes along!

src/systems/systems/src/pressurization/pressure_valve.rs
```diff
@@ -1,44 +1,99 @@
...
-                    .min(self.open_amount.get::<percent>() - self.target_open.get::<percent>()),
-            );
+    pub fn new_closed() -> Self {
+        Self {
+            open_amount: Ratio::new::<percent>(0.),
+            target_open: Ratio::new::<percent>(0.),
+            full_travel_time: Duration::from_secs(1),
+            manual_travel_time: Duration::from_secs(1),
+        }
+    }
```
beheh, line 38: I find it surprising that these constructors default to different `manual_travel_time` (60 seconds vs. 1 second).
Do you need both? Maybe this should be supplied as a parameter?

Also consider the `Default` trait if there is a "common" way of initializing the pressure valves (e.g. usually closed when starting).

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
```diff
@@ -0,0 +1,252 @@
...
+        context: &UpdateContext,
+        lgciu_gear_compressed: bool,
+        simulation_is_ground: bool,
+    ) -> Pressure {
+        if simulation_is_ground && !lgciu_gear_compressed {
+            // Formula to simulate pressure start state if starting in flight
+            let ambient_pressure: f64 = context.ambient_pressure().get::<hectopascal>();
+            Pressure::new::<hectopascal>(
+                -0.0002 * ambient_pressure.powf(2.) + 0.5463 * ambient_pressure + 658.85,
+            )
```
beheh, line 213: I'm not sure I understand this. What I'm thinking is: this code exists to set the system state when spawning in the air. However, are you sure that `simulation_is_ground` will be true in that case? My experience has usually been the opposite:  Even when spawning on ground, the simulator's own variables take a few moments to discover that it's on the ground, as the ground model loads in and the simulator initially believes it's hundreds of feet in the air.

On a grander scale (outside of the scope of this PR), this is not the first time that we're doing some weird things to deal with "spawn in situation X" cases. We should probably consider a generalized way (maybe using the SimulationVisitor) to deal with these special starts which we trigger once centrally, and then each system can decide how it wants to seed it's internal variables.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/ECAM/A320_Neo_UpperECAM.js
```diff
@@ -766,7 +766,54 @@ var A320_Neo_UpperECAM;
                                         action: "MAN ON"
                                     },
                                 ]
-                            }
+                            },
+                            {
+                                message: "EXCES RESIDUAL PR",
+                                level: 2,
```
beheh, line 772: This is actually a level 3 message with CRC:

```suggestion
                                level: 3,
```

beheh:
I left a few smaller comments. All in all this is really great progress and I'm looking forward to seeing this in the sim!

src/systems/systems/src/pressurization/pressure_valve.rs
mjuhe: The logic here was that the `new_open()` was applicable to the outflow valve, and the `new_closed()` was applicable to the safety valve. These both being pressure valves but with different physical mechanisms have different travel times. 

I've changed the names of the constructors to be clearer now. If in the future for any reason at all we decide to use this `PressureValve` type anywhere else we can rethink the constructor naming scheme.

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
mjuhe: Yes that's exactly it. This sets the cabin pressure when spawning in the air. I've just tried it a couple of times now and it does seem to work, but I agree it's not an elegant solution at all. I'm happy to change it to a more standard approach if you have any ideas on how to implement it, I'm afraid I would need some help on how to make that work 😃.

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -409,9 +450,8 @@ impl<S> PressureSchedule<S> {
         self
     }
 
-    fn new<T, U: Fn() -> S>(from: PressureSchedule<T>, ctor_fn: U) -> Self {
+    fn new<T, U: Fn() -> S>(_from: PressureSchedule<T>, ctor_fn: U) -> Self {
```
davidwalschots, line 453: **thought**: The linter told you one option for the unused variable was prefixing it with an underscore, another common option is to simply put an underscore. 
**suggestion**: I usually decide for the version with `_name` when providing an (empty) default trait function implementation, as in those situations I want implementors of said trait to use that name in their implementation. When working with a concrete function, I usually just use `_`, as there's no value in the variable having any name.

```suggestion
    fn new<T, U: Fn() -> S>(_: PressureSchedule<T>, ctor_fn: U) -> Self {
```

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -673,16 +719,24 @@ mod pressure_schedule_manager_tests {
 
     struct TestAircraft {
         cpc: CabinPressureController,
+        outflow_valve: PressureValve,
+        press_overhead: PressurizationOverheadPanel,
         engine_1: TestEngine,
         engine_2: TestEngine,
+        lgciu_gear_compressed: bool,
```
davidwalschots, line 726: **nitpick**: You could consider renaming this to `lgciu_gears_compressed`, as it's both the left and right gear that are compressed, so not a single one.

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
+            self.cpc[self.active_system - 1].should_open_outflow_valve()
+                && !press_overhead.is_in_man_mode(),
+        );
+
+        if !press_overhead.ldg_elev_is_auto() {
+            self.landing_elevation = Length::new::<foot>(press_overhead.ldg_elev_knob.value())
+        }
+        self.lgciu_gear_compressed = lgciu
+            .iter()
+            .all(|&a| a.left_and_right_gear_compressed(true));
```
davidwalschots, line 101: **thought**: We're likely to start repeating code like this. Currently there's no overarching type to query both LGCIUs. Maybe this type should be added, maybe not. I'm curious what @crocket63 and @beheh think.

I'm seeing that on the hydraulics side both LGCIUs are passed separately, as each individual LGCIU ends up in some controller. It also has code like this which basically calculates the same thing:
```rust
        let is_in_flight_gear_lever_up = !(lgciu1.left_and_right_gear_compressed(true)
            || lgciu2.left_and_right_gear_compressed(true)
            || self.is_gear_lever_down);
```

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
+        press_overhead: &PressurizationOverheadPanel,
+        engines: [&impl EngineCorrectedN1; 2],
+        lgciu: [&impl LgciuWeightOnWheels; 2],
+    ) {
+        self.cabin_pressure_simulation.update(
+            context,
+            self.outflow_valve.open_amount(),
+            self.safety_valve.open_amount(),
+            self.packs_are_on,
+            self.lgciu_gear_compressed,
```
davidwalschots, line 90: **question**: Is using `self.lgciu_gear_compressed` before modifying it below intentional? If so, what is the reason? Should that reason be added as a comment?

davidwalschots (commented):
Mostly minor comments and questions. I've not finished the full review yet, but looking good so far 😀.

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -106,6 +224,114 @@ impl Default for Pressurization {
     }
 }
 
+pub struct PressurizationOverheadPanel {
+    mode_sel: AutoManFaultPushButton,
+    man_vs_ctl_switch: SpringLoadedSwitch,
+    ldg_elev_knob: ValueKnob,
+    ditching: NormalOnPushButton,
```
davidwalschots, line 231: **note**: The same Law of Demeter violation mentioned earlier with regards to `ldg_elev_knob` also exists for other fields in this type. For example:

```rust
press_overhead.mode_sel.is_man()
```
could be changed to something like:
```rust
press_overhead.mode_is_manual()
```

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -106,6 +224,114 @@ impl Default for Pressurization {
...
+        (self.ldg_elev_knob.value() + 2000.).abs() < margin
+    }
+
+    fn is_in_man_mode(&self) -> bool {
+        !self.mode_sel.is_auto()
+    }
+
+    fn man_vs_switch_position(&self) -> usize {
+        self.man_vs_ctl_switch.toggle_position()
+    }
```
davidwalschots, line 255: **thought**: Different names to describe the same thing. `toggle_position` makes me think the position will be changed (it feels like an action to me). Should `SpringLoadedSwitch.toggle_position` maybe be renamed to `SpringLoadedSwitch.position` or `SpringLoadedSwitch.switch_position` (I like the first better as `Switch` is already part of the type name).

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -106,6 +224,114 @@ impl Default for Pressurization {
...
+
+impl ControllerSignal<PressureValveSignal> for PressurizationOverheadPanel {
+    fn signal(&self) -> Option<PressureValveSignal> {
+        if !self.is_in_man_mode() {
+            None
+        } else {
+            match self.man_vs_switch_position() {
+                0 => Some(PressureValveSignal::Open),
+                2 => Some(PressureValveSignal::Close),
+                _ => Some(PressureValveSignal::Neutral),
```
davidwalschots, line 282: **suggestion**: Of course catching any value but `0` and `2` and then returning `Some(PressureValveSignal::Neutral)` is possible. But having any other value than 1 for `PressureValveSignal::Neutral` might indicate a programming error. Maybe consider panicking?

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -106,6 +224,114 @@ impl Default for Pressurization {
...
+}
+
+impl ControllerSignal<PressureValveSignal> for ResidualPressureController {
+    fn signal(&self) -> Option<PressureValveSignal> {
+        if self.should_open_outflow_valve {
+            Some(PressureValveSignal::Open)
+        } else {
+            None
+        }
+    }
```
davidwalschots, line 332: **suggestion**: As `self.should_open_outflow_valve` is only read once in the `signal` code, consider moving the expression into the `signal` function and removing the field.

You could also consider adding a function `should_open_outflow_valve()` to make it more expressive.
```rust
    fn should_open_outflow_valve(&self) -> bool {
        self.timer > Duration::from_secs(60)
    }

    fn signal(&self) -> Option<PressureValveSignal> {
        if self.should_open_outflow_valve() {
            Some(PressureValveSignal::Open)
        } else {
            None
        }
    }
```

src/systems/systems/src/simulation/mod.rs
```diff
@@ -591,6 +591,12 @@ impl<T: Writer> Write<usize> for T {
     }
 }
 
+impl<T: Reader> Read<usize> for T {
+    fn read(&mut self, name: &str) -> usize {
+        self.read_f64(name) as usize
+    }
+}
```
davidwalschots, line 598: **note**: It is quite likely #5598 will merge into master earlier than this PR. When that happens, you can remove this code as an implementation is already provided: https://github.com/flybywiresim/a32nx/pull/5598/files#diff-59f7c495af8b955154c7a0b33b4322fd79efa182e130bebc23a8046a577af22fR602

src/systems/systems/src/pressurization/pressure_valve.rs
```diff
@@ -1,44 +1,99 @@
...
+            Some(PressureValveSignal::Close) => {
+                if self.open_amount > Ratio::new::<percent>(0.) {
+                    self.open_amount -= Ratio::new::<percent>(
+                        self.get_manual_change_for_delta(context)
+                            .min(self.open_amount.get::<percent>()),
+                    );
+                }
+            }
+            Some(PressureValveSignal::Neutral) => (),
+            _ => {
```
davidwalschots, line 63: **suggestion**: By making the `match` "complete" you ensure that whenever you add another enum value, you'll be notified of any places where that value isn't handled yet. Therefore, when dealing with a finite set of values I tend to write out all cases.
```suggestion
            None => {
```

src/systems/systems/src/pressurization/pressure_valve.rs
```diff
@@ -1,44 +1,99 @@
...
+                    self.open_amount +=
+                        Ratio::new::<percent>(self.get_valve_change_for_delta(context).min(
+                            self.target_open.get::<percent>() - self.open_amount.get::<percent>(),
+                        ));
+                } else if self.open_amount > self.target_open {
+                    self.open_amount -=
+                        Ratio::new::<percent>(self.get_valve_change_for_delta(context).min(
+                            self.open_amount.get::<percent>() - self.target_open.get::<percent>(),
+                        ));
+                }
```
davidwalschots, line 74: **question**: For my understanding. This code is only used when manual mode is on and only in manual mode does the valve not fully open or close all the time? It's interesting that in automatic mode the valve always opens fully and in manual mode it doesn't. 😀

davidwalschots (changes requested):
Okay. I finished the review. Some comments, some questions. Nothing really major. Nice work!

src/systems/systems/src/overhead/mod.rs
```diff
@@ -73,6 +73,67 @@ impl SimulationElement for OnOffFaultPushButton {
     }
 }
 
+pub struct AutoManFaultPushButton {
```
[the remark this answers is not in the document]
mjuhe: Unit tests added now. I completely overlooked this!

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -1,142 +1,116 @@
...
     length::{foot, meter},
-    pressure::{hectopascal, inch_of_mercury, pascal, psi},
+    pressure::{hectopascal, pascal, psi},
     ratio::{percent, ratio},
-    velocity::{foot_per_minute, foot_per_second, knot, meter_per_second},
+    velocity::{foot_per_minute, knot, meter_per_second},
+    volume_rate::cubic_meter_per_second,
 };
 
 #[derive(Copy, Clone)]
```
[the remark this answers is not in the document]
mjuhe: Implemented.

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -205,111 +182,142 @@ impl CabinPressureController {
...
+
+    pub fn is_excessive_residual_pressure(&self) -> bool {
+        (self.cabin_pressure - self.exterior_pressure) > Pressure::new::<psi>(0.01)
+    }
+
+    pub fn is_low_diff_pressure(&self) -> bool {
+        (self.cabin_pressure - self.exterior_pressure) < Pressure::new::<psi>(1.45)
+            && self.cabin_alt > (self.landing_elev + Length::new::<foot>(1500.))
+            && self.exterior_vertical_speed < Velocity::new::<foot_per_minute>(-500.)
+    }
```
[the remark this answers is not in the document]
mjuhe: I had the understanding that with the struct being `pub(super)` all functions within it had their visibility also limited to `pub(super)`. I did some reading and it seems like that's not entirely true, so I've now changed all function delcarations that were only used by the parent module.

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -321,6 +329,29 @@ impl PressureValveActuator for CabinPressureController {
     }
 }
 
+// Safety valve signal
+impl ControllerSignal<PressureValveSignal> for CabinPressureController {
+    fn signal(&self) -> Option<PressureValveSignal> {
+        if (self.cabin_pressure - self.exterior_pressure) > Pressure::new::<psi>(8.1) {
```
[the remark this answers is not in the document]
mjuhe: Implemented, I called it `cabin_delta_p()` for consistency, as the same function is declared in `cabin_pressure_simulation`. 😋

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -321,6 +329,29 @@ impl PressureValveActuator for CabinPressureController {
...
+            }
+        } else if self.safety_valve_open_amount > Ratio::new::<percent>(0.) {
+            Some(PressureValveSignal::Close)
+        } else {
+            Some(PressureValveSignal::Neutral)
+        }
+    }
+}
+
 #[derive(Copy, Clone)]
```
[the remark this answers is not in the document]
mjuhe: When removing it from here I get a _cannot move out of `self.pressure_schedule_manager` which is behind a mutable reference_. Any suggestions on how to get away from this without implementing `Copy`?

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
mjuhe: Changed to `_`.

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
mjuhe: Done 😄

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
```diff
@@ -0,0 +1,252 @@
...
+                    * (Self::GAMMA / (Self::GAMMA - 1.))
+                    * Self::RHO
+                    * self.cabin_pressure.get::<pascal>()
+                    * self.z_coefficient)
+                    .abs())
+                .sqrt(),
+        )
+    }
+
+    fn set_cabin_vs(&self) -> Velocity {
```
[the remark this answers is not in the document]
mjuhe: I changed this now to `calculate_cabin_vs` and then changed the function within `cabin_pressure_controller` with that name to `calculate_cabin_target_vs` for clarity.

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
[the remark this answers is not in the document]
davidwalschots: I'm going to assume this is caused by (if not, then please let me know which line generates the error):

```rust
        self.pressure_schedule_manager =
            self.pressure_schedule_manager
                .update(context, engines, lgciu_gear_compressed);
```

1. Change `self.pressure_schedule_manager` to an `Option<PressureScheduleManager>`.
2. Apply the following pattern:

```rust
        if let Some(manager) = self.pressure_schedule_manager.take() {
            self.pressure_schedule_manager = Some(
                manager.update(context, engines, lgciu_gear_compressed)
            );
        }
```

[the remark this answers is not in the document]
mjuhe: Thanks, this (and some additional tweaking) fixed it! Implemented now.

src/systems/systems/src/pressurization/mod.rs
mjuhe: It was, because it was used as a check to know if the aircraft had been initialized and then used it to set an "in the air" cabin pressure. This was far from ideal (see comment from @beheh above), so I've now changed this to add an `initialized` field to the `CabinPressureSimulation` type, and moved the update of the `self.lgciu_gear_compressed` before its use. 

Althought adding an extra field, I think this is a slightly more elegant solution and it seems to work well in the sim. Let me know your thoughts if you disagree!

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
```diff
@@ -0,0 +1,252 @@
...
+            .unwrap_or_else(|_| [context.ambient_pressure(); 20]);
+
+        let pressure_sum: f64 = self
+            .previous_exterior_pressure
+            .iter()
+            .map(|x| x.get::<hectopascal>())
+            .sum();
+
+        Pressure::new::<hectopascal>(pressure_sum / 20.)
+    }
```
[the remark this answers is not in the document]
mjuhe: Thanks for these suggestions, I've implemented all of them. I added the `average` function to the `shared` module, as I think it's quite neat and can be used by other systems as well in the future. For now I implemented it only on the `Pressure` `uom` type, but it can be easily implemented on `f64` (or any other type) if required.

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
```diff
@@ -0,0 +1,252 @@
...
+        let vertical_speed = (self.outflow_valve_open_amount.get::<ratio>()
+            * Self::OFV_SIZE
+            * self.flow_coefficient
+            * ((2.
+                * (Self::GAMMA / (Self::GAMMA - 1.))
+                * Self::RHO
+                * self.cabin_pressure.get::<pascal>()
+                * self.z_coefficient)
+                .abs())
+            .sqrt()
```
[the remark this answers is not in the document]
mjuhe: As far as I know this has no real life physical meaning by itself, but it helps avoid code repetition so I extracted this to a function now, as suggested.

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
+            self.safety_valve.open_amount(),
+            self.packs_are_on,
+            self.lgciu_gear_compressed,
+            self.cpc[self.active_system - 1].is_ground(),
+            self.cpc[self.active_system - 1].should_open_outflow_valve()
+                && !press_overhead.is_in_man_mode(),
+        );
+
+        if !press_overhead.ldg_elev_is_auto() {
+            self.landing_elevation = Length::new::<foot>(press_overhead.ldg_elev_knob.value())
```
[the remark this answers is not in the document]
mjuhe: Additional function added to the overhead type.

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
             controller.update(
                 context,
                 engines,
+                self.cabin_pressure_simulation.exterior_pressure(),
                 self.landing_elevation,
+                self.departure_elevation,
                 self.sea_level_pressure,
                 self.destination_qnh,
+                self.lgciu_gear_compressed,
+                self.cabin_pressure_simulation.cabin_pressure(),
```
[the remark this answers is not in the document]
mjuhe: Done!

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
             );
-            self.outflow_valve.update(context, controller);
+            controller.update_outflow_valve_state(&self.outflow_valve);
+            controller.update_safety_valve_state(&self.safety_valve);
```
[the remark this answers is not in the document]
mjuhe: I've eliminated these functions and added the update of the valves directly through the `update` function. The update of the safety valve is (indirectly) tested in `safety_valve_closes_when_condition_is_not_met()`, validating that the valve closes after the excess pressure differential condition disappears.

src/systems/systems/src/pressurization/mod.rs
mjuhe: Corrected, this was my mistake as I already had a function to do this: `is_in_man_mode()`

src/systems/systems/src/pressurization/mod.rs
mjuhe: Changed name to `position`.

src/systems/systems/src/pressurization/mod.rs
davidwalschots: I think adding the extra field is good! Makes the intent clearer!

src/systems/systems/src/pressurization/mod.rs
mjuhe: Added `panic!` for all cases not being 0, 1 or 2.

src/systems/systems/src/pressurization/mod.rs
mjuhe: Good spot, I moved the logic directly to the `signal` code.

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +58,96 @@ impl Pressurization {
...
+            &self.cpc[self.active_system - 1],
+            press_overhead,
+            &self.cabin_pressure_simulation,
+        );
+        if matches!(self.residual_pressure_controller.signal(), Some(_)) {
+            self.outflow_valve
+                .update(context, &self.residual_pressure_controller);
+        } else {
+            self.outflow_valve.update(context, press_overhead);
+        }
```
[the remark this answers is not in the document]
mjuhe: The idea here is that the `outflow_valve` signal is controlled by the overhead panel when in manual mode _except_ when the `ResidualPressureController` kicks in. Therefore, when `ResidualPressureController` is not active (i.e. sending a `None` signal), the signal sent to `PressureValve` can be `PressureValveSignal::Open`, `Close`, or `Neutral` when it's being controlled by the overhead panel, or `None` when it is not. 

A `None` signal triggers the CPC automatic control of the valve.

To summarise, the control of the outflow valve goes like this:
1. `ResidualPressureController`, if `None` ->
2. `PressurizationOverheadPanel`, if `None` ->
3. `CabinPressureController` through automatic operation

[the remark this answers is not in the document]
mjuhe: I've changed the line from `matches!` to `is_some()` but let me know what you think!

src/systems/systems/src/pressurization/pressure_valve.rs
mjuhe: No, it's the opposite 😄. As explained a few comments above, this part of the code only triggers when both the `ResidualPressureController` and the `PressurizationOverheadPanel` are sending a `None` signal (i.e., not active and not in manual mode, respectively).

When this happens, the `self.target_open` comes from `cpc.target_valve_position()`

src/systems/systems/src/pressurization/mod.rs
mjuhe: The way I understand the documentation, the CPCs get a discrete signal from each LGCIUs, then compare them and throw a mantenance error if they missmatch. Thinking now about the logic to set `self.lgciu_gear_compressed`, potentially it should check for `.any()` instead of `.all()`. But I don't know enough about how the LGCIUs work unfortunately, maybe @crocket63 can help? 

Rewording: what would we want to happen if both LGCIUs are sending different signals for compression? Should the aircraft assume it's on the ground, or still in the air?

src/systems/systems/src/pressurization/cabin_pressure_simulation.rs
mjuhe: As mentioned below to a different comment by @davidwalschots , I've changed this now to include an `initialized` field. This should make it clearer and more robust for now until we find a better way.

src/systems/systems/src/simulation/mod.rs
mjuhe: Removed when resolving merge conflicts 😃.

mjuhe:
Again thanks so much for the thorough review @davidwalschots ! 🙇  Lots of interesting comments and cool things to implement. Everything should've been addressed now!

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -205,123 +185,184 @@ impl CabinPressureController {
...
     }
 
-    pub fn cabin_delta_p(&self) -> Pressure {
-        self.cabin_pressure - self.exterior_pressure
+    pub(super) fn should_switch_cpc(&self) -> bool {
+        if let Some(manager) = &self.pressure_schedule_manager {
+            manager.should_switch_cpc()
+        } else {
+            false
+        }
```
davidwalschots, line 251:
```suggestion
        self.pressure_schedule_manager
            .as_ref()
            .map_or(false, |manager| manager.should_switch_cpc())
```

src/systems/systems/src/pressurization/cabin_pressure_controller.rs
```diff
@@ -205,123 +185,184 @@ impl CabinPressureController {
...
     }
 
-    pub fn should_switch_cpc(&self) -> bool {
-        self.pressure_schedule_manager.should_switch_cpc()
+    pub(super) fn should_open_outflow_valve(&self) -> bool {
+        if let Some(manager) = &self.pressure_schedule_manager {
+            manager.should_open_outflow_valve()
+        } else {
+            false
+        }
```
davidwalschots, line 259:
```suggestion
        self.pressure_schedule_manager
            .as_ref()
            .map_or(false, |manager| manager.should_open_outflow_valve())
```

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -30,26 +63,100 @@ impl Pressurization {
...
+    pub fn update(
+        &mut self,
+        context: &UpdateContext,
+        press_overhead: &PressurizationOverheadPanel,
+        engines: [&impl EngineCorrectedN1; 2],
+        lgciu: [&impl LgciuWeightOnWheels; 2],
+    ) {
+        self.lgciu_gears_compressed = lgciu
+            .iter()
+            .all(|&a| a.left_and_right_gear_compressed(true));
```
davidwalschots, line 95: **thought**: `self.lgciu_gears_compressed` is not used outside of this function, and the execution order is now such that this could be a local variable.
**suggestion**: Change to a local variable:

```suggestion
        let lgciu_gears_compressed = lgciu
            .iter()
            .all(|&a| a.left_and_right_gear_compressed(true));
```

src/systems/systems/src/pressurization/mod.rs
```diff
@@ -106,6 +233,118 @@ impl Default for Pressurization {
...
+impl ControllerSignal<PressureValveSignal> for PressurizationOverheadPanel {
+    fn signal(&self) -> Option<PressureValveSignal> {
+        if !self.is_in_man_mode() {
+            None
+        } else {
+            match self.man_vs_switch_position() {
+                0 => Some(PressureValveSignal::Open),
+                1 => Some(PressureValveSignal::Neutral),
+                2 => Some(PressureValveSignal::Close),
+                _ => panic!("Something went wrong in PressurizationOverheadPanel"),
```
davidwalschots, line 296: **suggestion**: Panic more descriptively.

```suggestion
                _ => panic!("Could not convert manual vertical speed switch position '{}' to pressure valve signal.", self.man_vs_switch_position()),
```

src/systems/systems/src/pressurization/pressure_valve.rs
```diff
@@ -101,21 +156,38 @@ mod pressure_valve_tests {
...
             self.target_valve_position = amount;
         }
     }
-    impl PressureValveActuator for TestValveActuator {
-        fn target_valve_position(&self) -> Ratio {
+    impl OutflowValveActuator for TestValveActuator {
+        fn target_valve_position(
+            &self,
+            _press_overhead: &PressurizationOverheadPanel,
+            _cabin_pressure_simulation: &CabinPressureSimulation,
```
davidwalschots, line 175:
```suggestion
            _: &PressurizationOverheadPanel,
            _: &CabinPressureSimulation,
```

davidwalschots (commented):
You're very welcome @mjuhe. I very much enjoy reviewing your code and helping out! I noted some minor last things. 

I would already approve the PR right now, but I see that there's still an open question to @crocket63 [here](https://github.com/flybywiresim/a32nx/pull/5504#discussion_r720810735). Do you want to wait for that answer before testing commences?

src/systems/systems/src/pressurization/mod.rs
mjuhe: True! Good spot.

mjuhe:
> 
> 
> You're very welcome @mjuhe. I very much enjoy reviewing your code and helping out! I noted some minor last things.
> 
> I would already approve the PR right now, but I see that there's still an open question to @crocket63 [here](https://github.com/flybywiresim/a32nx/pull/5504#discussion_r720810735). Do you want to wait for that answer before testing commences?

Thanks again 😄 and regarding the open question, I'm happy to move forward as is for now, we can always go back and change the interaction with the LGCIU at a later date if we feel it's needed.

davidwalschots (approved):
🚀

beheh (approved):

ShadowFrost3:


Object of testing: #5504
Tier of Testing:1
Date: 06/10/2021

Testing Process: 
Runway Spawn
Level off at 15000, packs off, -200 ft/min leakage 
@15000 packs back on leakage stops
open outflow valve - oxygen warning and increase of cabin altitude at a rate highest tested rate of 13,500 per minute.
entering climb with outflow valve open - cabin increases with climb 
entering climb with outflow valve open & time accel - calculation is broken increases with in real world time not aircraft  (if climbing at 3,000 minute and time accel climb rate will be 9000 a minute) * additional note, I had a descent under time accel that was reading correctly and I hit external view and then went back into cockpit and it went to the multiplied reading 
reduce altitude to 8,000 master caution disappears - checked 
cabin pressure behaves as expected in climb and descent - checked 
at 30,000ft aircraft pressurizes at ~2000 ft/min - checked 
Pack 1 off (2 on) cabin does not pressurize - fail?
Pack 2 off (1 on) cabin does pressurize - check 
After a period of flying with pack 1 off I was unable to get the aircraft to pressurize with both packs on and under the emergency descent request - fail? - from memory if needed to reproduce, pack 1 off, back on, 2 off back on, 1 off, descent to 8,000 when at 13,000 set climb to 15,000. At that point in time error was noticed 
----testing stopped -----



Negatives: 
Time accel pressure increase/decrease calculation is incorrect 
Pack 1 required to pressurize, Pack 2 irrelevant? - looked at manuals, I'm assuming they are redundant but you can tell me if incorrect
Broke the pressurization system some how... not sure 

Testing Results: Not Passed

Conclusions: 
Is there a max rate of change through the outflow valve? 
Hypoxia would be nice :) 
Media: 
image

Illuminati270:
> @Illuminati270 you are quite right. Unfortunately until the new model is ready I don't think we will be able to add a pull functionality to this knob. Regarding #5087, as per the issue expected behaviour:
> 
> * Auto landing elevation should be based on arrival airport on the MCDU. ✅
> * Manual landing elevation should become selectable once the rotary knob is pulled and turned on the overhead. ✅
> 
> Both of these issues are now fixed (minus the pull functionality as mentioned above). I have no problem with leaving the issue open of course, but potentially it needs to be reworded to capture what is now outstanding. Otherwise we can close this one and open a new issue.
> 
> PS: There is no airport in the world lower than -1266ft [https://en.wikipedia.org/wiki/List_of_lowest_airports], so in any case we shouldn't run into any issues with the knob as it is now 😄.

I've posted a reference regarding the scale (-2,0000ft ~ 14,000ft), but my expression of "should be pulled first to allow the turn" might have been inaccurate. There was a picture regarding the correct knob representation though.

I understand the knob limitations, so...Way to go! Well done!

PR #5591, opened 2021-08-08
fix(fcu): Workaround for selected HDG

### LINKED ISSUE #5479: Unable to engage HDG on AP
### Aircraft Version

Development

### Build info

```json
{
  "built": "2021-07-30T02:31:34+00:00",
  "ref": "master",
  "sha": "bab6a451346ee9fb2a28abe20bf4ec454933e371",
  "actor": "Benjozork",
  "event_name": "manual"
}
```

### Describe the bug

Unable to activate HDG mode inflight

### Expected behavior

Able to activate HDG mode inflight

### Steps to reproduce

1. Fly
2. Set direct
3. Press HDG knob

### References (optional)

_No response_

_No response_

_No response_

2hwk:
Fixes #5479 

## Summary of Changes

SU5 seems to have introduced a behaviour where if a K:Event is called before the flybywire WASM has been compiled, it will cause an exception and this event will be unabled to be triggered through javascript. K:Events will now not be sent until after the flight has been fully loaded.

## Screenshots (if necessary)

Old, bugged behaviour:

https://user-images.githubusercontent.com/15316958/128606682-631930f2-e508-40f6-9f70-bf90028e9432.mp4

New:

Works as expected

## Additional context

## Testing instructions

1. Run the deletenxwork.bat (Found in the pins in the #support channel on discord)
2. Start a new flight. This should take longer than a normal load in as it has to do some compilation first.
3. Verify that selected heading now works as expected.

### DIFF
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
@@ -486,7 +486,10 @@ class FMCMainDisplay extends BaseAirliners {
         SimVar.SetSimVarValue("L:A32NX_AP_CSTN_ALT", "feet", this.constraintAlt);
         SimVar.SetSimVarValue("L:A32NX_TO_CONFIG_NORMAL", "Bool", 0);
         SimVar.SetSimVarValue("L:A32NX_CABIN_READY", "Bool", 0);
+    }
 
+    connectedCallback() {
+        super.connectedCallback();
         SimVar.SetSimVarValue("K:A32NX.ATHR_RESET_DISABLE", "number", 1);
     }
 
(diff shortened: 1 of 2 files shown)


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -486,7 +486,10 @@ class FMCMainDisplay extends BaseAirliners {
         SimVar.SetSimVarValue("L:A32NX_AP_CSTN_ALT", "feet", this.constraintAlt);
         SimVar.SetSimVarValue("L:A32NX_TO_CONFIG_NORMAL", "Bool", 0);
         SimVar.SetSimVarValue("L:A32NX_CABIN_READY", "Bool", 0);
+    }
 
+    connectedCallback() {
+        super.connectedCallback();
```
MisterChocker, line 492: this basically removes `SimVar.SetSimVarValue("K:A32NX.ATHR_RESET_DISABLE", "number", 1);` from `initVariables()`.

Why is that?

MisterChocker (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
2hwk: image

This is one of 2 K:Events that no longer works after SU5. The issue happens when they are called before fbw.wasm is compiled, so I simply attached it to connectedCallback() instead which is only called when the element is attached to the DOM.

The event is still initialised, it just occurs late enough that it does not trigger the issue.

2hwk:
Andreas has an actual rework/fix and not a workaround that should put this issue to bed, keeping this up until that is drafted and then we can fall back to this if needed.

PR #5242, opened 2021-06-30
feat: air data inertial reference system

davidwalschots:
Partially addresses #3335.

## Summary of Changes
This PR provides an advanced ADIRS implementation. Currently the ADIRS' ADIRUs aren't separated: once a single ADIRU is aligned the PFD and ND on the left and right side will both fully work. In this PR, I modify the implementation such that each ADIRU is separated and has more functionality.

Note that certain features are out of scope for this PR. For example, realistic alignment, radio positioning mode, etc. will all be future work for another PR.

## Bug fixes

- [x] The longest instead of the shortest remaining alignment time is displayed on the upper ECAM.

## Features

- [x] ATT mode only supplies attitude and heading information<sup>1</sup>.

<sup>1</sup> Note that ATT mode isn't fully implemented yet. In the fully implemented version, ATT mode requires entering the HDG through the MCDU and frequently updating it.

### Overhead panel

- [x] IR1, IR3, or IR2 can be pushed OFF to disconnect output of IR information from that particular ADIRU.
- [x] ADR1, ADR3, or ADR2 can be pushed OFF to disconnect output of ADR information from that particular ADIRU.

### PFD

- [x] The selected ADR provides altitude (ALT), computer air speed (CAS), mach number (M), and barometric vertical speed (BARO V/S) information.
- [x] The selected IR provides attitude (ATT) and heading (HDG) information.
- [x] When the selected ADR does not provide CAS, the speed scale is replaced by a red SPD flag.
- [x] When the selected ADR does not provide M, the mach number indication is replaced by a red MACH flag<sup>1</sup>.
- [x] When the selected ADR does not provide ALT, the altitude indication is replaced by a red ALT flag.
- [x] When the selected IR does not provide ATT, the attitude indication is replaced by a red ATT flag.
- [x] When the selected IR does not provide V/S, ADR's BARO V/S is used.
- [x] When the selected IR does not provide HDG, the heading indication is replaced by a red HDG flag.

<sup>1</sup> The logic for when to show the MACH flag or the MACH number needs to be improved further as part of future work. For now the MACH flag will always be indicated when unavailable.

#### Heading

- [x] When the selected IR is available, the MAP or other selected mode is displayed.
- [x] When the selected IR is unavailable, a red HDG MAP NOT AVAIL flag is displayed.

#### True Airspeed

- [x] When the selected ADR is unavailable, ` ` is displayed.
- [x] When the selected ADR is available, and CAS < 60 knots, `---` is displayed.
[...]

PR #4998, opened 2021-06-04
refactor: replace f64::MIN with values closer to 0

davidwalschots:
## Summary of Changes
As part of #4912 I introduced `f64::MIN` as a value indicating "no value". As `f64::MIN` is a very large negative number not all other applications (such as yourcontrols) handle it well. I've now replaced `f64::MIN` with a new mechanism for reading and writing, using Rust traits.

The file in which I originally started writing `f64::MIN` is found in `src/systems/systems/src/apu/electronic_control_box.rs`. `src/systems/systems/src/simulation/mod.rs` contains the new traits and implementations.

## Additional context

Discord username (if different from GitHub): SjotgunSjonnie#9623

## Testing instructions
A basic test of ELEC, HYD, and APU functionality will suffice. The change affects the reading and writing of variables to the simulator. Thus look for any values being unexpected on ECAM pages, etc.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #4912, opened 2021-05-26
feat(apu): ecb powered by dc bat bus

davidwalschots:
## Summary of Changes
The APU's starter motor is already powered by the DC BAT BUS since #3625. However, the APU's Electronic Control Box and Air Intake Flap weren't yet covered by that PR. In this PR I'll add correct power/unpowered behaviour to those elements.

- Air intake flap:
    - [x] Doesn't move when DC BAT BUS is unpowered and the APU isn't running.
    - [x] Consumes 20W when moving.
- Electronic Control Box:
    - [x] Is powered through DC BAT BUS until N > 70%, at which point it's powered by the APU itself.
    - [x] Consumes 105W when on.
- [x] The APU start is aborted when DC BAT BUS is unpowered during APU start with 55% <= N <= 70%.
    - [x] When DC BAT BUS is powered again, no APU MASTER SW FAULT light is indicated, as the FAULT light is triggered by the ECB. As the ECB was unpowered, there was no way for it to trigger the FAULT.

### Code review

I've continued moving dependencies between multiple modules into `shared`. Mostly traits. One of these days I shall make a separate PR to finish that and add further documentation describing the module structure/design 😉.

### Example testing scenarios

1. BATs ONLY, start APU, abort start at < 55% N by turning OFF one BAT. Result: Start aborted, MASTER SW ON/FAULT lights extinguish as lights are no longer powered; press BAT ON again to see ON and FAULT light (ECB was still powered by other BAT).
1. BATs ONLY, start APU, abort start at < 55% N by turning OFF both BATs. Result: Start aborted, MASTER SW ON/FAULT lights extinguish as lights are no longer powered; press both BAT ON again to see ON light illuminated and FAULT not illuminated (ECB became unpowered and therefore didn't remember FAULT).
1. BATs ONLY, turn one BAT OFF at N between 55% and 70%. Result: Start continues.
1. BATs ONLY, turn both BATs OFF at N between 55% and 70%. Result: Start aborts, due to DC BAT BUS becoming unpowered (as scenario 2).
1. BATs ONLY, turn both BATs OFF at N between 70% and APU avail (basically before 95%). Result: Start continues, as APU itself powers ECB at N >= 70%.
1. EXT PWR ON, turn a BAT (or both) OFF at N between 55% and 70%. Result: Start continues.

**Please think of more scenarios** 😀

## References
@komp1821 

## Remaining work for another PR
[...]

PR #5598, opened 2021-08-09
feat: arinc 429 communication

davidwalschots:
## Summary of Changes
Over time various forms of communicating the absence of data have popped up. For example:

- In #4912 and #4998 I used zero degrees Kelvin to indicate the absence of data coming from the ECB to present on the APU ECAM page.
- All the ADIRS functionality added in #5242 uses `-1000000` to indicate the lack of a value.

There are various problems with this approach:

- There isn't any uniformity in what value indicates the absence of data, as a result lots of custom code to interpret the data is created.
- It doesn't support scenarios other than lack of data, e.g. there's no way to indicate the data is part of a functional test.

## ARINC 429

The aircraft itself uses an [ARINC 429 protocol](https://en.wikipedia.org/wiki/ARINC_429) bus system for large parts of its communication. The ARINC 429 protocol defines 32-bit words with the following information:

- **Label, 8 bits**: identifies the data type, for example "computed air speed".
- **Source/Destination Identifiers (SDI), 2 bits**: indicates the intended receiver or transmitting subsystem (it is usually the latter on the aircraft).
- **Data, 19 bits**: the actual value encoded as [bit-field](https://en.wikipedia.org/wiki/Bit_field), [BCD](https://en.wikipedia.org/wiki/Binary-coded_decimal) or [BNR](https://en.wikipedia.org/wiki/Two%27s_complement).
- **Sign/Status Matrix (SSM), 2 bits**: indicates the sign (positive or negative number) and/or status of a value. Refer to the table on the [wikipedia page](https://en.wikipedia.org/wiki/ARINC_429) for a full overview. Statuses that may be encoded are:
  - **Normal Operations**: the data is considered to be correct.
  - **Functional Test**: the data is provided by a test source.
  - **Failure Warning**: a failure causes the data to be suspect or missing.
  - **No Computed Data**: indicates data is missing/inaccurate for reasons other than a failure.
- **Parity, 1 bit**: bit to verify the data is not invalidated during transmission.

## Translating ARINC 429 to SimVar

We are limited to communication via SimVars. The SimVar's name can be considered equivalent to the ARINC 429 label and SDI, in that it can without restrictions indicate what value the SimVar contains, and additionally can indicate what is the source of the information (e.g. `"ADIRS_1_...", "ADIRS_2_..."). The parity bit isn't of concern, as we'll not simulate bits flipping up to that level.

This leaves the data and SSM. The only usable SimVar format is [double-precision floating-point number](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). These 64 bits give us precision far beyond what the aircraft itself can deal with, and we can thus capture both pieces of information in these bits easily, though it should be noted only 52 bits can reliably be used in JavaScript.

a diagram depicting the bits in a double-precision floating-point
<p align="right">Source: Codekaizen, Wikipedia</p>

To keep things simple, we will split the 64 bits in two: 32 bits will be used to represent the `Data` as a [single-precision floating-point](https://en.wikipedia.org/wiki/Single-precision_floating-point_format), and 32 bits will be used to hold the 2 bit `SSM`. For sake of simplicity, the sign will be contained in the 32-bit floating point, even for BCD data where this would traditionally be contained in the SSM.

One could argue that another method is possible: removing two of the fraction bits from the double-precision floating-point and using those two bits to contain the `SSM` data. It should however be noted that the JavaScript's bit manipulation API is somewhat limited and constrained, and that exponent bits cannot be used reliably for data. As a 32-bit floating point already captures data with a higher precision than in the real aircraft, I think it's a reasonable design.

## Additional context

## Testing instructions
Changes affect the PFD, ND, lower ECAM status area (SAT, TAT, etc), lower ECAM APU page, and lower ECAM FUEL page (specifically the APU fuel line and triangle indication). Thus I'd focus testing efforts on those areas.


flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/arinc429.js
```diff
@@ -0,0 +1,39 @@
...
+
+        this._ssm = Arinc429Word.u32View[0];
+        this._value = Arinc429Word.f32View[1];
+    }
+
+    static empty() {
+        return new Arinc429Word(0);
+    }
+
+    static fromSimVarValue(name) {
```
Benjozork, line 13: Are we sure that creating a new instance of this class every update is a good idea ? I don't know if this design is final or not, considering this is a draft, but it seems to me that creating a bunch of objects we immediately stop using and re-create might put a lot of pressure on GC.

Perhaps an instance should be able to be "recomputed" when new data is available, without instantiating this class again.

davidwalschots: For now I'm sure this is a good idea, because it leads to the simplest programming interface which nearly directly replaces `SimVar.GetSimVarValue()`. Of course this isn't yet programmed for React, and there we might be able to do something smarter in a hook combined with `useSimVar`.

cloudynetwork:
I think this approach is the only way forward to build the strong foundation for high quality failure modelling. Great to see.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/arinc429.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+Arinc429Word.f64View = new Float64Array(1);
+Arinc429Word.u32View = new Uint32Array(Arinc429Word.f64View.buffer);
+Arinc429Word.f32View = new Float32Array(Arinc429Word.f64View.buffer);
+
+Arinc429Word.SignStatusMatrix = Object.freeze({
+    FailureWarning: 0,
+    FunctionalTest: 1,
+    NoComputedData: 2,
```
beheh, line 41: One element that surprised me here is the order, encoding Functional Test as 10 instead of 01. ARINC has an interesting quirk where it's usually specified back to front (last bit first) unlike other serial formats, so the matrix I'm more familiar with looks like this:

| Bit 31 | Bit 30 | Meaning         |
|--------|--------|-----------------|
| 0      | 0      | Normal          |
| 1      | 0      | NCD             |
| 0      | 1      | Functional Test |
| 1      | 1      | Failure Warning |

This order also corresponds more closely to the status priority. If multiple statuses apply, the sending system should choose FW first, then FT, then NCD, and then finally Normal only if nothing else applies.

While I don't have a concrete suggestion here, I know little about how WebAssembly chooses it's endianness. All in all, maybe this quirky divergence from the ARINC spec warrants an extra comment, that it's due to a flipped bit order compared to the docs.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/arinc429.js
```diff
@@ -0,0 +1,43 @@
...
+        return new Arinc429Word(0);
+    }
+
+    static fromSimVarValue(name) {
+        return new Arinc429Word(SimVar.GetSimVarValue(name, "number"));
+    }
+
+    isNormal() {
+        return this.ssm === Arinc429Word.SignStatusMatrix.NormalOperation;
+    }
```
beheh, line 19: What about more utility methods? It seems likely we will need `isFailureWarning`.

src/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -241,7 +256,7 @@ export const MachNumber = ({ mach }) => {
         }
     }, [showMach, machPermille]);
 
-    if (Number.isNaN(mach)) {
+    if (!mach.isNormal()) {
```
beheh, line 259: So I actually suspect that in a lot of cases, Functional Test will be displayed. For example, I believe there is multiple test procedures where you can artificial values, e.g. a slowly descending RA value, and then can cross-check the validity by looking at the PFD. As I would assume that artificially generated values are flagged as Functional Test (FT), this check would not be accurate if we assume that a similar test exists for the Mach number. Maybe this (and a lot of other values) actually should only be hidden in the NCD and FW case.

Maybe we should check in on Discord to see if we can find some more evidence to corroborate this.

src/instruments/src/PFD/index.tsx
```diff
@@ -8,247 +10,218 @@ import { HeadingOfftape, HeadingTape } from './HeadingIndicator';
...
-    private AirspeedAccFilter: LagFilter;
-
-    private AirspeedAccRateLimiter: RateLimiter;
-
-    private LSButtonPressed: boolean;
+export const PFD: React.FC = () => {
+    const [displayIndex] = useState(() => {
+        const url = document.getElementsByTagName('a32nx-pfd')[0].getAttribute('url');
+        return url ? parseInt(url.substring(url.length - 1), 10) : 0;
+    });
```
beheh, line 23: In my opinion this could shouldn't even live in the PFD, but one level above and supplied via a prop, but I'm not sure our React setup has any components above the PFD.

This diplayIndex concern seems like a concern greater than the PFD alone and could also apply to the ND or other displays.

src/instruments/src/PFD/index.tsx
```diff
@@ -8,247 +10,218 @@ import { HeadingOfftape, HeadingTape } from './HeadingIndicator';
...
+    const [showSpeedBars, setShowSpeedBars] = useState(true);
 
-        this.isAttExcessive = false;
+    const inertialReferenceSource = getSupplier(displayIndex, getSimVar('L:A32NX_ATT_HDG_SWITCHING_KNOB', 'Enum'));
+    const airDataReferenceSource = getSupplier(displayIndex, getSimVar('L:A32NX_AIR_DATA_SWITCHING_KNOB', 'Enum'));
+    const computedAirspeed = useArinc429Var(`L:A32NX_ADIRS_ADR_${airDataReferenceSource}_COMPUTED_AIRSPEED`);
 
-        this.AirspeedAccFilter = new LagFilter(1.2);
-        this.AirspeedAccRateLimiter = new RateLimiter(1.2, -1.2);
+    const isOnGround = getSimVar('SIM ON GROUND', 'Bool');
```
beheh, line 50: Not directly relevant to this PR, but we actually have LGCIU inputs for this now based on gear compression, although I'm not sure those are exported as a SimVar already.

src/systems/systems/src/simulation/mod.rs
```diff
@@ -457,7 +489,40 @@ pub trait Write<T> {
...
+    /// struct MySimulationElement {
+    ///     n: f64,
+    /// }
+    /// impl SimulationElement for MySimulationElement {
+    ///     fn write(&self, writer: &mut SimulatorWriter) {
+    ///        writer.write_arinc429("MY_SIMULATOR_ELEMENT_N", self.n, SignStatus::NormalOperation);
+    ///     }
+    /// }
+    /// ```
+    fn write_arinc429(&mut self, name: &str, value: T, ssm: SignStatus)
```
beheh, line 517: I find it a bit weird that `read_arinc429` returns a `Arinc429Word`, however `write_arinc429` requires the raw parameters making up an `Arinc429Word`. As we're probably going to introduce additional parameters here, like SDI and the Label, how about we just write an `Arinc429Word` itself? Then, if we make `Arinc429Word`  more complicated, we don't have go back and refactor all the `write_arinc429` calls.

beheh (changes requested):
Not a lot of specific changes from me, however I have left a few comments.

All in all this is a great improvement.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/arinc429.js
davidwalschots: I referred to the [Wikipedia page](https://en.wikipedia.org/wiki/ARINC_429) on ARINC 429, where I believe I took the representation for BNR encoding instead of the representation for discrete data. All in all I have no opinion on which of those to pick, and thus will change it into the version you suggest as by all means I consider you to be the ARINC 429 specification master 😀.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/arinc429.js
davidwalschots: Yeah. Good point. I only coded the ones that I needed at this point in time, but it's probably better to provide methods for all of them.

src/systems/systems/src/simulation/mod.rs
davidwalschots: In the PR description I wrote:

> The SimVar's name can be considered equivalent to the ARINC 429 label and SDI, in that it can without restrictions indicate what value the SimVar contains, and additionally can indicate what is the source of the information (e.g. `"ADIRS_1_...", "ADIRS_2_...").

Do you feel this statement is not entirely correct for our situation?

I'd like to note that likely soon your Arinc429 bus changes will change the way we write these SimVars (component -> bus -> simvar). For that situation I think having a `write_arinc429` function which accepts an `Arinc429Word` makes sense, as you're dealing with that type within the bus itself. However, at this very moment having a `write_arinc429` function which is somewhat equivalent to the current `write` feels _okay_ to me. This area probably is where our implementations are "colliding".

src/instruments/src/PFD/index.tsx
davidwalschots: I agree. The `displayIndex` functionality is a recurring pattern in various displays. It's probably a good idea to standardise these patterns once all displays are migrated to React.

src/systems/systems/src/simulation/mod.rs
beheh: You're right, this is redundant.

davidwalschots:
@beheh Thank you for your review. I've made some modications based on your comments. Note that I haven't included changes for all comments as some of them (while valid) are out of scope for this PR.

src/instruments/src/Common/arinc429.tsx
```diff
@@ -0,0 +1,60 @@
...
+        NormalOperation: 0b11,
+    });
+
+    static f64View = new Float64Array(1);
+
+    static u32View = new Uint32Array(Arinc429Word.f64View.buffer);
+
+    static f32View = new Float32Array(Arinc429Word.f64View.buffer);
+
+    ssm: number;
```
beheh, line 17: There's a way to make this type fancier and stricter using Typescript:

```typescript
    static SignStatusMatrix = Object.freeze({
        FailureWarning: 0b00,
        NoComputedData: 0b01,
        FunctionalTest: 0b10,
        NormalOperation: 0b11,
    } as const);

    // ...

    ssm: (typeof Arinc429Word.SignStatusMatrix)[keyof typeof Arinc429Word.SignStatusMatrix];
```

To explain what's happening:

First, forget the typeofs, they're just needed because of Typescript's overlaying. Think of it this way:

```typescript
ssm: SignStatusMatrix[keyof SignStatusMatrix]
```

First, we're grabbing the union of keys from SignStatusMatrix:

```typescript
keyof SignStatusMatrix = "FailureWarning" | "NoComputedData" | "FunctionalTest" | "NormalOperation";
```

Now, we index the SignStatusMatrix with those keys:

```typescript
SignStatusMatrix[keyof SignStatusMatrix] = SignStatusMatrix["FailureWarning" | "NoComputedData" | "FunctionalTest" | "NormalOperation"]
```

...which gives us the union of values of SignStatusMatrix:

```typescript
SignStatusMatrix[keyof SignStatusMatrix] = SignStatusMatrix["FailureWarning"] | SignStatusMatrix["NoComputedData"] | SignStatusMatrix["FunctionalTest"] | SignStatusMatrix["NormalOperation"]
```

...which is resolved to:

```typescript
SignStatusMatrix[keyof SignStatusMatrix] = 0 | 1 | 2 | 3;
```

One little detail is that by default, Typescript will widen literals to be helpful. So for example `let foo = true;` will be helpfully widened to `let foo: bool = true;` instead of the "correct" `let foo: true = true`. The same things happens with numbers `let x = 5;` is `let x: number = 5` instead of `let x: 5 = 5`.
So by default, the last code snippet above would actually read:

```typescript
SignStatusMatrix[keyof SignStatusMatrix] = number | number | number | number;
```

...which doesn't help us at all. So to force Typescript to use the literal value, we add the little `as const` marker after your object above:

```typescript
    static SignStatusMatrix = Object.freeze({
        FailureWarning: 0b00,
        NoComputedData: 0b01,
        FunctionalTest: 0b10,
        NormalOperation: 0b11,
    } as const);
```

That way, we correctly get:

```typescript
SignStatusMatrix[keyof SignStatusMatrix] = 0 | 1 | 2 | 3;
```

Of course you could also just define:

```
    ssm: 0 | 1 | 2 | 3;
```

...but where's the fun in that? :-)


In either case, we'll end up where code like this:

```typescript
word.ssm == Arinc429Word.SignStatusMatrix.FailureWarning; // ok
word.ssm == 4;
word.ssm == 3; // ok
word.ssm == "5";
```

will error as expected:

```
test.ts:20:1 - error TS2367: This condition will always return 'false' since the types '0 | 1 | 2 | 3' and '4' have no overlap.

20 word.ssm == 4;
   ~~~~~~~~~~~~~

test.ts:22:1 - error TS2367: This condition will always return 'false' since the types 'number' and 'string' have no overlap.

22 word.ssm == "5";
   ~~~~~~~~~~~~~~~
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/A32NX_BaseNDCompass.js
```diff
@@ -408,12 +407,14 @@ class Jet_NDCompass extends HTMLElement {
...
 
-        const roundedHeading = fastToFixed(heading, 3);
+        const roundedHeading = fastToFixed(heading.valueOr(0), 3);
         this.setAttribute("heading_bug_rotation", roundedHeading);
-        if (groundSpeed <= 10) {
+
+        const groundSpeed = Arinc429Word.fromSimVarValue(`L:A32NX_ADIRS_IR_${inertialReferenceSource}_GROUND_SPEED`);
+        if (groundSpeed.isNormalOperation() && groundSpeed.value <= 10) {
             track = heading;
         }
```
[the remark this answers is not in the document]
davidwalschots: As discussed via Discord, you'll pick this up in a separate PR.

src/instruments/src/Common/arinc429.tsx
davidwalschots: A great thorough explanation on increasing type strictness. Thanks for that. I've applied your suggestion and added some error handling of my own.

jcalor:

Discord          : Jorge#6816
Object of testing: #5598 
Tier of Testing  : 2
Date             : 22/09/2021

Testing Process: 

Complete flight between EDDM and EDDS under normal procedures. Descended with LVL/CH FPA mode. Performed TEKSI5A arrival, ILS 07 with autoland.

Negatives:       

- While approaching to EDDS (once on STAR), I´ve noticed that aircraft was following the route but not the FD. Also, I´ve realized that the more the aircraft descends, the more precise the FD is (see screenshots). I only were able to recreate this issue with FPA mode, didn´t try V/S mode.

Testing Results: Passed (?)

Conclusions: Test is partially passed, but note the issue described above. Is not a major issue becouse airplane can fly but... issue is there.

Media:       
Notice how deviated is the FD on this first screenshot:
screenshot1
And now how less deviated is on this:
screenshot2

lukecologne:
That's just how the FPV/FPD (flight path vector/flight path director) in TRK/FPA works, the lateral deviation on the artificial horizon is the wind drift angle (the same can be observed on the heading tape, since you descended the wind probably weakened, so the drift angle gets less). The roll command in TRK/FPA is displayed by a rotation of the flight path director, instead of a lateral deviation like on the normal "bar-like" FD.

jcalor:
Ahh, I see. Didn´t use too much TRK/FPA. Then PR works as expected! I´ll edit my post :)

ShadowFrost3:


Object of testing: #5590
Tier of Testing:2
Date: 22/09/2021

Testing Process: Many things C&D flight KATL - KDCA

APU Batt start - checked
APU fuel happy triangle - checked
APU lower ECAM appears nominal - checked
Start ECAM statuses... no noticeable inaccuracies or missing items - checked
PFD response nominal - checked
Cabin Pressure page appears nominal - checked
ND pages CSTR, WPT, VORD, NDB, ARPT - Checked
Display brightness switches - checked checked checked checked (hint hint devs)
Upper ECAM has been behaving - checked
Upper ECAM tells me that I need to wear seatbelts - checked :)
Ram air bleed page indicating correct above 1 PSI and switch on - checked
Cross bleed indication nominal - checked
ENG Bleed switches and Bleed page are nominal - checked
ENG Fire test nominal, lower ECAM switches to ENG page, upper ECAM ENG fire - checked
RNAV Approach indication seemed happy - checked 



Managed to get four lower ECAM buttons to light up at once - bug... question is... whose bug - unable to recreate 
Bleed temps very HOT - failed - didn't recreate on base build
Pack fault 1/2 until takeoff mode recognized by aircraft - likely related to abnormal bleed temps
Bleed temp overheat audible warning missing - my guess is not implemented - not related I imagine 

One attempt at recreation of bleed/pack heating did not reproduce results. Not confident in 1-1 testing but writing this up for the moment so it is known. 





Negatives: Aircraft too hot (bleeds)
Testing Results: Failed 

Conclusions: More testing needed at this moment to confirm whether bleed issue exist only in this PR or is the result of a larger problem. 

Media: TOO HOT (it was a fair amount hotter than this as well)
image

tracernz:
The bleed system will be replaced with our own in https://github.com/flybywiresim/a32nx/pull/5384 so no need to worry about that.

ewold:

Discord: E-wold#8195
Object of testing: #5598
Tier of Testing: 1
Date:  23/09/2021
Testing Process: Doing to full flights > 1000nm
Checked screens for invalid readings / bugs when during full flight. 
Checked : 
* engines
* bleed
* pressure
* electrics
* hyrdraulics
* fuel
* apu
* cond
* doors
* wheels
* flight controls
* status 
Checked other screens. 
Remark : The initial fuel screen was at first a few orange pumps which then refreshed to green.  This was only the first time. 

Negatives: N/A
Testing Results: Passed

Conclusions: Everything seems fine in regards to reading values.

TheSledge-SLEJ:
OO

Stryderein:
Hi, anyone interested in making a software to convert ARINC 717 / 429 to a simulator required format. kindly get in touch with me

PR #5692, opened 2021-08-20
fix: new version of the SimVar provider to fix re-render issues

### LINKED ISSUE #4388: Memory leak and increase lag
**Mod Version**
Development e049d43 and ac9fbb4

**Describe the bug**
When fling with FlyPad it starts consuming more and more memory (after 40 minutes flight it consumes 16 GiB of RAM). It seems to be connected with fetching flypath from SimBrief.

**To Reproduce**
1. Switch on FlyPad
2. Fetch data from SimBrief on first page
3. Fly

**Expected behavior**
RAM usage is constant and system do not lag

**Actual behavior**
RAM usage increases and system increasingly lags (it's hard to say if it is due to memory pressure, GC overhead or something else).

**References**

Possibly duplicate?  #3949 

**Additional context**

Not to my knowledge (does not apply to master)

N/A. Unmodded game does not have fly pad

TheOftedal:
Fixes #4388

### New version of the SimVar provider with the following fixed/features
- RXJS based subscription event emitter for better control of updates.
- Ensures that only unique changes to variables result in update events being emitted.
- Code splitting for better and more logical structure.
- Correct typing and usage of arrow functions/function components.
- Better use of memoization with useMemo/useCallback.
- A per-instrument max update frequency for better control over performance.

###NOTE!!!
This is a PoC that touches all instruments with dependencies to SimVar's and must be thoroughly tested before being merged

## Additional context
The new SimVar provider is based on RXJS and uses subscriptions for each SimVar in order to listen to value updates instead of polling for updates as currently used. A push vs poll based system ensures much better control of updates which in turn gives much better control of re-renders. Currently, if a simvar value is being polled every frame, it will result in re-renders for the parent component and any child components for every single frame even when the variable has not changed.
Due to poor use of memoization throughout the React code this will in turn result in a massive amount of re-renders and consequently poor performance.

With the push based subscription system updates are only pushed when the variable value has actually changed, thus dramatically reducing the amount of re-renders in dependent components. In addition, it is now possible to limit the update frequency on a per-instrument basis to provide better control of how frequent these update events are to be emitted.

Discord username (if different from GitHub): DrPopsicle#8400

## Testing instructions
This is a major change that touches every part of the A32NX instruments and as such should be thoroughly tested before being accepted and merged. Consider this a PoC as changes may need to be implemented to satisfy all requirements. All instruments with dependencies to SimVars must be thoroughly tested.


TheOftedal:
##NOTE
This change currently does not work and requires changes in the way the SimVarProvider is mounted.
Please wait with testing until this has been resolved.

2hwk:
> ##NOTE
> This change currently does not work and requires changes in the way the SimVarProvider is mounted.
> Please wait with testing until this has been resolved.

If this is the case I'll mark this PR as draft until it is ready

src/instruments/src/Common/simvars/helpers.ts
```diff
@@ -0,0 +1,32 @@
...
+    case 'feets':
+    case 'Feets':
+        return 'feet';
+    case 'Knots':
+    case 'knots':
+        return 'knots';
+    default:
+        return unit;
+    }
+};
```
bouveng, line 32:
```suggestion
export const normalizeSimVarUnitName = (unit: SimVarUnitName): SimVarUnitName => {
    return unit.toLowerCase();
};
```

TheOftedal: We should probably add better typing for the simvar units and fix them where applicable to avoid having to sanitize/normalize the units. Is there a complete list of all accepted simvar units somewhere? I know there's better typing in the "simvar.d.ts" file under root typings but these are currently not exported.

tracernz: https://docs.flightsimulator.com/html/index.htm#t=Programming_Tools%2FSimVars%2FSimulation_Variable_Units.htm
Not sure on completeness of that list.

MisterChocker: I'd also suggest adjusting the unit names within the project to use the correct unit name straight of the batch.

TheOftedal: I have added all supported simvar typings and cleaned up all usage of simvars to conform with the correct units.
As these typings are currently only exported from the common folder within instruments I have added some temporary workaround for the globals typing for SimVars to accept the new typings in the get/set functions.

Benjozork:
I've made a branch `fix/rerender-issues` on this repo and rebased+fixed the current state of this.

tracernz (approved):

tracernz:
There are a couple of lint errors to be cleaned up.

src/instruments/src/Common/hooks.tsx
```diff
@@ -1,12 +1,31 @@
...
+ * inside the callback. This avoids unnecessary re-renders of your component if that variable is only used in this callback.
+ *
+ * Only use dependencies if you absolutely must keep up with a component state in the callback.
+ *
+ * @param handler           a callback function, called every time MSFS fires an update event. This varies based on
+ *                          the "GLASS COCKPIT REFRESH RATE" sim setting, but takes deltaTime for stability.
+ * @param additionalDeps    an array of dependencies. this acts just like the same parameter on {@link useCallback}, which is
+ *                          actually what is used internally.
+ */
+export const useUpdate = (handler: (deltaTime: number) => void, additionalDeps: any[] = []) => {
```
beheh, line 21: I like the new API, however this is not backwards-compatible:

Previously, a new handler function would have been used automatically because it would have been synced to the ref via useEffect. Now, if  additionalDeps is not passed, the value in useCallback will never be changed because the changes array is empty. While that is probably still useful, we will need to carefully inspect whether any previous usage of useUpdate passed a calback that depended on local variables (that would be stale after this change).

beheh (approved):
I wish this PR had separated out the functional (hook) changes from the boilerplate type renames. Reviewing it as-is is more annoying than it would have needed to be.

High-level this PR is a slight improvement over the current state. However, to me this PR is going in the wrong direction by reintroducing class instances representing SimVars, doubles down on the maxStaleness and shared state concepts, and as an additional cost pulls in two new dependencies.

In more detail, the issues to me are:
1. Getting away from SimVar classes was the entire reason these hooks were built in the first place. There were a bunch of subtle bugs that were happening because changes to these instances were not being consistenly synced to React. While this may be less of an issue now that people can just use hooks, we need to stay vigilant and make sure people don't start casually using these in React code directly in future.
2. The `maxStaleness` concept was a cute idea, but has turned out to be too confusing for users. The idea was you could specificy how often a value would be checked at least and then implement any kind of "redraw this value only once every second" on top of it, but it turns out that people just know that a value refreshes at a certain refresh rate and put that there. If another consumer in the same instrument happened to refresh that SimVar, it would also be refreshed in the unrelated place. So I think this concept should just go away in favour of a fixed refresh rate.
3. SimVar values being shared within an instrument also have not shown to be needed very often. The idea was that if 5 places in an instrument depend on one SimVar, you would not have to worry about syncing it down, but instead could access it in any place and we would call GetSimVar only once. However, in reality SimVars are rarely needed in multiple vastly different places at the same time, and we have shown that in cases with tight performance considerations we can always step in and manually hoist them (e.g. see our PFD work).

If this PR is indeed showing tangile real-world performance improvements, I'm okay with the changes, but I would avoid us landing this just for the sake of it.

Benjozork:
> Getting away from SimVar classes was the entire reason these hooks were built in the first place. There were a bunch of subtle bugs that were happening because changes to these instances were not being consistenly synced to React. While this may be less of an issue now that people can just use hooks, we need to stay vigilant and make sure people don't start casually using these in React code directly in future.

I'm not sure what you are referring to here. I've never used "SimVar classes" before and don't remember that being a pattern we ever used. Can you point me to what you mean exactly ?

> The `maxStaleness` concept was a cute idea, but has turned out to be too confusing for users. The idea was you could specificy how often a value would be checked at least and then implement any kind of "redraw this value only once every second" on top of it, but it turns out that people just know that a value refreshes at a certain refresh rate and put that there. If another consumer in the same instrument happened to refresh that SimVar, it would also be refreshed in the unrelated place. So I think this concept should just go away in favour of a fixed refresh rate.

While that was true of the old implementation, this new one happens to solve that issue. Is there anything about the general concept of `maxStaleness` that you don't like ?

> SimVar values being shared within an instrument also have not shown to be needed very often. The idea was that if 5 places in an instrument depend on one SimVar, you would not have to worry about syncing it down, but instead could access it in any place and we would call GetSimVar only once. However, in reality SimVars are rarely needed in multiple vastly different places at the same time, and we have shown that in cases with tight performance considerations we can always step in and manually hoist them (e.g. see our PFD work).

So what is being suggested here ? Avoiding the cache system altogether and instead updating every instance regardless of the number of them ?

src/instruments/src/Common/hooks.tsx
Benjozork: I don't aim for this to be backwards compatible. The reason this worked before is not really because of the ref, it's because any component would re-render at the highest `useSimVar` refresh rate (often 30Hz) causing this handler to be synced via that ref.

I actually added this array because removing the frequent re-renders actually broke the old system, as explained above.

beheh:
> I'm not sure what you are referring to here. I've never used "SimVar classes" before and don't remember that being a pattern we ever used. Can you point me to what you mean exactly ?

See the original [CachedSimVar.mjs](https://github.com/flybywiresim/a32nx/blob/b494bc3f9186c45ae0eaf3c918a993485476f2ca/src/instruments/src/RMP/Framework/CachedSimVar.mjs) that motivated #3270 (which removed them) because of sync issues and weird RMP bugs.

> While that was true of the old implementation, this new one happens to solve that issue.

Nothing has changed about that behaviour as far as I can see from reading this PR. If Component A says "the FWC flight phase should refresh every 3 seconds" and Component B says "the FWC flight phase should refresh every 10ms", then Component A will be redrawn with an updated FWC flight phase within 10ms. That's what maxStaleness guarantees, because every SimVar is hoisted to a single global representation.

>  Is there anything about the general concept of maxStaleness that you don't like ?

Yes! Quoting myself:

> [...] it turns out that people just know that a value refreshes at a certain refresh rate and put that there. If another consumer in the same instrument happened to refresh that SimVar, it would also be refreshed in the unrelated place. 

The original intent was that maxStaleness would be very rarely used, and we'd make global decisions like "in general we refresh SimVars every 100ms". In practice, everyone picks random numbers or refresh rates, as is evident through the useSimVar lines touched in this PR.

> So what is being suggested here ? Avoiding the cache system altogether and instead updating every instance regardless of the number of them ?

Right, which ultimately means refreshing the vars during useUpdate and/or a time threshold and doing away with all the subscription behaviour. At that point it's not a lot more than `useUpdate` coupled with a single `useState` for the variable you're watching.

Benjozork:
> Nothing has changed about that behaviour as far as I can see from reading this PR. If Component A says "the FWC flight phase should refresh every 3 seconds" and Component B says "the FWC flight phase should refresh every 10ms", then Component A will be redrawn with an updated FWC flight phase within 10ms. That's what maxStaleness guarantees, because every SimVar is hoisted to a single global representation.

Components no longer re-render if the actual simvar value isn't different. The SimVar will be queried, but the component won't re-render.

However one could argue, considering the above fact, that maxStalenesss is now not really useful. Except im VERY specific cases.

beheh:
> Components no longer re-render if the actual simvar value isn't different. The SimVar will be queried, but the component won't re-render.

My concern has never been about rerenders, Performance is always the secondary concern. The main concern is about the functional, simmer-facing implication that refresh intervals can (and have been) misused as refresh intervals, and values can update out-of-band because some unrelated component requests it.

Imagine a clock, e.g. on the SD, requesting the time with maxStaleness 1000ms. You would expect every second to last exactly a second.
Now image there's another component that for whatever reason refreshes the clock every 300ms. What happens is the following:

| Time                     | 0.0 | 0.3 | 0.6 | 0.9 | 1.2 | 1.5 | 1.8 | 2.1 | 2.4 | 2.7 | 3.0 | 3.3 | 3.6 | 3.9 | 4.2 |
|--------------------------|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|
| Shown Value on the Clock | 0   | 0   | 0   | 0   | 1   | 1   | 1   | 2   | 2   | 2   | 3   | 3   | 3   | 3   | 4   |
| Tick/Render Count        | 1   | 2   | 3   | **4**   | 1   | 2   | **3**   | 1   | 2   | **3**   | 1   | 2   | 3   | **4**   | 1   |

What happens is that the 1000ms "refresh interval" are completely ignored, and a completely unrelated component has now suddenly made your clock run non-linearly. You will see a long second, then two short seconds, followed by a long second again etc. etc.

> Should we then remove it in the future ?

I think the above case is a strong reason to make `maxStaleness` a true refresh `interval` instead. I think the current hooks are trying to be cleverer than they should be with the deduplication.

Benjozork:
We've discussed this and come to the conclusion it's better to split this up into smaller pieces.

Issue #5723, opened 2021-08-25
EFB Landing calculation page issues


donstim:
### Aircraft Version

Development

### Build info

```json
{
  "built": "2021-08-23T20:29:51+00:00",
  "ref": "merge",
  "sha": "fbd755b714259ced23b74315f272eaa80cad4582",
  "actor": "sidnov",
  "event_name": "manual"
}
```

### Describe the bug

After a landing calculation has been done once, the "Clear" button does not work and pressing the "Calculate" button after inserting new values in the wind direction and wind magnitude or rwy heading fields does not result in new landing distances being calculated. 

### Expected behavior

1) Pressing the "Clear" button should always clear all entry fields, and 
2) Changing the Wind Direction/Magnitude and Rwy Hdg inputs should result in a new landing distance calculation when changing from a headwind to a tailwind or vice versa, or when a tailwind magnitude changes.

### Steps to reproduce

Input entries in all fields of landing performance calculator. (Example that I used: 
![image](https://user-images.githubusercontent.com/70166617/130847231-6f9f6c77-fb3d-4c2d-a009-86100eefc966.png) - from a Discord user. (Although I did not get the crazy answers he got, the issue regarding clearing entries and changing wind speeds/direction and runway direction is present.)

Press "Clear" button

Change Wind Direction to match Rwy Hdg or vice versa. There is no change in the calculated landing distances even though the wind component went from a 15 knot tailwind to a 23 knot headwind. Changing the wind magnitude, which would change the tailwind component also does not change the calculated landing distances.

### References (optional)

![Screenshot (1117)](https://user-images.githubusercontent.com/70166617/130849097-a7007a5b-a2a4-4b12-ad1b-23b5adcb6101.png)

The calculated landing distances vary with the tailwind component.

### Discord Username (optional)

donbikes#4084

PR #5740, opened 2021-08-28
feat(efb): fix landing calculator issues and allow for unit toggle

### LINKED ISSUE #5734: flyPad: Landing Calculator Unit Toggles
### Aircraft Version

master

### Description

Add toggles to some/all of the units in the landing calculator. For example, QNH in mb to in Hg. Big one is LDA to feet, since all the US charts reference runway lengths in feet.

Ideally this would also toggle the output into feet as well.

The kg/LB toggle could probably use the setting already present in the flyPad settings.

### References (optional)

image

_No response_

Kennedy#6132

tracernz:
Fixes #5734 and #5723

## Summary of Changes

- General: Added a typescript unit conversion class similar to NXUnit that is synced to the `CONFIG_USING_METRIC_UNIT` setting.
- Landing calculator: Added unit toggles for "temperature", "pressure", "runway LDA" and "weight" inputs. The initial values and subsequent conversions make use of the new Units class.
- Landing calculator: Fixed the "Clear" button.
- Landing calculator: Fixed the "Wind Magnitude" input.
- Landing calculator: Made a separate button (from current *"Get METAR"*) to sync the weight.
- Landing calculator: Synced the displayed landing distance unit to the new "runway LDA" unit toggle *(m | ft)*. Also added a space between values and units, in the interests of good taste.
- Landing calculator: Made all inputs initially empty (by having *undefined* values), so that the placeholders can be seen.

## Screenshots (if necessary)
Initial landing calculator page (using metric system):
image
Filling in weight + runway LDA in ft:
image
Calculated values, length in m:
image

## Additional context

- @ghitier -> discord: `bk67`

## Testing instructions

Check that the landing performance page initially uses the units you select in Settings -> Aircraft -> Weight Unit (`kg` corresponds to metric units), when you change that setting the toggles should reflect that choice.

> Except for the Temperature which should initially always be in Celsius (it's the unit METARs use)

If you're able to use the Coherent debugger (if not don't worry), try to change the unit setting while looking at the EFB page:

- Open the debugger
- Choose the CDU
- In the console, run:
  - `NXDataStore.set('CONFIG_USING_METRIC_UNIT', '1')` to choose metric,
  - or `NXDataStore.set('CONFIG_USING_METRIC_UNIT', '0')` to choose imperial.
- Values should update in real-time on the EFB.

---

When you go to the landing calculator all inputs should initially be empty. The *"Calculate"* button will initially be greyed out, it should become enabled only once all fields are filled.

If you use the *"Get METAR"* button, make sure the weight doesn't get filled in anymore.

Check that the when you toggle between units, the value is converted accurately.

Use the button (down arrow icon) right of the weight label to auto-fill the input. Make the value is coherent with your currents aircraft weight. 

Make sure to play around with the different values to see if the landing distance gets updated coherently.

Use the *"Clear"* button and check all inputs get back to an empty state (placeholders visible).

2. Click on the **Checks** tab on the PR
3. On the left side, click on the bottom **PR** tab
4. Click on the **A32NX** download link at the bottom of the page


src/shared/src/units.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    static userToKilogram(value: number): KiloGram {
+        return Units.metricUnits ? value : value / 2.20462;
+    }
+
+    static kilogramToUser(value: KiloGram): number {
+        return Units.metricUnits ? value : value * 2.20462;
+    }
+
+    static get userWeightUnit(): 'kg' | 'lbs' {
```
Benjozork, line 29: **thought:** I'm not sure if the prefix should be the responsibility of this class.

If we don't want to refactor this a different way I would name this method `userWeightSuffixEis2` or something.

src/shared/src/units.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    static userToMetre(value: number): Metre {
+        return Units.metricUnits ? value : value / 3.28084;
+    }
+
+    static metreToUser(value: Metre): number {
+        return Units.metricUnits ? value : value * 3.28084;
+    }
+
+    static get userLengthUnit(): 'm' | 'ft' {
```
Benjozork, line 42: Ditto

src/shared/src/units.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    static userToCelsius(value: number): Celsius {
+        return Units.metricUnits ? value : (value - 32) * 5 / 9;
+    }
+
+    static celsiusToUser(value: Celsius): number {
+        return Units.metricUnits ? value : value * 9 / 5 + 32;
+    }
+
+    static get userTemperatureUnit(): '°C' | '°F' {
```
Benjozork, line 54: Ditto

src/shared/src/units.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    static hectopascalToUser(value: Celsius): number {
+        return Units.metricUnits ? value : value / 33.863886666667;
+    }
+
+    static hectopascalToUserString(value: Celsius): string {
+        return Units.metricUnits ? "" + Math.round(value) : (Math.round(value / 0.33863886666667) / 100).toFixed(2);
+    }
+
+    static get userPressureUnit(): 'hPa' | 'in.Hg' {
```
Benjozork, line 70: Ditto

Benjozork (changes requested):
Please fix lint

ErickSharp:
Can we please change metre to meter 😩

tracernz:
> Can we please change metre to meter 😩

We could, but the [SI unit for length is spelt metre](https://en.wikipedia.org/wiki/International_System_of_Units), and this is the spelling [used in all countries in the world except 2](https://en.wikipedia.org/wiki/Metre), so it wouldn't make a lot of sense.

src/shared/src/units.ts
ErickSharp: How come you would prefer for this to not be within the scope of this class? I think it makes sense to be able to access prefixes and conversions from a 'Unit' class instead of splitting prefixes off into a 'Prefix' class

ErickSharp:
Hiya, think we can address #5734 ? In addition to whats included in this pull request, it would be nice to have a toggle that initially assumes the aircraft's configured unit setting, but allows overriding.

ghitier:
@tracernz @ErickSharp As there's no progress on this PR and FlypadOS v3 is on it's way, I suggest closing this PR in favor of #6317 I just opened. It contains the same essential fixes, but without the unit and toggles features. It would probably be better to integrate any new needed EFB features directly into v3 imo.

Let me know if that's ok.

ErickSharp:
Closing this pull request as it has become stale in favor of #6317

ErickSharp:
}

tracernz:
Okay. The unit toggle was really the main point though. People in America don't like metric all that much. 🤣

PR #5786, opened 2021-09-05
feat: proportional–integral–derivative controller

crocket63:
Fixes #[issue_no]

## Summary of Changes

Simple PID implementation that should be imune to integral windup and provides non symetrical output bounds.
No filter added on D term so use d term at your own risk for noisy system responses

## Additional context

## Testing instructions

Autobrake controller is the guinea pig using the new PID controller.
Please check autobrake behaviour in LO MED and MAX in different CPU load conditions:
- huge airports with lower fps,
- small airports without traffic for best fps performance.

Autobrake should behave normally, and consistently between all cpu load conditions
-No abnormal brake oscillations in any autobrake mode
-Decel light not blinking forever (can blink once or twice as already observed on real plane)


src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn pid_init() {
+        let pid = Pid::new(1., 1., 1., 0., 1., 1.);
+
+        assert!(pid.output == 0.)
+    }
+
+    #[test]
+    fn proportional() {
+        let mut pid = Pid::new(2.0, 0.0, 0.0, 0.0, 100.0, 10.0);
+        assert!((pid.setpoint - 10.).abs() < f64::EPSILON);
```
BlueberryKing, line 108:
```suggestion
        assert_eq!(pid.setpoint, 10.);
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        let error = self.setpoint - measurement;
+
+        let p_term;
+        if self.error_k_1.is_some() {
+            p_term = (error - self.error_k_1.unwrap()) * self.kp;
+        } else {
+            p_term = error * self.kp;
+        }
```
BlueberryKing, line 64:
```suggestion
        let mut p_term = error * self.kp;
        if let Some(error_k_1) = self.error_k_1 {
            p_term -= error_k_1 * self.kp;
        }
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    #[test]
+    fn pid_init() {
+        let pid = Pid::new(1., 1., 1., 0., 1., 1.);
+
+        assert!(pid.output == 0.)
```
BlueberryKing, line 102:
```suggestion
        assert_eq!(pid.output, 0.)
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        // Test simple proportional
+        assert!((pid.next_control_output(0.0, None) - 20.).abs() < f64::EPSILON);
+    }
+
+    #[test]
+    fn derivative() {
+        let mut pid = Pid::new(0.0, 0.0, 2.0, -100.0, 100., 10.0);
+
+        //No derivative term for first two updates
```
BlueberryKing, line 118:
```suggestion
        // No derivative term for first two updates
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+        assert!((pid.next_control_output(0.0, None) - 0.).abs() < f64::EPSILON);
+        assert!((pid.next_control_output(0.0, None) - 0.).abs() < f64::EPSILON);
+
+        // Test that there's a derivative at 3rd update
+        assert!((pid.next_control_output(5.0, None) - -10.).abs() < f64::EPSILON);
+
+        //Then no more derivative term
```
BlueberryKing, line 125:
```suggestion
        // Then no more derivative term
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn reset_error(&mut self) {
+        self.error_k_2 = None;
+        self.error_k_1 = None;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
```
BlueberryKing, line 97:
```suggestion
    use super::*;

```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,173 @@
+use std::time::Duration;
+
+// Pid controller implementation
+// Implementation in a recursive form
+// u(k+1) = u(k) + (e(k)-e(k-1))*Kp + e(k) * Ki * dt + (e(k)- 2*e(k-1) + e(k-2)) * Kd / dt
```
BlueberryKing, line 5:
```suggestion
// u(k+1) = u(k) + (e(k) - e(k-1)) * Kp + e(k) * Ki * dt + (e(k) - 2 * e(k-1) + e(k-2)) * Kd / dt
```

BlueberryKing (changes requested):
Good stuff! Left some suggestions, mostly stylistic things.

I quickly tested it out in the pneumatic system and it was very easy to use. I did not run into any stability issues. I might need to re-tune some parameters but that's expected.

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,166 @@
...
+        Self {
+            kp,
+            ki,
+            kd,
+            min_output,
+            max_output,
+            setpoint,
+            error_k_1: None,
+            error_k_2: None,
+            output: 0.,
```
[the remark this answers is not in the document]
crocket63: This one is pretty much on purpose, as for None error we just skip d term evaluation whereas for first iteration the output really should be 0.  Yet I see how it can get confusing...... Will have a second look!

src/systems/systems/src/shared/pid.rs
crocket63: This is unreadable I know, but your suggestion would not compile because of direct f64 comparison sadly.... Unless I miss your point?

src/systems/systems/src/shared/pid.rs
crocket63: Same as the other assert comment :/

src/systems/systems/src/shared/pid.rs
crocket63: Wouldn't we add another multiplication here just for fun? I know cpu is cheap :D    Must admit I'm not a fan of the let Some sugar stuff.

src/systems/systems/src/shared/pid.rs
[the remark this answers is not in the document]
BlueberryKing: Ok, I guess I can see how you'd want to have `output` be `0` if `next_control_output` has not been called yet. 👍

src/systems/systems/src/shared/pid.rs
BlueberryKing: As you say, one multiplication is really not something we should be concerned about. I can live with the `is_some` -> `unwrap` stuff, but I wouldn't feel good if I hadn't point it out :D

src/systems/systems/src/shared/pid.rs
BlueberryKing: Really? You can't just compare two floats in Rust like that? Good to know. Sorry about that!

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,178 @@
+use std::time::Duration;
+
+// Pid controller implementation
+// Implementation in a recursive form
+// u(k+1) = u(k) + (e(k) - e(k-1)) * Kp + e(k) * Ki * dt + (e(k) - 2 * e(k-1) + e(k-2)) * Kd / dt
+// For variable dt add dt duration as argument, if fixed dt or scheduled synchronously with controlled system input None as dt
+pub struct Pid {
```
davidwalschots, line 7: **suggestion**: Maybe rename to `PidController`?
```suggestion
pub struct PidController {
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -0,0 +1,178 @@
+use std::time::Duration;
+
+// Pid controller implementation
+// Implementation in a recursive form
+// u(k+1) = u(k) + (e(k) - e(k-1)) * Kp + e(k) * Ki * dt + (e(k) - 2 * e(k-1) + e(k-2)) * Kd / dt
+// For variable dt add dt duration as argument, if fixed dt or scheduled synchronously with controlled system input None as dt
```
davidwalschots, line 6: **question**: Should this be a rust doc comment?

src/systems/systems/src/hydraulic/brake_circuit.rs
```diff
@@ -448,7 +444,8 @@ impl AutobrakeDecelerationGovernor {
     fn disengage(&mut self) {
         self.is_engaged = false;
         self.time_engaged = Duration::from_secs(0);
-        self.target = Acceleration::new::<meter_per_second_squared>(10.);
+        self.pid_controller
+            .reset(Self::RESET_ACCEL_VALUE_M_S_SQUARED);
```
davidwalschots, line 448: **question**: Would a particular instance of a `PidController` always be reset to the same value, or can the value be different?
**suggestion**: If it's always reset to the same value I'd suggest moving the reset value into the constructor function and then simply writing `self.pid_controller.reset()` here and at the other usage.

src/systems/systems/src/hydraulic/brake_circuit.rs
```diff
@@ -402,10 +403,7 @@ impl Default for AutobrakePanel {
 /// Deceleration governor is the PI controller computing the expected brake force to reach the target
 /// it's been given by update caller
 pub struct AutobrakeDecelerationGovernor {
-    target: Acceleration,
-    i_gain: f64,
-    p_gain: f64,
-    last_error: f64,
+    pid_controller: Pid,
```
davidwalschots, line 406: **thought**: It would be really awesome if we could write `Pid<Acceleration>`, such that we don't have the `.get::<meter_per_second_squared>()` calls everywhere. Sadly, uom's macro generic types are not so trivial and I've been unable to figure out which generic type constraints would need to be applied to the `Pid` type for it to work. Maybe @devsnek knows?

src/systems/systems/src/hydraulic/brake_circuit.rs
```diff
@@ -471,19 +471,14 @@ impl AutobrakeDecelerationGovernor {
...
-            let i_term = self.i_gain * target_error;
-            self.current_output += p_term + i_term;
-
-            self.last_error = target_error;
-
-            self.current_output = self.current_output.min(1.).max(0.);
+            self.current_output = -self.pid_controller.next_control_output(
+                self.filtered_acceleration.get::<meter_per_second_squared>(),
+                Some(context.delta()),
+            );
```
davidwalschots, line 477: **question**: Do you expect situations in which the delta wouldn't be available (and thus would be `None`)? Right now this only happens in your unit tests. Outside of unit tests it is always `Some`.

src/systems/systems/src/shared/pid.rs
davidwalschots: Not really the same though, as you're doing direct equality comparison here, so changing to `assert_eq!` is correct. The question here is whether one can safely compare for equality against 0.

src/systems/systems/src/hydraulic/brake_circuit.rs
davidwalschots: @devsnek Something _like_ that. But that isn't really the type I need. I think it's something along the lines of `Quantity<Dimension<???>, ???, ???>` but am not sure how to go from there.

> the trait bound `Quantity<(dyn uom::si::Dimension<L = PInt<UInt<UTerm, B1>>, J = Z0, M = Z0, T = NInt<UInt<UInt<UTerm, B1>, B0>>, N = Z0, I = Z0, Kind = (dyn Kind + 'static), Th = Z0> + 'static), (dyn Units<f64, electric_current = ampere, time = uom::si::time::second, luminous_intensity = candela, mass = kilogram, thermodynamic_temperature = uom::si::thermodynamic_temperature::kelvin, amount_of_substance = mole, length = meter> + 'static), f64>: uom::si::Unit` is not satisfied
the trait `uom::si::Unit` is not implemented for `Quantity<(dyn uom::si::Dimension<L = PInt<UInt<UTerm, B1>>, J = Z0, M = Z0, T = NInt<UInt<UInt<UTerm, B1>, B0>>, N = Z0, I = Z0, Kind = (dyn Kind + 'static), Th = Z0> + 'static), (dyn Units<f64, electric_current = ampere, time = uom::si::time::second, luminous_intensity = candela, mass = kilogram, thermodynamic_temperature = uom::si::thermodynamic_temperature::kelvin, amount_of_substance = mole, length = meter> + 'static), f64>`rustcE0277

src/systems/systems/src/shared/pid.rs
davidwalschots: @BlueberryKing Comparing floating point numbers for equality is tricky in any language. You can compare safely with `<` and `>`, but not with `<=`, `>=`, `==`, or `!=`. If you feel like hurting your brain, have a read [here](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/).

BlueberryKing: Yeah, I'm aware of this. I figured in this case there should be no problem as we are not doing any arithmetic before the comparison. Consequently `pid.setpoint` should indeed be precisely `10.` since this is perfectly representable in binary.

What caught me off guard was that `cargo clippy` actually complains about this.

src/systems/systems/src/hydraulic/brake_circuit.rs
crocket63: The pid library that Blueberry used at first had the "flaw" of not handling those delta times. 
So current implementation can use them, or not if None.

It should be used without delta in a fixed step context only (or adjusting gains depending on delta...).

src/systems/systems/src/hydraulic/brake_circuit.rs
crocket63: It might be useful to chose the reset point indeed for some cases. But your remark makes me realize code is wrong :D  The pid output is not necessarily same type as input, and in that case it's not, so it should be reset to 0 not to an acceleration value.

crocket63: Removed the parameter as this can be handled through a setpoint change when updating the controller.

davidwalschots (approved):

ShadowFrost3:


Object of testing: #5786
Tier of Testing:2
Date: 22/09/2021

Testing Process: C&D KDCA-KJFK
Rejected takeoff brakes work as expected high CPU
Landings conducted without spoilers or reversers to "feel" the brakes 
Med and low brakes work as expected high CPU
High, med, and low brakes work as expected at less CPU intensive places. 

Negatives: None found
Testing Results: Passed
Conclusions: 
Media:

jcalor:

Discord          : Jorge#6816
Object of testing: #5786 
Tier of Testing  : 1
Date             : 23/09/2021

Testing Process: 

ILS Practices:
- Spawned on RW25 EGGW. Calculated performance and did 2 rejected takeoffs
- Loaded aircraft again on RW25, calculated performance. Takeoff and started 2 ILS practices, the first one with autobrake LOW and the second one with MED.

Negatives: On the first landing, with autobrake LOW, inmediatly after touching down, got a stutter, with high CPU load and FPS dropped from ~42 to 25. FPS back to normal inmediatly, it was like 0.5sec.

Testing Results: Failed

Conclusions: My FPS were constant at every airport, so I couldn´t check if there´s more CPU load on huge apts than small apts. Rejected takoff + landings with autobrake MED went well. Maybe some checks from another QA Testers could find if this is a PR issue or my own issue. I´ll try to recreate the issue later.

crocket63:
> 
> Discord : Jorge#6816
> Object of testing: #5786
> Tier of Testing : 1
> Date : 23/09/2021
> 
> Testing Process:
> 
> ILS Practices:
> 
> * Spawned on RW25 EGGW. Calculated performance and did 2 rejected takeoffs
> * Loaded aircraft again on RW25, calculated performance. Takeoff and started 2 ILS practices, the first one with autobrake LOW and the second one with MED.
> 
> Negatives: On the first landing, with autobrake LOW, inmediatly after touching down, got a stutter, with high CPU load and FPS dropped from ~42 to 25. FPS back to normal inmediatly, it was like 0.5sec.
> 
> Testing Results: Failed
> 
> Conclusions: My FPS were constant at every airport, so I couldn´t check if there´s more CPU load on huge apts than small apts. Rejected takoff + landings with autobrake MED went well. Maybe some checks from another QA Testers could find if this is a PR issue or my own issue. I´ll try to recreate the issue later.

Had a quick test and can't reproduce a big performance hit in low AB mode. Would be nice to have some more people having a try. Was not expecting any performance change with this so I'll be surprised it's the PR itself. But well never knows :)

sidnov:
QA Report

Object of testing: #5786
Tier of Testing: 1
Date: 24/09/2021

Testing Process:
6 Landings/RTOs with different AB settings

1 on a fairly detailed airport, good/great FPS, no issues on touchdown
3 on KLAX, first landing had a significant 2-3 second pause, second and third were smooth without issues, with FPS being mediocre
2 on KJFK, no issues, mediocre FPS

All this was done with a YouTube stream on the second monitor, iTunes and a couple of browser tabs open (a bit taxing on the 8750H+1060 MaxQ combo)

Negatives: None
Testing Results: Passed

Conclusions: Apart from the one pause that I highly suspect to be linked with data streaming/rolling cache, I can't reproduce any stutters/pauses 100% accurately.

crocket63:
Did a small timing instrumentation on the autobrake update function and no problem shows related to this. Update keeps within 600 to 1500 nanoseconds.  There probably is a freeze problem on landing for some people but not related to AB change in this PR

PR #5792, opened 2021-09-07
feat: grid-layout same sized buttons on failure page

nathanmayall:
## Summary of Changes

EFB Failure buttons now look a lot more neat. Also uses a grid pattern, as more failures are added these will conform to the grid automatically.

## Screenshots (if necessary)

image

## Additional context

Discord username (if different from GitHub): AbsoluteNath#9406

## Testing instructions

Go to the Failures page in the EFB and check the buttons look neat and are in a grid 4 columns wide.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## 0.8.0
 1. [SOUND] Pack sounds use new APU model variables - @tracernz (Mike)
 1. [EFB] Moved setting for MCDU keyboard input into the EFB - @2hwk (2Cas#1022)
+1. [EFB] Neatened up failure buttons on failure page - @nathanmayall (AbsoluteNath#9406)
 
 ## 0.7.0
 1. [HYD] First building block, more to come. Hydraulics do not impact the sim YET. - @crocket6 (crocket)
--- a/src/instruments/src/EFB/Failures/FailureButton.tsx
+++ b/src/instruments/src/EFB/Failures/FailureButton.tsx
@@ -21,6 +21,7 @@ export const FailureButton: FC<FailureButtonProps> = ({ name, isActive, isChangi
             onClick={onClick}
             type={type}
             disabled={isChanging}
+            className="mx-2 my-2"
         >
             <span>{name}</span>
         </Button>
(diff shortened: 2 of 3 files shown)


ErickSharp:
Within this pull request, do you think you would be able to add some document event listeners to have some sort of dragging system? This would give us a lot of flexibility going forward but if not I can implement it shortly after this is merged.

nathanmayall:
@ErickSharp Sure, can do. Leave it with me

davidwalschots:
This view will over time have very different presentational requirements. It is merely a placeholder to build upon. I would deem it very unlikely the current way of failing things would persist, as having hundreds of buttons in a grid would not make sense. I would recommend reading through [this document](https://github.com/flybywiresim/rfcs/blob/1ed0dc339c3a43e95e6cd1636cf2619f007ee841/text/000-failures.md)'s requirements section to see what functionality we'd like to support at some point in the future. Does the grid change you propose now still fit with that, or would you take another approach?

I'd be curious to hear your thoughts!

> Within this pull request, do you think you would be able to add some document event listeners to have some sort of dragging system?
@ErickSharp this seems entirely unrelated to this PR and therefore shouldn't be added to it.

nathanmayall:
@davidwalschots Hi David, thank you for the feedback. Initially I just wanted to make the change to improve the current layout as it is. However after reading the doc you sent, I agree that the grid layout would not be ideal as time goes on.

For what you have at the moment, it's fine, but in the future I'd like to see some sort of tabbed display with the different categories of failure, and a section for engaging different scenarios such as random, real-world, pre-programmed and no-failure god mode etc. 

Although I'm more than happy for this to be merged as an improved placeholder, and once the failure system is more mature, I can work on another PR with the UI for said categories etc.

Cheers,

davidwalschots (approved):
Sounds good Nathan! I'll approve this because it is an improvement indeed 😀.

St54Kevin:
**Quality Assurance Tester**

Object of testing: #5792 
Tier of Testing: 1
Date: 08/09/2021

**Testing Process:**
1. Loaded onto RW and checked wether failures are shown in a 4 coumn grid

**Negatives:** N/A
**Testing Results:** Passed

PR #5909, opened 2021-09-25
fix(ap/fbw): improved behavior when using high sim rate and low performance

aguther:
## Summary of Changes
This PR aims to improve the behavior of the FBW/AP when using a high sim rate with low performance during cruise.

## Additional context
I think it should work now with sim rates up to 4x. I did a test with V-Sync limited to 20 fps and sim rate of 4 without seeing any oscillations.

## Testing instructions
- test during normal flights in lower altitudes but especially in high altitude with high speed and high sim rate.


Benjozork (approved):

ShadowFrost3:


Object of testing: #5909
Tier of Testing:1
Date: 25/09/2021

Testing Process: 
Recreation of bug due to lack of testing on previous relevant PRs on Dev build
KLGA 22 JUTE Departure - autopilot instability encountered at 15,000 ft at an average FPS of 44 (at time of incident) and time accel rate of 4x. Aircraft enters unrecoverable controllability state
Onto PR
KLGA 22 JUTE Departure - 44 fps (unlimited) - standard maneuvers (full on autopilot) - aircraft survives but with some instability - would be hesitant to assume it can survive all likely/plausible conditions under autopilot as it shows instability in level climb passing through ~20,000ft to cruising altitude. 
Manual heading change - 46-48 fps (unlimited) - aircraft will sometimes survive or enter upset state with 4x time accel
_Testing Stopped_

Negatives: 
Testing Results: Not Passed

Conclusions: ** Notes - No changes have been made to configuration files time accel limiter was active at times. Overall, some improvement in stability but not enough to consider the aircraft safe under x4 acceleration with ~45 fps. Another note, these issues seem to present themselves at altitude as lower FPS over the city did not produce issues. Could this be because the accel limiter was more likely to activate due to less margin on the FPS count? Or is it because higher density altitude has more margin for error?

aguther:
Officially there is still no support for sim rates higher than 1x.

i tried to improve situation as good as possible. In my tests this worked especially for cruise. If there is turbulence there is still the possibility that something happens.

Due to so many different situations and configurations it’s quite hard to do anything. Can you please share/provide video from what is happening?

ShadowFrost3:
Any operation where the FBW system is making inputs generally. Autopilot, manual input etc. The post above I was able to upset it only by heading changes and the video below I make a couple small left/right side stick inputs. All input from my side stops at or before 22 seconds. (Manually induced oscillation test) 

The situation is better than previously, however, if there are any extensive changes to any underlying systems they have not been tested to not cause other issues. I encountered a FMC/Autopilot issue but am leaning towards it being a generic issue rather than a PR issue. (~95% certain generic issue given I wasn't running "normal" SOPs and ended up hot spawning on runways to readily test different SIDS) General notes still apply, more likely to happen at high altitude rather than lower. Time accel protection is working but when the aircraft enters this state is it unrecoverable in my attempts. 

https://drive.google.com/file/d/1mcUV8FFzRDer-L8tB_0l-uncEld0__v9/view?usp=sharing

aguther:
I'm honestly confused, because for me it looks like this with around 40-50 fps and 4x:
https://user-images.githubusercontent.com/6629914/134801418-d09b0178-1f40-4489-b0fc-f029e15a989e.mp4

For limited fps to 20 and 4x (which normal limitation would not allow you to activate, it would be limited to 2x):
https://user-images.githubusercontent.com/6629914/134801444-81b3a13e-8f86-4e7f-99b6-e35c8ef363c9.mp4

ShadowFrost3:
Adding additional notes here, on development version. Even at 2x time accel with 55-60fps (should be over 17x2 by a margin) the aircraft is susceptible to the issues documented above. Did not try heading mode, only banking the aircraft left/right within 10 degrees. Wind vs no wind tested no difference. Aircraft entered an oscillation that quickly built an entered it into an upset condition like the above demonstrations.

lukecologne:
I am not able to reproduce your issues. At 38000ft, and around 40 fps (with occasional drops to 35), I have no issue whatsoever with oscillations, even at 4x simrate, both in manual flight and with AP engaged. (Although controlling the aircraft at 4x time acceleration is difficult in general). Also,

aguther:
Feedback is as such that the PR does not fix the issue.

PR #5971, opened 2021-10-05
refactor(fbw): Move elevator elasticity table, correct elevator maximum deflection

lukecologne:
## Summary of Changes

This PR simply removes the elevator elasticity tables (which are basically just gains scaling the elevator deflection as a function of q) from the flight model config file, and moves it to the FBW model.
This gives us more control over these gains, especially for example in upcoming control law reconfigurations such as direct law, where the gains will be different, and no longer depending on speed.

Further, this PR also corrects the elevator maximum deflection values to reflect the real life values. This should not affect the behavior of the law in any way.

The overall behaviour of the aircraft should be exactly the same as before, as this is only a refactor.

## Additional context

Discord username (if different from GitHub): lukecologne#1156

## Testing instructions

Simply fly the aircraft, and verify that it behaves the same as before in the pitch axis, in as many flight conditions as possible (i.e. high altitude, low altitude, high speed, low speed, takeoff, landing, abnormal situations such as high speed protection, high angle of attack protection etc.)


tracernz (approved):

IbrahimK42:
I have checked the previous model with this PR and did some pitching maneuvers in low speed, high speed.  The behavior seems the same.

jcalor:

Discord          : Jorge#6816
Object of testing: #5971
Tier of Testing  : 2
Date             : 6/10/2021

Testing Process: 

Full flight from LEIB to LEVC

All messages displayed are screeshoted on media section. No issues were found.

Negatives: N/A

Testing Results: Passed

Conclusions: PR works as expected, FMGC messages works as before.

Media:       

image
image
image
image
image
image
image
image

lukecologne:
Umm.. That's not the right PR you commented on. This one is not about the FMS, but the flight controls

lukecologne:
The law sets in at 70kn in the takeoff roll, and disengages as soon as you are in the air, at which point the pitch normal law engages. The maximum pitch rate that can be commanded in the rotation law is 12°/s. I don't understand what you are saying with the stall in other flight conditions. The maximum elevator deflection could obviously have always stalled the aircraft even before this change, but that is why the normal law exists, it will prevent excessive pitch rate and load factor, no matter the maximum deflection.

aguther:
When you want to achieve the SOP requested 3°/s around 1/4 of the Sidestick range it makes 12 °/s as limit for full deflection. Let‘s on top not forget that on CEO it‘s completely direct law with full authority…

aguther:
Yeah, it has tailstrike protection, but we deactivated it because people constantly got into it, maybe something related to how the airplane comes in the air (too slowly). 

I mean, if pilots tell me IRL they have to pull more on the stick I’m happy to do so. But having 320 Sim Pilot in mind that always gets excited on TO and no other feedback that it’s wrong that’s my guess why it seems to be really ok right now.

donstim:
Fair enough. I'll ask in the pilot's channel.

What do you mean the airplane comes into the air too slowly? According to the Airbus "A Focus on the Takeoff Rotation" document, the airplane should lift off 4-5 seconds after the pilot initiates rotation and the pitch should be about 10 deg.
Screenshot (1276)_LI

Now, that seems a little inconsistent to me that it would take 4-5 seconds to reach a pitch attitude of 10 degrees when rotating at 3 deg/s (it should take a little over 3 seconds), but maybe there is some inertia included in there to start the pitch rate after the sidestick is first moved.  I did 3 takeoffs with changing rotation rates slightly to try to hit 3 deg/s - the first one is highest, then decreasing for the next 2. The first takeoff resulted in lifting off about 3 seconds after the VR of 128 KCAS at nearly 10 deg pitch attitude. The second was 3-4 seconds after the VR of 136 KCAS at a little over 8 deg pitch attitude, and the third was 4 seconds at about 8 degrees of pitch attitude.

Seems pretty good to me.
https://1drv.ms/v/s!Ajsy7Ey4_8uFz2thzMEYuOnb--pH?e=FEBBOu
https://1drv.ms/v/s!Ajsy7Ey4_8uFz2xO9ZYbQMu2RBG9?e=P8Wl3L
https://1drv.ms/v/s!Ajsy7Ey4_8uFz23-lvyl4lsJ6KA0?e=f0XMTH

beheh:
To make sure this keeps moving, this is currently one of the oldest PRs marked as ready for review, and *technically* it would seem this is ready for QA. However as it's so outdated we should probably rebase this...?

@lukecologne could you take care of that, or otherwise mark this as WIP again? It also seems like we should pull in the latest model changes as suggested by @aguther before proceeding here.

aguther:
I honestly see no benefit of moving the table from the flight_model.cfg to the code _right now_. As it turned out the change is not as easy as it was assumed. I just had a round of 200 Autolands behind me, I will not do this again in the next time. So if the project decides this is important, I'm not the only one that can do this.

This will come up anyhow when single failing surfaces etc. should be simulated. Why not doing it then?

beheh:
Okay, then how about we just move this PR back to draft and then it's ready for us to pick up whenever we feel ready for it again.

lukecologne:
Superseded by #7516, which takes the EFCS PR into account. It also only removes the elasticity table, and doesn't change the max deflection, to reduce possible side effects.

PR #6004, opened 2021-10-12
feat: activate fmgc phase approach with LOC*/LOC engagement

aguther:
## Summary of Changes
Based on A320 type rated pilot feedback, this PR introduces a switch to FMGC approach phase when:
- plane altitude is below 9500 ft, and DECEL waypoint is reached within 1 nm, and NAV is engaged
- plane altitude is below 9500 ft, and LOC* or LOC engages

## Testing instructions
- test both above conditions


aguther:
Ok, after getting some more feedback and discussion I take this PR back. It seems to be wrong that way.

PR #6013, opened 2021-10-13
feat(athr/engines): improved CLB thrust limit for improved FLEX N1 numbers

aguther:
## Summary of Changes
This PR aims to _improve_ the N1 when performing a FLEX take-off. In parallel, this PR limits the rate the N1 thrust limits can change.

## Additional context
The N1 that is used for a FLEX TO is limited by the CLB thrust limit which means that FLEX N1 cannot be lower than CLB thrust. Up to now, the CLB thrust limit was taken 1:1 from the FCOM tables without any mach adaption. From some numbers that IRL pilots have taken about the CLB thrust limit the reduction in the lower speed area was estimated and is now used.

Important is that this PR aims to improve the overall situation but quite sure the resulting numbers are not realistic in all conditions and without more data its even quite hard to judge.

## Testing instructions
* Perform FLEX take-offs in different conditions


John-UK:
Flex has minimum and maximum values. I can’t imagine an OAT of 11 C and being able to use flex 20 C as above. If there a way to limit to acceptable values on ie > TREF & < TMAX?

aguther:
Have you tried it? I think I have already a limitation in.

marcman86:
I set the flex temp. in the MCDU. Then i set the thrust levers to the detend and looked at the engine gauges at the upper ECAM. These are the written values. The calc value is from the FLX value in the upper right corner of the upper ECAM.

donstim:
You seem to have one of the limits on FLEX temp on the low end in that if a FLEX temp less than the OAT is input, then TOGA and the related TOGA N1 are shown on the upper ECAM. Does it work that way IRL? Since a  flex temp was entered wouldn't it show FLX, but with the same N1 as TOGA?  Whatever it does there, shouldn't it also do the same thing if a FLEX temp of TREF or below is input?  The engine is flat rated to TREF (ISA+29 C), so it will put out TOGA power at any temp up to TREF.

The FCOM says that FLX will still be displayed on the upper ECAM for flex temps as low as OAT-10 C. It doesn't say what is displayed if the entered flex temp is even lower. For CEOs, if the OAT is higher than the flex temp, dashes are shown in cyan for the flex temp on the upper ECAM.

On the high temp side, FLEX TEMP is limited to ISA+55 C, but I don't know what, if any, safeguards there are IRL other than SOP to prevent entering a higher flex temp. The Honeywell FMS document says the acceptable range of values is -99 to 99, and the Smith Thales document does not show a range.

Maybe these items could be investigated for a future clean up PR so as not to delay this one any further?

PR #6021, opened 2021-10-15
perf: decrease systems variable read/write and hashing time

davidwalschots:
## Summary of Changes
Decreases time spent on reading and writing variables from the Rust system simulation code to MSFS. This resulted in one frame's system calculation time reducing from 400 microseconds to 135 microseconds on my machine. More importantly, it scales better as the size of systems code grows. Replaced `HashMap` and `HashSet` usages with `FxHashMap` and `FxHashSet` for another decrease to 112 microseconds.

Performance is improved by removing lots of `String` allocations. This is done by moving the `String` to MSFS variable mapping to the initialisation phase. Determining the target of a read/write is thereafter done by an index lookup in two `Vec`s. Thus removing the need for `HashMap<String, AircraftVariable/NamedVariable>` (very expensive) or `HashMap<VariableIdentifier, AircraftVariable/NamedVariable>` (slightly expensive).

**In summary, this PR provides a 3.57x performance increase on system simulation code.**

Thanks to @SparkyPotato for pointing out the variable read/write issue and giving their thoughts on it.

## Code review
- `systems/src/simulation/mod.rs` and `systems_wasm/src/lib.rs` contain the most important changes.
- Hydraulic test style is different from other tests. There I've done some hasty changes to make it work. In the future these tests should start using the `SimulationTestBed` throughout. I've discussed this with @crocket63. I've also flagged the `UpdateContext::new` function as deprecated, to serve as a continuous reminder.
- I've made sure to add an `InitContext` this time around, to make any future cross-cutting concern changes in the initialisation phase easier to add. Having to modify function signatures every time is a pain!

## Additional context

## Testing instructions
Just use the aircraft a bit like you always would 😀.


Gabriellscorreia0001:

Discord          : Gabriellscorreia#0001
Object of testing: #6021 
Tier of Testing  : 1
Date             : 20/10/2021

Testing Process: 
Flight SBBE SBKP 3 Hours complete, same flight made with the trial version. the fps difference is small but there was a relative difference in rederization. The screens and all systems observed throughout the flight. PR 6029 FPS average in cruise: 59 Ultra settings.
Latest trial version, FPS average at cruise 56.
Between the flight of the two versions I changed to several different climate settings, nothing affected the fps

Negatives:       
Nothing

Testing Results: 
Passed

Conclusions: 
More stable fps but the difference is not noticeable on my system (On some weaker it might be a noticeable one)

PR #6024, opened 2021-10-16
refactor(pfd): refactor PFD structure and optimize react usage

Benjozork:
## Summary of Changes

**Highly WIP. Feel free to suggest changes as this is a work-in-progress and I am looking for opinions from fellow React devs.**

This is a significant refactor of the PFD react structure that aims to achieve two outcomes:

- (primarily) improve the usage of react, through idiomatic usage of hooks, removal of `SimPlane` calls
- optimize performance through various methods noted below.

#### Removal of `SimPlane` calls

`SimPlane` is an API present in the core sim JS libraries that more often than not simply provides a SimVar call wrapped in a pretty function name.

Sim APIs are not considered public and therefore breakage from updates is likely. SimVar changes are rarer, and therefore it is preferable to use raw SimVar names instead.

Additionally, `SimPlane` is an imperative API which is not idiomatic to react's declarative methodology. One can also argue that SimVar calls are more transparent and obvious about their one-line logic.

#### Idiomatic conditional rendering

Previously, a lot of conditional rendering was achieved by switching over the SSM provided by a prop ARINC 429 value state inside a component displaying a certain part of the PFD. This causes two issues:

- Usually, the more idiomatic way of conditionally rendering from prop values is to use inline JSX expressions with ternaries or boolean AND operators.
- While that is a chicken-and-egg problem, passing in ARINC 429 words as prop values, as noted in the section about class instances below, interferes with memoization.

### Performance Optimizations

The old PFD structure and some of the patterns used caused a lot of unnecessary re-renders / DOM changes of a large number of components.

This can be boiled down to four higher level reasons:

- Lack of rounding / precision fixing of noisy prop values;
- Lack of rounding / precision fixing of noisy JSX attribute values;
- Passing of class instances that are re-created every frame down to multiple layers of components;
- (minor) Inappropriate mixing of static and dynamic JSX.

#### Lack of rounding / precision fixing of noisy prop values

Values that contain background noise generated from unstable calculations, when passed as props, will force re-renders. This by itself was not an issue since the previous structure did not use memoization, and relied on imperative updates instead. But when introducing memoization (combined with passing raw values to layers not concerned about the ARINC 429 SSM), this causes unnecessary re-renders.

#### Lack of rounding / precision fixing of noisy JSX attribute values

Position values like `x` and `y` or `transform` will not see a difference with more than one decimal place of precision at a normal scaling of 1. It is therefore wasteful to force DOM updates by passing in un-rounded noisy values.

#### Passing of class instances that are re-created every frame down to multiple layers of components

An example of this is the usage of `Arinc429Word` instances. While re-creating those every frame is a good way to force re-consideration of the SSM value, those instances should not be passed down through layers that do not care about the SSM anymore. This, even with memoization, that will cause unnecessary re-renders, since a new object instance is always a different reference than the previous one. This is completely wasted if the layer accepting this prop only uses the `value` field of the class.

#### Inappropriate mixing of static and dynamic JSX

It is fine to render static backgrounds / symbology via JSX. However, it should not be mixed with dynamic values as that JSX has no reason being run every time the dynamic part changes. Putting static JSX inside of a different, memoized, prop-less component causes that component to render exactly once for the lifetime of the display. 

#### **Measured performance impact**

Lower GC spikes and slightly higher framerate were noticed on a low-end system (Ryzen 5 2600X). Higher end systems are likely to be unaffected as their JS/GC execution will be more efficient. Regardless, this PR should not be considered an optimization PR, and a lot of the performance optimizations are side effects of structure refactoring.

## Additional context


Benjozork:
TODO: replace `Number` constructor with a `fastToFixed` function as in `autopilot-custom-fpm`.

Benjozork:
Waiting for #5692

Kimbyeoungjang:
While I import this PR into dfd a350, I had a tape issue that moving like 10 fps, and I fix it like this

I hope this helpful

`
interface VerticalTapeProps {
    displayRange: number;
    valueSpacing: number;
    distanceSpacing: number;
    graduationElementFunction: (elementHeading: number, offset: number) => JSX.Element | null;
    bugs?: [(offset: number) => JSX.Element, number][];
    tapeValue: number;
    lowerLimit?: number;
    upperLimit?: number;
    xOffsets?: number;
}

export const VerticalTape = memo<VerticalTapeProps>(({
    displayRange, valueSpacing, distanceSpacing, graduationElementFunction,
    bugs, tapeValue, lowerLimit = -Infinity, upperLimit = Infinity, xOffsets = 0,
}) => {
    const numTicks = Math.round((displayRange * 2) / valueSpacing);
    const clampedValue = Math.max(Math.min(tapeValue, upperLimit), lowerLimit);

    let lowestGraduation = Math.max(Math.round((clampedValue - displayRange) / valueSpacing) * valueSpacing, lowerLimit);

    if (lowestGraduation < tapeValue - displayRange) {
        lowestGraduation += valueSpacing;
    }

    const graduationElements = useMemo(() => {
        const graduationElements: JSX.Element[] = [];

        for (let i = 0; i < numTicks; i++) {
            const elementValue = lowestGraduation + i * valueSpacing;
            if (elementValue <= upperLimit) {
                const offset = (-elementValue * distanceSpacing) / valueSpacing;
                const element = graduationElementFunction(elementValue, offset);
                if (element) {
                    graduationElements.push(element);
                }
            }
        }

        return graduationElements;
    }, [displayRange, distanceSpacing, graduationElementFunction, lowestGraduation, upperLimit, valueSpacing]);

    const bugElements: JSX.Element[] = useMemo(() => {
        const bugElements: JSX.Element[] = [];

        bugs?.forEach((currentElement) => {
            const value = currentElement[1];
            const offset = (-value * distanceSpacing) / valueSpacing;
            bugElements.push(currentElement[0](offset));
        });

        return bugElements;
    }, [bugs, distanceSpacing, valueSpacing]);

    const yOffsets = (clampedValue * distanceSpacing) / valueSpacing + 260;

    return (
        <g transform={`translate(${xOffsets} ${yOffsets})`}>
            {graduationElements}
            {bugElements}
        </g>
    );
});

interface HorizontalTapeProps {
    displayRange: number;
    valueSpacing: number;
    distanceSpacing: number;
    graduationElementFunction: (elementHeading: number, offset: number) => JSX.Element;
    bugs?: [(offset: number) => JSX.Element, number][];
    heading: number;
    yOffset?: number;
}

export const HorizontalTape = memo<HorizontalTapeProps>(({
    displayRange, valueSpacing, distanceSpacing, graduationElementFunction,
    bugs, heading, yOffset = 0,
}) => {
    const numTicks = Math.round((displayRange * 2) / valueSpacing);

    let leftmostHeading = Math.round((heading - displayRange) / valueSpacing) * valueSpacing;
    if (leftmostHeading < heading - displayRange) {
        leftmostHeading += valueSpacing;
    }

    const graduationElements: JSX.Element[] = useMemo(() => {
        const graduationElements: JSX.Element[] = [];

        for (let i = 0; i < numTicks; i++) {
            const elementHeading = leftmostHeading + i * valueSpacing;
            const offset = (elementHeading * distanceSpacing) / valueSpacing;
            const element = graduationElementFunction(elementHeading, offset);
            if (element) {
                graduationElements.push(element);
            }
        }

        return graduationElements;
    }, [numTicks, leftmostHeading, graduationElementFunction, valueSpacing]);

    const bugElements: JSX.Element[] = useMemo(() => {
        const bugElements: JSX.Element[] = [];

        bugs?.forEach((currentElement) => {
            const angleToZero = getSmallestAngle(heading, 0);
            const smallestAngle = getSmallestAngle(currentElement[1], 0);
            let offset = currentElement[1];
            if (Math.abs(angleToZero) < 90 && Math.abs(smallestAngle) < 90) {
                if (angleToZero > 0 && smallestAngle < 0) {
                    offset = currentElement[1] - 360;
                } else if (angleToZero < 0 && smallestAngle > 0) {
                    offset = currentElement[1] + 360;
                }
            }

            offset *= distanceSpacing / valueSpacing;
            bugElements.push(currentElement[0](offset));
        });

        return bugElements;
    }, [bugs, distanceSpacing, valueSpacing]);

    const xOffset = (-heading * distanceSpacing) / valueSpacing;

    return (
        <g transform={`translate(${xOffset} ${yOffset})`}>
            {graduationElements}
            {bugElements}
        </g>
    );
});
`

PR #6069, opened 2021-10-23
feat: aevc sound and others

hotshotp:
Fixes #[issue_no]

## Summary of Changes

This PR is huge and took me a while, I hope everyone enjoys it. Here are the changes:

-The AEVC now has override configurations. It has all of the configurations modeled on interior and exterior, including correct fan and valve locations
-A more minor change - I changed up the volume curve for the engines and wind. The engines should be audible a bit later into climb and the wind should only really start being louder at around 250+ knots indicated. 
-The pack sound in the cabin on interior has two different random sounds now, not much difference between the two, but nevertheless adds some randomness and spice.

## References

many. I cannot share them publicly, but I have confirmed with a few engineers that these configurations are very accurate.

## Additional context

KNOWN ISSUES:

If you press the override buttons on and off then immediately start the aircraft, you might hear a residual sound of when the override turns off and the aevc self tests, and then the normal sound immediately after. I cannot fix this without a proper system implementation - however its not a huge issue as its an edge case scenario.

Another thing: the yellow electric pump pitch might be messed up in this PR - I did this as a baseline for #4799 as the RPM simvar is changing a bit. For now the yellow pump might be the wrong pitch.

## Testing instructions

AEVC: Start the external power - listen to the slightly different sounding regular self test, go to the galley/front passenger views and listen to how its different there with the extractor (high pitch whine). 

Usually after the self test the extractor will go to low power and stay there, unless the ambient temperature will rise above 25 degrees where it will stay at high power. Close and open cockpit door and listen to the change in the extractor whine. Now on the outside, you should be able to hear a difference when standing on the left and right of the airplane. The right side will have the outlet (extractor) and the left side will have the inlet (blower).

Set the temperature above 25 degrees and wait a few seconds. On both inside and outside you should hear a change in sound where the extractor runs at higher sleed. Set the temperature to 11 degrees and listen on the OUTSIDE - the sound of the extract should change to a more windy noise due to it now going to a partially open configuration. Set it to 8 degrees, and listen to the valve close - the fans will basically be barely audible on the outside as the valves are now in a closed configuration. Go back to 13+ degrees and you will hear it return to normal. 

Press the blower pushbutton - you should hear a quick burst of air and then a quieter overall fan sound. The extractor will run at high speed and the blower will stop. Outside you should be able to hear a valve closing and the blower fan stopping (left side of aircraft will be quiet)

Release the blower pushbutton - you should be able to hear a self test of the AEVC once again - high flow, then low flow. 

Press the extractor pushbutton - you should be able to hear a different sound of the fan ambience and the extract fan go to high. On the outside you should also hear a valve closing.

Release the extractor pushbutton - you should hear an AEVC self test again.

Press both the blower and extractor pushbuttons - you should hear a different fan ambience and the extractor fan will run at high. On the outside you should hear a very windy sound as the valves are partially open.

Release both the pushbuttons - self test should happen.

To test the new pack sounds in cabin, start the APU, and press the apu bleed and instantly go to cabin. Then turn it off, and turn if on again and go to cabin once more to check if its different.

To test new engine/wind volume curve - set the EFB volume curves to default (0 for both engine and wind, doesn't matter for exterior) - Take off and listen to how the engines sound during takeoff, climb, and late climb. Try to pay attention to how much slower and smoother they fade out.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -51,6 +51,7 @@
 1. [ECAM] Fixed misaligned items on ecam memos - @tyler58546 (tyler58546)
 1. [ECAM] Revised fuel page to match neo - @RichardPilbery (tricky_dicky#3571)
 1. [RMP] Added 25 kHz VHF spacing option - @tyler58546 (tyler58546)
+1. [MCDU] Scratchpad rewrite - @derl30n (Leon)
 
 ## 0.7.0
 
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_ATSU.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_ATSU.js
@@ -182,28 +182,6 @@ const getSimBriefOfp = (mcdu, updateView, callback) => {
         });
 };
 
-const getSimBriefUser = (value, mcdu, updateView) => {
-    if (!value) {
-        throw new Error("No SimBrief username/pilot ID provided");
-    }
-
-    SimBriefApi.getSimBriefUser(value)
-        .then(data => {
-            if (data.fetch.status === "Error: Unknown UserID") {
-                mcdu.sendDataToScratchpad(value);
-                mcdu.addNewMessage(NXFictionalMessages.noSimBriefUser);
-            } else {
-                NXDataStore.set("CONFIG_SIMBRIEF_USERID", data.fetch.userid);
-            }
-
-            updateView();
-        })
-        .catch(_err => {
-            console.log(_err.message);
-            updateView();
-        });
-};
-
 /**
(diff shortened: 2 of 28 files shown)


AviationSpotter:
Only issue I ran into is the sounds from the APU Bleed aren't working. Along with the rest of the sounds, an amazing update as usual.

AviationSpotter:
Non QA Tester Report

Object of testing: #6069
Tier of Testing: 1
Date: 23/10/2021

Testing Process:

1. Spawned C&D 
2. Tested ventilations explicitly
3. Tested APU Bleed
4. Performed circuit (at KTUS)

Negatives: N/A
Testing Results: Passed

Conclusions: Previous issue has been fixed; no further issues were found. Amazing sound update as always!

PR #6088, opened 2021-10-27
refactor: use coherent events to transfer printer data

Saschl:
Fixes #[issue_no]

## Summary of Changes

Use a Coherent Event instead of LocalVars to communicate printer data to the printer. This should make the system more robust and *maybe* stops the CTDs when trying to print an ATIS or METAR.

## References

Lots of patience

## Additional context

As the printer CTD is hard to reproduce there's no reliable way to say if this PR actually improves things. But should there be no reports of CTDs by QA Testers we can at least try and see how it behaves with a larger user base.

## Testing instructions

1. Request some METAR or ATIS reports (preferrably in intensive areas with lots of traffic)
2. Print and ensure multiple reports can be printed, switched and discarded
3. ???
4. Profit


flybywire-aircraft-a320-neo/html_ui/Pages/VLivery/Liveries/Printer/Printer.js
```diff
@@ -13,7 +13,7 @@ class LiveryPrinter extends TemplateElement {
         const url = document.getElementsByTagName("livery-printer-element")[0].getAttribute("url");
         this.index = parseInt(url.substring(url.length - 1));
         if (this.index == 0) {
-            localStorage.setItem("pages", "[]");
+            NXDataStore.set("pages", "[]");
```
Benjozork, line 16: **question:** the use of either `localStorage` or `NXDataStore` here is a bit sus. Why do we need that kind of persistence if we only access this state here, in this class ?

If it is needed, why was it changed to use `NXDataStore` ?

I'm aware this is not the scope of the PR, but since it ended up changing it, I think we could remove that if there is no use for it.

Benjozork (changes requested):

flybywire-aircraft-a320-neo/html_ui/Pages/VLivery/Liveries/Printer/Printer.js
Saschl: This is indeed very sus. The reason it was implemented is because there's effectively two instances of the printer "paper" (one one the printer itself and one one the pedestal after clicking on it). Maybe we can sync those as well via events or find a way to eliminate the need of two instances. But I'd prefer having that in a separate PR. For now I can revert to localStorage again, I mainly did it for consistency as I didn't see any other direct usage of localStorage in the code.

tracernz: sessionStorage also works, and is erased by a trip through the main menu (localStorage is not).

Benjozork: Actually, now that I think of it. Didn't we find out `xyzStorage` is very slow in Coherent ?

Saschl: I'll give sessionstorage a go. I think speed is not that much of an issue as the refresh interval of this is very low (afaik every 500 or even 1000ms)

Saschl: I've managed to remove persistence entirely.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_MainDisplay.js
```diff
@@ -27,6 +27,7 @@ class A320_Neo_CDU_MainDisplay extends FMCMainDisplay {
             Prog: 2000,
             Dyn: 1500
         };
+        this.fmgcMesssagesListener = RegisterViewListener('JS_LISTENER_SIMVARS');
```
Saschl, line 30: The name is misleading but it exists on the acfpm branch so I thought I might reduce conflicts here

tracernz (approved):
LGTM

Benjozork (approved):

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #6088
Tier of Testing: 1
Date: 9.11.2021

Testing Process: 
Various starts in different phases (gate, runway, approach)
Different airports, liveries, etc. 
METAR, ATIS - single and multiple airports - several prints
About 10+ tests.

Negatives: None - no CTD and printing worked fine. 
(but I never had a CTD before either :) )

Testing Results: Passed

oim1:

Discord: oim#0001
Object of testing: 6088
Tier of Testing: 1
Date: 11/11/2021

Testing Process: 
1. Spawn cold and dark at YMML.
2. Request 2 airport's METAR reports. 
3. Print several times.
4. Request 4 airport's METAR reports. 
5. Print several times.
6. Request the METAR report of a non-real airport.
7. Print several times. 

Negatives: N/A
Testing Results: Passed!

Conclusions: I printed several times under numerous conditions and have not received a CTD. I have received a CTD prior to this PR from the printer so this seems to have fixed it. LGTM!

FlyingCarpet7:
Also had CTDs before. Printed like 10 times now on a flight with this new update and no CTD. So far so good!

PR #6096, opened 2021-10-28
perf: LCD overlay causing excessive re-painting

Benjozork:
## Summary of Changes

This patch improves performance when the opacity filter is active on LCDs. Opacity values are fixed to two decimal places to avoid repaints every frame when the opacity is nonzero.

Frame drops are still present when moving or zooming (since that causes a lot of repaints), however, this will be addressed by the removal of our own pixel filter to instead use the native MSFS pixel filter (#6097).

## Additional context


2hwk:
Suggest closing in favour of https://github.com/flybywiresim/a32nx/pull/6110

PR #6127, opened 2021-11-02
chore: move Arinc429Word out of React context

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

To use the `Arinc429Word` class to decode/encode variables outside of React, we moved it to a `.ts` file from the `.tsx` file it was previously in. This enables us to use it in the FMGC code for example which is written in raw TypeScript

## Additional context

## Testing instructions

- Start in different contexts and observe that all the displays and systems still work correctly.


davidwalschots (approved):

src/shared/src/arinc429.ts
```diff
@@ -0,0 +1,65 @@
...
+type SignStatusMatrixRange = (typeof Arinc429Word.SignStatusMatrix)[keyof typeof Arinc429Word.SignStatusMatrix];
+
+export class Arinc429WordSsmParseError extends Error {
+    constructor(public ssm: number) {
+        super();
+    }
+}
+
+export class Arinc429Word {
+    static SignStatusMatrix = Object.freeze({
```
Benjozork, line 11: If we're gonna move this to TS, I would make this an enum

src/shared/src/arinc429.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    static u32View = new Uint32Array(Arinc429Word.f64View.buffer);
+
+    static f32View = new Float32Array(Arinc429Word.f64View.buffer);
+
+    ssm: SignStatusMatrixRange;
+
+    value: number;
+
+    constructor(word: number) {
```
davidwalschots, line 28: **suggestion**: As this is now also usable outside of a React context, consider adding the following functions (as found in `arinc429.js`) to make it easier to use:
```typescript
    static empty() {
        return new Arinc429Word(0);
    }

    static fromSimVarValue(name) {
        return new Arinc429Word(SimVar.GetSimVarValue(name, "number"));
    }
```

davidwalschots (approved):

tracernz:
Can we not just add typings for `arinc429.js` rather than duplicating the implementation?

BlueberryKing:
> Can we not just add typings for arinc429.js rather than duplicating the implementation?

This is actually a good idea, I didn't even think of that. I've never done this, so I suppose I would create a file `typings/fs-base-ui-html-ui/Pages/arinc429.d.ts` and declare types like it is done in the other files of that folder?

And does this still allow the React `.tsx` files to import this JS file with types? I guess I will play around for a bit.

davidwalschots:
This would mean we'd have to include the path in the compilation configuration. It means going outside of the regular (React) directories to a totally unrelated directory. I'd not recommend doing this. `arinc429.js` is there for legacy purposes and intended to die with the ancient non-React displays. By creating these cross-references we make it harder to get rid of it.

tracernz:
> 
> 
> This would mean we'd have to include the path in the compilation configuration. It means going outside of the regular (React) directories to a totally unrelated directory. I'd not recommend doing this. `arinc429.js` is there for legacy purposes and intended to die with the ancient non-React displays. By creating these cross-references we make it harder to get rid of it.

Ideally then make the new impl. accessible from the old code, like the `Fmgc.*` stuff and kill the .js then...? Or just have both if easier.

davidwalschots:
> Ideally then make the new impl. accessible from the old code, like the Fmgc.* stuff and kill the .js then...? Or just have both if easier.

Yes. That would also be an option. Downside would probably be that it doesn't really fit in any of the existing "packages" being built, and thus would require setting up a whole package just for one thing.

Saschl:
Non-QA associated report:

Discord          : saschl#9432
Object of testing: #6127
Tier of Testing  : 
Date             : 08/11/2021

Testing Process: 
 - Spawned in air, displays were working, the ADIRS aligned and no exceptional values were observed
 - Spawned on runway, displays were working and the ADIRS aligned as usual and no exceptional values were observed
 - Spawned C&D, turned on external power. PFD and ND showed the unaligned status, all 3 ADIRS Knobs tuned to nav, displays behave as expected after the alignment.

Negatives:       
N/A, looks good!

Testing Results: 
Passed

Conclusions: 
No changes observed :)

PR #6155, opened 2021-11-07
feat(hyd): Actuator position control

crocket63:
Fixes #[issue_no]

## Summary of Changes

Currently the position control implemented on cargo door is ON-OFF behaviour, meaning we can only require full open or full closed.

This PR aims at providing a full actuator position control loop so it can achieve fine position control for any control surface.

This is mainly to prepare the future, and cargo door is expected to be unchanged or slightly better.

- [x] Add a generic flow open loop function (position_error^3 * GAIN)
   - [x] Check that new open loop characteristics can still operate door normally
- [x] Ability to init actuator in any position
- [x] Use shared PID controller
- [x] Improved stability of damping mode
- [x] Adding an hydraulic closed loop mode with retract/extend port valves connected together alowing slow move of actuator (Gravity extension....)
- [x] Variable flow limiter depending on extension position (a lot of actuators have damping holes at start/end of travel)
- [x] Preliminary work on gear door physics

## Additional context

## Testing instructions

Cargo door operation should look realistic, same as before.

Behaviour should be normal if requiring closing while opening, or requiring opening while closing.

Crazy stuff can be tried too like disconnecting ground power unit while door is moving while at a place where there's no external power available. This should end up in a state equivalent to aileron droop mode due to no hyraulics.


src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -646,17 +815,12 @@ impl LinearActuatedRigidBodyOnHingeAxis {
         new_body.actuator_extension_gives_positive_angle =
             Self::initialize_actuator_force_direction(new_body);
```
BlueberryKing, line 816: I know this wasn't addressed in this PR, but I don't see why `initialize_actuator_force_direction` would be static. Personally, would prefer having `new_body.initialize_actuator_force_direction()`.

Furthermore, it looks like you are moving `new_body` when you are calling this function. This only works in this case, because you mark it as `#[derive(PartialEq, Clone, Copy)]` meaning you are actually creating a copy.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -679,10 +843,11 @@ impl LinearActuatedRigidBodyOnHingeAxis {
     }
 
     /// Indicates correct direction of the rigid body when an actuator would be extending or compressing.
-    /// If compressing actuator would give a rising rigid body angle, sets TRUE
+    /// If extending actuator would give a rising rigid body angle, sets TRUE
```
BlueberryKing, line 846: I don't think this change was intentional. 🙂

src/systems/a320_systems/src/hydraulic.rs
```diff
@@ -1686,9 +1697,9 @@ impl A320DoorController {
                 self.control_position_request = if self.position_requested > Ratio::new::<ratio>(0.)
                     || self.duration_in_hyd_control < Self::UP_CONTROL_TIME_BEFORE_DOWN_CONTROL
                 {
-                    Ratio::new::<ratio>(1.)
+                    Ratio::new::<ratio>(1.0)
                 } else {
-                    Ratio::new::<ratio>(0.)
+                    Ratio::new::<ratio>(-0.1)
```
BlueberryKing, line 1702: What is the reason for this being initialized at `-0.1` instead of `0`?

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -70,51 +84,87 @@ struct CoreHydraulicForce {
...
     last_control_force: Force,
 
-    force: Force,
+    force_raw: Force,
+    force_filtered: LowPassFilter<Force>,
+    max_force: Force,
+
+    pid_controller: PidController,
 }
 impl CoreHydraulicForce {
```
BlueberryKing, line 93: This piece of code does a lot. I can see that you realize this since you describe that this aims to encapsulate - what in reality - consists of many different components. In fact, it looks like you have already created this struct simply to separate some of the internal physics of the `LinearActuator` in this, which is good.

However, I believe this can be optimized a bit. Part of the reason why this might be hard to understand is that you're supporting four different modes from `LinearActuatorMode` and dynamic transitions between them. I cannot give you a concrete suggestion as to how to optimize this right now, but if we can somehow separate those four modes, that would be a step in the right direction.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -70,51 +84,87 @@ struct CoreHydraulicForce {
...
-    const DEFAULT_I_GAIN: f64 = 0.2;
-    const DEFAULT_P_GAIN: f64 = 0.05;
-    const DEFAULT_FORCE_GAIN: f64 = 200000.;
+    const OPEN_LOOP_GAIN: f64 = 1.;
+
+    const MIN_PRESSURE_TO_EXIT_POSITION_CONTROL_PSI: f64 = 500.;
+    const MIN_PRESSURE_TO_ALLOW_POSITION_CONTROL_PSI: f64 = 700.;
 
     fn new(
+        _: &mut InitContext,
```
BlueberryKing, line 100: This is not necessary. You can add the `InitContext` back in in the future when you require it.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -70,51 +84,87 @@ struct CoreHydraulicForce {
...
+        slow_hydraulic_damping_constant: f64,
         fluid_compression_spring_constant: f64,
         fluid_compression_damping_constant: f64,
         max_flow: VolumeRate,
         min_flow: VolumeRate,
         bore_side_area: Area,
         rod_side_area: Area,
+        flow_open_loop_modifier_map: [f64; 6],
+        flow_open_loop_position_breakpoints: [f64; 6],
+        flow_control_proportionnal_gain: f64,
```
BlueberryKing, line 112:
```suggestion
        flow_control_proportional_gain: f64,
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -628,7 +797,7 @@ impl LinearActuatedRigidBodyOnHingeAxis {
             control_arm_actual: control_arm,
             actuator_extension_gives_positive_angle: false,
             anchor_point,
-            position: min_angle,
+            position: init_angle,
             speed: AngularVelocity::new::<radian_per_second>(0.),
             acceleration: AngularAcceleration::new::<radian_per_second_squared>(0.),
```
BlueberryKing, line 802: While it is pretty clear to me now, it took me some time to realize that position is actually an angle. I think we could improve clarity a bit by renaming it everywhere.

I can see how `position` might be more concrete than `angle`, but maybe with `current_angle`? It would then also make sense to rename `speed` and `acceleration` to their respective rotational counterparts.

BlueberryKing (commented):
Good improvements as always! I have left some initial comments. I haven't looked at it in detail and some of my suggestions might be beyond the scope of this PR, but some of my thoughts might be worth considering when you're working on the further design.

src/systems/systems/src/shared/pid.rs
```diff
@@ -92,6 +108,19 @@ impl PidController {
         self.error_k_2 = None;
         self.error_k_1 = None;
     }
+
+    pub fn set_p_i_gain(&mut self, kp: f64, ki: f64) {
+        self.ki = ki;
+        self.kp = kp;
+    }
```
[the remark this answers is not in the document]
crocket63: Yeaaaah, using it for tuning purpose and pretty sure I'll use it for each actuator added.  But this one is simple to rewrite indeed :)

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Was initially used while creating object itself thus object was not existing at call time. Now can definitly make this a method for consistency indeed!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Oh it was! I thought it was easier to read if using extending in both case so we don't get lost reading it. I might have been wrong considering this remark :O
Logic was also reverted from what I was expecting in documented convention: positive force when extending

src/systems/a320_systems/src/hydraulic.rs
crocket63: First PR only implemented ON/OFF position control, meaning full force until position reached.
Now I use a general formula that is doing position control at requested position.  So for cargo door for example, I expect a ON/OFF behaviour, and don't want the actuator to try to exactly match close position. This is ensured by asking a position further than close position.
Will add documentation about this somewhere......

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Yeah this is a difficult part... Initially I didn't have any mode transition in there. But added the closed circuit mode, and to use it properly any user would have to know that they need to go to that mode if pressure is low.

Adding the transition here hides the required logic, and if you need to code a door controller, you don't have to know that when there's too low pressure, you'd have to switch to "closed circuit" mode.

That said, if I can find a magic formula the closed circuit mode might disapear again and everything could be handled by the closed loop control. We'll see if brain gets illuminated some day....   But user code in the controller wouldn't have to be changed though.

Modes are also really intricated as for example I need to save the last force output to start from here in the next mode to keep continuity and coherency...  Welcome in the core of the sorcery!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Added this for tuning purpose only, and thought I would get mad if I had to remove it to again cascade it again if we add new init stuff....   Will remove but if I have to add it again I'll find you!!!! :D

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Yeah I'm only rotating things so to me it's clear :D

Clearer using angular_xxx naming I hope now!

BlueberryKing (approved):

Gabriellscorreia0001:

Discord          : Gabriellscorreia#0001
Object of testing: #6155 
Tier of Testing  : 1 
Date             : 17/12/2021

Testing Process: 
Spawnada in SBCG, connected to External Power. everything works normally, when the port is moving and I withdraw the External power it keeps moving. (NOTE: batteries connected)

Negatives:       
No issues found 

Testing Results: 
tested

Conclusions: 
Works great, congrats to the team

mehoygg:

Object of testing: #6155
Tier of Testing  : 1
Date             : 17/12/21

Testing Process: 
Spawned C/D at KLAX.
First round, I started ground power and started to open the AFT Door.  Worked as it should and looked realistic. Closed as it should as well. 
I shut off EXT Power and was able to close the door and open it once more. That time, on closing, it did not shut all the way and stayed with it poking out a bit.
Second round, I spawned C/D at KLAX again. 
I opened the door, moved the A/C to an area that does not have EXT Power and closed it.  Door very slowly closed and stay open at about 1/3 of the way. 

Negatives:       
N/A

Testing Results: 
Passed.

Conclusions: 

Media:       Round 1:
Microsoft Flight Simulator Screenshot 2021 12 17 - 13 31 01 95

Round 2:

Microsoft Flight Simulator Screenshot 2021 12 17 - 14 21 20 18

FlightSimmer7700:
Quality Assurance Tester

Discord: FlightSimmer7700#5438
Object of testing: #6155 
Tier of Testing: 1
Date: 19/12/21

Testing Process: 
Loaded in C&D and tested cargo door, tried getting rid of electrical power while the door was moving and so on
Negatives: N/A
Testing Results: Passed

Conclusions: LGTM good work!

PR #6220, opened 2021-11-21
feat(cond): temperature control system

mjuhe:
## Summary of Changes

Implementation of the first pieces of the air conditioning system on Rust. The Air Conditioning system of the A320 consists of the following subsystems:
- Distribution system (Ref. 21-20-00-00)
- Pressurization Control System (Ref. 21-30-00-00) _(implemented)_
- Air Cooling system (Ref. 21-50-00-00)
- **Temperature control system (Ref. 21-60-00-00)**

This PR focuses on parts of 21-60. I tried to keep it as small as possible, and it still ended up being almost 3k lines long. However, this will set the foundation for future work on the air cond system, specially when pneumatics and bleed is implemented and the packs can be modelled. 

I've approached this model backwards, instead of following the flow of air from the engines, I've started the model from the cabin. This creates some inaccuracies at the moment, because the duct air just "magically" appears to have the right temperature, but it also allows for easier validation, and it decouples this from the engine bleed air. I've structured the code approximately like this (in blue the scope of this PR):

image

And this is how the system looks like on the aircraft:

image

## Features
- Zone control:
     - Duct air temperature reacts to selected and current temperatures as per the real aircraft
     - The cabin air warms up when packs are off and there's passengers in the cabin (the more passengers, the hotter it'll get)
     - There is heat tranfer between the cabin air and the outside (don't leave passengers inside of the cabin with packs off on a hot day!)
     - Opening doors accelerates the heat transfer between cabin and outside as it would on the real plane
- Flow control
     - Pack flow in the cabin is as per the selection on the overhead knob except in these conditions (each line has precedence over the line before);
     - Pack flow is minimum NORM if any of the packs is in start conditon (30s from start)
     - Pack flow is HI if APU bleed is on
     - Pack flow is HI if only one pack is on (of course the total flow will still be lower than with both packs on)
     - Pack flow is LO if the aircraft is taking off or landing
- Flow control valve
     - Flow control valve opens if there's air pressure differential (engine on or APU bleed on) and (all need to be valid):
     - The corresponding pack pushbutton is on
     - The corresponding engine is not in start mode
     - The opposite engine is not in start mode OR crossbleed is not on
     - The corresponding engine is not on fire
     - The overhead `Ditching` pushbutton is not on

## References

For aircraft references I used FCOM, AMM, Air Cond Requirement Documents and a bunch of irl photos and pilot feedback... If you find anything innacurate feel free to let me know.
For the heat transfer equations and data, I mainly used:
    Holman, J. P. 1986. _Heat Transfer_. 6th Edition. McGraw-Hill.
    Long, C.; Sayma, N. 2009. _Heat Transfer_. Ventus Publishing.
    Cengel, Y; Boles M.; Kanoglu, M. 2019. _Thermodynamics: an engineering approach_. McGraw-Hill.

## Additional context

Known issues / future work:
- <del>There is no integration yet with the engine bleed system, so the logic is basically based on engines being on or off at the moment<del> There is very basic integration with the engine bleed system.
- The heat transfer into and outside of the cabin only simulates conduction and convection. Radiaton has been left out due to not having the ability of measuring sun intensity (or even if it's day and night). Therefore, only the outside temperature has an effect on the variation of inside temperature, not rain, clouds or sunshine.
- One of the characteristic features of the air conditioning system is that if the packs can't supply the required temperature, they would send a signal to the FADEC to raise the idle rpm of the engines. This has not yet been modelled.
- No faults or overheating
- No connection to the electrical system
- The pressurization system is still not integrated with this. Future work will be to integrate it all together, to have a single simulation of the cabin air.

## Testing instructions

1. Start cold and dark. As you start switching on the APU and the engines, validate the flow control valve behaviour, as above.
2. When the packs are on, validate the flow control behaviour, as above.
3. Spawn in different parts of the world (or play around with the weather), load passengers, change the temperature on the overhead, open and close doors... Validate the duct temperatures stay in range and behave as expected. Validate after a few minutes (warning, it can take quite long!) the cabin temperature reaches the desired temperature.


davidwalschots (commented):
Just started with the review, still a lot of files to go through. 😀

flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/sound/sound.xml
```diff
@@ -514,48 +514,48 @@
...
+        <Sound WwiseData="true" WwiseEvent="pack2ext" NodeName="SOUND_PACK_OUTFLOW_RIGHT" FadeOutType="2" FadeOutTime="1" LocalVar="A32NX_COND_PACK_FLOW_VALVE_1_IS_OPEN" Units="BOOL" Index="0">
             <Range LowerBound="1.0" />
             <WwiseRTPC SimVar="TURB ENG N1" Units="percent" Index="1" RTPCName="SIMVAR_TURB_ENG_N1" />
             <WwiseRTPC SimVar="TURB ENG N1" Units="percent" Index="2" RTPCName="SIMVAR_TURB_ENG_N1" />
             <WwiseRTPC SimVar="AMBIENT TEMPERATURE" Units="celcius" Index="0" RTPCName="SIMVAR_AMBIENT_TEMPERATURE"/>
             <WwiseRTPC LocalVar="A32NX_SOUND_EXTERIOR_MASTER" Units="number" Index="0" RTPCName="LOCALVAR_A32NX_EXTERIOR_VOLUME" />
         </Sound>
 
-        <Sound WwiseData="true" Continuous="false" WwiseEvent="pack2extshut" NodeName="SOUND_PACK_OUTFLOW_LEFT" LocalVar="A32NX_PACKS_1_IS_SUPPLYING" Units="BOOL" Index="0">
+        <Sound WwiseData="true" Continuous="false" WwiseEvent="pack2extshut" NodeName="SOUND_PACK_OUTFLOW_LEFT" LocalVar="A32NX_COND_PACK_FLOW_VALVE_1_IS_OPEN" Units="BOOL" Index="0">
```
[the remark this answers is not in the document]
hotshotp: Can't the pack valve be open without the pack running? I can discuss this later as it's a bit more complex.

BlueberryKing:
Nice work! Some of the variable changes will break #5384. Assigning myself here as I will want to have a look through it before this is merged 😉.

src/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn new(context: &mut InitContext, number: usize) -> Self {
+        Self {
+            pack_flow_valve_id: context.get_identifier(Self::pack_flow_valve_id(&number)),
+            number,
+            is_open: false,
+            timer_open: Duration::from_secs(0),
+        }
+    }
+
+    fn pack_flow_valve_id(number: &usize) -> String {
```
davidwalschots, line 171:
```suggestion
    fn pack_flow_valve_id(number: usize) -> String {
```

src/systems/systems/src/shared/mod.rs
```diff
@@ -83,6 +83,11 @@ pub trait EngineUncorrectedN2 {
     fn uncorrected_n2(&self) -> Ratio;
 }
 
+pub trait CabinAltitude {
+    fn cabin_altitude(&self) -> Length;
+    fn cabin_pressure(&self) -> Pressure;
+}
```
davidwalschots, line 89: **thought**: I'm wondering if we can improve this name a bit. `CabinAltitude` with `cabin_altitude` and `cabin_pressure` has a few (minor) problems:
- "cabin" is repeated a lot.
- pressure is not initially expected to be in this type.

**suggestion**: Maybe the following would be a better option? Thoughts?

```suggestion
pub trait Cabin {
    fn altitude(&self) -> Length;
    fn pressure(&self) -> Pressure;
}
```

src/systems/systems/src/shared/pid.rs
```diff
@@ -39,6 +40,14 @@ impl PidController {
         self.setpoint = new_setpoint;
     }
 
+    pub fn change_min_output(&mut self, new_min: f64) {
+        self.min_output = new_min;
+    }
+
+    pub fn change_max_output(&mut self, new_max: f64) {
+        self.max_output = new_max;
+    }
```
davidwalschots, line 49: **thought**: Should these perhaps be named `set_...` instead of `change_...` to be in line with common "getter/setter" patterns?

davidwalschots (commented):
The second part of my review. There will be a third part later this week.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    const UPPER_DUCT_TEMP_LIMIT_HIGH: f64 = 343.15; // K
+    const LOWER_DUCT_TEMP_LIMIT_LOW: f64 = 275.15; // K
+    const LOWER_DUCT_TEMP_LIMIT_HIGH: f64 = 281.15; // K
+    const KI_DUCT_DEMAND_CABIN: f64 = 0.05;
+    const KI_DUCT_DEMAND_COCKPIT: f64 = 0.04;
+    const KP_DUCT_DEMAND_CABIN: f64 = 3.5;
+    const KP_DUCT_DEMAND_COCKPIT: f64 = 2.;
+
+    fn new(context: &mut InitContext, zone_id: &'static str) -> Self {
+        let pid_controller = if zone_id == "CKPT" {
```
davidwalschots, line 360: **suggestion**: I'm not sure to what extent you'll apply the index-based suggestion I made earlier. Here you could also consider using an enum instead. Something like:

```rust
let pid_controller = match zone_type {
    ZoneType::Cockpit => { ... },
    ZoneType::Cabin(_) => { ... }
}
```

You could then add an implementation for `Display` to the `ZoneType` enum for the `zone_temp_id` and other assignments. See `ElectricalBusType` for an example.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    zone_measured_temperature: ThermodynamicTemperature,
+    pid_controller: PidController,
+}
+
+impl ZoneController {
+    const K_ALTITUDE_CORRECTION: f64 = 0.0000375; // deg/feet
+    const UPPER_DUCT_TEMP_LIMIT_LOW: f64 = 323.15; // K
+    const UPPER_DUCT_TEMP_LIMIT_HIGH: f64 = 343.15; // K
+    const LOWER_DUCT_TEMP_LIMIT_LOW: f64 = 275.15; // K
+    const LOWER_DUCT_TEMP_LIMIT_HIGH: f64 = 281.15; // K
```
davidwalschots, line 353: **suggestion**: I'd include the unit in the variable name.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    fn calculate_duct_temp_upper_limit(&self) -> ThermodynamicTemperature {
+        if self.zone_measured_temperature > ThermodynamicTemperature::new::<degree_celsius>(19.) {
+            ThermodynamicTemperature::new::<kelvin>(Self::UPPER_DUCT_TEMP_LIMIT_LOW)
+        } else if self.zone_measured_temperature
+            < ThermodynamicTemperature::new::<degree_celsius>(17.)
+        {
+            ThermodynamicTemperature::new::<kelvin>(Self::UPPER_DUCT_TEMP_LIMIT_HIGH)
+        } else {
+            let interpolation =
+                (Self::UPPER_DUCT_TEMP_LIMIT_LOW - Self::UPPER_DUCT_TEMP_LIMIT_HIGH) / (19. - 17.)
```
davidwalschots, line 450: **thought**: It seems 17 and 19 have some significance in this code. Maybe a local const would be useful?

As a general remark with magic numbers used in this code (and the functions around it): try to give them a name where it isn't immediately obvious what the number represents.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    }
+
+    fn write(&self, writer: &mut SimulatorWriter) {
+        // TODO: Replace this with actual duct temperature when mixer is modelled, not duct demand temperature
+        writer.write(&self.zone_duct_temp_id, self.duct_demand_temperature);
+    }
+}
+
+#[derive(Clone, Copy)]
+enum OvhdFlowSelector {
```
davidwalschots, line 498:
```suggestion
enum OverheadFlowSelector {
```

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    fcv_2_open_allowed: bool,
+    should_open_fcv: [bool; 2],
+
+    apu_bleed_on: bool,
+    crossbleed_is_on: bool,
+    ditching_is_on: bool,
+    engine_1_in_start_mode: bool,
+    engine_2_in_start_mode: bool,
+    engine_1_on_fire: bool,
+    engine_2_on_fire: bool,
```
davidwalschots, line 542: **thought**: I'm not quite sure if the current logic accurately represents the engine being on fire.
**thought**: Maybe you care about the engine fire pb being released more than a fire indication?

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    const FLOW_REDUCTION_LIMIT: f64 = 80.;
+    const BACKFLOW_LIMIT: f64 = 80.;
+
+    const FLOW_CONSTANT_C: f64 = 0.5675; // kg/s
+    const FLOW_CONSTANT_XCAB: f64 = 0.00001828; // kg(feet*s)
+
+    fn new(context: &mut InitContext) -> Self {
+        Self {
+            apu_bleed_valve_open_id: context.get_identifier("APU_BLEED_AIR_VALVE_OPEN".to_owned()),
+            apu_rpm_id: context.get_identifier("APU_N_RAW".to_owned()),
```
davidwalschots, line 563: **issue**: Data which is calculated within Rust shouldn't be read through LVars. Instead, pass the necessary information down through `update` calls.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+            crossbleed_id: context.get_identifier("KNOB_OVHD_AIRCOND_XBLEED_Position".to_owned()),
+            ditching_id: context.get_identifier("OVHD_PRESS_DITCHING_PB_IS_ON".to_owned()),
```
davidwalschots, line 565: **thought**: These should probably be contained in one or more Overhead types. Same goes for `ovhd_flow_selector_id`.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        engines: [&impl EngineCorrectedN1; 2],
+        pressurization: &impl CabinAltitude,
+        pack_flow_valve: &[PackFlowValve; 2],
+    ) {
+        self.flow_control_valve_open = pack_flow_valve.iter().any(|fcv| fcv.fcv_is_open());
+        self.single_pack_operation =
+            pack_flow_valve[0].fcv_is_open() != pack_flow_valve[1].fcv_is_open();
+        self.pack_in_start_condition = self.pack_start_condition_determination(pack_flow_valve);
+        // TODO: Add overheat protection
+        self.flow_demand = self.flow_demand_determination(aircraft_state);
```
davidwalschots, line 608: **suggestion**: `pack_in_start_condition` can be passed as an argument, instead of being a field on the struct.
```suggestion
        let pack_in_start_condition = self.pack_start_condition_determination(pack_flow_valve);
        // TODO: Add overheat protection
        self.flow_demand = self.flow_demand_determination(aircraft_state, pack_in_start_condition);
```

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        if self.flow_control_valve_open {
+            if self.single_pack_operation {
+                self.absolute_flow
+            } else {
+                self.absolute_flow * 2.
+            }
+        } else {
+            MassRate::new::<kilogram_per_second>(0.)
+        }
+    }
```
davidwalschots, line 704: **suggestion**: Consider computing `pack_flow` once in the `update` function and storing it in a field. That way you can remove the `self.flow_control_valve_open`, `self.single_pack_operation`, and `self.absolute_flow` fields.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        self.absolute_flow = self.absolute_flow_calculation(pressurization);
+        self.fcv_open_allowed_determination(acs_overhead);
+        self.should_open_fcv = self.should_open_fcv_determination(engines);
+    }
+
+    fn pack_start_condition_determination(&mut self, pack_flow_valve: &[PackFlowValve; 2]) -> bool {
+        // Returns true when one of the packs is in start condition
+        pack_flow_valve
+            .iter()
+            .any(|fcv| fcv.fcv_timer() <= Duration::from_secs_f64(Self::PACK_START_TIME))
```
davidwalschots, line 618:
```suggestion
            .any(|fcv| fcv.fcv_timer() <= Duration::from_secs_f64(Self::PACK_START_TIME_SECONDS))
```

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    }
+
+    fn flow_demand_determination(&self, aircraft_state: &AcStateManager) -> Ratio {
+        let mut intermediate_flow =
+            Ratio::new::<percent>((self.flow_selector_position as u32) as f64);
```
davidwalschots, line 623: **suggestion**: Add a `From` implementation:

```rust
impl From<OvhdFlowSelector> for Ratio {
    fn from(value: OvhdFlowSelector) -> Self {
        Ratio::new::<percent>((value as u8) as f64)
    }
}
```
and then change this code to:
```suggestion
        let mut intermediate_flow: Ratio = self.flow_selector_position.into();
```

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+
+    fn write(&self, writer: &mut SimulatorWriter) {
+        // If both flow control valves are closed, the flow indication is in the Lo position
+        if self.should_open_fcv.iter().any(|&x| x) {
+            writer.write(&self.pack_flow_id, self.flow_demand);
+        } else {
+            writer.write(
+                &self.pack_flow_id,
+                Ratio::new::<percent>((OvhdFlowSelector::Lo as u32) as f64),
+            );
```
davidwalschots, line 742: **suggestion**: Use the `From` implementation suggested earlier.

```suggestion
            let flow: Ratio = OvhdFlowSelector::Lo.into();
            writer.write(&self.pack_flow_id, flow);
```

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+            && !self.engine_1_on_fire
+            && !self.ditching_is_on;
+        // && ! pack 1 overheat
+        // Flow Control Valve 2
+        self.fcv_2_open_allowed = acs_overhead.pack_2_pb_is_on()
+            && !self.engine_2_in_start_mode
+            && (!self.engine_1_in_start_mode || !self.crossbleed_is_on)
+            && !self.engine_2_on_fire
+            && !self.ditching_is_on;
+        // && ! pack 2 overheat
```
davidwalschots, line 673: **thought**: This is declared in the `systems` crate and thus should also work for other aircraft types (e.g. A380 with 4 engines). Is there a way to abstract this a bit?

davidwalschots (changes requested):
Lots of cool features! 🚀

BlueberryKing:
Incredible work! You really did a good job capturing the complexities of the air conditioning system while maintaining concise and readable code.

mjuhe:
Thanks a lot @davidwalschots for the in depth review! I've just gone through all comments and will be starting implementing them soon. I see a few common themes; this is super helpful for me to avoid making the same mistakes in the future and continue learning cool tricks.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+    passengers: usize,
+    true_airspeed: Velocity,
+    fwd_door_is_open: bool,
+    rear_door_is_open: bool,
+}
+
+impl CabinZone {
+    const TRUE_AIRSPEED: &'static str = "AIRSPEED TRUE";
+    const FWD_DOOR: &'static str = "INTERACTIVE POINT OPEN:0";
+    const READ_DOOR: &'static str = "INTERACTIVE POINT OPEN:3";
```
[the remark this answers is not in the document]
mjuhe: Indeed a typo, good spot!

src/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+}
+
+impl SimulationElement for PackFlowValve {
+    fn write(&self, writer: &mut SimulatorWriter) {
+        writer.write(&self.pack_flow_valve_id, self.is_open);
+    }
+}
+
+pub struct Air {
```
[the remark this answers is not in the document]
mjuhe: Yes, the idea is that this will be the elemental type that will be used to simulate the air moving through the packs and mixing unit, so more elements of the Air Conditioning system will use it in the future.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+    ) {
+        if !self.is_initialised {
+            self.internal_air
+                .set_temperature(context.ambient_temperature());
+            self.flow_out.set_temperature(context.ambient_temperature());
+            self.is_initialised = true;
+        }
+        self.internal_air
+            .set_pressure(pressurization.cabin_pressure());
+        let number_of_open_doors: usize = fwd_door_is_open as usize + rear_door_is_open as usize;
```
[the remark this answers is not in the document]
mjuhe: Replaced all `usize`s for `u8`s except in array indices or iteration counts in the testing module.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+struct ZoneAir {
+    flow_out: Air,
+    internal_air: Air,
+    is_initialised: bool,
+}
+
+impl ZoneAir {
+    const PASSENGER_HEAT_RELEASE: f64 = 0.1; // kW - from Thermodynamics: An Engineering Approach (Cengel and Boles)
+    const A320_CABIN_DIAMETER: f64 = 4.14; // m
+    const FLOW_RATE_THROUGH_OPEN_DOOR: f64 = 0.6; // kg/s
```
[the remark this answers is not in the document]
mjuhe: Added units to const names except for very verbose units (like kJ/kg*K or m2/s2/K).

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+            flow_out: Air::new(),
+            internal_air: Air::new(),
+            is_initialised: false,
+        }
+    }
+
+    fn update(
+        &mut self,
+        context: &UpdateContext,
+        flow_in: &Air,
```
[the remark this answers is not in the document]
mjuhe: Changed!

src/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        self.acsc.update(
+            context,
+            &self.acs_overhead,
+            &self.pack_flow_valve,
+            engines,
+            pressurization,
+            lgciu,
+        );
+
+        for fcv in self.pack_flow_valve.iter_mut() {
```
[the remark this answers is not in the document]
mjuhe: That is me mixing naming conventions in my head! `fcv` stands for Flow Control Valve, which is the same thing as the `pack_flow_valve`. Changed now to `pack_fv` to avoid confusion.

src/systems/systems/src/shared/pid.rs
mjuhe: Agreed!

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: As per comment above, implemented.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+                pressurization,
+            )
+        }
+        self.pack_flow_controller.update(
+            &self.aircraft_state,
+            acs_overhead,
+            engines,
+            pressurization,
+            pack_flow_valve,
+        );
```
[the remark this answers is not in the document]
mjuhe: Not intended 😄 fixed now.

src/systems/a320_systems/src/lib.rs
```diff
@@ -133,6 +137,12 @@ impl Aircraft for A320 {
     }
 
     fn update_after_power_distribution(&mut self, context: &UpdateContext) {
+        self.air_conditioning.update(
+            context,
+            [&self.engine_1, &self.engine_2],
+            &self.pressurization,
+            [&self.lgciu1, &self.lgciu2],
+        );
```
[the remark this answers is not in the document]
mjuhe: Not intentional 😄 fixed now.

src/systems/a320_systems/src/air_conditioning.rs
```diff
@@ -0,0 +1,140 @@
...
+                    Volume::new::<cubic_meter>(Self::A320_CABIN_VOLUME / 2.),
+                    0,
+                ),
+                CabinZone::new(
+                    context,
+                    "AFT",
+                    Volume::new::<cubic_meter>(Self::A320_CABIN_VOLUME / 2.),
+                    0,
+                ),
+            ],
```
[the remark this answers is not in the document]
mjuhe: Really cool idea. I've implemented this and moved the reading of the seat numbering to the `CabinZone` type. I've made the ROWS const a `usize` however, the compiler didn't seem to like `u8`.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+    zone_air: ZoneAir,
+    zone_volume: Volume,
+    passengers: usize,
+    true_airspeed: Velocity,
+    fwd_door_is_open: bool,
+    rear_door_is_open: bool,
+}
+
+impl CabinZone {
+    const TRUE_AIRSPEED: &'static str = "AIRSPEED TRUE";
```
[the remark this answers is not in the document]
mjuhe: Done :)

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+        // Temperature differential should be based on the wall temperature instead of the inside vs outside. This is a simplification
+        let temperature_differential: f64 = self.internal_air.temperature().get::<kelvin>()
+            - context.ambient_temperature().get::<kelvin>(); // Kelvin
+
+        // Convection coefficient for horizontal cylinder in air
+        let convection_coefficient: f64 =
+            Self::CONVECTION_COEFFICIENT_CONSTANT_FOR_NATURAL_CONVECTION
+                * (temperature_differential.abs() / Self::A320_CABIN_DIAMETER).powf(1. / 4.);
+        convection_coefficient
+    }
```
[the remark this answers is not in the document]
mjuhe: Unfortunately, addition and subtraction are not implemented for the `ThermodynamicTemperature` type. I'm assuming this is because of the different temperature scales, for example between Celsius and Fahrenheit, where subtracting in one unit and then converting is not the same as converting first and then subtracting.

https://docs.rs/uom/0.30.0/uom/si/thermodynamic_temperature/index.html

src/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn should_open_fcv(&self) -> [bool; 2];
+}
+
+pub trait PackFlow {
+    fn pack_flow(&self) -> MassRate;
+}
+
+pub struct AirConditioningSystem {
+    acs_overhead: AirConditioningSystemOverhead,
+    acsc: ACSController,
```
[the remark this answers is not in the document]
mjuhe: Done.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+use uom::si::{
+    f64::*,
+    length::foot,
+    mass_rate::kilogram_per_second,
+    ratio::{percent, ratio},
+    thermodynamic_temperature::{degree_celsius, kelvin},
+    velocity::knot,
+};
+
+pub(super) struct ACSController {
```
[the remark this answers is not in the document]
mjuhe: Duplicate from above, done :).

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+        accept_iterable!(self.zone_controller, visitor);
+        self.pack_flow_controller.accept(visitor);
+
+        visitor.visit(self);
+    }
+}
+
+#[derive(Copy, Clone)]
+enum AcStateManager {
```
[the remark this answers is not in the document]
mjuhe: Renamed.

src/systems/systems/src/shared/pid.rs
```diff
@@ -4,6 +4,7 @@ use std::time::Duration;
 /// Implementation in a recursive form
 /// u(k+1) = u(k) + (e(k) - e(k-1)) * Kp + e(k) * Ki * dt + (e(k) - 2 * e(k-1) + e(k-2)) * Kd / dt
 /// For variable dt add dt duration as argument, if fixed dt or scheduled synchronously with controlled system input None as dt
+#[derive(Clone)]
```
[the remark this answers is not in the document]
mjuhe: I honestly have no idea how I thought I needed this before but actually I didn't 😕. Oh well, thanks for spotting it!

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        }
+    }
+
+    fn step(
+        self: AcState<Initialisation>,
+        lgciu: [&impl LgciuWeightOnWheels; 2],
+    ) -> AcStateManager {
+        if lgciu
+            .iter()
+            .all(|&a| a.left_and_right_gear_compressed(true))
```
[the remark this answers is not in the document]
mjuhe: Removed `&`! Not sure how it ended up in there in the first place 😄 .

src/systems/systems/src/shared/mod.rs
mjuhe: I followed your suggestion and renamed the trait to `Cabin`.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Added! This is cool.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Done! This simplified the code but maybe it made it a bit less readible.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Yes good spot, instead of keeping it in the `update` function I've moved it inside of `flow_demand_determination` altogether.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        {
+            AcStateManager::InFlight(self.into())
+        } else if engines
+            .iter()
+            .all(|&x| x.corrected_n1() > Ratio::new::<percent>(70.))
+            && lgciu
+                .iter()
+                .all(|&a| a.left_and_right_gear_compressed(true))
```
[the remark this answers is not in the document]
mjuhe: Good idea, done!

src/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        for id in cabin_zone_ids.iter() {
+            let knob_id = format!("COND_{}_SELECTOR", id);
+            temperature_selectors.insert(id, ValueKnob::new_with_value(context, &knob_id, 24.));
+        }
+        Self {
+            pack_1_pb: OnOffFaultPushButton::new_on(context, "COND_PACK_1"),
+            pack_2_pb: OnOffFaultPushButton::new_on(context, "COND_PACK_2"),
+            temperature_selectors,
+        }
+    }
```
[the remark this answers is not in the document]
mjuhe: This is a good point, I didn't think about the speed of the `Hashmap`s. I've now refactored this and a couple of other instances where I was using `Hashmap` to use const generics, vectors and enums instead.

In this particular instance, I failed to use your suggestion because I couldn't find a way to implement Default for `ValueKnob` type (even if the default uses `None`, the type needs to implement `Copy`). I've used a vector instead initialised with a const generic array.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+            engines,
+            pressurization,
+            pack_flow_valve,
+        );
+    }
+}
+
+impl DuctTemperature for ACSController {
+    fn duct_demand_temperature(&self) -> HashMap<&'static str, ThermodynamicTemperature> {
+        let mut duct_temperature: HashMap<&str, ThermodynamicTemperature> = HashMap::new();
```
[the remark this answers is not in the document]
mjuhe: As per the answer a few comments above, I've changed this to use `Vec` instead.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: As per one of the other comments above. I've now replaced all `Hashmap`s for vectors, const generic arrays and enums. I like it because it makes the code much more "adaptable" to other types of aircraft that might come in the future.

In this particular instance I've implemented your suggestion, but I failed to extract the `Display` implementation from the `systems` module. This is the only bit of code specific to the A320 still living in systems. Do you have any idea on how I can get this out to the `a320_systems` module?

davidwalschots: Maybe by including the name into the enum? E.g. `ZoneType::Cabin(1, "FWD".to_owned())`. Then in pattern matching most of the time you can just ignore the string (`ZoneType::Cabin(number, _)`), unless you need it.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+        test_bed.set_passengers(0);
+        test_bed.set_flow_in_flow_rate(MassRate::new::<kilogram_per_second>(0.));
+        test_bed.run();
+        test_bed.set_ambient_temperature(ThermodynamicTemperature::new::<degree_celsius>(0.));
+        let initial_temp = test_bed.cabin_temperature();
+        test_bed = test_bed.iterate_with_delta(100, Duration::from_secs(10));
+        let first_temperature_differential = initial_temp.get::<degree_celsius>()
+            - test_bed.cabin_temperature().get::<degree_celsius>();
+
+        let mut test_bed2 = CabinZoneTestBed::new();
```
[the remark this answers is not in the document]
mjuhe: Changed.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+        test_bed2.set_flow_in_flow_rate(MassRate::new::<kilogram_per_second>(0.));
+        test_bed2.set_ambient_temperature(ThermodynamicTemperature::new::<degree_celsius>(24.));
+        test_bed2.run();
+        test_bed2.set_ambient_temperature(ThermodynamicTemperature::new::<degree_celsius>(0.));
+        test_bed2.set_true_airspeed(Velocity::new::<meter_per_second>(130.));
+        test_bed2.set_ambient_pressure(Pressure::new::<pascal>(45000.));
+        let initial_temp2 = test_bed2.cabin_temperature();
+        test_bed2 = test_bed2.iterate_with_delta(100, Duration::from_secs(10));
+        let second_temperature_differential = initial_temp2.get::<degree_celsius>()
+            - test_bed2.cabin_temperature().get::<degree_celsius>();
```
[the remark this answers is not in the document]
mjuhe: I've overhauled the testing suite in `cabin_air` and tweaked the one in `acs_controller`. I've also implemented the idea of "storing" a temperature in the `TestBed`, it's a really good idea thanks!

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Eliminated magic numbers.

mjuhe: Eliminated magic numbers.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+    fn pack_flow(&self) -> MassRate {
+        self.pack_flow_controller.pack_flow()
+    }
+}
+
+impl FlowControlValveSignal for ACSController {
+    fn should_open_fcv(&self) -> [bool; 2] {
+        self.pack_flow_controller.should_open_fcv()
+    }
+}
```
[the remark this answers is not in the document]
mjuhe: I've now wrapped this in `ControllerSignal<S>`. I wanted to integrate/merge it with the `PackFlowValveController` in the Bleed system, but after spending a good amount of time looking at it I didn't find an easy straight forward way to do this within this PR without increasing the scope considerably. 

At the moment both systems work independently from each other. I think we can leave this for a future PR since the packs also need to be modelled at some point. What do you think @davidwalschots and @BlueberryKing ?

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: I changed this to pull from the overhead `EngineFirePushButtons` instead.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Absolutely true. I've replaced all "reads" within `AirConditioningSystemController` by pulling data through the `update` functions within it.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: As above, I've now removed this and pulled them directly from their respective overhead types. I've moved `ovhd_flow_selector` to `AirConditioningSystemOverhead`

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: I've changed this slightly to read "right" and "left" engines instead of 1 and 2. The A380 also has 2 packs, so we should be able to use the same code just by passing the state of the corresponding engines (i.e. `pneumatic.left_engine_state() == EngineState::Starting` means engine 1 on the A320, and will mean any of engines 1 or 2 in the A380).

Having said this, I don't know enough about the A380 yet to promise this won't need any tweaks.

flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/sound/sound.xml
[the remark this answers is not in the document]
mjuhe: @davidwalschots yes, looking at it now I think you are right. I've changed the `pack2` named events to depend on `A32NX_COND_PACK_FLOW_VALVE_2_IS_OPEN`. Only pack 1 was used previously though.

@hotshotp as far as I know, no. The valve opening and the pack running are basically the same thing unless there is a fault of course. Note that the packs are not fully modelled yet.

src/systems/systems/src/air_conditioning/cabin_air.rs
```diff
@@ -0,0 +1,697 @@
...
+        fwd_door_is_open: bool,
+        rear_door_is_open: bool,
+        pressurization: &impl CabinAltitude,
+    ) {
+        if !self.is_initialised {
+            self.internal_air
+                .set_temperature(context.ambient_temperature());
+            self.flow_out.set_temperature(context.ambient_temperature());
+            self.is_initialised = true;
+        }
```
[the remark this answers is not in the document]
mjuhe: That would be neat, but I'd rather leave it for a different PR to not mix too many things in this one 😄.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+impl AcState<BeginTakeOff> {
+    fn step(
+        self: AcState<BeginTakeOff>,
+        context: &UpdateContext,
+        engines: [&impl EngineCorrectedN1; 2],
+    ) -> AcStateManager {
+        if (engines
+            .iter()
+            .all(|&x| x.corrected_n1() > Ratio::new::<percent>(70.))
+            && context.indicated_airspeed().get::<knot>() > 70.)
```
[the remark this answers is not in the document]
mjuhe: This is a good spot. After doing some more digging through the documentation I found this actually depends on the ground speed, not IAS or CAS.

I've extracted the ground speed from the `Adirs` through a trait, and passed it to the `acs_controller`. 

CMM 21-63 states this uses data from ADIRU 1.

image

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,1808 @@
...
+        lgciu: [&impl LgciuWeightOnWheels; 2],
+    ) -> AcStateManager {
+        if !lgciu
+            .iter()
+            .all(|&a| a.left_and_right_gear_compressed(true))
+            || self.timer > Duration::from_secs(10)
+        {
+            AcStateManager::InFlight(self.into())
+        } else {
+            AcStateManager::EndTakeOff(self.increase_timer(context))
```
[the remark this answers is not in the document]
mjuhe: This is absolutely correct 😄 . Using the same reference as above:

image

src/systems/systems/src/air_conditioning/acs_controller.rs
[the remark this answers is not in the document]
BlueberryKing: My `PackFlowValveController` is more of a placeholder than anything else. Feel free to remove it and emit the signal from the ACSC instead. It's also perfectly fine to leave the systems entirely separate for now.

src/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -82,6 +82,7 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
             .provides_aircraft_variable("ACCELERATION BODY Y", "feet per second squared", 0)?
             .provides_aircraft_variable("ACCELERATION BODY Z", "feet per second squared", 0)?
             .provides_aircraft_variable("AIRSPEED INDICATED", "Knots", 0)?
+            .provides_aircraft_variable("AIRSPEED TRUE", "Knots", 0)?
```
beheh, line 85: This line is redundant, it's repeated two lines below.

mjuhe: You are right, my bad. Fixed now!

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -288,6 +288,22 @@ impl A320Pneumatic {
...
+    }
+}
+impl EngineStartState for A320Pneumatic {
+    fn left_engine_state(&self) -> EngineState {
+        self.fadec.engine_state(1)
+    }
+    fn right_engine_state(&self) -> EngineState {
+        self.fadec.engine_state(2)
+    }
+}
```
BlueberryKing, line 306: I might have brought this upon myself, but I am not sure if we should be getting the engine state from the FADEC via the bleed system. Ideally, the FADEC should be independent (perhaps created in `A320` and passed into the pneumatic and the air conditioning system. This might not be within the scope of this PR though.

src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -0,0 +1,2054 @@
...
+        // && ! pack 2 overheat
+    }
+
+    fn should_open_fcv_determination(
+        &self,
+        engines: [&impl EngineCorrectedN1; 2],
+        pneumatic: &(impl PneumaticBleed + EngineStartState),
+        pneumatic_overhead: [bool; 2],
+    ) -> [bool; 2] {
+        // Pneumatic overhead represents engine bleed pushbutton for left and right engine(s)
```
BlueberryKing, line 690: Is there a reason you chose to pass down `pneumatic_overhead` as an array of booleans instead of the actual overhead panel type or a trait implementing the bleed pushbutton states?

BlueberryKing (commented):
LGTM! Just two questions :)

src/systems/systems/src/air_conditioning/acs_controller.rs
davidwalschots: Twice even! 😀

davidwalschots (commented):
Once again: very nice work! I'll approve this one after @BlueberryKing's questions and remarks are completed.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Yup, mainly because this is in the `systems` crate and the overhead panel type is A320 specific. My reasoning is that the A380 (and I believe any other aircraft type) also has 2 packs but of course with 4 engines the overhead type will be different. By passing booleans we can extract the logic at a higher level, since the system itself is only concerned with whether the pack is receiving bleed air from the engines or not.

Of course when the packs are fully modelled this logic will move from here to the packs themselves.

mjuhe: And... as soon as I said this I realised I could get the same result with a trait! Forget what I said, implementing a trait instead.

src/systems/a320_systems/src/pneumatic.rs
mjuhe: I think you are right, we already have a `LeapEngine` module within `systems` that we will most likely have to extract into the `A320` specific area of the code at some point. Then it would be a good idea to get the FADEC out of `pneumatic` and merge it with the rest of the engine module. Might be a good idea for a small PR in the near future, maybe I have a look at it myself after this one.

src/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: Done 😃

mjuhe:
Thanks again all for the reviews 😄. This should now be ready to roll!

BlueberryKing (approved):
Awesome! Let's get this into testing 🚀

davidwalschots (approved):

mico975:

Discord          : mico#3145
Object of testing: #6220
Tier of Testing  : 1
Date             : 05/02/2022

Testing Process: 
Started C/D, outside temp 3°C.

Zone control check:
Duct air temperature reacts to selected and current temperatures as per the real aircraft - **OK**
The cabin air warms up when packs are off and there's passengers in the cabin (the more passengers, the hotter it'll get) - Cabin temperature got above 40°C with outside air temperature of 3°C which is unreasonable, some temperature differential limiting function is probably needed. **NOT OK**
There is heat tranfer between the cabin air and the outside (don't leave passengers inside of the cabin with packs off on a hot day!) - **OK**
Opening doors accelerates the heat transfer between cabin and outside as it would on the real plane - **OK**

Flow control checks:
Pack flow according to overhead selection - **OK**
Start condition, flow min NORM - **OK**
APU bleed, flow HI - **OK**
Single pack, flow HI - **OK**
Take off/landing, flow LO - Logic seems a bit weird to me as flow go from LO to NORM on rotation, which seems too early in TO phase to me. Also flow goes from NORM to LO on touchdown and not any earlier. If this is correct then **OK**

Flow control valve checks - **OK**

Also tested in various temperature conditions and system behaves as expected.

Negatives:       
Cabin warms up too much with packs not running.

Testing Results: 
Not passed

Conclusions: 
Works nicely, just needs limiting of temperature differential without packs running.

Media:       
image

image
Temperature numerical fail, sorts itself out after some time.

mjuhe:
Thanks for the test @mico975 ! I've pushed an update now to improve the simulation of the passenger heat generation to be much more accurate (ref https://engineer-educators.com/topic/3-heat-transfer-from-the-human-body/). With this I've also increassed the aircraft skin insulation to compensate.

mico975:
Updated the test report above. PR passed. 🚀

ShadowFrost3:


Object of testing: #6220
Tier of Testing:1
Date: 13/02/2022

Testing Process: 
Cold and Dark 
APU bleed cycle, pack cycles, and temperature cycles - checked 
Zone control - doors, passengers, temperature, outside air - checked 
Flow control - packs and engines seem happy - checked 
Flow control valve seems happy - checked 

Airspawn 
Aircraft operates nominally - checked
Slight note, cabin temperature rises from approximately 10C due to initialization I imagine - room for improvement but not an issue in my opinion
Potentially hot bleeds/packs but likely related to bleed simulation instead of PR - checking - passed



Cross compare against Dev 
image
(this pr - live weather)
image
(this PR - clear skies) 
image
(dev - live weather)
image
(dev - clear skies) 

Negatives: 
Testing Results: Passed

PR #6228, opened 2021-11-23
fix: transition alt/level

tracernz:
Fixes #[issue_no]

## Summary of Changes

Refactor transition altitude and level handling a little to accomodate secondary and alternate flight plan (future). Fix trans alt and level not being sent to EFIS when from DB (currently only Navigraph OFPs).

Note: no changelog as there are no user visible changes since the previous stable release.

## Additional context

## Testing instructions

Fetch an OFP from SimBrief and note the transition altitude and level are automatically filled from the database. Climb through the transition altitude and ensure QNH blinks on the PFD. Activate STD, reach cruise, then descend through the transition level and ensure STD blinks on the PFD.

### DIFF
--- a/src/fmgc/src/components/EfisLabels.ts
+++ b/src/fmgc/src/components/EfisLabels.ts
@@ -9,9 +9,8 @@ export class EfisLabels implements FmgcComponent {
 
     private flightPlanManager: FlightPlanManager;
 
-    init(): void {
-        // FIXME we need a better way to fetch this... maybe just make it a singleton
-        this.flightPlanManager = FlightPlanManager.DEBUG_INSTANCE;
+    init(flightPlanManager: FlightPlanManager): void {
+        this.flightPlanManager = flightPlanManager;
     }
 
     update(_deltaTime: number): void {
--- a/src/fmgc/src/components/EfisLabels.ts
+++ b/src/fmgc/src/components/EfisLabels.ts
@@ -1,0 +1,32 @@
+import { FlightLevel } from '@fmgc/guidance/vnav/verticalFlightPlan/VerticalFlightPlan';
+import { FmgcComponent } from '@fmgc/lib/FmgcComponent';
+import { FlightPlanManager } from '@fmgc/wtsdk';
+
+export class EfisLabels implements FmgcComponent {
+    private lastTransitionAltitude: Feet;
+
+    private lastTransitionLevel: FlightLevel;
+
+    private flightPlanManager: FlightPlanManager;
+
+    init(): void {
+        // FIXME we need a better way to fetch this... maybe just make it a singleton
+        this.flightPlanManager = FlightPlanManager.DEBUG_INSTANCE;
+    }
+
+    update(_deltaTime: number): void {
+        const transitionAltitude = this.flightPlanManager.originTransitionAltitude;
+        const transitionLevel = this.flightPlanManager.destinationTransitionLevel;
+
+        // FIXME ARINC429 when the PR adding a TS impl. lands...
+        if (transitionAltitude !== this.lastTransitionAltitude) {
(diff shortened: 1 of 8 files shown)


Benjozork (approved):

src/fmgc/src/components/EfisLabels.ts
```diff
@@ -0,0 +1,32 @@
...
+    private lastTransitionLevel: FlightLevel;
+
+    private flightPlanManager: FlightPlanManager;
+
+    init(): void {
+        // FIXME we need a better way to fetch this... maybe just make it a singleton
+        this.flightPlanManager = FlightPlanManager.DEBUG_INSTANCE;
+    }
+
+    update(_deltaTime: number): void {
```
davidwalschots, line 17: **thought**: Variable is unused.
**thought**: Variable is prefixed with `_` for no reason.
**suggestion**: Remove the unused variable.

src/fmgc/src/components/EfisLabels.ts
```diff
@@ -0,0 +1,32 @@
...
+export class EfisLabels implements FmgcComponent {
+    private lastTransitionAltitude: Feet;
+
+    private lastTransitionLevel: FlightLevel;
+
+    private flightPlanManager: FlightPlanManager;
+
+    init(): void {
+        // FIXME we need a better way to fetch this... maybe just make it a singleton
+        this.flightPlanManager = FlightPlanManager.DEBUG_INSTANCE;
```
davidwalschots, line 14: **suggestion**: If the `FlightPlanManager` is essential to most `FmgcComponent` implementations, you could consider moving the responsibility of providing the instance "up" to `loop.ts`. It would at the very least reduce the amount of FIXME things that will come out of this when more `FmgcComponent` implementations are added.

src/fmgc/src/components/EfisLabels.ts
tracernz: Our lint rules require the prefix. Since this is a function defined by the super class I prefer to keep the signature the same, even if not required by JS. What do you think?

src/fmgc/src/components/EfisLabels.ts
tracernz: Good suggestion. I think a great many components will indeed require access to the flight plan.

src/fmgc/src/components/EfisLabels.ts
davidwalschots: An unused argument apparently is required to have a `_` prefix. Supposedly because a argument named `_` is sometimes used to pattern match any value, or ignore the value. A problem I see in various spots within our code base is the `_variableName` pattern staying around when the argument _is_ used. For example, `FmsMessages.ts` defines a used `_deltaTime`.

Personally I'd still remove the argument when unused, because there's nothing in the type system that forces you to keep the argument. If however you feel strongly about keeping it. I'd rename it to `_`.

St54Kevin:
**Quality Assurance Tester**

Object of testing: #6228
Tier of Testing: 1
Date: 24/11/2021

**Testing Process:**
1. Short hop to test PFD QNH blinking on TA/TL passing

**Negatives:** N/A
**Testing Results:** Passed. Fix is working

Saschl:
Non Quality Assurance related report

Object of testing: #6228
Tier of Testing: 1
Date: 28/11/2021

Testing Process:
Imported flight plan from simbrief with TA 6000 and 11000, not entering any manual value

After passing 6000 QNH starts flashing until set to STD.
On descend, after passing 11000 feet, QNH on STD is blinking until knob is pushed.


Negatives: N/A
Testing Results: Passed.

Saschl:
Functionality wise this seems good to merge for me. Are the comments for future reference or to being addressed in this PR?

tracernz:
> Functionality wise this seems good to merge for me. Are the comments for future reference or to being addressed in this PR?

Addressed now.

PR #6240, opened 2021-11-24
feat(ap): FINAL mode

lukecologne:
## Summary of Changes

This PR adds the FINAL common AP/FD mode, but only the actual mode logic, and not the FMS changes required for it.
This means that the PR should not change any existing functionality at the moment, but is only a part of the changes required for RNAV approaches, the other part being FMS changes, which will be implemented at a later date.

## Additional context


IbrahimK42 (approved):
🚀

aguther:
I tested the above commits and found no impact on the AP modes itself.

The RNAV function is not active yet anyway.

PR #6242, opened 2021-11-24
fix(typings): add boolean SetSimVarValue overload

Benjozork:
## Summary of Changes

This patch fixes type checking errors that would reject boolean values being used in `SetSimVarValue` with boolean units, which is a valid operation. The same is not done on `GetSimVarValue` because that function does not return a boolean, even with the right unit.

## Additional context

## Testing instructions

N/A. No runtime changes.

### DIFF
--- a/typings/fs-base-ui/html_ui/JS/simvar.d.ts
+++ b/typings/fs-base-ui/html_ui/JS/simvar.d.ts
@@ -2,9 +2,11 @@
 /// <reference path="./Simplane.d.ts" />
 /// <reference path="../../../types.d.ts" />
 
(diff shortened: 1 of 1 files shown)


bouveng:
Hi. This is already part of https://github.com/flybywiresim/a32nx/pull/6238 , commit https://github.com/flybywiresim/a32nx/pull/6238/commits/43eeefd0c4c4e44fd22d0267b8567f1f991ad29e to be more exact.

:mag: :eyes:

Benjozork:
😆

davidwalschots:
Brilliant! 😀

PR #5804, opened 2021-09-10
fix(adirs): use heading as track at low ground speeds

beheh:
## Summary of Changes

When the aircraft has less than 50 knots of ground speed, the IRSs should use the  current heading as track, which they currently do not. This commit fixes that.

This is especially noticeable when spawning in on the ground, or when pushing back, as the track bug at the bottom of the PFD might be not visible at all or initialized in a weird place.

## Screenshots (if necessary)

Spawning or Pushback on Ground (**Old**):
image
The green track diamond at the bottom of the PFD is not visible because it is behind the aircraft or out of view.

Spawning or Pushback on Ground (**New**):
image
The green track diamond at the bottom of the PFD is forced to the current heading below 50 knots, which ensures it's always visible.

## References

reference

1) Video: https://www.youtube.com/watch?v=cv41hm4JEB8#t=16s
Notice how the green track diamond is visible at the center of the heading scale, despite the aircraft moving backwards due to the push back (and the actual track should thus not be visible because it points to behind the aircraft).

2) Video: https://www.youtube.com/watch?v=ECV3IBkzZLc#t=8s
Notice how the green track diamond jumps to the left once 50 knots are reached.

## Additional context

This PR does not introduce the missing true track and true heading values, and only fixes the magnetic track (via the magnetic heading). This fix will also have to be applied to true track when it is added in future.

This PR also does not update the MCDU IRS MONITOR page which is currently still using the vars from the sim itself, so this fix will have no effect. It will be fixed in future once that page is hooked up to the LVars from the (Rust) ADIRS.

Discord username (if different from GitHub):

## Testing instructions

1. Spawn in on the ground either cold and dark, or on a runway. The track diamond should be centered at the current heading like in the screenshot above.
2. Perform push back (e.g. via the EFB). The track diamond should remain where it as, at the center of the scale, moving with the heading.
3. Takeoff while turning, or abort a takeoff while turning. Notice how above 50 kts ground speed the track indicator lags behind, whereas below 50 kts ground speed it is attached to the center of the scale.
4. Fly around with varying winds. Your track should be displayed as usual.

Note that the IRS MONITOR pages on the MCDU are not hooked up to the ADIRS yet, so they may still show an incorrect value below 50 knots. This will be hooked up in a future PR.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
[...]

PR #6274, opened 2021-11-30
feat(nd): track up mode

Benjozork:
## Summary of Changes

Credit @tracernz for the implementation.

- [x] Introduces ND Track Up mode pin-program, enabled via the "Aircraft Configuration" section of the flyPad settings
- [x] Necessary map rotation and needle changes
- [x] TRK flag (when ADIRS not aligned for now; when track invalid later)

## Screenshots (if necessary)

image

image

image

## References

https://www.pprune.org/archive/index.php/t-566053.html

FBW-31-03

## Additional context

## Testing instructions

- Make sure heading up mode on the ND is accurate in regards to:
  - aircraft symbol and lubber (yellow) line do **not** rotate
  - aircraft points towards heading
  - track is represented by the green diamond 
  - needles are accurate
  - LS bug is accurate
  - "TRK" flag is **not** shown when ADIRS unaligned
 
- Make sure track up mode on the ND is accurate in regards to (better to test with a crosswind):
  - aircraft symbol and lubber (yellow) line **do** rotate
  - aircraft points towards heading
  - track is represented by the green diamond, which never moves
  - needles are accurate
  - LS bug is accurate
  - "TRK" flag **is** shown when ADIRS unaligned


src/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -64,15 +71,17 @@ export class EfisSymbols {
             lat: SimVar.GetSimVarValue('PLANE LATITUDE', 'degree latitude'),
             long: SimVar.GetSimVarValue('PLANE LONGITUDE', 'degree longitude'),
         };
-        const trueHeading = SimVar.GetSimVarValue('PLANE HEADING DEGREES TRUE', 'degrees');
+        const groundSpeed = SimVar.GetSimVarValue('GPS GROUND SPEED', 'knots');
+        // TODO in true mode, this should be true track/heading... should get the values from FM position when implemented
+        const upBearing = (this.trackUpMode && groundSpeed >= 40) ? SimVar.GetSimVarValue('GPS GROUND MAGNETIC TRACK', 'degrees') : SimVar.GetSimVarValue('PLANE HEADING DEGREES MAGNETIC', 'degrees');
```
davidwalschots, line 76: Should t these either use or be added to the ADIRS implementation?

beheh, line 76: So I think the groundSpeed check is because of #5804? If so, that number is wrong and should be >= 50 knows.

However, ideally this should simply use the ADIRS value instead, because that already handles this ([see PFD code here on how to access the ARINC429 data](https://github.com/davidwalschots/a32nx/blob/79b77e98e6d60b1f5c40e13e84605770a3b30df7/src/instruments/src/PFD/index.jsx#L109)).

tracernz:
@beheh Not sure why but I can’t reply to your comment above. The track code is a copy of the existing code in the ND, and the logic needs to match the ND track bug or things will look very weird indeed. We would want to change in all places if doing something different. Not entirely sure if the FM uses an ADIRU track directly, or it’s own ground track calculation, for the map…

beheh:
@tracernz I think GitHub does not let you comment if you have an outstanding PR review draft.

I see, we can fix this later. Re the data source, didn't we talk about this on Discord and this is directly from the ADIRS?

tracernz:
> @tracernz I think GitHub does not let you comment if you have an outstanding PR review draft.
> 
> I see, we can fix this later. Re the data source, didn't we talk about this on Discord and agree this is directly from the ADIRS?

For the ND itself, yes… not so sure about the FM part that generates the symbols and path for the map though. 😆

beheh:
I was wondering whether the DMC just matched those two, but looking at single FMGC failures it looks like both NDs must be set to the same mode and range, so chances seem good that the vectors are rotated by the FMGC (rather than the DMC).

However, a situation where the FMS assembles one image and the ND superimposes the heading directly from ADIRS seems really dangerous if the two mismatch, as the overall ND visualisation could be actively misleading. So probably the source is the same either way, and . We should (in future) look at how the FMS picks a track from an ADIRU, and I suspect the DMC will do exactly the same or it will just be retransmitted by the FMS.

Benjozork:
Should we go forward with this or does anyone wanna look at the things noted above ?

wpine215:
> Should we go forward with this or does anyone wanna look at the things noted above ?

I say we go ahead with merging this PR and we can look into implementing the feedback above in the future

beheh:
@Benjozork Would you mind rebasing this so we can properly review it and move forward? It's currently one of the oldest PRs marked as Ready for Review.

Benjozork:
Done

beheh (approved):

mico975:

Discord          : mico#3145
Object of testing: #6274
Tier of Testing  : 1
Date             : 22/02/2022

Testing Process: 
Make sure heading up mode on the ND is accurate in regards to:
aircraft symbol and lubber (yellow) line do not rotate - **OK**
aircraft points towards heading - **OK**
track is represented by the green diamond - **OK**
needles are accurate - **OK**
LS bug is accurate - **OK**
"TRK" flag is not shown when ADIRS unaligned - **OK**

Make sure track up mode on the ND is accurate in regards to (better to test with a crosswind):
aircraft symbol and lubber (yellow) line do rotate - **OK**
aircraft points towards heading - **OK**
track is represented by the green diamond, which never moves - **OK**
needles are accurate - **OK**
LS bug is accurate - **MAYBE OK**, it looks weird
"TRK" flag is shown when ADIRS unaligned - **OK**

Negatives:       
N/A

Testing Results: 
Undecided as of now 😆 

Conclusions: 
None atm

Media:       
image
HDG mode
image
TRK mode
image
HDG mode
image
TRK mode, this looks weird as ILS CRS is 283

Benjozork:
Will fix.

src/instruments/src/ND/pages/RoseMode.tsx
```diff
@@ -40,6 +41,7 @@ export const RoseMode: FC<RoseModeProps> = ({ symbols, adirsAlign, rangeSetting,
     const [fmaLatMode] = useSimVar('L:A32NX_FMA_LATERAL_MODE', 'enum', 200);
     const [fmaLatArmed] = useSimVar('L:A32NX_FMA_LATERAL_ARMED', 'enum', 200);
     const [groundSpeed] = useSimVar('GPS GROUND SPEED', 'Meters per second', 200);
+    const [trackUpMode] = usePersistentNumberProperty('A32NX_ND_TRACK_UP', 0);
```
tracernz, line 44:
```suggestion
    const [trackUpModeSetting] = usePersistentNumberProperty('A32NX_ND_TRACK_UP', 0);
    const trackupMode = trackUpModeSetting && mode === Mode.ROSE_NAV;
```

alexr4339:

Discord : Alex R#6663
Object of testing: PR #6724
Tier of Testing : 1
Date : 04/03/2022

Testing Process:
Departed EDDS in HDG mode and flew direct DKB (please not that no dep/arr was inserted on the Init page) Check all points for HDG mode, except LS bug and TRK flg when unaligned. Everything was OK, except the VOR needle was NOT accurate (reference will be found under Media)
Switched to TRK mode and tested all points except LS bug and TRK flag, everything worked except the VOR needle.
Then I truned back to EDDS and manually inserted the ILS' frequency. The ILS bearing cross on the ND was incorrect for both HDG and TRK up.
After having landed I turned off the IRS and check for the TRK flag, which only appeared in TRK up mode, as intended.

Negatives:
-Inaccurate VOR needle and ILS bearing cross
-Drawing issue (Perhaps beacause no dep/arr was entered

Testing Results:
Not passed

Media:
VOR needle bug

HDG up


TRK up


LS bug

HDG up


TRK up



On this pic, the drawing was not as expected (Would have excpected a connection between DKB and EDDS. The Init page was not filled with any information, I entered EDDS via the next waypoint section from DKB) Perhaps that is the reason why there was no runway/airport indicated after UNSER in the screenshots from the ILS approach.

Benjozork:
This will be done as part of the new ND.

PR #5934, opened 2021-09-30
feat(tcas): TCAS TA/RA logic and annunciations (ReactJS/TS)

2hwk:
Fixes #130

## Summary of Changes

This PR builds upon #2242 and #5162

### Additional note:
This PR also includes some improvements to FDR (more fields to be available) and also a new parameter to control minimum simulation rate and maximum simulation rate. The reason is simply that FDR interface version needs to be managed and due to TCAS a new interface version is anyway needed. So to reduce number of interface versions those changes will come with this PR.

### Autopilot features: (beside TCAS)
- fixed issue with V/S speed protection law (turned out during testing for TCAS)
- fixed issue with A.FLOOR not being able to deactivate by disconnecting the A/THR system (was made as this logic needed revisit anyway for TCAS)

### Features:
- [x] Add TA (traffic advisory) logic and annunciation
- [x] Add RA (resolution advisory) logic and annunciation for all possible RA callouts/variants
- [x] Integrate TA and RA annunciations with core sound manager for proper sound sequencing
- [x] Do not change sensitivity level during an active RA
- [x] Use more up-to-date TAU calculation
- [x] Inhibit RA's depending on altitude/other systems
    - [x] Inhibit "increase descent" RA's below 1,550 ft AGL
    - [x] Inhibit all "descend" RA's below 1,200 ft AGL at takeoff, and 1,000 ft AGL during approach
    - [x] Inhibit all RA's below 1,100 ft AGL in climb, 900 ft AGL in descent ("TA ONLY" message on ND)
    - [x] Inhibit all RA's as well as aural TA's below 600 ft. in climb, 400 ft. in descent
    - [x] Inhibit all climb RA's at or above 39,000 ft
    - [x] Inhibit all RA's when in ~~windshear/stall~~ condition or during GPWS "glideslope" warning ("TA ONLY" message on ND)
    - ~~Inhibit all RA's for non-altitude reporting aircraft above 15,500 ft MSL (currently not possible given sim data)~~
- [x] Ignore aircraft on ground (below 380 ft. AGL when they are climbing, 400 ft. AGL when they are descending) when own altitude is less than 1650 AGL when descending, and 1750 AGL when climbing
- [x] Implement TVTHR support
- ~~Implement HMD (horizontal miss distance) filtering~~ (Probably not in this PR)
- [x] Add green/red zones to PFD vertical speed display during an active RA which indicate the desired/forbidden vertical speed ranges, respectively
[...]

PR #6290, opened 2021-12-03
feat: autopilot and PFD preparation for TCAS

aguther:
## Summary of Changes
This PR adds / fixes:
* TCAS functionality to Autopilot and PFD
* fixed issue with V/S speed protection law (turned out during testing for TCAS)
* fixed issue with A.FLOOR not being able to deactivate by disconnecting the A/THR system (was made as this logic needed revisit anyway for TCAS)
* ability to configure minimum sim rate via config file
* ability to temporarily change minimum and maximum sim rate via LVAR
* additional data into the FDR (flight data recorder) to help in support cases

**The TCAS functions are not used right now and will not be visible to users until PR #5934 will be merged.**

## Additional context
It seems TCAS function in JS will need some additional work and will take some time to merge on master. This blocks any other AP functions to be worked on due to our creation/development process for AP etc. Therefore the main goal of this PR is to unblock it while users will not have an negative impact. It should have positive impact by the two fixes that are included.

### Overall
* do a short flight and use AP functions as usual; overall expecting no differences to master

### Fixes
* bring the plane into V/S speed protection:
  * set FCU to 15000 ft and climb to this altitude (by OP CLB or V/S)
  * when altitude is reached and ALT is shown on FMA, set V/S to +2000 ft and pull the V/S knob to engage this mode
  * now limit thrust with the thrust levers (or even disengage ATHR and control it manually) so that the plane starts to decelerate
  * when reaching VLS this speed shall be kept (with a bit of tolerance at the beginning depending on how fast this speed is approached)
  * The V/S on the FMA will have an orange rectangle around and you will hear a triple click; V/S will decrease to keep speed
  * now give a bit more thrust so that the plane will increase V/S again
  * do this a bit and a bit down, when plane will reach +2000 V/S it should speed up again
  * the import things is that this transition is always smooth and no jumps of the FD occur (this happened before)


src/instruments/src/PFD/FMA.tsx
```diff
@@ -537,6 +546,82 @@ const BC1Cell = () => {
...
+    } else if (false) {
+        text = 'SET HOLD SPEED';
+        className = 'White';
+    } else if (false) {
+        text = 'VERT DISCONT AHEAD';
+        className = 'Amber';
+    } else if (false) {
+        text = 'FINAL APP SELECTED';
+        className = 'White';
+    } else {
```
tracernz, line 607: This would be better off using the new message framework from CFPM (which is intended for MCDU, ND and PFD). There is at least no need to add any FM messages as they will be handled by that system.

aguther: How should we move forward? Right now it's integrated with AP in the way above. Can we do that with an upcoming PR?

IbrahimK42 (approved):
I tested the change regarding the AP and the fix works fine.

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #6290
Tier of Testing: 2
Date: 5.12.2021

Testing Process: 
Full flight - with tests at 15.000ft and other alts.

Negatives: none regarding the VLS protection - everything smooth. 
Testing Results: Passed (VLS protection)

Conclusions:
I did notice that after disconnecting A/THR that if I move the throttles to TOGA A/THR is armed and when pulling back to CL A/THR is activated again. I'm actually not sure if this is as it should be after a manual disconnect in flight. 

Media:

https://user-images.githubusercontent.com/16833201/144727852-1aacc67b-889a-4ebb-a947-cc652b15f90a.mp4

frankkopp:
Re-test: the A/THR arm and activation after TOGA in-air is now gone - great fix!

PR #6291, opened 2021-12-03
refactor: derive macro for visitor system

SparkyPotato:
## Summary of Changes

Added derive macro `#[derive(NestedElement)]` to automatically visit sub-elements of a struct, as well as the struct itself.

This is tied to the trait `simulation::NestedElement`, which has been separated from `simulation::SimulationElement` to allow for the derive macro.

(Yes it is a pretty big PR because almost every file had to be refactored to work with the new system)

## Testing instructions

All systems have been refactored, but since all tests pass, this shouldn't cause any problems.
All you need is a normal flight to ensure that everything touched by the Rust code still functions correctly in the simulator.


davidwalschots:
I added myself as a reviewer, but if someone else feels like reviewing then please feel free to do so, my review isn't mandatory 😀.

davidwalschots (commented):
There seems to be a bug in JetBrains IDEs which causes macro expansion of this change to fail. Opening the project CLion shows me errors like this:

> the trait bound `AirIntakeFlap: NestedElement` is not satisfied [E0277] the trait `NestedElement` is not implemented for `AirIntakeFlap`

As JetBrains IDEs are commonly used by developers of this project, I don't think we can merge this PR until this issue is solved. I would suggest reporting a bug over at [IntelliJ Rust](https://github.com/intellij-rust/intellij-rust).

aguther:
Closing due to inactivity. The PR can be re-opened at any time if work is continued.

PR #6307, opened 2021-12-05
feat(hyd):  Wheel chocks braking

crocket63:
Fixes #[issue_no]

## Summary of Changes

Applies braking force in sim when chocks are on.

## Additional context

## Testing instructions

When chocks are on the wheels, plane should not roll.
Good test case could be:
- Find a slopped surface (Courchevel? :D). 
- All hyd pumps off. 
- Empty brake accumulator by multiple OFF/ON cycles on park brake handle.
- With or without park brake, plane should stay steady with wheel chocks. 
- Without wheel chocks plane should start rolling.


bouveng:
Closed, since impl. in `feat-chocks`

PR #3581, opened 2021-02-17
feat(efb): EFB Landing Performance ⚡

matt-winfield:
## Summary of Changes
Adds a landing performance page to the Performance section of the EFB. Previously raised in a PR with takeoff performance, but TO perf needed more work so merging seperately to keep the level of changes that needs reviewing small.

## Screenshots (if necessary)
image
image
image
image
image

## References
Quick Reference Handbook, In-Flight Performance, Landing Distance Without Failure (Applicable to: D-AVVI, D-AXAJ, F-WWTQ, PR-OBJ)
image
image
image

15% safety margin is determined from the recommended safety margin in the QRH:
image

## Additional context
Discord username: TheMightyMat#0585

## Testing instructions
* Ensure landing distances vary as expected with changes in input values.
* Ensure landing distances on runway visualisation match output data
* Ensure distances that are longer than LDA are highlighted in red
* Test for display issues for extreme values and realistic values
* Ensure inputs are limited to sensible values

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created and uploaded.
The build script will have already been run with the latest changes, so no need to rerun it once you download the zip.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the right side, click on the **Artifacts** drop down and click the **A32NX** link

PR #6317, opened 2021-12-07
fix(efb): landing calculator

### LINKED ISSUE #5723: EFB Landing calculation page issues

ghitier:
Fixes #5723

## Summary of Changes

- Landing calculator:
  - Fixed the "Calculate" button's *disabled* state (make it not clickable when disabled).
  - Made all inputs initially empty (by having undefined values), so that the placeholders (/units) can be seen. It also avoids being able to press "Calculate" initially (*which avoids calculating with, previously, out of range values*).
  - Fixed the "Clear" button.
  - Fixed the "Wind Magnitude" input.
  - Improved the input placeholders.

## Screenshots (if necessary)

EFB-empty

EFB-filled

## Additional context

This PR contains the same fixes as PR #5740 , but without the toggle feature. It has a comparatively reduced scope and should therefore be easier to review. The mentioned PR will likely be closed in favor of this one and the upcoming FlypadOS v3.

Discord username (if different from GitHub): `bk67#6465`

## Testing instructions

1. Go to the landing calculator page in the EFB: all inputs should **initially be empty**, the "Calculate" button should be **disabled** (somewhat transparent and not clickable).
2. Input an ICAO in the "Airport ICAO" field (for ex: `LFST`).
3. Click "Get METAR".
4. "Wind Direction", "Wind Magnitude", "Temperature", "QNH" and "Weight" inputs should be filled in.
5. Fill in all "Rwy ..." inputs and the "Approach Speed" inputs (the missing ones).
6. The "Calculate" button should be enabled now (fully opaque).
7. Click "Calculate".
8. Make sure the distances seem consistent with the input values.
9. Change the wind direction to get a tailwind (relative to the runway heading), set the wind magnitude to 0. Click "Calculate". Set the wind magnitude to 10. Click "Calculate". **Make sure the landing distance had increased with the higher tailwind**.
10. Click "Clear".
11. Make sure all inputs are empty again (as initially).


BlueberryKing (approved):
LGTM! Nice work. 😄

xZreai:

Discord: Cleon#5802
Object of testing: #6317 
Tier of Testing: (1)
Date: (08/12/2021)

Testing Process: 

- Started up aircraft, turned on EFB
- Checked landing calculator page, all empty fields initially and calculate button was unavailable [shaded color]
- Inputted "LIRF" ICAO.
- Clicked "Get METAR", Wind direction, Wind Magnitude, Temperature, QNH, and Weight were auto-filled in. 
- Filled in runway data and approach speed 
- Calculate button is now green and available after all fields are filled in.
- Calculated landing distance, values look normal
- Changed the wind direction to get a tailwind and set the wind magnitude to 0 
- Changed wind magnitude to 10 and landing distance had changed, normal behavior.
- Clicked the clear button, all input fields were empty.

Negatives: Nil


Testing Results: PASS


Conclusions:


Media: 
image
image

St54Kevin:
**Quality Assurance Tester**

Object of testing: #6317
Tier of Testing: 2
Date: 09/12/2021

**Testing Process:**
1. Loaded in and checked calc empty
image
2. Got METAR for a couple of airports
image
3. Input data and calculated
image
4. change tailwind component
image
image
5. Cleared inputs
image


**Negatives:** 
- Landing distance does not change with different headwind components (0-40kts)
image
image

- Landing distance is shorter with higher weight
image

- tailwind from 0 to 10 does nearly double landing distance (see above point 4)

**Testing Results:** Not Passed, something is wrong with calculations

ghitier:
The landing calculations shouldn't have changed. Other than that the wind magnitude value can now be updated (as opposed to being stuck to the value of "Get METAR").

As can be seen in the reference image of the linked issue #5723 (it's a calculations table), it looks to me that only tailwind should affect the distance output.

St54Kevin:
> The landing calculations shouldn't have changed. Other than that the wind magnitude value can now be updated (as opposed to being stuck to the value of "Get METAR").
> 
> As can be seen in the reference image of the linked issue #5723 (it's a calculations table), it looks to me that only tailwind should affect the distance output.

I just checked our companies Ladning Calculator and I can confirm Headwind doesnt directly effect landing distance calculation, BUT it does effect vApp which in turn effects landing distance again. So when changing it, you also need to change the landing speed

ghitier:
> I just checked our companies Ladning Calculator and I can confirm Headwind doesnt directly affect landing distance calculation, BUT it does affect vApp which in turn effects landing distance again. So when changing it, you also need to change the landing speed.
 
On this landing calculator the approach speed has to be entered manually ("Approach Speed" input) it isn't currently a feature to calculate it automatically.

It would be a nice feature though.

> Landing distance is shorter with higher weight

I didn't touch any of that code, but I'll have a  look at what's going on.

Benjozork:
As the reported isssue is out of scope for the PR, we can do this at a later stage. Usage bug fixes are more important for now.

donstim:
I think I found the problem, and it is not in this PR. It is in PR #3581, the one that added the landing calculator to the EFB. Our calculator uses the QRH in-flight landing distance tables to calculate the landing distances to use with the before arrival landing performance check. Those tables contain corrections for different conditions, including tailwind. The tailwind correction is given as a distance to be added per 5 knots of tailwind. In checking calculated tailwind distances and the code from PR #3581, I found that it is adding the prescribed distance per knot of tailwind instead of per 5 knots of tailwind.

In PR #3581, the error is in the file, src/instruments/src/EFB/Performance/Calculators/LandingCalculator.ts. Line 360 says `const windCorrection = tailWind * landingData.windCorrection;`. It should say `const windCorrection = (tailWind/5) * landingData.windCorrection;`

@St54Kevin FYI, as noted above this EFB calculator uses the QRH in-flight landing distance tables, which will be more conservative than the 1st principles calculations in your company's landing distance calculator. I think this is probably the best we can do since we don't have access to the exact parameters used by the IRL EFB landing distance calculators. It is similar to the issue we have with developing a takeoff distance calculator, but at least in this case we do have generalized tabular data to use instead. Do you agree with this approach, or should we try to better match the IRL EFB landing distance calculators somehow?

donstim:
@ghitier Would you like me to put the correction to the tailwind effect on landing distance in a separate PR, or are you going to add that to this PR?

ghitier:
@donstim I'd say it would be better to add it to this PR. I'll give access to you on my fork if you'd prefer to add it yourself, just let me know.

jcalor:

Discord          : Calo#6816
Object of testing: #6317 
Tier of Testing  : 1
Date             : 11/12/2021

Testing Process: 

- Go to the landing calculator page in the EFB: all inputs should initially be empty, the "Calculate" button should be disabled (somewhat transparent and not clickable) - checked
- Make sure the distances seem consistent with the input values - checked
- Make sure the landing distance had increased with the higher tailwind - **NOT** working.
- Click "Clear". Make sure all inputs are empty again (as initially) - checked


Negatives:       

- `Get METAR` text is quite unreadable becouse of contrast between light blue background and white text. Same with `Calculate` button (only when clickable).
- Landing distance does **NOT** increase with higher tailwind. **It does** with higher headwind.

image
image
image
image

Testing Results: Not passed

Conclusions: Almost working, landing distance is increased with headwind not tailwind. Also some visual fixes needs to be improved just to do it better visual. Rest is ok.

donstim:
Your screenshots show that the landing distance is increasing with tailwind and staying the same with headwind, just like the QRH data (except that the effect of tailwind is magnified as noted in my earlier comment.

Wind direction is reported as the direction the wind is coming **from**. So, if you are landing on a runway heading of 323 with a reported wind direction of 320, that is a headwind. Landing on the same runway with a wind direction of 180 is a tailwind.

jcalor:
> Your screenshots show that the landing distance is increasing with tailwind and staying the same with headwind, just like the QRH data (except that the effect of tailwind is magnified as noted in my earlier comment.
> 
> Wind direction is reported as the direction the wind is coming **from**. So, if you are landing on a runway heading of 323 with a reported wind direction of 320, that is a headwind. Landing on the same runway with a wind direction of 180 is a tailwind.

Ohhh sorry about that then! My bad.

donstim:
@ghitier Okay, I added the tailwind correction fix (and also modified the pressure altitude equation to be consistent with what is being used by the aviation industry).

However, I found another error in the code that I did not know about before. Since I am not a coder, I am hesitant to try to change it, but I can describe exactly what is wrong and how it should be fixed.  The error is in the correction to landing distance for temperature. A distance correction is applied whenever the temperature is above the standard day temperature (from the International Standard Atmosphere, or ISA). 

The code currently calculates a distance correction for temperatures above 15 C, which is the ISA temperature for sea level. This means that the temperature correction will be wrong for any pressure altitude different than sea level. To fix this, a new parameter should be defined, ISAtemperature = 15 - (0.0019812 * pressureAltitude. I assume this would be done where the pressure altitude calculation is done in the code (lines 706-712).

Then lines 692-4 should be changed to:

const temperatureCorrection = temperature > ISAtemperature
            ? ((temperature - ISAtemperature) / 10) * landingData.tempCorrection
            : 0;

Would you be able to do this?

donstim:
> 
> Ohhh sorry about that then! My bad.

np

ghitier:
> Would you be able to do this?

@donstim 

I'd have to take a closer look at it... you seem way more knowledgeable than me on this though 😯

I'll try to do some research and come up with something tomorrow (Sunday, Paris timezone).

Thanks for contributing the tailwind component changes 👍

donstim:
I'm knowledgeable on what should be coded, but not the exact syntax on how to code it!

For example, I would put the ISAtemperature calculation down here like this, but I don't know the meaning of the different symbols }, /, //, *, so I don't know whether this is correct (pressure altitude calculation is in the existing code) :

 }

    /**
     * Converts a given pressure to equivalent pressure altitude
     * @param pressure Pressure in mb
     * @returns Pressure altitude in feet
     */
    private getPressureAltitude(pressure: number): number {
        // Equation from Boeing Jet Transport Performance Methods document
        return 145442.15 * (1 - ((pressure / 1013.25) ** 0.190263));
    }
   
    /**
     * Calculates ISA temperature for a given pressure altitude
     * @param PressureAltitude is pressure altitude in feet
     * @returns ISA temperature in degrees C
     * /
    private get ISAtemperature(pressureAltitude: number): number {
         return 15 - (0.0019812 * pressureAltitude)
     }

Then the temperature correction equation in current lines 692-4 would be changed as I indicated in my last comment. (Actually, only 692-3 would be changed.)

ghitier:
Thanks @Saschl for adding that.

---

> Landing distance is shorter with higher weight

@St54Kevin I second that observation in some cases. From the reading the code it seems though that it's tied to the "Approach Speed". For example if you select a high "Approach Speed" value say 180. The landing distance increases  as weight increases.

The code makes a diff between the VLS (at a certain weight and flaps config), and the "Approach Speed" value. It applies a correction based of that....

Code details:

```typescript
/**
 * VLS speed (kts) for full flap configuration
 * Index 0 = 40T, Index 8 = 80T, 5T increment
 */
const confFullVls = [116, 116, 116, 120, 125, 130, 135, 139, 143];

/**
 * VLS speed (kts) for conf 3 flaps
 * Index 0 = 40T, Index 8 = 80T, 5T increment
 */
const conf3Vls = [116, 116, 124, 130, 136, 141, 146, 151, 155];

/**
 * Converts mass into an index from 0-8 for use with VLS tables
 * @param mass Mass in tons
 */
function getVlsTableIndex(mass: number): number {
    const index = Math.ceil(((mass > 80 ? 80 : mass) - 40) / 5);
    return index >= 0
        ? index
        : 0;
}
```

Then in the calculate function:

```typescript
let targetApproachSpeed: number;
const vlsTableIndex = getVlsTableIndex(weight / 1000);
if (flaps === LandingFlapsConfig.Full) {
    targetApproachSpeed = confFullVls[vlsTableIndex];
} else {
    targetApproachSpeed = conf3Vls[vlsTableIndex];
}

// approachSpeed is "Approach Speed" input value
let speedDifference = approachSpeed - targetApproachSpeed;
if (speedDifference < 0) {
    speedDifference = 0;
}

const speedCorrection = (speedDifference / 5) * landingData.speedCorrection;
```

I don't understand well why this difference with the vls is made, I can't say for sure if this behavior is as expected....

donstim:
> I don't understand well why this difference with the vls is made, I can't say for sure if this behavior is as expected....

Yes, it is as intended. This landing distance calculator is based on a set of tables in the Quick Reference Handbook (QRH). There are separate tables for each runway condition (dry, good, good to medium, etc.). Each table contains a reference landing distance for each braking level (max manual, a/b medium, and a/b low) and each landing configuration (full landing flaps and CONF 3). The reference landing distance is for 68T, VAPP=VLS, sea level, ISA, no wind, no slope, no reverse thrust, and no autoland. The distance needs to be corrected for any other condition.  (Note: We are currently missing the correction of autoland. I have made an issue (#6343) for this.)

Getting back to the approach speed correction, since the reference landing distance is based on VAPP = VLS, then for any approach speed above VLS a correction must be made. The correction value in the table is for every 5 knot increase in VAPP over VLS. That is why the correction equation uses (the input approach speed minus VLS) / 5 and multiplies that by the correction value from the table.

In an earlier comment, you said that it would be a nice feature for the calculator to input the approach speed automatically. The reason this is not done currently is that the actual approach speed to be used is up to the flightcrew. The minimum it can be is VLS. There are recommended speed additives for wind, but the flightcrew may add their own increment based on the conditions and their experience. (They should also be entering this VAPP in the PERF APPR page of the MCDU if it is different than what is calculated by the FMS.)

For sim pilots especially, it may indeed be helpful to automatically enter the approach speed into the EFB. But it should be the approach speed from the PERF APPR page rather than an independent calculation within the EFB landing calculator code. And we should think about whether some message should be displayed somewhere if the sim pilot changes the approach speed (o flap setting for that matter) in the MCDU.

ShadowFrost3:


Object of testing: #6317
Tier of Testing:1
Date: 30/12/2021

Testing Process: Hotstart
KATL
Get METAR works - checked
Calculate unclickable - checked 
Headwind does not affect landing distance - understood to be intentional - checked?
Tailwind working correct - checked 
Clear button works - checked 

Temperature does not affect landing performance below ~15C - correct/incorrect?

Negatives: 
***** (Not necessarily this PR) *****
RWY Slope - not operational
Thrust reverser setting is unreliable - probably broken 
Medium poor condition max braking is less effective than medium - assuming correct given the potential for excessive ABS reducing landing performance 
Poor landing surface has better performance than medium-poor 
Medium-Poor to Poor flaps CONF 3 appears broken (see referenced)


Testing Results: Partial pass while waiting for clarification on temperature 

Conclusions: 
Media: 
image
image

donstim:
1 Correct that no credit is given for improved landing distance with headwind. So headwind effectively has no effect.
2. Only temperatures above ISA affect landing distance (no credit for lower temperatures)
3. Runway slope should be operational. Like the above, only a downslope (negative slope) will affect the landing distance. No credit for positive slope.
4. Congratulations, you found an error in one of the tables providing the landing performance for a poor runway surface condition, using autobrakes medium, and CONF 3.

donstim:
Fixed the typo in the landing distance table that caused the erroneous result in @ShadowFrost3's QA test. Did a thorough review of all the tables and found other minor typos and fixed them as well.

ShadowFrost3:


Object of testing: #6317
Tier of Testing:1
Date: 31/12/2021

Testing Process: Quick retest. 
Fairly happy with everything, exception below. 
The only item that is concerning is (and I don't have the documents to easily reference this) with RWY condition poor. Thrust reversers equal a near 1,000 meter performance boost. 

Other than that, as far as the references I have available, nothing appears to be outside of plausible parameters in terms of operation. 

*** Bug Found 
Clear does not reset RWY condition visibly but does reset it internally. So the landing performance calculator will give you false information if not corrected/noticed. 

Steps to reproduce
Enter runway condition poor
hit clear button 
enter airport identifier, grab METAR, and input other relevant information
the runway has defaulted to dry conditions but the EFB indicates poor and displays calculations for dry conditions 

Negatives: 
Testing Results: Not Passed

https://streamable.com/rh0pbd

donstim:
Great job testing once again. Yes, reversers make that much of a difference in poor runway surface conditions. That part is correct. However, your finding that clearing the calculator reverts to dry while not changing the input field accordingly is a show stopper. In confirming this, I also found the reverser and overweight procedure fields to be similar. Upon clearing the calculator, it calculates a distance based on no reversers and no overweight procedure even though the input fields are not changed to this if they were previously with reversers and with overweight procedure.

I found another error as well. Like the temperature, wind, slope, and approach speed corrections, there is a correction for altitude, but it is only supposed to add distance (and it does this if the pressure altitude is above sea level). If it is never supposed to subtract distance, but currently the calculator will yield shorter distances if the pressure altitude is below sea level.

donstim:
@ghitier @Saschl See the previous comments. I fixed the pressure altitude issue (discussed in the 2nd paragraph of my comment above), but I'll need one of you guys to fix the issue found by @ShadowFrost3 and expanded upon in the 1st paragraph of my comment above.

ghitier:
@ShadowFrost3 thanks for further having investigated the core of the issue. Since it has to do with the reset this is definitely my bad, I'll take a look to fix it.

ghitier:
@donstim all the selectors (rwy condition, flaps, overweight and reverse) were indeed affected. It was an issue with the base React component (efb only though), I just fixed that.

ShadowFrost3:


Object of testing: #6317 
Tier of Testing:1
Date: 31/12/2021

Testing Process: Retested clears, get METAR, and landing calculations 
LGTM
Negatives: 
Testing Results: Passed

Conclusions: Nice work to everyone involved in diagnosing and reacting quickly.

donstim:
Thanks for AQ testing so quickly!  I made a few more minor changes:
1. Changed the initial (and after clearing) state of the landing flaps to full landing flaps. That is the normal landing configuration;
2. Changed the capitalization of Full to FULL to be consistent with the MCDU; and
3. Added the runway condition codes in parentheses after each runway condition. This will be helpful to pilots translating a runway condition code from ATIS to the corresponding runway surface condition.

ShadowFrost3:


Tested one last time (I hope). LGTM

PR #6370, opened 2021-12-16
feat(nd): Add XTE display in PLAN mode, reduce XTE display precision

lukecologne:
## Summary of Changes

This PR adds the XTE indication, which was missing in PLAN mode, and also changes the XTE in all modes to only display 1 digit of precision.

## References

Image credit: @tshomas 
IMG_2187

grafik

## Additional context

## Testing instructions

Make sure that the XTE indication shows correctly, both in PLAN and also in the other ND modes.

### DIFF
--- a/src/instruments/src/ND/elements/CrossTrack.tsx
+++ b/src/instruments/src/ND/elements/CrossTrack.tsx
@@ -15,8 +15,8 @@ export const CrossTrack: React.FC<CrossTrackProps> = memo(({ x, y, isPlanMode })
     let crossTrackX = x;
     const crossTrackAbs = Math.abs(crossTrackError);
 
-    if (crossTrackAbs > 0.02) {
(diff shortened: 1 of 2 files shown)


Benjozork:
While we are here, we were gonna change the min XTE to 0.1 NM instead of 0.02, can we do this in this PR ?

lukecologne:
Sure, I don't see why not

beheh (approved):
Cross-checked some more references, looks good for now. In future we might want two digits back for RNP approaches.

Benjozork (approved):

tracernz:
Should not overlap chrono. Not sure if that's been checked.

lukecologne:
I have, and it doesn't overlap. I've actually assumed that it was directly below the chrono, so I based the positioning on the Chrono

PR #6427, opened 2021-12-28
refactor: declarative msfs aspects

davidwalschots:
## Summary of Changes
The Rust code has a layer which bridges MSFS and our system simulation. Some of this bridging is imperative and convoluted, making it hard to understand what is going on. I'm in the process of adapting these pieces to a more declarative style, which hopefully is easier to understand.

### Definitions

The _simulator_ is MSFS.

The _simulation_ is the system simulating code written in Rust.

An _aspect_ is a concern that should be handled by the bridging layer.

There are three types of _variables_ that aspects can handle:
1. `Variable::Aircraft`: the variable type we already know from MSFS itself (`A:` or no prefix).
1. `Variable::Named`: the variables we already know from MSFS itself (`L:`). Note that the common prefix (in our case `"A32NX_"` is added to all of these.
1. `Variable::Aspect`: the variables which are accessible by all aspects and the simulation.

It should be noted that the simulation is completely unaware of what provides a variable's value.

### Methods

The `MsfsSimulationBuilder` provides a new `with_aspect` function. Each aspect that is added through `with_aspect` is unaware of other aspects. The function passed to `with_aspect` receives a `MsfsAspectBuilder` through which one declares the aspect:

```rust
MsfsSimulationBuilder::new(...).with_aspect(brakes)?;

fn brakes(builder: &mut MsfsAspectBuilder) -> Result<(), Box<dyn Error>> {
    // ...
}
```

The declarative interface is still work in progress. At the moment the following methods have been declared on a builder type:

#### event_to_variable

This method converts the occurrence of an event to a variable. It returns the event's id to use with `variable_to_event_id` if necessary.

```rust
builder.event_to_variable(
    "PARKING_BRAKES",
    EventToVariableMapping::CurrentValueToValue(|current_value| {
        from_bool(!to_bool(current_value))
    }),
    Variable::Named("PARK_BRAKE_LEVER_POS".into()),
    |options| options.mask(),
)?;
```

The `"PARKING_BRAKES"` event is masked and captured. When it occurs the given `EventToVariableMapping` is executed after which the resulting value is written to the `"A32NX_PARK_BRAKE_LEVER_POS"` named variable.

There are many forms of `EventToVariableMapping`, for example:

```rust
EventToVariableMapping::Value(1.),
EventToVariableMapping::EventDataToValue(|event_data| from_bool(event_data == 1))
EventToVariableMapping::EventData32kPosition,
EventToVariableMapping::SmoothPress(KEYBOARD_PRESS_SPEED, KEYBOARD_RELEASE_SPEED),
```

Sometimes one might want to ignore an event for a little while after it occurred, or one might want to only set a value for one tick. This is possible through the `leading_debounce` and `afterwards_reset_to` options:

```rust
builder.event_to_variable(
    "AUTOBRAKE_LO_SET",
    EventToVariableMapping::Value(1.),
    Variable::Named("OVHD_AUTOBRK_LOW_ON_IS_PRESSED".into()),
    |options| {
        options
            .leading_debounce(Duration::from_millis(1500))
            .afterwards_reset_to(0.)
    },
)?;
```

Leading debounce immediately executes the action associated with the event, but ignores any subsequent event until no event triggered for the given duration. `afterwards_reset_to` awaits for the debounce wait period to pass before resetting the value. If no leading debounce was declared, the value will be reset after the next simulation tick.

#### variable_to_event

Writes the variable's value to an event. If you use `event_to_variable` for the same event, then you should use `variable_to_event_id` instead.

```rust
    builder.variable_to_event(
        Variable::Aspect("STEERING_ANGLE".into()),
        VariableToEventMapping::EventData32kPosition,
        VariableToEventWriteOn::EveryTick,
        "STEERING_SET",
    )?;
```

`VariableToEventWriteOn` provides a way to declare whether the variable should be written every tick, or only when it has changed compared to the last write.

#### variable_to_event_id

Writes the variable's value to an event with the given id. This function should be used when you previously acquired an event id by using `event_to_variable`.

```rust
    builder.variable_to_event_id(
        Variable::Aspect("BRAKE LEFT FORCE FACTOR".into()),
        VariableToEventMapping::EventData32kPosition,
        VariableToEventWriteOn::EveryTick,
        axis_left_brake_set_event_id,
    );
```

`VariableToEventWriteOn` provides a way to declare whether the variable should be written every tick, or only when it has changed compared to the last write.

#### copy

This method copies a variable to another variable. It is primarily used to replace the mechanism previously provided by `provides_aircraft_variable_with_additional_names`.

```rust
builder.copy(
    Variable::Aircraft("EXTERNAL POWER AVAILABLE".into(), "Bool".into(), 1),
    Variable::Aspect("OVHD_ELEC_EXT_PWR_PB_IS_AVAILABLE".into()),
);
```

#### map

This method maps a variable to another variable.

```rust
builder.map(
    UpdateOn::PreTick,
    Variable::Named("FLAPS_HANDLE_INDEX".into()),
    |value| value / 4.,
    Variable::Named("FLAPS_HANDLE_PERCENT".into()),
);
```

Prior to every simulation tick, the variable is read, the function is applied to it and the resulting value is written to `"FLAPS_HANDLE_PERCENT"`.

#### map_many

This method maps a set of variable values to another variable.

```rust
builder.map_many(
    UpdateOn::PostTick,
    vec![
        Variable::Aspect("NOSE_WHEEL_POSITION_RAW".into()),
        Variable::Aspect("RUDDER_POSITION_RAW".into()),
    ],
    |values| {
        let nose_wheel_position = values[0];
        let rudder_position = values[1];

        steering_demand_to_msfs_from_steering_angle(nose_wheel_position, rudder_position)
    },
    Variable::Aspect("STEERING_ANGLE".into()),
);
```

#### reduce

This method reduces variable values to a single value and writes it to the given variable.

```rust
// The maximum braking demand of all keyboard and controller inputs
// is calculated and made available as a percentage.
builder.reduce(
    UpdateOn::PreTick,
    vec![
        Variable::Aspect("BRAKES".into()),
        Variable::Aspect("BRAKES_LEFT".into()),
        Variable::Aspect("BRAKE LEFT FORCE FACTOR".into()),
    ],
    0.
    max,
    Variable::Named("LEFT_BRAKE_PEDAL_INPUT".into()),
);
```

Prior to every simulation tick, the variables are read and the `max` function is applied to every element, resulting in the maximum value out of all the input variables to be written to the `"LEFT_BRAKE_PEDAL_INPUT"` named variable.

#### variables_to_object

Writes a set of variables to a SimConnect data definition object.

```rust
builder.variables_to_object(Box::new(FlapsSurface {
    left_flap: 0.,
    right_flap: 0.,
}));
```

The above is seemingly missing the necessary information on what data should be written to the object. One has to implement `VariablesToObject` on the type:

```rust
impl VariablesToObject for FlapsSurface {
    fn variables(&self) -> Vec<Variable> {
        vec![
            Variable::Named("LEFT_FLAPS_POSITION_PERCENT".into()),
            Variable::Named("RIGHT_FLAPS_POSITION_PERCENT".into()),
        ]
    }

    fn write(&mut self, values: Vec<f64>) {
        self.left_flap = values[0];
        self.right_flap = values[1];
    }

    set_data_on_sim_object!();
}
```

After every simulation tick, the `write` function is invoked with a `Vec` containing the variable values. The order of the `values` is equal to the definition order in the `variables` function.

## Additional context

## Testing instructions
The PRs changes mainly will affect the following items. These should work as they used to.

- AUTOBRAKE through the controller/keyboard.
- BRAKES through the controller/keyboard.
- FLAPS using the controller/keyboard, amongst other things ensuring that the right animations (flaps/slats angle) is depicted on the outside of the aircraft.
- NOSE WHEEL STEERING using the controller/keyboard, amongst other things ensuring the animations for rudder pedals, rudder, tiller and nose wheel are correct.


src/systems/systems_wasm/src/aspects.rs
```diff
@@ -0,0 +1,852 @@
...
+    /// Convert event occurrences to a variable.
+    ///
+    /// If you want to write a variable back to the same event, then use the
+    /// event id returned by this method and pass it to [Self::variable_to_event_id].
+    pub fn event_to_variable(
+        &mut self,
+        event_name: &str,
+        mapping: EventToVariableMapping,
+        target: Variable,
+        configure_options: fn(EventToVariableOptions) -> EventToVariableOptions,
```
BlueberryKing, line 134: Why did you opt for making this a function from `EventToVariableOptions` to the same type? I think just using `option: EventToVariableOptions` would reduce complexity, rather than allowing a `configure_options` function and always passing in the default value.

src/systems/systems_wasm/src/aspects.rs
```diff
@@ -0,0 +1,852 @@
...
+        self.actions
+            .push((Map::new(inputs, func, output).into(), execute_on));
+    }
+
+    /// Map a set of variable values to another variable.
+    pub fn map_many(
+        &mut self,
+        execute_on: ExecuteOn,
+        inputs: Vec<Variable>,
+        func: fn(&[f64]) -> f64,
```
BlueberryKing, line 85: I assume it is not trivial to use const generics here to ensure that `func` takes an array with the same number of elements as `inputs` has? Would be nice though. 😄

src/systems/systems_wasm/src/lib.rs
```diff
@@ -93,8 +103,7 @@ impl<'a, 'b> MsfsSimulationBuilder<'a, 'b> {
         mut self,
         aircraft_ctor_fn: U,
     ) -> Result<(Simulation<T>, MsfsHandler), Box<dyn Error>> {
-        let mut aspects: Vec<Box<dyn SimulatorAspect>> =
-            vec![Box::new(self.electrical_buses.unwrap())];
+        let mut aspects: Vec<Box<dyn Aspect>> = vec![Box::new(self.electrical_buses.unwrap())];
         if self.apu.is_some() {
             aspects.push(Box::new(self.apu.unwrap()));
         }
```
BlueberryKing, line 109: I'm aware that this is not directly related to this PR, but I am wondering why are the APU and the electrical busses treated differently from other `MsfsAspect`s. Why are they not built from and `MsfsAspectBuilder` like the brakes for example?

src/systems/systems_wasm/src/aspects.rs
```diff
@@ -0,0 +1,852 @@
...
+    /// Converts the event occurrence to a value which increases and decreases
+    /// by the given factors.
+    SmoothPress(f64, f64),
+}
+
+#[enum_dispatch]
+enum MessageHandler {
+    EventToVariable,
+}
+
```
BlueberryKing, line 654: Are you expecting other types of `MessageHandler` in the future? Why does this need to be an enum?

src/systems/systems_wasm/src/aspects.rs
```diff
@@ -0,0 +1,852 @@
...
+        if matches!(variable, Variable::Aircraft(..)) {
+            eprintln!("Writing to variable '{}' is unsupported.", variable);
+        }
+    }
+}
+
+pub struct MsfsAspect {
+    message_handlers: Vec<MessageHandler>,
+    actions: Vec<(VariableAction, ExecuteOn)>,
+}
```
BlueberryKing, line 201: It is a bit unclear to me what the difference between a message handler and an action is. It seems like a message handler is an action to handle SimConnect events while an action is typically some generic variable/event mapping.

Is there a fundamental reason why we cannot treat message handlers as an action as well?

src/systems/systems_wasm/src/lib.rs
```diff
@@ -299,139 +319,275 @@ impl SimulatorReaderWriter for MsfsHandler {
...
+
+        if let Some(variable_registry) = &mut self.variables {
+            variable_registry.write(identifier, value);
+        }
+    }
+}
+
+/// Declares a variable of a given type with a given name.
+#[derive(Clone)]
+pub enum Variable {
```
BlueberryKing, line 345: This is a broad question: We now have the concepts of `Variable`, `VariableType`, `VariableValue`, `VariableIdentifier`. There is quite a bit of code involving `From` and `Into` between these types. Particularly the first three are essentially enums differentiating between aircraft variables, named variables, and aspect variables. Would it not be more feasible to introduce these types (`AircraftVariable`, `NamedVariable`, `AspectVariable`) as structs directly? Shared behavior could be specified via a `Variable` trait perhaps. I think we could avoid some `match` expressions using polymorphism like this.

BlueberryKing (commented):
Very cool! Left some comments, including a few that might not be only related to the area of code they annotate.

I haven't looked over everything in detail yet.

src/systems/systems_wasm/src/aspects.rs
davidwalschots: This lets one write:

```rust
    builder.event_to_variable(
        "BRAKES",
        EventToVariableMapping::SmoothPress(KEYBOARD_PRESS_SPEED, KEYBOARD_RELEASE_SPEED),
        Variable::aspect("BRAKES"),
        |options| options.mask(),
    )?;
```

Whereas changing it to `EventToVariableOptions` would require the following code:

```rust
    builder.event_to_variable(
        "BRAKES",
        EventToVariableMapping::SmoothPress(KEYBOARD_PRESS_SPEED, KEYBOARD_RELEASE_SPEED),
        Variable::aspect("BRAKES"),
        EventToVariableOptions::default().mask(),
    )?;
```

I like having the type not repeated on every usage.

src/systems/systems_wasm/src/aspects.rs
davidwalschots: It is not possible. The reason for that is that it would require all usages of `map_many` to have the same number of input variables known at compile time, which is unlikely to be the case.

src/systems/systems_wasm/src/lib.rs
davidwalschots: Well spotted. While working on this I realised that probably the APU and electrical buses can be replaced by what I built here as well. However, I did not want to increase the scope of the PR, so I'll make those changes in a later PR.

src/systems/systems_wasm/src/aspects.rs
davidwalschots: Yes. There are different types of SimConnect data coming from a message. `SimConnectRecv::Event(e)` is the one handled by `EventToVariable` now. I'd consider it likely that e.g. `SimConnectRecv::SimObjectData(data)` would later also be handled through another declarative function. There might also be other forms of event handling coming, e.g. `EventToArinc429Bus` which would connect to @beheh's work.

By adding the structure for this now, I'm making sure that any future modification will not require too much uprooting of what's already there.

src/systems/systems_wasm/src/aspects.rs
davidwalschots: Yes, the fundamental reason is that `actions` are executed every frame (either before or after a simulation tick), whereas `message_handlers` only handle things when the SimConnect message it handles is received.

src/systems/systems_wasm/src/lib.rs
davidwalschots: The `AircraftVariable` and `NamedVariable` types are in the `msfs-rs` crate we use. I could of course implement a trait for these types and use these types directly. However, this leads to problems for when I add automated tests, which I'm hoping to do soonish as well.

The main reason for that is that `msfs-rs` only exposes the `AircraftVariable` and `NamedVariable` types when build targeting `wasm32`, which is not what I'd be running the tests with. My hope is that I can flag the current `From<&Variable> for VariableValue` implementation to only compile when targeting `wasm32`, and to create another `From<&Variable> for VariableValue` implementation which would be used for testing, thus removing my dependency on the `msfs-rs` types. If instead I'd use `AircraftVariable` and `NamedVariable` directly, that would be a lot harder, as those types would be scattered throughout.

`AircraftVariable::from` returns a `Result<AircraftVariable, SomeErrror>` which would need to be handled at every usage as well and would likely obscure things.

src/systems/a320_systems_wasm/src/autobrakes.rs
```diff
@@ -0,0 +1,39 @@
+use std::error::Error;
+use std::time::Duration;
+use systems_wasm::aspects::{EventToVariableMapping, EventToVariableOptions, MsfsAspectBuilder};
+use systems_wasm::Variable;
```
BlueberryKing, line 4: We might want to group imports from the same crate, but this is not important to me. This is also relevant for the other files.

src/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -162,6 +106,51 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
...
+                    Variable::aircraft("EXTERNAL POWER ON", "Bool", 1),
+                    Variable::aspect("OVHD_ELEC_EXT_PWR_PB_IS_ON"),
+                );
+
+                builder.copy(
+                    Variable::aircraft("GENERAL ENG MASTER ALTERNATOR", "Bool", 1),
+                    Variable::aspect("OVHD_ELEC_ENG_GEN_1_PB_IS_ON"),
+                );
+
+                builder.copy(
```
BlueberryKing, line 136: Any reason there is an extra line break between some of these and not others?

BlueberryKing (changes requested):

src/systems/a320_systems_wasm/src/autobrakes.rs
davidwalschots: I agree. What I've found is that CLion and vscode handle these things differently and all have their quirks. For example, if I manually group these in CLion then any additional import added through the IDE within the `::aspects::` bit would go to a new line. Probably because it doesn't understand nested imports to that many levels, at least not for formatting/code completion purposes.

For now I guess we'll just have to accept this is how it is. I'll be sure to keep an eye out for improvements in this area. If at some point the editors work fine, we can create a PR that makes all imports throughout the project nicer.

src/systems/a320_systems_wasm/src/lib.rs
davidwalschots: No. The line breaks were clearly confusing. I've now combined those items that are related. Should make more sense now 😀.

BlueberryKing (approved):
Good to go from my side!

D4rknessCodez:

Discord: Darkness🍁#7734
Object of testing: #6427
Tier of Testing: 2
Date: (19/01/2022)

Testing Process:
- Loaded into aircraft and did normal start procedures
- Once started I took off and did a flight around the airfield using both manual controls and the autopilot
- In between, I tested the autobrake, brake, flaps and NWS all using my joystick module (counted as controller)

Negatives:
- Some frame loss on startup and taxi, frames went back to normal at cruise

Testing Results:
Passed 

Conclusions:
Everything worked as it was intended **on controller**, however, an investigation into performance would be appreciated as there was a loss of framerate as explained in the negatives. Nice work and let me know if you guys need any more tests.

mico975:

Discord          : mico#3145
Object of testing: #6427 
Tier of Testing  : 2
Date             : 20/01/2022

Testing Process: 
Started C/D at LJLJ and did a quick flight to LJMB.
In the process tested all mentioned controls with TCA pack as well as keyboard and all worked fine. Animations all worked as intended.
Completely shutdown the aircraft and repowered it. Worked well.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
Nothing broken, no performance loss observed.

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #6427
Tier of Testing: 1
Date: 20.1.2022

Testing Process: 
- C&D at gate
- testing according to instructions
- additional testing and short flight

Negatives: none

Testing Results: Passed

Conclusions: LGTM

Protocol:
<details>

- AUTOBRAKE through the controller/keyboard.
  - OK via TCA AddOn Switch and SPAD.neXt: LVAR L:A32NX_AUTOBRAKES_ARMED_MODE (0-3)
- BRAKES through the controller/keyboard.
  - OK - tested via pedals and button
- FLAPS using the controller/keyboard, amongst other things ensuring that the right animations (flaps/slats angle) is depicted on the outside of the aircraft.
  - OK on ground and in the air - animation good - tested keyboard (F5-F8) and TCA AddOn with SPAD.neXt mapping: FLAPS_SET (0-16384)
- NOSE WHEEL STEERING using the controller/keyboard, amongst other things ensuring the animations for rudder pedals, rudder, tiller and nose wheel are correct.
  - OK (with activated tiller separation) and standard mapping in MSFS

Full flight: no issues - all worked as expected. 

</details>

Issue #6271, opened 2021-11-29
Engine out acceleration altitude is not using airfield elevation in the calculation on the MCDU perf page


John-UK:
### Aircraft Version

Development

### Build info

```json
"built": "2021-11-26T19:06:28+00:00",
  "ref": "master",
  "sha": "6803283d85f97decbccf60eb3162d701ec71b217",
  "actor": "sidnov",
  "event_name": "manual"
}
```

### Describe the bug

The engine out acceleration value on the Perf page defaults to the EFB value rather than the EFB value + airfield elevation.  The code appears to add the two values in the same way as it does for thrust reduction / acceleration.
  

### Expected behavior

The MCDU take off page displays the EFB default THR/ACC ENG OUT ACC values + the airfield elevation.

### Steps to reproduce

Load into any stand. Check the values set in the EFB normally 1500 for THR RED/ACC and ENG OUT ACC. Check the airfield elevation. Note that THR RED / ACC shows the two figures added together and that ENG OUT ACC shows only the value in the EFB.

### References (optional)

Worked in a previous build.

### Additional info (optional)

In file A320_NEO_PerformancePage.js from line 1060 this code appears to add the two values. 

A similar functions exist for Thrust reduction altitude and that does display the two values added together in the MCDU. On line 1037  let elevation = SimVar.GetSimVarValue("GROUND ALTITUDE", "feet");  Should this line also be in the below function?

  static UpdateEngOutAccFromOrigin(mcdu) {
        if (mcdu.engineOutAccelerationAltitudeIsPilotEntered) {
            return;
        }
        const origin = mcdu.flightPlanManager.getOrigin();
        const elevation = origin ? origin.altitudeinFP : SimVar.GetSimVarValue("GROUND ALTITUDE", "feet");

        const offset = +NXDataStore.get("CONFIG_ENG_OUT_ACCEL_ALT", "1500");
        const alt = Math.round((elevation + offset) / 10) * 10;

        mcdu.engineOutAccelerationAltitude = alt;
        mcdu.engineOutAccelerationAltitudeIsPilotEntered = false;
        SimVar.SetSimVarValue("L:A32NX_ENG_OUT_ACC_ALT", "feet", alt);

### Discord Username (optional)

john P Maguire#9027

PR #6450, opened 2021-12-31
fix(mcdu): take off page THR RED ACC and ENG OUT ACC calculations and restrictions

John-UK:
Fix for issue #6271 Engine out acceleration altitude is not using airfield elevation in the calculation on the MCDU perf take off page.
It also adds restrictions to the THR RED ACC values that can be entered.

## Summary of Changes
 origin.altitudeinFP no longer returns the elevation of the orgin airfield.  It returns zero. This results in the minimum altitude restriction for THR RED ACC and ENG OUT ACC to not take account of airfield elevation.  Additionally no restrictions were imposed on the valid values for THR RED and ACC. This allowed ACC to be set lower than THR RED. This PR sets the minimum performance altitudes to FROM airfield elevaion + 400 and imposses that ACC => THR RED. Only the take off page has been modified in this PR. The go around also has the same issues, I suggest these should be addressed when the GA track is implemented.
 
## Screenshots (if necessary)
image

## References
FCOM. Verified in discussions with type rated A320 pilots. 

## Additional context
Other parts of the code make use of xxxxx.altitudeInFP these should be checked as unlikely they work as intended.

john P Maguire#9027

## Testing instructions
Load in on the stand and on the runway. (take off page performance altitudes should not be editable after take off).

1. In the EFB check / set an offset value for THR RED ACC and ENG OUT ACC. Defaults to 1500. Change this during the test to check that the entered offset is used.

2. Check that without a FROM/TO in INIT A the performance attitudes display ----/---- and ---- 

3. On entering a FROM/TO on INIT A the THR RED ACC and ENG OUT ACC altitudes are calculated using the EFB offset + the airfield elevation of the FROM airfield.

4. Enter a different value for each of THR RED ACC and ENG OUT ACC (a pilot entered value). Check that four digits are required eg eng out acc 900, 0900 must be entered.  Check the text size is changed to large. Use CLR to clear the value and check it reverts back to the calculated altitude and small text. 

5. Check that a lower altitude cannot be entered for ACC than THR RED or a higher altitude cannot be entered for THR RED than ACC. These values should be able to be entered as a pair or individually. 

6. Change the FROM/TO in INIT A check that the new FROM airfield altitude is used and an updated THR RED ACC and ENG OUT ACC is correctly calculated for the new FROM airfield.

7. Enter a pilot value for the THR RED ACC and ENG OUT ACC. Change the FROM/TO airports on INIT A and check that the pilot entered value is not changed until the CLR is used to reset the calculation.  This should revert to small text and the correct altitude (airfield elevation + EFB offset).

8. Check that an altitude lower than the FROM airfield elevation + 400 cannot be entered. That a value higher than 45 000 cannot be entered in any of the fields THR RED ACC or ENG OUT ACC.

9. Check for no regressions.


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -2338,69 +2340,76 @@ class FMCMainDisplay extends BaseAirliners {
...
+                this.addNewMessage(NXSystemMessages.formatError);
                 return false;
+            } else if (accAlt && accAlt.length > 0) {
+                accAlt = parseInt(accAlt);
+                accAlt = Math.round(accAlt / 10) * 10;
+                checkAccAlt = accAlt;
             }
 
-            newThrRedAlt = thrRedAlt;
         }
```
BlueberryKing, line 2385: I think this block is a bit convoluted. I believe we can clean this up a bit by removing some redundant null checks on `thrRedAlt` and `accAlt` for example. Furthermore, it seems that this block does not check whether the altitudes are in the required range, and that this check is done further down. Are you sure this is intended?

BlueberryKing (commented):
Glad to see you reopened your PR! Good fixes, I just noticed a few things which I'm curious about.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -2338,69 +2340,76 @@ class FMCMainDisplay extends BaseAirliners {
...
-        let newThrRedAlt = null;
-        let newAccAlt = null;
+        let checkThrRedAlt = this.thrustReductionAltitude;
+        let checkAccAlt = this.accelerationAltitude;
 
         let [thrRedAlt, accAlt] = s.split("/");
 
-        if (thrRedAlt && thrRedAlt.length > 0) {
-            if (!/^\d{3,5}$/.test(thrRedAlt)) {
+        if (thrRedAlt && thrRedAlt.length > 0 || accAlt && accAlt.length > 0) {
```
BlueberryKing, line 2365: I think this `if` block is a bit convoluted. I believe we can clean this up a bit by removing some redundant null checks on `thrRedAlt` and `accAlt` for example. Furthermore, it seems that this block does not check whether the altitudes are in the required range, and that this check is done further down. Are you sure this is intended?

John-UK:
> Glad to see you reopened your PR! Good fixes, I just noticed a few things which I'm curious about.

I wanted to add the constraints as well as fix the eng out acc not using the airfield elevation. I will get to these comments 
tomorrow.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -254,9 +254,10 @@ class FMCMainDisplay extends BaseAirliners {
         this.updateFuelVars();
         this.updatePerfSpeeds();
 
-        CDUPerformancePage.UpdateThrRedAccFromOrigin(this, true, true);
+        CDUPerformancePage.UpdateThrRedAccFromOrigin(this);
         CDUPerformancePage.UpdateEngOutAccFromOrigin(this);
         SimVar.SetSimVarValue("L:AIRLINER_THR_RED_ALT", "Number", this.thrustReductionAltitude);
+        SimVar.SetSimVarValue("L:A32NX_ENG_OUT_ACC_ALT", "feet", this.engineOutAccelerationAltitude);
```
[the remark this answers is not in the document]
John-UK: I wondered at these myself. The are repeated in line 512-513. Without knowing if the simVar needs to initialised early I added in the var L:A32NX_ENG_OUT_ACC_ALT  as the two simVars go together.  I would think that line 259 and 260 could be deleted. I have run a build without them, but lost my nerve on submitting the PR with them deleted in case they are required and I could not find where from.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
John-UK: The reason this may look a bit overdone is because the values can be input as a pair  1200/1200 or as indivuals 1200 or /1200. It is just to check how the data is being input. Then the checks on the validity of the values is checked.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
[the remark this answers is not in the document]
John-UK: Having thought about the inputs. It is possible that 1200/q200 could be entered. If the if block separated the two parts 1200 would be past as valid. When in fact the whole line should be returned as invalid format.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
John-UK: Having thought about the inputs. It is possible that 1200/q200 could be entered. If the if block separated the two parts 1200 would be past as valid. When in fact the whole line should be returned as invalid format.

John-UK: > I think this `if` block is a bit convoluted. I believe we can clean this up a bit by removing some redundant null checks on `thrRedAlt` and `accAlt` for example. Furthermore, it seems that this block does not check whether the altitudes are in the required range, and that this check is done further down. Are you sure this is intended?

I am doing a lot more pilot entry validation. Also exiting the function earlier and checking if can exit with more often so the wait and heavy lifting code to update the perf page is only done when it is needed.

async trySetThrustReductionAccelerationAltitude(s) {
        if (s === FMCMainDisplay.clrValue) {
            this.thrustReductionAltitudeIsPilotEntered = false;
            this.accelerationAltitudeIsPilotEntered = false;
            CDUPerformancePage.UpdateThrRedAccFromOrigin(this);
            return true;
        }
        if (s.length == 0 || s.length > 11) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }
        if (s.length > 6 && !/^\d{3,5}\/\d{3,5}$/.test(s)) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }

        let [thrRedAlt, accAlt] = s.split("/");

        if (thrRedAlt === undefined || thrRedAlt == "") {
            thrRedAlt = 0;
        }

        if (accAlt === undefined || accAlt == "") {
            accAlt = 0;
        }

        if (isNaN(thrRedAlt) || isNaN(accAlt)) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }

        if (thrRedAlt > 45000 || accAlt > 45000) {
            this.addNewMessage(NXSystemMessages.entryOutOfRange);
            return false;
        }

        if (thrRedAlt > 0 && !/^\d{3,5}$/.test(thrRedAlt)) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }

        if (accAlt > 0 && !/^\d{3,5}$/.test(accAlt)) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }

        let currentThrRedAlt = this.thrustReductionAltitude;
        let currentAccAlt = this.accelerationAltitude;
        let updateAccAlt = false;
        let updateThrRedAlt = false;

        const origin = this.flightPlanManager.getOrigin();
        let elevation = SimVar.GetSimVarValue("GROUND ALTITUDE", "feet");
        if (origin) {
            elevation = await this.facilityLoader.GetAirportFieldElevation(origin.icao);
        }

        const minimumAltitude = elevation + 400;

        if (thrRedAlt > 0) {
            thrRedAlt = parseInt(thrRedAlt);
            thrRedAlt = Math.round(thrRedAlt / 10) * 10;
            currentThrRedAlt = thrRedAlt;
        }

        if (accAlt > 0) {
            accAlt = parseInt(accAlt);
            accAlt = Math.round(accAlt / 10) * 10;
            currentAccAlt = accAlt;
        }

        if (thrRedAlt > 0 && (thrRedAlt < minimumAltitude || thrRedAlt > currentAccAlt)) {
            this.addNewMessage(NXSystemMessages.entryOutOfRange);
            return false;
        } else if (thrRedAlt > 0) {
            updateThrRedAlt = true;
        }

        if (accAlt > 0 && (accAlt < minimumAltitude || accAlt < currentThrRedAlt)) {
            this.addNewMessage(NXSystemMessages.entryOutOfRange);
            return false;
        } else if (accAlt > 0) {
            updateAccAlt = true;
        }

        if (updateThrRedAlt == true) {
            this.thrustReductionAltitude = thrRedAlt;
            this.thrustReductionAltitudeIsPilotEntered = true;
            SimVar.SetSimVarValue("L:AIRLINER_THR_RED_ALT", "Number", thrRedAlt);
        } else {
            this.thrustReductionAltitude = currentThrRedAlt;
            SimVar.SetSimVarValue("L:AIRLINER_THR_RED_ALT", "Number", currentThrRedAlt);
        }

        if (updateAccAlt == true) {
            this.accelerationAltitude = accAlt;
            this.accelerationAltitudeIsPilotEntered = true;
            SimVar.SetSimVarValue("L:AIRLINER_ACC_ALT", "Number", accAlt);
        } else {
            this.accelerationAltitude = currentAccAlt;
            SimVar.SetSimVarValue("L:AIRLINER_ACC_ALT", "Number", currentAccAlt);
        }
        return true;
    }

John-UK: PR updated to force pilot input to use 4 digits. eg if eng out acc is 900 then 0900 must be entered. This is IAW the manual.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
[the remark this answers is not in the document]
John-UK: I deleted these two lines no adverse effect seen.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
```diff
@@ -1071,7 +1077,7 @@ class CDUPerformancePage {
         mcdu.engineOutAccelerationAltitudeIsPilotEntered = false;
         SimVar.SetSimVarValue("L:A32NX_ENG_OUT_ACC_ALT", "feet", alt);
     }
-    static UpdateThrRedAccFromDestination(mcdu) {
+    static async UpdateThrRedAccFromDestination(mcdu) {
         const destination = mcdu.flightPlanManager.getDestination();
         const elevation = destination ? destination.altitudeinFP : SimVar.GetSimVarValue("GROUND ALTITUDE", "feet");
```
tracernz, line 1082:
```suggestion
        let elevation = SimVar.GetSimVarValue("GROUND ALTITUDE", "feet");
        if (destination) {
            elevation = await mcdu.facilityLoader.GetAirportFieldElevation(destination.icao);
        }
```

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -2338,82 +2338,129 @@ class FMCMainDisplay extends BaseAirliners {
...
             return true;
         }
+        if (s.length == 0 || s.length > 11) {
+            this.addNewMessage(NXSystemMessages.formatError);
+            return false;
+        }
+        if (s.length > 6 && !/^\d{4,5}\/\d{4,5}$/.test(s)) {
+            this.addNewMessage(NXSystemMessages.formatError);
+            return false;
+        }
```
tracernz, line 2355: Deal with the whole formatting issue with the one regex... job done.
```js
        const [match, thrRedAlt, slash1, , slash2, accAlt] = s.match(/^([0-9]{4,5})?(\/?)?((\/)([0-9]{4,5}))?$/) || [];
        if (!match || (slash1 && slash2) || (!thrRedAlt && !accAlt)) {
            this.addNewMessage(NXSystemMessages.formatError);
            return false;
        }
        
        // use thrRedAlt and accAlt
        // check if undefined first, which one will be if only one was entered
        // if not undefined parseInt is guaranteed to work because the regex already verified their format
```

John-UK: Thanks this looks very neat. I will put this in and test all likely pilot inputs against the code later today. Then update the PR.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_PerformancePage.js
John-UK: All destination GA reverted back to original code. The GA (destination) will be implemented when the secondary flight plan and GA track are complete.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
John-UK: Lovely code for checking the format, thanks.  I changed the thrRedAlt and accAlt to "checkThrRedAlt" & "checkAccAlt". This is so match is kept a const.  Then kept thrRedAlt and accAlt to store and change value as needed.
All the modifications have been check for the values and errors that I think would be input.

.github/CHANGELOG.md
```diff
@@ -107,6 +107,7 @@
 1. [FBW] Improved sideslip controller - @IbrahimK42 (IbrahimK42)
 1. [AP] Further improved ALT CST* and ALT CST conditions - @aguther (Andreas Guther)
 1. [MCDU] Added support for connecting MCDU to external devices - @tyler58546 (tyler58546)
+1. [MCDU] Fix Take off page THR RED ACC and ENG OUT ACC caluculations - @John-UK  (John-UK)
```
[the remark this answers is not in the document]
BlueberryKing: I don't think this is quite resolved yet 😄 It looks like there is still two spaces there.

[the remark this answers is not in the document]
John-UK: > 

I think this reverted back when I could not fetch upstream. Should be back to one space,

BlueberryKing (approved):

mico975:

Discord          : mico#3145
Object of testing: #6450 
Tier of Testing  : 1
Date             : 24/01/2022

Testing Process: 
Checked EFB for 1500 default - OK
Checked if empty perf TO page is displaying ----/---- and ---- - OK
Correct altituded calculated default + airport altitude - OK
Manual input - OK
Require 4 digits - OK
ACC !< THR RED - OK
THR RED !> ACC - OK
CLR the altidued and revert to auto calc - OK
Change INIT A for different departure airport, auto calc updates correctly - OK
Manually input altitudes and then change INIT A, altitudes not changing until CLR and then auto calc - OK
No lower than airfield + 400 and no higher than 45000 - OK, possible up to 45004 (intended round down to 45000), 45005 not posible (intended round up to 45010)

Tried this process C/D and on the runway. - both OK
Departed with auto calc and custom inputs - AP flies THR RED and ACC correctly
Engine out departure auto calc and manual input - OK, AP transitions to CLB phase correctly but doesn't accelerate after passing ENG OUT ACC altitude but after reaching the FCU set altitude, probably not yet implemented. 

Manipulating values after setting TO thrust not possible - OK


Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM, thanks for very detailed testing instructions :))

Media:       
image
Min and max possible, default ENG OUT ACC of 10+1500
image
ENG OUT ACC should probably not be allowed to be lower than THR RED, but probably not implemented yet.

John-UK:
Thanks for very detailed testing. 
Due to rounding a value of 45004 can be input which is then reduced to 45000.
Engine out would not have a thrust reduction as would require maximum continuous thrust. So this altitude is separate from the THR RED ACC pairing.

PR #6510, opened 2022-01-09
feat(pfd): PFD visual improvements

lukecologne:
## Summary of Changes

This PR:
* adds a box around the FD indication on the FMA, when the FD engagement status changes
* Adjusts the positioning of the FD indication text, and also small adjustments most other text fields, since as a result of the new font, they have been slightly shifted to the left
* Adjust some of the text's font sizes, for example the speed- and alttape graduation texts, and most FMA texts
* A/THR text on FMA will now have a smaller font size when A/THR is armed, but not active (i.e. when the text is colored cyan instead of white
* Change some lines (roll scale, V/S graduation, roll protection markers) stroke width to a new, smaller width
* Improve the appearence of the V/S needle, it now has a rounded linecap, originally it was cut off sharply at the border of the screen

The adjustments may be very hard to notice for some texts, and more noticeable for others.

## Screenshots (if necessary)

FD text boxing:
grafik

Rounded V/S linecap:
grafik

PFD with old font:
OldFont
PFD with new font, but before this PR:
NewFontBefore
PFD with new font, after this PR:
grafik

## References

grafik
grafik

AthrArmedActiveFont

## Additional context

## Testing instructions

* Observe the PFD in as many situations as possible, and see if any text looks weird or wrong, i.e. the positioning, size, or otherwise.
* Test that the other features mentioned in Summary of Changes work as described


beheh (approved):

lukecologne:
Sorry for adding all these changes after the fact, I promise I'm done now

tracernz (approved):
🚀

lukecologne:
Hm. Looks pretty good to me, can you specifically point out what looks wrong?

GlitchedYT:
It looks like the 1 is just the tiniest bit far to the right over the "A", and the 2 is the tiniest bit to the left over the "R". like i said though could just be perspective

lukecologne:
Here is a more close-up pic, looks more or less OK, maybe the 1 is a few pixels too far to the right, the 2 looks pretty much perfect, but all in all I think it's good enough for now
IMG_20210912_224753

lukecologne:
Ok, made some adjustments after looking at the ref pic again. Should be pretty much pixel perfect now, will push soon after looking at it in the sim

PR #6521, opened 2022-01-12
fix: update cargo hold limits to standard A320Neo configuration

sidnov:
Fixes #[issue_no]

## Summary of Changes

Update/revert changes merged in the previous W/B PR where the max payload carrying capacity was reduced below standard maximum available on an A320Neo. This would also help set a baseline for future W/B PRs such as the EFB UI and complex boarding and loading algorithms.

## References

image

## Additional context

## Testing instructions

Check if you are able to load a maximum of 9.4 tons of cargo


davidwalschots (approved):

flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/flight_model.cfg
```diff
@@ -37,7 +37,7 @@ activate_cg_limit_based_on_mach = 0 ; Activate cg limitation depending on mach v
 ; MLDW = 67400 kg/148591 lbs
 ; MZFW = 64300 kg/141757 lbs
 ; OEW = 42500 kg/93697 lbs
-; Max Payload = 21800 kg/48060 lbs
```
donstim, line 40: Why was this changed?  You can't have a payload that, when added to the OEW is more than the MZFW. So, max payload == 64300 kg - 42500 kg = 21800 kg. What this means if that you carry 174 passengers with bags at 104 kg/passenger, the cargo weight will be limited to 3704 kg. If you carry fewer passengers (or they weigh less), you can carry more payload.

Is this variable used anywhere?

frankkopp:

Discord: Cdr_Maverick#6475
Object of testing: #6521
Tier of Testing: 1
Date: 13.1.22

Testing Process: 
C&D at gate 


Negatives: 
Changing cargo over max or passengers resets the cargo value completely. See video [0]

Could overload max payload (see donstim's comment) by entering 174 pax and 9.4t cargo into MCDU. See screenshot [1]

Testing Results: Not passed

Media:
[0]
https://user-images.githubusercontent.com/16833201/149331738-64af1bba-5e0f-4ece-beec-055a4d59c1e7.mp4
[1]
image

Protocol:

<details>
Simbrief loading (max - used 174 Pax and 10t Freight "Payload/Cargo Limited y MZFW)

image

Simbrief loading (max - used 0 Pax and 10t Freight)
image

Simbrief loading (max - used 100 Pax and 7.4t Freight)
image

Manual loading (several combinations) - NOT OK see negatives

image


</details>

mehoygg:

Object of testing: 6521
Tier of Testing  : 1
Date             : 13/01/21

Testing Process: 
Spawned in C/D.  Started the plane via EXT Power to get power to the MCDU.
Went in and loaded 9.4Ts of Cargo into the Cargo Hold.
It shows up to 9.4T. Anything over, it states "Entry Out of Range."

Negatives:       
------

Testing Results: 
As per testing instructions, imo it Passed.  It shows the 9.4Ts rather than the before Value or around 3.X tons. 

Conclusions: 
As per testing instructions, it does Pass, imo.  Now if there is needed more testing via seeing a difference between PAX and Cargo for MTOW or any other variables, I am not testing that atm as I don't believe its apart of this PR.

Media:       

image

tracernz (dismissed):

flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/flight_model.cfg
sidnov: Pushed a fix for this, along with the hard limit for pax, bag and cargo not exceeding max payload

tracernz:
General question: why do we try to stop people from overloading? It should be possible to overload just as it's possible to overload the real aircraft, no?

sidnov:
> General question: why do we try to stop people from overloading (MZFW)? It should be possible to overload just as it's possible to overload the real aircraft, no?

Ideally, we should allow overloading. But the main reason for not allowing it with the current implementation is the overload scenarios can only be spotted in a few cases (when GW is visible/using Simconnect or Simvar Watcher tools), the rest of the time an off-limit C/G would just create problems for support-ops.

tracernz:
Okay, this can be improved when we have a proper load manager in the EFB.

donstim:
> General question: why do we try to stop people from overloading (MZFW)? It should be possible to overload just as it's possible to overload the real aircraft, no?

Ideally, maybe yes. But IRL there are systems/procedures in place to prevent overloading. In the sim, I'll bet most people don't even know that the MZFW is, so they would not know if they were overloading it. If they use Simbrief, Simbrief does this check for them and will offload cargo or passengers as necessary to prevent overloading. I've gone back and forth in my head about this myself, but I think we should mirror what Simbrief and IRL loading processes do and prevent overloading the airplane (for both MZFW and MTOW).

tracernz:
> Edit: Another option could be to put up a big warning somewhere that the load exceeds the MZFW (or MTOW) somewhere.

Yeah, we could do this in future when we have a proper load manager UI in the EFB. At least until then I agree it's best to limit the load.

frankkopp:
Update to docs:

Preview: https://docs-git-fork-frankkopp-updt-cargo-limit-pr6521-flybywire.vercel.app/fbw-a32nx/feature-guides/loading-fuel-weight/#a32nx-configuration
PR: https://github.com/flybywiresim/docs/pull/416

@sidnov @donstim could you give this a check as well?

donstim:
Although technically not a part of this PR, it would be nice to fix the fuel tank capacity information given as a comment in the flight_model.cfg file.  This is what it has currently:
; Max Fuel = 23723 ltrs (19045 kg)/6267 us gal (41987 lbs) @ specific gravity of 0.8 ISA conditions

The fuel tank capacities are specified in the flight_model.cfg file in terms of US gallons. MSFS then does its own conversions to liters, pounds, and kilograms, which may be different than are used to provide information in different units in type certification data sheets and airplane documents. We should be consistent with the MSFS unit conversions. I think the above line should be changed to:
; Max Fuel = 23721 ltrs (19046 kg)/6267 US gal (41989 lbs) @ specific gravity of 0.8 ISA conditions

(Note: All values other than the US gal quantity should change.) This would then be consistent with what is in the EFB as well. The Simbrief profile also needs to be changed.

For more information on how these values were determined using the MSFS Fuel and Payload page, see https://github.com/flybywiresim/a32nx/pull/3545#issuecomment-779526838

PR #6560, opened 2022-01-18
feat: improved thrust limits

aguther:
## Summary of Changes
This PR introduces improved thrust limits that are based on CN1 tables instead of N1 tables.

### Known issues
It can happen that N1 of the limit cannot be reached. This was also the case before this PR. The engine model will be further tuned in a separate PR.

## Testing instructions
- do several TO with calculated performance / known N1 limit (needs IRL reference)
- overall observe that thrust limits are not inversed (i.e. MCT larger than TO)


Taz5150 (commented):
Thrust limits defined as expected per FCOM tables. 

The discrepancies between actual N1 and thrust limits are due to the engine model itself (which needs adapting and fine tuning).

IbrahimK42 (approved):

PR #6640, opened 2022-02-03
fix: Remote MCDU CommandLine Improvements

frankkopp:
## Screenshots (if necessary)

- Fixes an issue with the "Press Key To Continue" function which did not pause the script before and started the server anyway. 
- Also users have reported that when starting the server with shortcuts there is no time to see any error messages as the command window closes too quickly. To improve this the "Press Key To Continue" function is called after any error message. 
- Margin now allows any value even negative ones. Limiting it to >0 is not required functionally  and user feedback was to remove this limitation. 
- Check to avoid users configuring http and websocket port to the same port. This was aught later with a "port in use" message already but this check explicitly tells the reason. 

## Testing instructions
- Provoke error messages with deliberately wrong command line options and check that Press Any Key comes up and actually stops starting the server.
  - https://docs.flybywiresim.com/fbw-a32nx/feature-guides/web-mcdu/#advanced-configuration
- Do this in a Command Line Terminal and in a Windows Shortcut.
- Use a negative margin - output will be ugly but the option should work
- Try to start the server with http and websocket port options identical (could implicitly by using the default of one of the two for the other)


2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #6640
Tier of Testing  : 1
Date             : 25/02/2022

Testing Process: 
Fixes an issue with the "Press Key To Continue" function which did not pause the script before and started the server anyway. - OK
Also users have reported that when starting the server with shortcuts there is no time to see any error messages as the command window closes too quickly. To improve this the "Press Key To Continue" function is called after any error message. - OK
Margin now allows any value even negative ones. Limiting it to >0 is not required functionally and user feedback was to remove this limitation. - OK
Check to avoid users configuring http and websocket port to the same port. This was aught later with a "port in use" message already but this check explicitly tells the reason. - OK

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #6658, opened 2022-02-06
fix(nd): Center 'TA ONLY' message

andrebrito16:
## Summary of Changes
Changed 'TA ONLY' message on ND to show on center. I know this is positioned to left due to ATSAW but while this feature is not implemented I checked with a lot of FBW users and they said that liked 'TA ONLY' on center.

## Screenshots (if necessary)
Microsoft Flight Simulator Screenshot 2022 02 06 - 09 24 55 62 Thumbnail

## References
I have talked with some brazilian A320Neo Pilots and they told me that 'TA ONLY' message shows on center. We also can see this on this [real  A320 NEO video](https://youtu.be/cv41hm4JEB8?t=158).

Discord username: **afbb#0987**

## Testing instructions
Just open aircraft and chose one situation when 'TA ONLY' message is displayed, liked before takeoff.

### DIFF
--- a/src/instruments/src/ND/elements/messages/TcasWxrMessages.tsx
+++ b/src/instruments/src/ND/elements/messages/TcasWxrMessages.tsx
@@ -1,7 +1,7 @@
 import React, { FC } from 'react';
 import { Layer } from '@instruments/common/utils';
-import { Mode } from '@shared/NavigationDisplay';
 import { useSimVar } from '@instruments/common/simVars';
+import { Mode } from '@shared/NavigationDisplay';
 
 /*
 Messages in priority order, from 1-12 (full set with ATSAW and nice weather radar)
@@ -18,12 +18,6 @@ Messages in priority order, from 1-12 (full set with ATSAW and nice weather rada
 [                 |     ADS-B (amber)     ]
 */
 
-enum TcasPosition {
-    Standby = 0,
-    Ta = 1,
-    TaRa = 2,
-}
-
 interface TcasWxrMessage {
     text: string;
     color: 'White' | 'Amber';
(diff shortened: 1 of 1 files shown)


src/instruments/src/ND/elements/messages/TcasWxrMessages.tsx
```diff
@@ -42,7 +43,7 @@ export const TcasWxrMessages: FC<{ modeIndex: Mode }> = ({ modeIndex }) => {
         leftMessage = { text: 'TA ONLY', color: 'White' };
```
beheh, line 43: This should probably be `centerMessage` instead?

andrebrito16: Fixed! :)

beheh (approved):
I think this makes sense just purely on the fact that AFAIK even newer EIS standards still support the non-ATSAW mode with the centered messages.

beheh:
> I checked with a lot of FBW users and they said that liked 'TA ONLY' on center.

Sidenote: This is not a good argument for a PR, especially as I can't find any such discussion in Discord. Always prefer actual references over anecdotal "I'd like XYZ".

andrebrito16:
> > I checked with a lot of FBW users and they said that liked 'TA ONLY' on center.
> 
> Sidenote: This is not a good argument for a PR, especially as I can't find any such discussion in Discord. Always prefer actual references over anecdotal "I'd like XYZ".

Ok, got it! Thank you!

Samueleonard:

Discord: 1samueleonard1
Object of testing: (#6658 )
Tier of Testing: (1)
Date: (06/02/2021)

Testing Process: 
. loaded on runway
. TA only was centered lol
Negatives: n/a
Testing Results: (Passed)

Conclusions: it works
Media: n/a

Benjozork:
Considering the ATSAW configuration is specified as part of our hardware spec, and that this is something that is meant to be implemented in the future, I don't really see the benefit of this change.

The ND was developed by studying specific FMS revisions and configurations, all slightly affecting display graphics.

I think we should stick with our configuration for now. When older planes are made or an option can be implemented, this would be part of it.

davidwalschots:
Overall I think there's agreement within the team to stick to one configuration for now. Therefore I'm going to close this PR. Should we ever include more configurations then of course this particular feature might be included at that time.

PR #6659, opened 2022-02-06
fix(bleed): Improve bleed system stability on long time steps

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

A few people reported issues of the blue hydraulic pump reporting a fault light on the blue pump. The fault light was explained by the entire bleed system being full of `NaN` values.

While the exact nature of this problem is not entirely clear, we suspect this to be an issue with large timesteps causing issues due to unphysically large pressure and temperature corrections. This PR limits the timestep size by splitting them up into smaller updates.

## Additional context

## Testing instructions

- Open the Local variable debugger window.
  - Enable dev mode -> Windows -> Behaviors -> Local Variables -> Filter for "PNEU"
- Make sure the bleed and the hydraulic system still works
  - Try starting an engine, enabling and disabling packs, bleed pushbuttons, things like that
  - If you see `nan` for any of the variables during the test, please tell me :)
- Report any noticeable performance degradation.
- Do an engine startup and shutdown


src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -12,6 +12,7 @@ use uom::si::{
 
 use systems::{
     accept_iterable,
+    hydraulic::update_iterator::MaxStepLoop,
```
crocket63, line 15: Should we move those loop updaters to shared?

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -128,8 +131,11 @@ pub struct A320Pneumatic {
     packs: [PackComplex; 2],
 }
 impl A320Pneumatic {
+    const PNEUMATIC_SIM_MAX_TIME_STEP: Duration = Duration::from_millis(33);
```
crocket63, line 134: Is the 33ms something you evaluated or just ported from hydraulics?
There are pretty fast things causing troubles in hydraulics, but I don't know what's the status for bleed?

Saying this because 33ms would cause any system running below 30fps to always run two bleed loops (and there are lot of people running less than 30fps including me :D).
If you think it's safe to have steps a bit higher, a 50ms updater would allow only one pass above 20fps, which is already less systems, or even 100ms for really lot of margin.

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: Good point, we probably should. I can do that 👍

src/systems/a320_systems/src/pneumatic.rs
BlueberryKing: Yeah, this is just something I copied over from hydraulics without too much thought. I will run some tests with 50ms and 100ms to see what impact this has.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -893,8 +903,13 @@ mod tests {
 
     #[test]
     fn pressurised_reservoir_behaves_like_open_valve() {
+        let mut electricity = Electricity::new();
```
crocket63, line 906: Its possible to get rid of all of these hacks in tests. Check the tests just added in the loop_updater we just removed those there

BlueberryKing: Oh nice, will add it to my list. I don't think this should be part of this PR.

davidwalschots (approved):

crocket63 (approved):

mico975:

Discord          : mico#3145
Object of testing: #6659 
Tier of Testing  : 1
Date             : 07/02/2022

Testing Process: 
Started C/D at KFBW and started plane up regularly.
Take off and landing. Shut down and restarted engines,
Did another circle and landed.

Hyd and bleed systems work.
No undefined values in localvar window.
No performance loss observed.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM, I have never had the issue before.

Media:       
image
After testing, with everything spooling down.

igor8518:
I feel that division by zero occurs somewhere, for example, division by "dt".

PR #6673, opened 2022-02-08
feat: improved pitch stability in turbulence

aguther:
## Summary of Changes
This PR increases gyroscopic pitch stability and therefore improves / makes the pitch reaction in turbulence more realistic. Up to now the plane has reacted very heavily in pitch in turbulence.

## Testing instructions
- fly through turbulent weather (i.e. a flight over the Alps in Europe with lot of winds)
- check if the remaining flight model is not negatively impacted


IbrahimK42 (approved):

donstim:
Test instructions seem a bit vague. When flying through turbulent weather (if you can find it), what should the outcome be? How can it be compared with the pre-PR situation?  Sorry for these questions, but it seems like this might be hard to QA as written. Is there anything specific they can look for in terms of finding turbulence (e.g., is it updrafts/downdrafts, changes in wind direction, changes in pressure or temperature?), and how should the airplane response be evaluated (e.g., changes in airspeed, altitude, vertical speed)?

aguther:
Unfortunately this is very much based on feelings.

The extreme pitch reactions on turbulence definitively feel not right to me, but I agree real data is missing.

So the most important test to me is the confirmation that the core flight model is not impacted.

PR #6689, opened 2022-02-10
feat(hyd): aerodynamic model for moving parts

crocket63:
Fixes #[issue_no]

## Summary of Changes

Implements an aerodynamic drag and lift model that can be applied to any moving part, be it cargo door, gear doors, control surfaces....

Any hydraulically controlled component will have to fight aerodynamic forces, providing a realistic feedback for all control surfaces.

- [x] Recreate local relative wind and add it to systems context
- [x] Basic drag model
- [x] Basic lift model
- [x] Refactoring aero forces calculations
- [x] Creating an "aero testbed" in unit tests
- [x] Rework of hydraulic controllers in edge case situations with massive aerodynamic forces
- [x] Tuning of control surface actuator

## Additional context

## Testing instructions

Standard operations:
1. Check that current aileron droop function on power off works properly
2. Check that aileron control when hydraulic power is available works properly (no wobble, positioning is correct)
3. Shutting down all hydraulic power in flight should result in aileron staying close to neutral position due to relative wind

Ludicrous operations:
   Use custom weather with editable wind
1. Wind from back of the plane should accelerate droop down if aileron surface is lower than horizontal position.
2. Wind from the back of the plane should droop an aileron UP if hydraulic power is lost while aileron is above horizontal position.
3. Same as 1. & 2. with plane upside down
4. Max wind (150kts) from any direction: aileron position control should still be satisying: good positioning. No lag. No wobbling.

### DIFF
--- a/src/systems/a320_systems_wasm/src/lib.rs
+++ b/src/systems/a320_systems_wasm/src/lib.rs
@@ -78,10 +78,14 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
     .provides_aircraft_variable("AIRSPEED INDICATED", "Knots", 0)?
     .provides_aircraft_variable("AIRSPEED MACH", "Mach", 0)?
     .provides_aircraft_variable("AIRSPEED TRUE", "Knots", 0)?
+    .provides_aircraft_variable("AMBIENT DENSITY", "Slugs per cubic feet", 0)?
     .provides_aircraft_variable("AMBIENT PRESSURE", "inHg", 0)?
     .provides_aircraft_variable("AMBIENT TEMPERATURE", "celsius", 0)?
     .provides_aircraft_variable("AMBIENT WIND DIRECTION", "Degrees", 0)?
     .provides_aircraft_variable("AMBIENT WIND VELOCITY", "Knots", 0)?
+    .provides_aircraft_variable("AMBIENT WIND X", "meter per second", 0)?
+    .provides_aircraft_variable("AMBIENT WIND Y", "meter per second", 0)?
+    .provides_aircraft_variable("AMBIENT WIND Z", "meter per second", 0)?
     .provides_aircraft_variable("ANTISKID BRAKES ACTIVE", "Bool", 0)?
     .provides_aircraft_variable("EXTERNAL POWER AVAILABLE", "Bool", 1)?
     .provides_aircraft_variable("FUEL TANK LEFT MAIN QUANTITY", "Pounds", 0)?
@@ -103,6 +107,7 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
     .provides_aircraft_variable("PLANE PITCH DEGREES", "Degrees", 0)?
     .provides_aircraft_variable("PLANE BANK DEGREES", "Degrees", 0)?
     .provides_aircraft_variable("PLANE HEADING DEGREES MAGNETIC", "Degrees", 0)?
+    .provides_aircraft_variable("PLANE HEADING DEGREES TRUE", "Degrees", 0)?
     .provides_aircraft_variable("PLANE LATITUDE", "degree latitude", 0)?
     .provides_aircraft_variable("PLANE LONGITUDE", "degree longitude", 0)?
     .provides_aircraft_variable("PUSHBACK STATE", "Enum", 0)?
@@ -118,6 +123,9 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
     .provides_aircraft_variable("TURB ENG CORRECTED N2", "Percent", 2)?
     .provides_aircraft_variable("TURB ENG IGNITION SWITCH EX1", "Enum", 1)?
     .provides_aircraft_variable("UNLIMITED FUEL", "Bool", 0)?
+    .provides_aircraft_variable("VELOCITY BODY X", "feet per second", 0)?
+    .provides_aircraft_variable("VELOCITY BODY Y", "feet per second", 0)?
+    .provides_aircraft_variable("VELOCITY BODY Z", "feet per second", 0)?
     .provides_aircraft_variable("VELOCITY WORLD Y", "feet per minute", 0)?
     .with_aspect(|builder| {
         builder.copy(
--- a/src/systems/systems/src/apu/mod.rs
+++ b/src/systems/systems/src/apu/mod.rs
@@ -1456,10 +1456,12 @@ pub mod tests {
 
         #[test]
         fn max_starting_egt_at_or_above_25000_feet_is_982_degrees() {
+            // We test using some decimals above 25000 because foot conversion of 25000 feet
+            // ends up being lower than 25000 feet in uom 0.32.0
             let mut test_bed = test_bed_with()
                 .starting_apu()
                 .and()
-                .indicated_altitude(Length::new::<foot>(25000.))
+                .indicated_altitude(Length::new::<foot>(25000.00001))
                 .run(Duration::from_secs(1));
 
             assert_about_eq!(
--- a/src/systems/systems/src/hydraulic/aerodynamic_model.rs
+++ b/src/systems/systems/src/hydraulic/aerodynamic_model.rs
@@ -1,0 +1,669 @@
+use nalgebra::{Rotation3, Vector3};
+
+use uom::si::{
+    area::square_meter, f64::*, force::newton, length::meter,
+    mass_density::kilogram_per_cubic_meter,
+};
+
+use crate::{shared::interpolation, simulation::UpdateContext};
+
+/// Projects a 3D box with 3D size along a projection vector.
+/// Outputs the projected area on the plane normal to projection vector
+fn area_projected(size: Vector3<Length>, projection_vector: Vector3<f64>) -> Area {
+    let norm_projection_vector = projection_vector.normalize();
+
+    Area::new::<square_meter>(
+        size[0].get::<meter>() * size[2].get::<meter>() * norm_projection_vector[1]
+            + size[0].get::<meter>() * size[1].get::<meter>() * norm_projection_vector[2]
+            + size[1].get::<meter>() * size[2].get::<meter>() * norm_projection_vector[0],
+    )
+}
+
+pub struct AerodynamicModel {
+    max_drag_normal: Vector3<f64>,
+    lift_axis: Vector3<f64>,
+    lift_normal: Vector3<f64>,
+
+    lift_area: Area,
+    drag_area: Area,
+
+    has_drag: bool,
+    has_lift: bool,
+}
+impl AerodynamicModel {
+    const DEFAULT_LIFT_ANGLE_MAP_DEGREES: [f64; 7] = [-30., -18., -10., 0., 10., 18., 30.];
+    const DEFAULT_LIFT_COEFF_MAP: [f64; 7] = [0., -0.5, -1.2, 0., 1.2, 0.5, 0.];
+
+    const DEFAULT_DRAG_COEFF: f64 = 1.2;
+
+    const MIN_WIND_SPEED_TO_COMPUTE_AERO_FORCES_M_S: f64 = 0.5;
+
+    pub fn new(
+        body: &impl AerodynamicBody,
+        max_drag_normal: Option<Vector3<f64>>,
+        lift_axis: Option<Vector3<f64>>,
+        lift_normal: Option<Vector3<f64>>,
+    ) -> Self {
+        let mut obj = Self {
+            max_drag_normal: max_drag_normal.unwrap_or_default(),
+            lift_axis: lift_axis.unwrap_or_default(),
+            lift_normal: lift_normal.unwrap_or_default(),
+
+            lift_area: Area::default(),
+            drag_area: Area::default(),
+            has_drag: max_drag_normal.is_some(),
+            has_lift: lift_axis.is_some(),
+        };
+        obj.initialize_areas(body.size());
+        obj
+    }
+
+    fn initialize_areas(&mut self, size: Vector3<Length>) {
+        self.drag_area = if self.has_drag {
+            area_projected(size, self.max_drag_normal)
+        } else {
+            Area::default()
+        };
+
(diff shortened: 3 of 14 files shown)


omrygin:
Since you calculate it in a custom way, can it cause consistency issues with the true force the sim uses?

crocket63:
> Since you calculate it in a custom way, can it cause consistency issues with the true force the sim uses?

Nope, they do not interact each other directly. The only link will be the effect our custom forces will have on our surfaces, thus impacting flight model if deflections of surfaces are impacted.

Of course I'd prefer doing nothing and just retrieve actual forces from msfs :p

omryg2:
> > Since you calculate it in a custom way, can it cause consistency issues with the true force the sim uses?
> 
> Nope, they do not interact each other directly. The only link will be the effect our custom forces will have on our surfaces, thus impacting flight model if deflections of surfaces are impacted.
> 
> Of course I'd prefer doing nothing and just retrieve actual forces from msfs :p

What I meant is that whether the force you calculate on the aileron would be consistent with the force the sim calculates to roll the plane. I don't know if that's a thing to care about, just a thought.

crocket63:
> > > Since you calculate it in a custom way, can it cause consistency issues with the true force the sim uses?
> > 
> > 
> > Nope, they do not interact each other directly. The only link will be the effect our custom forces will have on our surfaces, thus impacting flight model if deflections of surfaces are impacted.
> > Of course I'd prefer doing nothing and just retrieve actual forces from msfs :p
> 
> What I meant is that whether the force you calculate on the aileron would be consistent with the force the sim calculates to roll the plane. I don't know if that's a thing to care about, just a thought.

We really don't need something super precise, but it would be nice if we could at least be not too far from the real values. In nominal cases it won't change much if it's good or totally off, but in degraded hydraulic case it's there that we'd like to have the forces as close as possible to the real value. Dunno if we can find that kind of data...  Else we'll just hope my basic formulas are valid :D

tracernz:
It's a moot point anyhow as the sim doesn't provide such data... if it does in the future we can revisit.

.gitignore
```diff
@@ -61,3 +61,4 @@ src/instruments/buildSrc/custom/*
 /flybywire-aircraft-a320-neo/MCDU SERVER/
 /src/mcdu-server/client/build
 !/flybywire-aircraft-a320-neo/CORS SERVER/*
+flybywire-aircraft-a320-neo/html_ui/JS/sentry-client/sentry-client.js
```
beheh, line 64: This can be dropped again as it has been solved differently in master (it also has nothing to do with this PR).

src/systems/systems/src/apu/mod.rs
```diff
@@ -1459,7 +1459,7 @@ pub mod tests {
             let mut test_bed = test_bed_with()
                 .starting_apu()
                 .and()
-                .indicated_altitude(Length::new::<foot>(25000.))
+                .indicated_altitude(Length::new::<foot>(25000.01))
```
beheh, line 1462: I guess this requires a bit of explanation. I understand it has something with the UOM upgrade, but shouldn't we rather update the assertion below or even better the code behind this? I think reading the name of the test, we have failed the intent of this test now.

At the very least it deserves a comment.

src/systems/systems/src/simulation/update_context.rs
```diff
@@ -21,10 +26,15 @@ pub trait DeltaContext {
 pub struct Attitude {
     pitch: Angle,
     bank: Angle,
+    true_heading: Angle,
```
beheh, line 29: Just definition-wise I find it slightly questionable to consider true heading part of the attitude.

src/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -72,10 +72,14 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
             .provides_aircraft_variable("AIRSPEED INDICATED", "Knots", 0)?
             .provides_aircraft_variable("AIRSPEED MACH", "Mach", 0)?
             .provides_aircraft_variable("AIRSPEED TRUE", "Knots", 0)?
+            .provides_aircraft_variable("AMBIENT DENSITY", "Slugs per cubic feet", 0)?
             .provides_aircraft_variable("AMBIENT PRESSURE", "inHg", 0)?
             .provides_aircraft_variable("AMBIENT TEMPERATURE", "celsius", 0)?
             .provides_aircraft_variable("AMBIENT WIND DIRECTION", "Degrees", 0)?
             .provides_aircraft_variable("AMBIENT WIND VELOCITY", "Knots", 0)?
```
beheh, line 79: Do we still need these? If we do and they have issues, can we just switch to your new method?

beheh (changes requested):

src/systems/systems/src/apu/mod.rs
crocket63: Yeah I failed to investigate the exact cause, but I'd suspect something like added precision on the value of a foot in new uom package. The test being "at or above", it didn't felt as killing the test purpose though.
If my rust analyzer stops failing I might have a chance at finding the root cause....

src/systems/systems/src/simulation/update_context.rs
crocket63: Yeah agreed, I grouped the three here because you need those 3 to position the plane fully in space. I can just move heading at a single separate Angle indeed!

src/systems/a320_systems_wasm/src/lib.rs
crocket63: Not sure those have issues I did not test. Only used in ADIRS right now, I'd say we can keep them as reading those is simple enough, so we avoid recomputing those from our wind 3D vector.  I'll have a check when possible to make sure those 2 are indeed really valid ingame

src/systems/systems/src/hydraulic/aerodynamic_model.rs
```diff
@@ -0,0 +1,669 @@
...
+            + size[1].get::<meter>() * size[2].get::<meter>() * norm_projection_vector[0],
+    )
+}
+
+pub struct AerodynamicModel {
+    max_drag_normal: Vector3<f64>,
+    lift_axis: Vector3<f64>,
+    lift_normal: Vector3<f64>,
+
+    lift_area: Area,
```
aguther, line 27: I personally think it would be good to have unit names in the variable name to be clear what it is. but maybe Rust covers that already

aguther (commented):
LGTM overall:
- there are tests
- it seems to me that new code is embedded as other code before
- no obvious anti-patterns seen

But I say this as absolute non-rust developer and therefore no format approval. I hope someone with more rust background can look over it, too.

If no longer code review is happening it will get a go from my side though.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
crocket63: We use a package that hides units indeed.  Then you just do lift_area.get::<square_meter>() or any other unit

aguther (approved):
Just seen that Beheh already provided feedback and it was incorporated.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
```diff
@@ -0,0 +1,669 @@
...
+/// Outputs the projected area on the plane normal to projection vector
+fn area_projected(size: Vector3<Length>, projection_vector: Vector3<f64>) -> Area {
+    let norm_projection_vector = projection_vector.normalize();
+
+    Area::new::<square_meter>(
+        size[0].get::<meter>() * size[2].get::<meter>() * norm_projection_vector[1]
+            + size[0].get::<meter>() * size[1].get::<meter>() * norm_projection_vector[2]
+            + size[1].get::<meter>() * size[2].get::<meter>() * norm_projection_vector[0],
+    )
+}
```
davidwalschots, line 20: **suggestion**: Always consider what is most important for the person reading the code in this module. Should `area_projected` be the first thing they see, or are other functions or types more important to understand the module? I'd suggest moving this down in the file.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
```diff
@@ -0,0 +1,669 @@
...
+            lift_normal: lift_normal.unwrap_or_default(),
+
+            lift_area: Area::default(),
+            drag_area: Area::default(),
+            has_drag: max_drag_normal.is_some(),
+            has_lift: lift_axis.is_some(),
+        };
+        obj.initialize_areas(body.size());
+        obj
+    }
```
davidwalschots, line 59: **suggestion**: As all usages currently pass `Some` value, and you'll use the default if not provided, consider not having an `Option` type.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
```diff
@@ -0,0 +1,669 @@
+use nalgebra::{Rotation3, Vector3};
```
davidwalschots, line 1: **question**: Should this module be a child of the `hydraulic` module? Is it only related to hydraulics or can these types be reused for other means?

src/systems/systems/src/simulation/update_context.rs
```diff
@@ -224,19 +345,72 @@ impl UpdateContext {
...
+    /// speed is reached.
+    /// As we want the real relative wind including standing still on ground we recompute it here.
+    ///
+    /// World coordinate wind is first rotated to local plane frame of reference
+    /// Then we substract local plane velocity to obtain relative local wind velocity.
+    ///
+    /// X axis positive is left to right
+    /// Y axis positive is down to up
+    /// Z axis positive is aft to front
+    fn update_relative_wind(&mut self) {
```
davidwalschots, line 394: **suggestion**: Add tests for this function.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -47,6 +50,16 @@ pub enum LinearActuatorMode {
...
+impl Display for LinearActuatorMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            LinearActuatorMode::ClosedValves => write!(f, "ClosedValves"),
+            LinearActuatorMode::PositionControl => write!(f, "PositionControl"),
+            LinearActuatorMode::ActiveDamping => write!(f, "ActiveDamping"),
+            LinearActuatorMode::ClosedCircuitDamping => write!(f, "ClosedCircuitDamping"),
+        }
+    }
+}
```
davidwalschots, line 62: **thought**: Seems to be unused; implementation seems to be equal to deriving from `Debug`.
**suggestion**: Remove this implementation. Derive from `Debug` if you need `println!` capabilities. Note that `Display` is for user-facing output.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -356,15 +370,26 @@ impl CoreHydraulicForce {
     }
 
     /// Computes a hydraulic flow request based on required actuator position and its current position
-    /// Flow is computed through the formula flow = position_error^3 * Gain
+    /// Flow is computed through the formula flow = position_error^2 * maxflow / OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW^2
+    /// This formula means max flow will be applied above OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW position error
+    /// Below OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW position error, a squared law defines flow vs positioning error
+    ///
```
davidwalschots, line 376: **suggestion**: Refer to code elements such that `cargo doc` can detect broken documentation.

```suggestion
    /// Flow is computed through the formula flow = position_error^2 * maxflow / [Self::OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW]^2.
    /// This formula means max flow will be applied above [Self::OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW] position error.
    /// Below [Self::OPEN_LOOP_POSITION_ERROR_FOR_MAX_FLOW] position error, a squared law defines flow vs positioning error.
    ///
```

davidwalschots: The documentation does seem to be repeating what's in the code. It might be the documentation isn't adding any benefit over reading the code itself.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
crocket63: As a clever piece of trick from @omryg2 it should deserve to be on top :D  But yeah I can have it lower in the file :p

src/systems/systems/src/hydraulic/aerodynamic_model.rs
crocket63: Generally speaking it could be used for other things, but for now don't really see any use case for other modules than hydraulics. 
Any evil ideas there? :p

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Only println! needed there indeed.

src/systems/systems/src/hydraulic/aerodynamic_model.rs
crocket63: I'd rather keep the option just for now until some more tests are done on spoilers and gears. Lift might not be needed there but I'm still unsure :p

src/systems/systems/src/hydraulic/aerodynamic_model.rs
davidwalschots: Nope. We could always move it later when we see a use for it elsewhere.

src/systems/systems/src/hydraulic/linear_actuator.rs
davidwalschots: I'd like to note that adding `Debug` as part of `#[derive(PartialEq, Clone, Copy, Debug)]` will result in the same output as your current custom implementation. Probably you want to remove it?

crocket63: Always forget about that macro sorcery ..... Thanks!

davidwalschots (approved):

PR #6698, opened 2022-02-11
fix: ALT CST and DES logic

aguther:
## Summary of Changes
This PR fixes an issue when flying in ALT CST and the altitude constraint jumping to the FCU altitude. Now the DES mode is correctly engaged, and the aircraft descends to the FCU altitude. When nothing further changes, it will capture the altitude with ALT* and will end in ALT.

The broken behavior before was that DES only disarmed and the plane incorrectly remained in ALT CST. This error only happened when FCU altitude and new altitude constraints were equal. This applied only to DES and not to CLB.

## Testing instructions
- test the above-described scenario:
  - set 6000 ft on the FCU altitude and climb/descend to it
  - engage NAV
  - set an altitude constraint with +5000 ft
  - set FCU altitude to 4000 ft
  - push ALT knob to engage DES
  - plane should capture the constraint of 5000 ft (and end in ALT CST)
  - for next waypoint set altitude constraint to +4000 ft
  - wait for waypoint to cycle
  - plane should engage DES when altitude constraint changes
  - plane should capture 4000 ft in ALT* and end in ALT

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -146,6 +146,7 @@
 1. [ELEC] Fix emergency elec on init in flight wrongly triggering the RAT - @Crocket63
 1. [ATHR] Fix ATHR Speed Undershoot - @IbrahimK42 (IbrahimK42)
 1. [FLIGHTMODEL] Improved pitch stability in turbulence - @aguther (Andreas Guther)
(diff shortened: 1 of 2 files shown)


IbrahimK42 (approved):
Tested and LGTM. Autopilot behaves as described.

PR #6701, opened 2022-02-11
feat: w&b integration with advanced Simbrief API

sidnov:
Fixes #[issue_no]

## Summary of Changes

Makes PAX_WEIGHT and BAG_WEIGHT dynamic to support the updated Simbrief API's ``<weights>`` section.
We are now able to pull separated weight values that can be customised on individual Simbrief airframes as per pilot/company requirements. 
Default values are set to 84KGs (Pax weight+Cabin luggage) and 20KGs (check-in baggage per pax). Option to customize the default weight values without using Simbrief would be planned for the EFB load management interface.

## Screenshots (if necessary)

image
image

## Additional context

## Testing instructions

1. Spawn C/D
2. Prepare a Simbrief Flightplan with the new airframe or use your own custom airframe.
3. Launch devmode->behaviors->local variables and filter using "WB"
4. Observe that your pax and bag weight from Simbrief is being loaded correctly on pressing OFP REQUEST
5. Input pax numbers manually to see the weights change to default (84,20)
6. Start boarding and observe the payload value after completion. Cross check with Simbrief OFP values. 


flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
```diff
@@ -70,13 +70,21 @@ class CDUAocOfpData {
...
                     await SimVar.SetSimVarValue(`L:${station.simVar}_DESIRED`, "Number", value);
                     await setTargetCargo(value, '');
                     updateView();
                 }
             );
         }
 
+        async function setDefaultWeights(simbriefPaxWeight, simbriefBagWeight) {
+            const paxWeight = (simbriefPaxWeight === 0) ? 84 : simbriefPaxWeight;
+            const bagWeight = (simbriefBagWeight === 0) ? 20 : simbriefBagWeight;
```
donstim, line 83: Shouldn't these lines be if the corresponding simbrief weights are NaN, i.e., simbrief was not used? If someone wants to set the passenger weight or per passenger baggage weight to zero in order to get some desired payload number, shouldn't we be using that zero value?  Or am I misunderstanding what these lines do?

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
```diff
@@ -99,7 +107,7 @@ class CDUAocOfpData {
         }
 
         async function setTargetCargo(numberOfPax, simbriefCargo) {
-            const bagWeight = numberOfPax * 20;
+            const bagWeight = numberOfPax * BAG_WEIGHT;
             const maxLoadInCargoHold = 9435; // from flight_model.cfg
             const maxTotalPayload = 21800; // from flight_model.cfg
             let loadableCargoWeight = undefined;
```
donstim, line 113: Does something need to be changed in the loadable cargo weight loop to allow for the case that someone can pick a per passenger bag weight high enough that the bag weight alone can exceed the max cargo load hold, or when combined with the passenger weight, the max total payload?

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
sidnov: The thing here is we are using the same function to set default weights in both scenarios. When taking values from Simbrief, it is never a NaN, and when manually inputting values, we cannot modify the per pax/bag values till we have further changes on the EFB side. As a fallback we simply set the values to our current default values

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
sidnov: When loading manually: the per pax/bag values are still the same as before, which wouldn't allow it to exceed
When loading via Simbrief using custom per pax/bag values: the OFP generation takes care of excess load as mentioned by Derek from SB- " But beware: In case the aircraft is in an overweight situation and payload has to be removed to respect the MTOW or MZFW, cargo will be the first thing that gets removed. First, <freight_added> gets removed, and if that isn't enough, baggage weight goes next. So in these cases it is possible for (<pax_count> * <bag_weight>) to not match up with the <cargo> value. If you need to, you can perform (<cargo> / <bag_weight>) to figure out approximately how many bags actually remain on board. "

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
donstim: Ahh, of course.

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/ATSU/A320_Neo_CDU_AOC_OFPData.js
sidnov: > But Simbrief doesn't know the max cargo hold load, only the MTOW or MZFW.

Yes, but Simbrief also added something which restricts the max per baggage weight. If trying to set unrealistic values, this is what is displayed on the airframe page
image

sidnov: @donstim I thought about it again, and this seems like a cleaner way. We restrict max weight of baggage and freight in cargo hold to 9435KGs and not max payload (21800KGs). Have pushed the changes accordingly

donstim:
@sidnov There is something strange going on when you try to load 16 passengers or more. Suddenly there seems to be some kind of calculation of roundoff error occurring in the load station simVars that doesn't happen with 15 or fewer passengers. Here is a screenshot with 15 passengers loaded (doesn't matter if it is 15 spread amongst the rows by the loading algorithm or 15 put specifically in one row):
Screenshot (1827)

Now, see what happens with 16 passengers:
Screenshot (1828)
Why is each passenger now 83.914779 kg instead of 84 kg like they were with 15 passengers? Try 174 passengers and the roundoff error becomes significant.

I also have a question regarding manually adding cargo weight. If I load passengers only, there is nothing added to the cargo hold for their bags. But if I add cargo, it then adds passengers bags to the cargo weight I've requested. Is this intentional?

sidnov:
> @sidnov There is something strange going on when you try to load 16 passengers or more. Suddenly there seems to be some kind of calculation of roundoff error occurring in the load station simVars that doesn't happen with 15 or fewer passengers. 
Will check this soon

donstim:
I am assuming this is caused by the same issue: https://github.com/flybywiresim/a32nx/issues/6802

John-UK:
Does this change affect the calculation for CG? With the hard coded 84-20 the automatic pax distribution resulted in a CG at/near 25.

sidnov:
> Does this change affect the calculation for CG? With the hard coded 84-20 the automatic pax distribution resulted in a CG at/near 25.

In the current logic, we focus on uniform loading, and not a CG-targeted loading. However, this will be rewritten with flyPadOS v3 as we'll remove the WB page and logic completely from the MCDU

davidwalschots:
@sidnov What's your current expectation on the next step in this PR? Are you awaiting code review?

sidnov:
> @sidnov What's your current expectation on the next step in this PR? Are you awaiting code review?

Affirmative. Once the basic changes are in from this PR, I've been planning along with Erick to refactor and use the base for the upcoming EFB page. Also have a bunch of feedback from ghosh9691 (who has also contributed to the changes earlier to the W&B systems) through Discord to consider in the upcoming PR.

For now, this is ready to patch all the existing bugs and be the bridge between the upcoming release version and the future release version with flyPadOS v3

davidwalschots:
I see. I think this statement might have stopped people from reviewing thus far:

> Will check this soon

I'll tag the PR to make it extra clear this is ready 😀.

sidnov:
> I see. I think this statement might have stopped people from reviewing thus far:
> 
> > Will check this soon
> 
> I'll tag the PR to make it extra clear this is ready 😀.

Should've mentioned that particular issue is something that's happening on the sim-side, although I've requested reviews a number of times through dev-chat, especially when talks of a new release come up, since the current W/B page isn't too robust IMO. Thanks for the label though😄

donstim:
I confirm the issue appears to be something occurring on the sim side. It was present before this PR. I was hoping it would be resolvable by something that could be changed within this PR, but it looks like that is not the case.  This PR should not be held up for that.

beheh (approved):
How about still adding a CHANGELOG.md entry?

sidnov:
> How about still adding a CHANGELOG.md entry?

I had it in mind, will add it once back from vacation tomorrow

mico975:

Discord          : mico#3145
Object of testing: #6701
Tier of Testing  : 1
Date             : 21/03/2022

Testing Process: 
Observe that your pax and bag weight from Simbrief is being loaded correctly on pressing OFP REQUEST - **OK**
Input pax numbers manually to see the weights change to default (84,20) - **OK**
Start boarding and observe the payload value after completion. Cross check with Simbrief OFP values. - **OK**, but ZFW 300kg too low.

Negatives:       
Payload is correct but ZFW is 300kg off - see media.
Simbrief profile used
https://www.simbrief.com/system/dispatch.php?sharefleet=439746_1638100543721

Testing Results: 
Passed, payload is correct but ZFW is not.

Conclusions: 
LGTM with a small

Media:       
image
image
image
image

sidnov:
> Testing Results: Passed, payload is correct but ZFW is not.

Had the chance to test this, it is related to the loading issue we were discussing earlier here, coming from the sim's side <https://github.com/flybywiresim/a32nx/pull/6701#issuecomment-1053231241>

donstim:
> > Testing Results: Passed, payload is correct but ZFW is not.
> 
> Had the chance to test this, it is related to the loading issue we were discussing earlier here, coming from the sim's side [#6701 (comment)](https://github.com/flybywiresim/a32nx/pull/6701#issuecomment-1053231241)

This particular issue can be "fixed" though, can't it?  I think he was noting just the difference between the ZFW on the MCDU Fuel Pred page vs the ZFW on the MCDU ATSU W&B page. The one on the FUEL PRED page is screwed up because it is using the airplane total weight simvar minus the fuel load. The total weight simvar is wrong because of the weird sim error. If that page used the ZFW from the MCDU ATSU W&B page directly, then the 2 ZFWs would agree, and this very noticeable issue would be fixed. There would still be the issue that the individual loading station weights would differ from the MSFS individual loading station weights, so the payload, ZFW, and airplane total weight would differ from the MSFS values, but these differences would be pretty transparent to the user. Only those who know how to read the simvars would be able to spot the differences.

sidnov:
> This particular issue can be "fixed" though, can't it? I think he was noting just the difference between the ZFW on the MCDU Fuel Pred page vs the ZFW on the MCDU ATSU W&B page. The one on the FUEL PRED page is screwed up because it is using the airplane total weight simvar minus the fuel load. The total weight simvar is wrong because of the weird sim error. If that page used the ZFW from the MCDU ATSU W&B page directly, then the 2 ZFWs would agree, and this very noticeable issue would be fixed. There would still be the issue that the individual loading station weights would differ from the MSFS individual loading station weights, so the payload, ZFW, and airplane total weight would differ from the MSFS values, but these differences would be pretty transparent to the user. Only those who know how to read the simvars would be able to spot the differences.

This imo is the best we can do for now, have pushed the changes required so we display the OFP values for ZFW instead of actuals onboard.

mico975:
Updated the test report above, correction implemented.

PR #6736, opened 2022-02-16
feat(navigation): add radio altimeter systems

### LINKED ISSUE #5883: Take radio altimeter device location into account on rotation
### Aircraft Version

Well, actually every version

### Request Describe

A320's Low Range Radio Altimeter device is at bottom of the aircraft, near tail side.
let me explain it, coz  PFD shows 0(Radio Altimeter's height) before you rotate, and it still shows 0 when you rotate, 
well. it turns to be 1 , 2 ... after you rotate(of course, coz plane has off the ground).
the problem is when you rotate the plane, it still shows 0( if plane is still running on the ground).
it should be showed -1 when you rotate the plane, coz radio altimeter device is under the aircraft near its tail.
when you rotate, nose goes up, but tail goes down. so does that radio altimeter device. that's the exact reason why it should be showed -1 in the bottom of PFD.
yeah, A32nx's radio altimeter device is at the bottom of the aircraft but not close enough to its tail.
Flybywire guys and pilots of MSFS, I ain't here messing around, I wanna do something helpful that improve A32nx to be the best in future. sorry about my English, I ain't good at speaking English, but I will do my best to explain to you all if you have questions about my report.
thank you, I wish everybody the best.

beheh:
Fixes #5883

## Summary of Changes

This PR introduces a proper dual radio altimeter system featuring more realistic readings during (de)rotation, an accurate refresh rate, and ARINC 429 outputs. It improves the accuracy of the PFD display, and puts a system in place ready for the upcoming FWC to use.

The new radio altimeters are based on the Honeywell ALA-52B, require their appropriate bus to be powered and can be failed using the standard failure system in the EFB. They take a few seconds to start up after being powered.

The existing PFD, FWC, TCAS and GPWS have been updated to use the new radio altimeter sources, including fallback to the other system on failure. Some systems like TCAS will fail outright if no RA is functioning.

## Screenshots (if necessary)

Video of new rotation: https://streamable.com/zpb7j3

image
image
image

## References

- https://fccid.io/AOIALA-52B2/User-Manual/Users-Manual-901606.pdf
- https://avsi.aero/wp-content/uploads/2021/12/Radar-Altimeter-Overview-of-Design-and-Performance.pdf
- https://www.sciencedirect.com/science/article/pii/S1474667015350436/pdf
- https://www.youtube.com/watch?v=ECV3IBkzZLc#t=25s

## Additional context

The RAs have a small startup delay after being powered, usually around 5 seconds, but never more than 6. They are hooked up to AC BUS 1 (RA 1) and AC BUS 2 (RA 2), so if you unpower and repower these busses it will take a moment for the RAs to come online. You might get an intermittent RA 1(2) failure after repowering a bus, this is expected as the RA isn't emitting data until it's completely started and ready to go.

Changes were made to the following systems:
- PFD
  - RA annunciation at bottom of PFD
  - Rising Runway symbol near center of PFD
  - Red ribbon in ALT band
  - Threshold for V/S indicator border to change color to amber (easier to trigger amber box near with lower V/S the ground)
- TCAS: uses the new Radio Altimeters to inhibit "Descend" RAs.
- FWC: uses the new Radio Altimeters to calculate flight phases, most noticeable throug T.O INHIBIT and LDG INHIBIT
- GPWS: uses the new Radio Altimeters to generate height callouts and GPWS callouts ("Too Low, Terrain")
- ECAM (EWD) and Status Page: In some cases the FAIL messages are inhibited if they're due to another obvious failure. For example, RA 1 FAULT is inhibited when AC BUS 1 is unpowered, and TCAS FAULT is inhibited when both RAs are unavailable. However, affected systems will always appear on the status page.

The following systems have **not** been touched and are still using the SimVars directly, and will not respond to the new behaviour:
- FCU (special behaviour below 30ft)
- FBW & APs (Flare Law, Autoland etc.)

While this PR doesn't fully model all the impacts and fully accurate ECAM messages, it nevertheless puts the system in place ready to be consumed by the new FWC later.

## Testing instructions

Before testing, please make sure to read the list of affected systems above.

With RAs normal:
1. On ground, observe values near 0. Some slight discrepancy between the RAs (around 1ft) is allowed.
2. During takeoff at rotation, observe the radio altitude on both PFDs going briefly negative (an extra -1 to -2ft). The F/O PFD is slightly more sensitive to this, as the RAs are further at the back.
3. When flying at high altitudes, ensure TCAS stays available even while at extreme attitudes. You should not see TA ONLY pop up at high altitudes just because you entered an extreme attitude.
4. Make sure TCAS can still be triggered both at high altitudes and if possible, near the ground (Radio Altimeter should inhibit "Descend" at very low altitudes between 1000ft and 2350ft - however TCAS does never issue RAs below 1000ft) - do note however that flying at extreme attitudes near the ground may cause a false reading, and the TCAS would then issue a "Descend" regardless (no inhibitions).
5. When flying over rugged terrain or sharp cliffs, observe the slightly smoothed radio altitude indications on the PFD. There should be no extreme jumps, everthing should be smoothly animated. This includes:
    - RA value at the bottom of the PFD
    - Rising Runway in the center of the PFD near the ground
    - Red ribbon in ALT band
6. When flying close to the terrain, observe the usual GPWS callouts ("Too Low, Terrain") unless in an extreme attitude
7. During approach, observe the LDG MEMO appearing as usual (below 2000ft AGL)
8. During final approach, observe normal "Minimum" callouts (when using RADIO)
9. When landing normally, observe near 0ft RA annunciation on the PFDs at touchdown. During and after derotation you may see 1ft briefly, especially as the aircraft is lurching forward during decellaration and the RAs are slightly farther away from the ground.

With one RA failed:
*RAs can be failed using the EFB's failure page*
1. Observe NAV RA 1(2) FAULT on ECAM - this may be inhibited in certain flight phases
2. All behaviours should be as above, because the systems fall over to the non-failed RA
3. If an RA fails and you unfail it, it'll take approximately 5-6 to come back. If an RA has only failed because it lost power, it'll come back instantly as long as power is restored within about 30 seconds. After that it'll take 5-6 seconds for the full startup.

With dual RA failure:
1. Observe the same warnings as above
2. Observe a flashing RA flag on the PFD when below the transition altitude - note: If you don't have a transition altitude set, it won't flash.
3. Observe a flashing TCAS flag on the PFD, amber TCAS on the ND, and NAV TCAS FAULT on ECAM - this may be inhibited in certain flight phases. This is a knock-on effect of the TCAS failing without RAs.
4. LDG MEMO will only appear once gear is downlocked
5. No Minimums callout when landing with RADIO minimum set, BARO minimum should be unaffected

### DIFF
--- a/src/instruments/src/PFD/VerticalSpeedIndicator.tsx
+++ b/src/instruments/src/PFD/VerticalSpeedIndicator.tsx
@@ -3,11 +3,12 @@ import { useSimVar } from '@instruments/common/simVars';
 import React from 'react';
 
 interface VerticalSpeedIndicatorProps {
-    radioAlt: number,
     verticalSpeed: Arinc429Word
+    radioAltitude: Arinc429Word,
+    filteredRadioAltitude: number,
 }
 
-export const VerticalSpeedIndicator = ({ radioAlt, verticalSpeed }: VerticalSpeedIndicatorProps) => {
+export const VerticalSpeedIndicator = ({ verticalSpeed, radioAltitude, filteredRadioAltitude }: VerticalSpeedIndicatorProps) => {
     if (!verticalSpeed.isNormalOperation()) {
         return (
             <>
@@ -22,9 +23,14 @@ export const VerticalSpeedIndicator = ({ radioAlt, verticalSpeed }: VerticalSpee
     }
 
     const absVSpeed = Math.abs(verticalSpeed.value);
+    const radioAltitudeValid = !radioAltitude.isNoComputedData() && !radioAltitude.isFailureWarning();
 
     let isAmber = false;
-    if (absVSpeed > 6000 || (radioAlt < 2500 && radioAlt > 1000 && verticalSpeed.value < -2000) || (radioAlt < 1000 && verticalSpeed.value < -1200)) {
+    if (
+        absVSpeed >= 6000
+        || (verticalSpeed.value <= -2000 && radioAltitudeValid && filteredRadioAltitude <= 2500 && filteredRadioAltitude >= 1000)
+        || (verticalSpeed.value <= -1200 && radioAltitudeValid && filteredRadioAltitude <= 1000)
+    ) {
         isAmber = true;
     }
 
--- a/src/systems/a320_systems/src/lib.rs
+++ b/src/systems/a320_systems/src/lib.rs
@@ -4,6 +4,7 @@ mod air_conditioning;
 mod electrical;
 mod fuel;
 mod hydraulic;
+mod navigation;
 mod pneumatic;
 mod power_consumption;
 
@@ -16,10 +17,11 @@ use electrical::{
     A320Electrical, A320ElectricalOverheadPanel, A320EmergencyElectricalOverheadPanel,
     APU_START_MOTOR_BUS_TYPE,
 };
-
 use hydraulic::{A320Hydraulic, A320HydraulicOverheadPanel};
+use navigation::A320RadioAltimeters;
 use power_consumption::A320PowerConsumption;
 use systems::simulation::InitContext;
+
 use systems::{
     apu::{
         Aps3200ApuGenerator, Aps3200StartMotor, AuxiliaryPowerUnit, AuxiliaryPowerUnitFactory,
@@ -63,6 +65,7 @@ pub struct A320 {
     pressurization: Pressurization,
     pressurization_overhead: PressurizationOverheadPanel,
     pneumatic: A320Pneumatic,
+    radio_altimeters: A320RadioAltimeters,
 }
 impl A320 {
     pub fn new(context: &mut InitContext) -> A320 {
@@ -106,6 +109,7 @@ impl A320 {
             pressurization: Pressurization::new(context),
             pressurization_overhead: PressurizationOverheadPanel::new(context),
             pneumatic: A320Pneumatic::new(context),
+            radio_altimeters: A320RadioAltimeters::new(context),
         }
     }
 }
@@ -162,6 +166,9 @@ impl Aircraft for A320 {
             &self.landing_gear,
             self.ext_pwr.output_potential().is_powered(),
         );
+
+        self.radio_altimeters.update(context);
+
         self.pressurization.update(
             context,
             &self.pressurization_overhead,
@@ -236,6 +243,7 @@ impl SimulationElement for A320 {
         self.ext_pwr.accept(visitor);
         self.lgciu1.accept(visitor);
         self.lgciu2.accept(visitor);
+        self.radio_altimeters.accept(visitor);
         self.autobrake_panel.accept(visitor);
         self.hydraulic.accept(visitor);
         self.hydraulic_overhead.accept(visitor);
--- a/src/systems/systems/src/navigation/radio_altimeter.rs
+++ b/src/systems/systems/src/navigation/radio_altimeter.rs
@@ -1,0 +1,64 @@
+use crate::shared::arinc429::Arinc429Word;
+use uom::si::f64::{Length, Time};
+
+/// This struct describes the installation of a physical Antenna in relation to an aircraft's
+/// center of gravity.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct AntennaInstallation {
+    y: Length, // negative towards top (from CG), positive towards bottom (from CG)
+    z: Length, // negative towards front (from CG), positive for aft
+    electric_length: Length, // The electric length of the cables and antenna
+}
+
+impl AntennaInstallation {
+    pub fn new(y: Length, z: Length, electric_length: Length) -> Self {
+        Self {
+            y,
+            z,
+            electric_length,
+        }
+    }
+
+    pub fn y(&self) -> Length {
+        self.y
+    }
+
+    pub fn z(&self) -> Length {
+        self.z
+    }
+
+    pub fn electric_length(self) -> Length {
+        self.electric_length
+    }
+}
+
+/// This trait describes a transceiver pair as used in radio altimeters, usually in the C band.
+/// It models the physical transmission of the signal between the transceivers and returns the
+/// response.
+pub trait TransceiverPair {
+    fn response(&self) -> Option<TransceiverPairResponse>;
+}
+
+/// This struct encodes a response from a pair of transceivers. Currently it only encodes a travel
+/// time, which in a modulated continuous wave radar (FMCW) can already be considered the filtered
+/// duration between the transmission and the response. In the future it may also return the
+/// remaining signal strength, which can be used to model the signal loss over various surfaces, or
+/// model an actual frequency response.
+#[derive(Debug, Clone, PartialEq)]
+pub struct TransceiverPairResponse {
+    travel_time: Time,
+}
+
+impl TransceiverPairResponse {
+    pub fn new(travel_time: Time) -> Self {
+        Self { travel_time }
+    }
+
+    pub fn travel_time(&self) -> Time {
+        self.travel_time
+    }
+}
+
+pub trait RadioAltimeter {
+    fn radio_altitude(&self) -> Arinc429Word<Length>;
+}
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -173,6 +173,7 @@
 1. [MCDU] Added button press sound to remote mcdu  - @frankkopp (Cdr_Maverick#6475)
 1. [MCDU] Improved Remote MCDU with CLR hold and light-dark mode - @frankkopp (Frank Kopp)
 1. [MCDU] Improved mcdu server command line handling  - @frankkopp (Cdr_Maverick#6475)
+1. [RA] Add dual Radio Altimeter simulation - @beheh (Benedict Etzel)
 
 ## 0.7.0
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The following aircraft configuration is currently simulated or targeted:
  APU         APS3200
  FMS         Honeywell Release H3
  FWC Std.    H2F9C
+ RA          Honeywell ALA-52B
  TAWS        Honeywell EGPWS
  ACAS        Honeywell TPA-100B
  ATC         Honeywell TRA-100B
--- a/docs/a320-simvars.md
+++ b/docs/a320-simvars.md
@@ -15,6 +15,7 @@
 1. [Pneumatic](#pneumatic)
 1. [Landing Gear (ATA 32)](#landing-gear-ata-32)
 1. [ATC (ATA 34)](#atc-ata-34)
+1. [Radio Altimeter (ATA 34)](#ra-ata-34)
 
 ## Uncategorized
 
@@ -2518,3 +2519,12 @@ In the variables below, {number} should be replaced with one item in the set: { 
     - {number}
         - 0
         - 1
+
+## Radio Altimeter (ATA 34)
+
+- A32NX_RA_{number}_RADIO_ALTITUDE
+    - `Arinc429Word<Feet>`
+    - The height over ground as measured by the corresponding radio altimeter towards the aft of the aircraft
+    - {number}
+      - 0
+      - 1
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_FWC.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_FWC.js
@@ -47,7 +47,7 @@ class A32NX_FWC {
 
         // ESDL 1. 0.190
         this.memoLdgMemo_conf01 = new NXLogic_ConfirmNode(1, true); // CONF 01
-        this.memoLdgMemo_memory1 = new NXLogic_MemoryNode(false);
+        this.memoLdgMemo_inhibit = new NXLogic_MemoryNode(false);
         this.memoLdgMemo_conf02 = new NXLogic_ConfirmNode(10, true); // CONF 01
         this.memoLdgMemo_below2000ft = new NXLogic_MemoryNode(true);
 
@@ -115,7 +115,9 @@ class A32NX_FWC {
     }
 
     _updateFlightPhase(_deltaTime) {
-        const radioHeight = SimVar.GetSimVarValue("RADIO HEIGHT", "Feet");
+        const radioHeight1 = Arinc429Word.fromSimVarValue("L:A32NX_RA_1_RADIO_ALTITUDE");
+        const radioHeight2 = Arinc429Word.fromSimVarValue("L:A32NX_RA_2_RADIO_ALTITUDE");
+        const radioHeight = radioHeight1.isFailureWarning() || radioHeight1.isNoComputedData() ? radioHeight2 : radioHeight1;
         const eng1N1 = SimVar.GetSimVarValue("ENG N1 RPM:1", "Percent");
         const eng2N1 = SimVar.GetSimVarValue("ENG N1 RPM:2", "Percent");
         // TODO find a better source for the following value ("core speed at or above idle")
@@ -125,7 +127,7 @@ class A32NX_FWC {
         );
         const eng1Or2Running = this.eng1OrTwoRunningConf.write(oneEngRunning, _deltaTime);
         const engOneAndTwoNotRunning = !eng1Or2Running;
-        const hFail = false;
+        const hFail = radioHeight1.isFailureWarning() && radioHeight2.isFailureWarning();
         const adcTestInhib = false;
 
         // ESLD 1.0.60
@@ -137,8 +139,8 @@ class A32NX_FWC {
         const acSpeedAbove80kts = this.speedAbove80KtsMemo.write(ias > 83, ias < 77);
 
         // ESLD 1.0.90
-        const hAbv1500 = radioHeight > 1500;
-        const hAbv800 = radioHeight > 800;
+        const hAbv1500 = radioHeight.isNoComputedData() || radioHeight.value > 1500;
+        const hAbv800 = radioHeight.isNoComputedData() || radioHeight.value > 800;
 
         // ESLD 1.0.79 + 1.0.80
         const eng1TLA = SimVar.GetSimVarValue("L:A32NX_AUTOTHRUST_TLA:1", "number");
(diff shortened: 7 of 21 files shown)


src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,722 @@
...
+            .sqrt();
+        let shortest_path_length: Length =
+            path_in_air + self.transmitter.cable_length + self.receiver.cable_length;
+
+        // At this point we've determined the length of the shortest path between the two
+        // transceivers. We now convert it into a travel time.
+
+        // It would be nice if this could be a const, but unfortunately the Velocity constructor is
+        // too complicated for what Rust deems acceptable for a const.
+        let speed_of_light: Velocity = Velocity::new::<meter_per_second>(299_792_458.);
```
tracernz, line 162: If we want to nitpick the propagation velocity in the cables will be approx 2/3 * c for typical coax. The effect will be too small to be significant though when airborne.😁

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,722 @@
...
+
+impl TransceiverPair for Ala52BTransceiverPair {
+    /// Returns the time the signal took to travel from one of the transceivers to the other,
+    /// to the receiver, or None if there is no path (or the signal is too weak).
+    /// While it is simplified compared to a proper frequency simulation, you can imagine that the
+    /// measured runtime is already based on the filtered difference between the sent and received
+    /// frequency in an FMCW-based radar.
+    fn response(&self) -> Option<TransceiverPairResponse> {
+        if self.pitch.abs() > Angle::new::<degree>(45.)
+            || self.bank.abs() > Angle::new::<degree>(44.)
```
tracernz, line 125: Is this a hard limit in the RA software, or approximated as the point where the return is too weak?

beheh: No, these were just picked randomly with a good margin outside of the specified operating range. I presume it's mostly a factor of the ground material and incidence angle. In future we could model the terrain you're over to vary these, but the sim unfortunately only provides the terrain exactly below you, and I need to think about how we'll make it feel good rather than randomly disappearing when you fly over a tiny patch of trees. I didn't want to grow the scope too much though.

crocket63 (changes requested):

src/instruments/src/PFD/VerticalSpeedIndicator.tsx
```diff
@@ -22,9 +23,14 @@ export const VerticalSpeedIndicator = ({ radioAlt, verticalSpeed }: VerticalSpee
     }
 
     const absVSpeed = Math.abs(verticalSpeed.value);
+    const radioAltitudeValid = !radioAltitude.isNoComputedData() || !radioAltitude.isFailureWarning();
```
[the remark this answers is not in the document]
beheh: You're right. I think I was trying to do !(NCD || FW).

src/systems/systems/src/navigation/radio_altimeter.rs
tracernz: Yep, fair enough. I was thinking maybe of multiplying the cosine of the two angles and a factor for the square of distance to give a rough approximation of return strength... but you'd end up tuning those to achieve something similar to this anyway. Good for initial impl. at the very least. 👍

beheh: So I started doing that but I cut myself short and removed it again just because I didn't want to grow the scope.
Exactly, we can easily add this in future and this PR puts the architecture in place to update the system transparently once we want to spend more time on it.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,722 @@
...
+            Ala52BAircraftInstallationDelay::FortyFeet => Length::new::<foot>(40.),
+            Ala52BAircraftInstallationDelay::FiftySevenFeet => Length::new::<foot>(57.),
+            Ala52BAircraftInstallationDelay::EightyFeet => Length::new::<foot>(80.),
+        }
+    }
+}
+
+impl Ala52BRadioAltimeter {
+    const ARINC429_REFRESH_INTERVAL_MILLIS: u64 = 80;
+    const MAXIMUM_STARTUP_TIME_MILLIS: u64 = 6_000;
```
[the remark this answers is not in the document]
beheh: For the REFRESH_INTERVAL I think it makes sense, however the startup time I want to be able to randomize, and that's much easier if I have the millis directly.

crocket63 (approved):

tracernz (approved):

ShadowFrost3:


Object of testing: #6736
Tier of Testing:2
Date: 19/02/2022
Testing Process: 

C&D Full Flight
Cold and Dark Startup normal - checked 
Pilot 0ft FO -1ft - checked 
Rotation/Pitch angle changes - checked 
~1000 ft TA only - checked 
LDG Memo - checked
+100 callout - failed
minimums audio callout - failed - MSFS bug?


Hot Start Testing Below 

ILS Approach 01 KDCA 200 Radio Live Weather 

1000 - checked 
500/400/300 - checked
300 is not called out as 100 above 
200 is not called out as minimums 
+100 - failed
Minimums - failed - DH shows just no audio callout 



TCAS high altitude  - not checked 
TCAS low altitude - functioning - I can get maintain descent at ~1100 AGL until RA inhibit - Need to consult manuals and possibly TCAS PR  

RA errors due to aircraft angle/attitude - checked - only looking down sim limitation   
GPWS warnings - failed - Terrain/Terrain, Pull up not functional
Sink Rate - failed 

RA stops working at approximately 10-15 degree pitch down over flat surfaces - correct? 


RA Individual failures - checked
RA dual failures - checked 
TCAS Standby - checked
Landing Memo Gear Downlock - not checked 


RA can entered a bug state when unknown conditions are entered at cruising altitudes - fail 
- probable recreation related to any of the following conditions: above 5000ft, RA 1/2 or both failed, TCAS mode switch 

Negatives: 
Testing Results: Not Passed

lukecologne:
The GPWS Terrain Terrain/Pull Up alert is not implemented AFAIK, so that is not a problem with this PR.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+trait TransceiverPair {
+    fn response(&self) -> Option<TransceiverPairResponse>;
+}
+
+/// This struct encodes a response from a pair of transceivers. Currently it's only encodes as a
+/// travel time, which in an modulated continuous wave radar (FMCW) can already be considered the
+/// filtered duration between the transmission and the response. In future it may also return the
+/// remaining signal strength, which can be used to model the signal loss over various surfaces, or
+/// model an actual frequency response.
+#[derive(Debug, Copy, Clone, PartialEq)]
```
davidwalschots, line 47: **thought**: This thought only relates to our codebase being a "library" for others; which isn't the case yet (and in this case actually the type is not even `pub`, so it doesn't matter much). `Copy`, `Clone` and `PartialEq` are not used. While having `Clone` and `PartialEq` on a type is fine. One has to be careful with `Copy` as future changes might make the type no longer `Copy` (i.e. if it starts to contain any type which is not `Copy`). Removing it later is a breaking change. I think you'll only add things that will be `Copy` to this type, but in general this is something to be aware of.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        self.travel_time
+    }
+}
+
+/// This struct represents a highly simplified pair of the transceivers as used by the ALA-52B
+/// system and models some of the rudimentary physical properties like signal travel time, maximum
+/// distance and attitude cutoffs. The pair refers to 1 transmitter system and 1 receiver system.
+/// Both of these systems include their radio wave generator/receiver respectively, their
+/// respective antenna installation, and the cables between the generator/receiver and the antenna.
+#[derive(Copy, Clone)]
```
davidwalschots, line 67: **thought**: These seem unnecessary here.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        self.pitch.cos() * self.distance_between_antennas()
+    }
+}
+
+impl TransceiverPair for Ala52BTransceiverPair {
+    /// Returns the time the signal took to travel from one of the transceivers to the other,
+    /// to the receiver, or None if there is no path (or the signal is too weak).
+    /// While it is simplified compared to a proper frequency simulation, you can imagine that the
+    /// measured runtime is already based on the filtered difference between the sent and received
+    /// frequency in an FMCW-based radar.
```
davidwalschots, line 131: **suggestion**: Link to items by adding `[` and `]`. You can see the result of this particular change by executing: `cargo doc --open --document-private-items`.

```suggestion
    /// Returns the time the signal took to travel from one of the transceivers to the other,
    /// to the receiver, or [None] if there is no path (or the signal is too weak).
    /// While it is simplified compared to a proper frequency simulation, you can imagine that the
    /// measured runtime is already based on the filtered difference between the sent and received
    /// frequency in an FMCW-based radar.
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+            None
+        }
+    }
+}
+
+impl SimulationElement for Ala52BTransceiverPair {
+    fn read(&mut self, reader: &mut SimulatorReader) {
+        self.alt_above_ground = Length::new::<foot>(reader.read(&self.alt_above_ground_id));
+        self.pitch = Angle::new::<degree>(reader.read(&self.pitch_id));
+        self.bank = Angle::new::<degree>(reader.read(&self.bank_id));
```
davidwalschots, line 196: **thought**: These are equal to the default unit declared in `src/systems/systems/src/simulation/mod.rs`:
```rust
read_write_uom!(Length, foot);
read_write_uom!(Angle, degree);
```
**suggestion**: Remove the unit declaration:
```suggestion
        self.alt_above_ground = reader.read(&self.alt_above_ground_id);
        self.pitch = reader.read(&self.pitch_id);
        self.bank = reader.read(&self.bank_id);
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+    }
+}
+
+pub trait RadioAltimeter {
+    fn radio_altitude(&self) -> Arinc429Word<Length>;
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Ala52BAircraftInstallationDelay {
```
davidwalschots, line 205: **thought**: Adding a comment as to what this represents might be helpful. I'm not quite able to determine it based on the usage. (Maybe `Delay` feels like something of `Time`, but it's `Length` here, which confuses me a bit 😀.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+
+    fn new_running(aid: Ala52BAircraftInstallationDelay) -> Self {
+        Self {
+            remaining_self_check: Duration::ZERO,
+            is_tracking_ground: false,
+            stale_for: Self::ARINC_429_REFRESH_INTERVAL,
+            aircraft_installation_delay: aid,
+            radio_altitude: Length::new::<foot>(0.),
+        }
+    }
```
davidwalschots, line 250:
```suggestion
    fn new_running(aid: Ala52BAircraftInstallationDelay) -> Self {
        Self::new(Duration::ZERO, aid)
    }
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        // First, check if we're still starting up and if so, simulate a wait until all self tests
+        // have completed
+        if let Some(new_remaining) = self.remaining_self_check.checked_sub(context.delta()) {
+            self.remaining_self_check = new_remaining;
+        } else {
+            self.remaining_self_check = Duration::ZERO;
+        }
+
+        // If there's any startup time remaining, do nothing
+        if self.remaining_self_check > Duration::ZERO {
```
davidwalschots, line 262: I'm super happy that `Duration::ZERO` is finally usable since 1.53. Great to know!

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+            return;
+        }
+
+        // Update only once per refersh interval
```
davidwalschots, line 266:
```suggestion
        // Update only once per refresh interval
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        if self.stale_for >= Self::ARINC_429_REFRESH_INTERVAL {
+            // Interrogate the transceivers for a response (which might fail if no physical path exists
+            // or some other kind of failure).
+            if let Some(response) = transceivers.response() {
+                let travel_time = response.travel_time();
+
+                let speed: Velocity = Velocity::new::<meter_per_second>(299_792_458.)
```
davidwalschots, line 273: **suggestion**: Consider adding a const to the module and using it in both places where you define it:
```rust
const SPEED_OF_LIGHT_METER_PER_SECOND: f64 = 299_792_458.;
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+    const MINIMUM_POWER_HOLDOVER: u64 = 30_000;
+    const MAXIMUM_POWER_HOLDOVER: u64 = 35_000;
+
+    pub fn new(
+        context: &mut InitContext,
+        number: usize,
+        aircraft_installation_delay: Ala52BAircraftInstallationDelay,
+        powered_by: ElectricalBusType,
+    ) -> Self {
+        let was_powered = context.has_engines_running();
```
davidwalschots, line 350: **thought**: I find going to past tense slightly confusing. `has_engines_running` is present tense.
**suggestion**: Consider `is_powered`.

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+            let runtime = self
+                .runtime
+                .get_or_insert_with(|| Ala52BRadioAltimeterRuntime::new(self_check, aid));
+            runtime.update(context, transceivers);
+        }
+    }
+
+    pub fn number(&self) -> usize {
+        self.number
+    }
```
davidwalschots, line 417:
```suggestion
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        } else {
+            Arinc429Word::new(Length::new::<foot>(0.), SignStatus::FailureWarning)
+        }
+    }
+}
+
+impl SimulationElement for Ala52BRadioAltimeter {
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T)
+    where
+        Self: Sized,
```
davidwalschots, line 447:
```suggestion
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+        fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+            self.system_1_transceivers.accept(visitor);
+            self.radio_altimeter_1.accept(visitor);
+
+            visitor.visit(self);
+        }
+
+        fn read(&mut self, _reader: &mut SimulatorReader) {}
+
+        fn write(&self, _writer: &mut SimulatorWriter) {}
```
davidwalschots, line 582: **suggestion**: As default empty implementation exists for these. There's no need to repeat it here.
```suggestion
```

src/systems/systems/src/navigation/radio_altimeter.rs
```diff
@@ -0,0 +1,864 @@
...
+
+    // Outputs
+    radio_altitude_id: VariableIdentifier,
+}
+
+impl Ala52BRadioAltimeter {
+    const MINIMUM_STARTUP_TIME_MILLIS: u64 = 4_000;
+    const MAXIMUM_STARTUP_TIME_MILLIS: u64 = 6_000;
+    const MINIMUM_POWER_HOLDOVER: u64 = 30_000;
+    const MAXIMUM_POWER_HOLDOVER: u64 = 35_000;
```
davidwalschots, line 342: It's interesting to see that it has such a long time it can stay "alive" without power!

davidwalschots (commented):
Minor comments. I really really like the implementation. It's great work! 🚀
Note that I've only reviewed the Rust portion of the PR.

src/systems/systems/src/navigation/radio_altimeter.rs
beheh: I added this and kept it for now because the RA does in fact know which RA it is, thanks to pin-programming.
This is especially important later on because the RA would set the SDI bits (Source Destination Identifier) in any ARINC words based on whether it's RA 1 or 2 (or 3 for the A380). If we drop this fn, then the `number` field becomes unused, and it seemed nicer to explicitly make it part of the external API rather than `#[allow(dead_code)]` the field above.

Thoughts?

src/systems/a320_systems/src/lib.rs
```diff
@@ -1,9 +1,11 @@
+extern crate core;
 extern crate systems;
```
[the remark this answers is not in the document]
beheh: I'm dropping the one my IDE (presumably) added, but I'll leave the other one just because it doesn't feel like this is the PR to make that change.

src/systems/systems/src/navigation/radio_altimeter.rs
davidwalschots: If you expect usages outside of the module, then I'd leave it in. If you only expect usages inside the module, I'd add `#[allow(dead_code)]` for now.

beheh: Alright, then to keep the API clean I'll just go with allowing the dead code. Thanks for your input.

src/systems/systems/src/navigation/radio_altimeter.rs
beheh: Yeah, it's a bit confusing, but I'm just following the industry here. Apparently Aircraft Installation Delay / AID is standard terminology (presumably defined in ARINC 707), and the pins are also defined in terms of Lengths.

See for example [this manual](https://www.viavisolutions.com/ja-jp/literature/alt-8000-operations-manual-manuals-user-guides-en.pdf) (search for "AID") and [this book](https://soaneemrana.org/onewebmedia/AIRCRAFT%20RADIO%20SYSTEM%20BY%20J%20POWEL.pdf):

> Part of the two-way travel time is accounted for by the aircraft installation delay (AID). Since, ideally,
the radio altimeter should read zero feet on touchdown, the residual altitude which accounts for the AID is made up of cable length multiplied by a factor (typically 1.5) to allow for propagation speed, and the sum of the antenna heights. 

Tracer already raised the cable propagation speed further up which I didn't want to model, but I don't think that changes anything about the interface.

beheh:
While this can (and should!) be tested as-is, I'm waiting for some final numbers to drop for the power holdover times.

davidwalschots (approved):

beheh:
Thanks, I fixed the altitude callouts, they slipped through last-minute as I was tweaking the GPWS fail behaviour.

beheh:
While this is reviewed and everything, just marking it as Draft again while I wait for some final numbers.

beheh:
Okay, the numbers we have for now are the best we'll get.

PR #6738, opened 2022-02-16
fix(athr): improved condition to transition CLB limit from FLX TO

aguther:
## Summary of Changes
There has been at least one report in experimental version that the CLB thrust limit was not restored to normal from FLX reduction. It ended for the user not being able to climb beyond FL2070.

To avoid this situation the condition to trigger the transition has been revisited.

## Testing instructions
- do a FLX TO
- after thrust reduction altitude and lever being set from FLX/MCT detent to CLB detent, the transition should start after 10 s (thrust will increase)

### DIFF
--- a/src/fadec/src/EngineControl.h
+++ b/src/fadec/src/EngineControl.h
@@ -951,7 +951,7 @@ class EngineControl {
       isFlexActive = false;
     }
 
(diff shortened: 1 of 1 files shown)


IbrahimK42 (approved):

aguther:
Was tested on exp

PR #6740, opened 2022-02-16
fix: Prevent default stall warning from occurring

### LINKED ISSUE #6724: Stall Reporting
### Aircraft Version

Stable

### Build info

```json
The bug is in all stable, developer and experimental. Am reporting on behalf of a Virtual Airline that constantly logs pilot stalls using your aircraft.
```

### Describe the bug

In flight logging programs that are used for Virtual Airlines, A32NX continuously reports false stalls on takeoff because the default stall variable that Asobo and TFDi smartCARS  is not updated. The logging programs report stalls climbing at 145+ knots with flaps 1+F even though V2 is way less than that.

We are aware that the A32NX has its own stall variable, but you be able to update the default Asobo stall variable so this false reporting doesn't happen.

### Expected behavior

The default Asobo stall variable should be updated to be the same as the A32NX custom stall variable that is used.

### Steps to reproduce

Use a flight logging program that reads the default stall variable from Asobo and takeoff with flaps 1+F or 2 and follow flight director.

### References (optional)

_No response_

_No response_

_No response_

donstim:
Fixes #6724 

## Summary of Changes

This change is intended to prevent the MSFS default stall warning simvar bool from being set to "on" (i.e., 1). We will be using a custom A32NX variable for stall warning activation. As evidenced by issue #6724, some 3rd party apps use the Asobo default stall warning simvar to determine if the airplane has stalled. Although stall warning activation does not mean the airplane has stalled, we can resolve the issue with this PR.
## Additional context

Currently, the MSFS SDK does not document how the default stall warning is triggered. Through testing, it was found that it triggers at approximately 10.5 degrees AOA for the A32NX. It triggers at a much higher AOA for the Asobo default A320. It was found that if the stall protection system is disabled for the Asobo A320 (as it is for the A32NX since we have our own custom envelope protection system), stall warning would also be activated for the Asobo A320 at approximately 10.5 degrees AOA.

By re-enabling the MSFS stall protection system in the A32NX, but setting the AOA and timing triggers to values that will prevent it from activating (and interfering with our custom envelope protection system), this PR will result in the default stall warning Simvar from showing that the stall warning system has been activated.

## Testing instructions

1. Conduct a CONF 2 takeoff with the airplane without this PR. Using SimVar watcher (available in C:\MSFS SDK\Samples\SimvarWatcher\bin\x64\Release after installing the MSFS SDK. Select the STALL WARNING Simvar from the dropdown list, set the units to `Bool,` and click on "Add request." Then click on "Connect" after starting MSFS.), observe the `STALL WARNING` Simvar value change to "1" during the takeoff initial climb. (You may have to pitch a bit aggressively to make this happen.) Optionally, you can also enable the SIM TUNING debug window in the aircraft editor in MSFS dev mode to observe the current airplane AOA. Make sure you get an AOA above 10.5 degrees.)
2. Do the same thing with this PR applied. Pitch aggressively to verify that the `STALL WARNING` Simvar value does not change to 1. (If you pitch too aggressively, you should activate alpha protection, but the `STALL WARNING` Simvar value should not change to 1.
3. Verify that ALPHA FLOOR and TOGA LK are not affected. That is, it is still possible to activate these protection features as normal.


tracernz (approved):

sidnov:
QA/Devteam Report

Discord : sidnov#8337
Object of testing: Disabled default stall triggers
Tier of Testing : 1
Date : 17/02/2022

Testing Process:
Spawn on the runway in 2 aircrafts, with and without PR. Without changes from the PR, stall warning is easily triggered in several cases. With the PR, the stall warnings remain on 0 with several maneuvers. Checked with Simvar Watcher.

Testing Results:
Passed

PR #6767, opened 2022-02-20
feat(atsu): Event based atsu

svengcz:
## Summary of Changes
The first version of the ATSU was interval-driven. Every 100ms were the synchronization between the DCDU and the ATSU checked.

This PR increases the overall A32NX performance by removing this interval-driven synchronization.
Except to ATC MSG button and ring tone timers are all shared information event driven.
The required timers are started and stops as needed to reduce the time consuming, but unnecessary overhead.

## Additional context

## Testing instructions

- Validate that DCL and OCL are still working
- Validate that instructions from ATC are well handled
  - every 10 seconds a ring tone until the pilot pressed ATC MSG or preselected an answer in the DCDU
  - All answers and requests are sent only once to ATC


beheh:
As a note, the ATC MSG flash and ring will also be dealt with by the FWCs at some point in future. The only thing the ATSU would have to expose is an LVar, and the FWC will take care of things like inhibitions depending on the flight phase.

svengcz:
> As a note, the ATC MSG flash and ring will also be dealt with by the FWCs at some point in future. The only thing the ATSU would have to expose is an LVar to enable the visual or the aural message, and the FWC will take care of things like inhibitions depending on the flight phase.

Will be a easy change as soon as we have the FWC. System should be flexible enough

beheh (approved):
Great job! I think this is all a very nice pattern - especially with the callbacks instead of direct SimVar access everywhere. It centralizes the SimVar names and deduplicates some logic and makes it testable in future. Let's get this into QA!

PR #6769, opened 2022-02-20
feat: add developer state protection from payload reset

sidnov:
Fixes #[issue_no]

## Summary of Changes

When A32NX_DEVELOPER_STATE is true, MCDU reloads will not affect payload or pax values.

## Additional context

## Testing instructions

Developer-side changes, no QA needed.

### DIFF
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
@@ -22,10 +22,10 @@ class A32NX_Boarding {
         const inDeveloperState = SimVar.GetSimVarValue("L:A32NX_DEVELOPER_STATE", "Bool");
         if (!inDeveloperState) {
             // Set default pax (0)
-            this.setPax(0);
-            this.loadPaxPayload();
-            this.loadCargoZero();
-            this.loadCargoPayload();
+            await this.setPax(0);
+            await this.loadPaxPayload();
+            await this.loadCargoZero();
+            await this.loadCargoPayload();
         }
     }
 
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
@@ -19,11 +19,14 @@ class A32NX_Boarding {
     }
 
     async init() {
-        // Set default pax (0)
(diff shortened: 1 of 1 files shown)


flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
```diff
@@ -19,11 +19,14 @@ class A32NX_Boarding {
...
-        await this.loadCargoZero();
-        await this.loadCargoPayload();
+        const inDeveloperState = SimVar.GetSimVarValue("L:A32NX_DEVELOPER_STATE", "Bool");
+        if (!inDeveloperState) {
+            // Set default pax (0)
+            this.setPax(0);
+            this.loadPaxPayload();
+            this.loadCargoZero();
+            this.loadCargoPayload();
+        }
```
davidwalschots, line 29: **thought**: There used to be lots of `await` usages there. Now there are none. That _might_ be fine. But then probably you also want to remove the `async` modifier on the function?

davidwalschots (commented):
Could you explain why specifically we'd not want this in this situation?

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
sidnov: This case (reset payload to 0) simply sets the values, while some other usages of the same fn would ideally require an ``await`` before proceeding to cargo loading

davidwalschots: I'm not quite sure what you mean. As there are no `await`s, one can remove `async` from `async init() {`.

davidwalschots: I'd definitely keep `await` in when the function exposes a Promise (is async). You shouldn't consdier the internals of the functions from the call site, as those might change in the future.

sidnov: I think I misunderstood from the first comment about async, I get what you meant 👍🏻

tracernz (approved):
Legend! 😁

tracernz:
> Could you explain why specifically we'd not want this in this situation?

When developing e.g. the FMS... the payload gets reset to zero each time you reload the FMGC+MCDU instrument prior to this PR.

PR #6778, opened 2022-02-21
feat(pfd): improved FPV and FPD

lukecologne:
## Summary of Changes

This PR aims to improves the FPV (flight path vector, bird) and FPD (flight path director) in the following ways:
* Firstly, the FPA is now no longer the aerodynamic FPA (AOA + pitch), but the kinematic FPA (atan(VS / GS)).
* Improved behaviour in gusty weather: The bird no longer jumps around as erratically
* The roll and pitch commands of the FPD are improved:
  * The roll command now actually displays the desired roll angle, instead of being scaled down slightly like before this PR
  * The pitch command now actually points to the desired pitch angle.
* Improved drawing priority: The FPA/FPD is now rendered above the fixed aircraft symbol (the yellow/black aircraft shape) and the rising ground (the lower "earth-colored" sector that rises as the radio height decreases). I'm not entirely sure about the exact drawing order here, but this should pose an improvement for now.

Also there is a small secondary improvement: The glideslope yellow reference line now also shows, when the LS button is pressed and the attitude is invalid. Previously, it only showed when the attitude was valid. While I don't have a 100% sure reference on this, it would make sense that the LS data is available even with invalid attitude data, which includes the reference line.

The FPA and DA should normally be calculated directly in the IR, but as this isn't implemented yet, I'm doing it in the PFD.

## Screenshots (if necessary)

Before: https://streamable.com/awoxiq
After (The changed drawing priority is not yet implemented here): https://streamable.com/dpl498

## Additional context

### DIFF
--- a/src/instruments/src/PFD/AttitudeIndicatorFixed.tsx
+++ b/src/instruments/src/PFD/AttitudeIndicatorFixed.tsx
@@ -1,6 +1,11 @@
 import { Arinc429Word } from '@shared/arinc429';
+import { getSmallestAngle } from '@instruments/common/utils.js';
 import React from 'react';
+import { calculateHorizonOffsetFromPitch } from './PFDUtils';
 import { getSimVar } from '../util.js';
+
+const DistanceSpacing = 15;
+const ValueSpacing = 10;
 
 interface AttitudeIndicatorFixedUpperProps {
     pitch: Arinc429Word;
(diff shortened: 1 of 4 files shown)


tracernz (approved):

mico975:

Discord          : mico#3145
Object of testing: #6778
Tier of Testing  : 1
Date             : 22/02/2022

Testing Process: 
Firstly, the FPA is now no longer the aerodynamic FPA (AOA + pitch), but the kinematic FPA (atan(VS / GS)). - **OK**

Improved behaviour in gusty weather: The bird no longer jumps around as erratically - **OK**

The roll and pitch commands of the FPD are improved:
The roll command now actually displays the desired roll angle, instead of being scaled down slightly like before this PR - **OK**
The pitch command now actually points to the desired pitch angle. - **OK**

Improved drawing priority: The FPA/FPD is now rendered above the fixed aircraft symbol (the yellow/black aircraft shape) and the rising ground (the lower "earth-colored" sector that rises as the radio height decreases). I'm not entirely sure about the exact drawing order here, but this should pose an improvement for now. - **NOT OK**, yellow middle line is gone

Negatives:       
Yellow horizontal line is gone on right side of the PFD for some reason

Testing Results: 
Not passed.

Conclusions: 
None at the moment.

Media:
image

PR #6783, opened 2022-02-22
feat: option to use calculated ILS

aguther:
## Summary of Changes
This PR introduces the possibility to use a calculated ILS signals instead of the signals provided by the sim. It also improves conditions for `RWY` and `RWY TRK`.

## Additional context
On many runways there is an issue with the LOC or G/S signal to be lost. This happens when the plane is getting lower than the antenna position. Therefore there are a lot of runways where IRL the LOC signal is available and `RWY` would engage but in MSFS it does not happen due to the unavailability of the LOC signal.

## Testing instructions
- tbd


beheh:
Would you mind also updating the GPWS Mode 5 ("Glideslope" callout) with the new LVars?

https://github.com/flybywiresim/a32nx/blob/cb1cf4de784e0e81734a44ff906de2bc53db03bb/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_GPWS.js#L414-L418

aguther:
> Would you mind also updating the GPWS Mode 5 (for the "GLIDE SLOPE" callout) with the new LVars?
> 
> https://github.com/flybywiresim/a32nx/blob/cb1cf4de784e0e81734a44ff906de2bc53db03bb/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_GPWS.js#L408-L431

Done

tracernz (approved):

PR #6795, opened 2022-02-23
refactor(tcas): Improve traffic draw

vkrizan:
## Summary of Changes

Improves the draw response of TCAS traffic on ND on mode and range change.

Received traffic from the computer is kept in state. The `alive` attribute is kept and used for mode and range changes.

## References

Range change: https://youtu.be/8QzlOwrUDDs?t=439

## Additional context

Discord username (if different from GitHub): coolll.sk#0022

## Testing instructions

Observe ND traffic markers position change:
* immediately on range change
* very shortly after mode change (ARC to ROSE, ROSE to ARC)

To ensure that the previous draw functionality is preserved the traffic markers should keep ~1s cadence update when:
* flying
* changing heading

Observe performance.

### DIFF
--- a/src/instruments/src/ND/elements/Traffic.tsx
+++ b/src/instruments/src/ND/elements/Traffic.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable camelcase */
 import { useSimVar } from '@instruments/common/simVars';
-import React, { FC, useState, useReducer, memo } from 'react';
+import React, { FC, useState, memo } from 'react';
 import { Layer } from '@instruments/common/utils';
 import { TCAS_CONST as TCAS, TaRaIntrusion, TaRaIndex } from '@tcas/lib/TcasConstants';
 import { Coordinates } from '@fmgc/flightplanning/data/geo';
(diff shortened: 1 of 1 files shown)


tracernz:
The ACAS map only rotates once per second (with each update from the TCAS computer) on the real ND... it does not rotate smoothly. Just checking that you retain that behaviour?

aguther:
Please also keep an eye on the following PR:
https://github.com/flybywiresim/a32nx/pull/6766

vkrizan:
Yeah, sorry I did not had these references. This makes it smoother, but that's not how it is IRL. I'll look into more references and see if this PR would add any value at all. Otherwise it can be closed or be skimmed to have just the refactoring parts.

@aguther Yep, thanks. I was looking at it prior to making changes.

src/tcas/src/components/TcasComputer.ts
```diff
@@ -9,7 +9,6 @@ import { TcasComponent } from '@tcas/lib/TcasComponent';
 import { LatLongData } from '@typings/fs-base-ui/html_ui/JS/Types';
 import { LocalSimVar } from '@shared/simvar';
 import { NXDataStore } from '@shared/persistence';
-import stringify from 'safe-stable-stringify';
```
tracernz, line 12: This was a performance optimisation, as the Coherent JSON.stringify is very slow.

src/tcas/src/components/TcasComputer.ts
```diff
@@ -9,7 +9,6 @@ import { TcasComponent } from '@tcas/lib/TcasComponent';
 import { LatLongData } from '@typings/fs-base-ui/html_ui/JS/Types';
 import { LocalSimVar } from '@shared/simvar';
 import { NXDataStore } from '@shared/persistence';
```
aguther, line 11:
```suggestion
import stringify from 'safe-stable-stringify';
```

aguther, line 11:
```suggestion
import { NXDataStore } from '@shared/persistence';
import stringify from 'safe-stable-stringify';
```

src/tcas/src/components/TcasComputer.ts
```diff
@@ -1105,7 +1104,7 @@ export class TcasComputer implements TcasComponent {
             }
         });
 
-        this.sendListener.triggerToAllSubscribers('A32NX_TCAS_TRAFFIC', stringify(this.sendAirTraffic));
```
aguther, line 1108: This needs to be kept.

src/tcas/src/components/TcasComputer.ts
```diff
@@ -1125,7 +1124,7 @@ export class TcasComputer implements TcasComponent {
         if (this.tcasMode.getVar() === TcasMode.STBY) {
             if (this.sendAirTraffic.length !== 0) {
                 this.sendAirTraffic.length = 0;
-                this.sendListener.triggerToAllSubscribers('A32NX_TCAS_TRAFFIC', stringify(this.sendAirTraffic));
```
aguther, line 1128: This needs to be kept.

src/tcas/src/components/TcasComputer.ts
vkrizan: It seems like that the `JSON.stringify` is called in `ViewListener.triggerToAllSubscribers`. If you know that this implementation is more optimized, then to avoid double serialization calls of `triggerToAllSubscribers` can be replaced with
```
Coherent.trigger('TRIGGER_EVENT_TO_ALL_SUBSCRIBERS", this.sendListener.m_name, 'A32NX_TCAS_TRAFFIC', stringify(this.sendAirTraffic))
```

`JSON.stringify` is likely implemented in C, so I wonder how the JS implementation of stringify is better. It must cut some corners. Just curious...

vkrizan:
@aguther  @tracernz I see that you guys have removed the commit that was removing the stringify optimization. It can move on as is now.

aguther (approved):

PR #6800, opened 2022-02-25
fix(lnav): fix HX entry RAD, FC and FD legs, and enroute runway

tracernz:
Fixes #[issue_no]

## Summary of Changes

- I made a small mistake in rewriting HX entries, forgetting to account for turn direction in the roll anticipation calculations. This caused a turn in the wrong direction before turning back to the correct direction for left turn holds.
- FC and FD legs were mapped incorrectly from the source data... now the geometry is correct (perfect implementation will need to wait for cfms-20, as I don't want to do two refactors of the magnitude required prior to that).
- Enroute runways weren't shown quite right on the ND (they should be centred on the runway location).
- There was an exception thrown in the FMGC when the map reference point was undefined in plan mode (happens briefly while creating a new flightplan... but whichever part of the FMGC decides the MRP should actually set MRP to PPOS when no F-PLN pages are in use... future PR?)
- The leg length of database holds specified as a distance was not scaled from metres to nautical miles

## Screenshots (if necessary)

FD legs:
EGPE ILS23 via INS before:
image
After:
image

ELLX ILS24 via DIK before:
image
After (note the VI onto the FACF leg is borked for some reason):
image

Enroute range runway after:
image

## Additional context

## Testing instructions

Test some approaches with FD legs, suggestions (I haven't tested these ones myself):
image

and FC legs:
image

### DIFF
--- a/src/fmgc/src/efis/EfisSymbols.ts
+++ b/src/fmgc/src/efis/EfisSymbols.ts
@@ -84,6 +84,7 @@ export class EfisSymbols {
         this.lastNearbyFacilitiesVersion = this.nearby.version;
         const fpChanged = this.lastFpVersion !== this.flightPlanManager.currentFlightPlanVersion;
         this.lastFpVersion = this.flightPlanManager.currentFlightPlanVersion;
+        // FIXME map reference point should be per side
         const planCentreIndex = SimVar.GetSimVarValue('L:A32NX_SELECTED_WAYPOINT', 'number');
         const planCentre = this.flightPlanManager.getWaypoint(planCentreIndex)?.infos.coordinates;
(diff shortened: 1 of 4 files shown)


Benjozork (approved):

Benjozork:
The VI is bored because that's a VI IF CF sequence, which won't work until I push leg skipping

mico975:

Discord          : mico#3145
Object of testing: #6800
Tier of Testing  : 1
Date             : 25/02/2022

Testing Process: 
Tried some of suggested approaches. All OK.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #6810, opened 2022-02-27
feat(bleed): Use mass instead of volumes for a better pneumatic simulation

Gurgel100:
Fixes #[issue_no]

## Summary of Changes

This PR changes the simulation of the pneumatic system to use air masses instead of air volumes. This should provide a more accurate simulation of the pneumatic system and also provides a framework for the PACK simulation (will come in a seperate PR). Also it fixes a small issue where the wrong simvar was read to determine if the packs are enabled.
This PR does not implement the BLEED ECAM page.
I think the bleed pressure and temperature from the engine needs some tweaking but this can be done in a separate PR.

## Additional context

The current pneumatic system was implemented with #5384. This PR enhances this system by the simulation of mass.

## Testing instructions

Generally check the simvars (filter for `PNEU_`) and see if there is something unusual or hugely different from master. Also check that `PNEU_ENG_{1,2}_PRECOOLER_OUTLET_PRESSURE` converges to 46 psi.
Play with the system. Try different outside temperatures and pressures in different flight phases and engine power.


src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -49,11 +50,78 @@ pub trait ControllablePneumaticValve: PneumaticValve {
...
+            other.temperature(),
+            transfer_factor,
+        ));
+    }
+
+    /**
+     * Transfer heat to the container depending on the temperature difference.
+     * Returns the temperature difference which would be applied to the opposite container.
+     */
+    fn heat_conduction_single(
```
BlueberryKing, line 108: I think it should not be the responsibility of a `PneumaticContainer` to compute the temperature changes to apply to this and the other container. This calculation should be performed on the connecting element, which is presumably the `PneumaticContainerConnector` or `PneumaticExhaust`.

In the future, this calculation can then be replaced by the computation of the energy exchange. The effect of an addition of heat to a container can then be handled by the container, but not the calculation of how much heat should be exchanged.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -49,11 +50,78 @@ pub trait ControllablePneumaticValve: PneumaticValve {
...
+            .pressure()
+            .get::<pascal>()
+            .powf(1. / Self::HEAT_CAPACITY_RATIO);
+        let m1 = self.mass().get::<kilogram>();
+        let m2 = other.mass().get::<kilogram>();
+        Mass::new::<kilogram>(m1 * m2 * (p2 - p1) / (m2 * p1 + m1 * p2))
+    }
+
+    /// Transfer heat between two containers depending on the temperature difference
+    fn heat_conduction(
```
BlueberryKing, line 91: This is an interesting function. From what I see, it only exists to apply a temperature change to this container, after the change has been applied to the connected container?

I think the function is obsolete when we implement my suggestion on `heat_conduction_single`.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -75,12 +143,38 @@ impl PneumaticContainer for PneumaticPipe {
...
-    fn change_fluid_amount(&mut self, volume_change: Volume) {
-        let pressure_change = self.calculate_pressure_change_for_volume_change(volume_change);
+    fn mass(&self) -> Mass {
+        /*
+        We calculate the mass instead of storing it because floating points are
+        not exact and after some time the true mass drifts from the calculated
+        mass which then results in some negative pressures and temperatures
+        when doing heat conduction. This later leads to NaNs.
+        Another possibility would be to calculate the volume instead.
+        */
```
BlueberryKing, line 153: Interesting observation 👍 What would you think about eliminating `pressure` instead and computing it from mass, volume and temperature? In my mind, using `mass` as a form of "source of truth", rather than pressure makes more sense for some reason.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -128,32 +218,53 @@ impl PneumaticPipe {
     }
 
     fn change_volume(&mut self, new_volume: Volume) {
+        let new_pressure = self.calculate_pressure_for_new_volume(new_volume);
+        self.update_temperature_for_target_pressure(new_pressure);
+        self.pressure = new_pressure;
         self.volume = new_volume;
     }
```
BlueberryKing, line 225: Is there a situation in which we wish to change the volume of a `PneumaticPipe` at runtime, now that we use mass as a measure for fluid quantity? We have a `VariableVolumeContainer` which should serve this exact purpose.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -128,32 +218,53 @@ impl PneumaticPipe {
...
+            * (self.volume.get::<cubic_meter>() / new_volume.get::<cubic_meter>())
+                .powf(Self::HEAT_CAPACITY_RATIO)
+    }
 }
 
-pub struct TargetPressureSignal {
+pub struct TargetPressureTemperatureSignal {
     target_pressure: Pressure,
+    target_temperature: ThermodynamicTemperature,
 }
```
BlueberryKing, line 237: I like the idea of having a temperature here as well 👍.

src/systems/a320_systems/src/pneumatic.rs
```diff
@@ -814,7 +814,7 @@ impl EngineBleedAirSystem {
             ),
             engine_starter_exhaust: PneumaticExhaust::new(3e-2, 3e-2, Pressure::new::<psi>(0.)),
             engine_starter_valve: DefaultValve::new_closed(),
-            precooler: Precooler::new(5.),
+            precooler: Precooler::new(180. * 2.),
```
BlueberryKing, line 817: What does this `* 2` signify? I assume there is a reason why this is done explicitly.

src/systems/systems/src/pneumatic/mod.rs
```diff
@@ -335,11 +466,12 @@ impl Precooler {
                 - container_one.temperature().get::<degree_celsius>(),
         );
 
-        let temperature_change =
-            temperature_gradient * (1. - (-self.coefficient * context.delta_as_secs_f64()).exp());
+        let temperature_change = temperature_gradient / Self::HEAT_CAPACITY_CONSTANT_PRESSURE
+            * (self.heat_transfer_coefficient * context.delta_as_secs_f64());
```
BlueberryKing, line 470: Hmm, this is really an energy to transfer. I think this is a good idea, but we should name the variable appropriately.

BlueberryKing (commented):
Nice work! As always, let me know if you have any questions or disagree with any of my suggestions, happy to discuss them!

src/systems/a320_systems/src/pneumatic.rs
Gurgel100: The `* 2` signifies the area of exchange and the `180.` is the heat transfer coefficient.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: Yes you are right.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: Previously both `PneumaticContainerConnector` and `PneumaticExhaust` had an implementation of the heat conduction. I wanted to remove the code duplication by putting it here.
I could move it back or maybe a better solution would be to move it out of the trait into a separate function?

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: To be honest I looked at the performance perspective to decide which component to calculate. I simply checked how often the mass is queried. Both pressure and temperature are queried relatively often (also for all the simvars). But I agree with you to better calculate something else. Maybe we can remove the volume? It's not used anymore anyway.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: I think the only location where it's used is in `VariableVolumeContainer`. I didn't want to change that. But I think it should be possible.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: It's used because the air already has this temperature when it comes out of the engine.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: I know. The thing is that it's neither energy nor a temperature but something in between. I try to find another variable name.

crocket63:
Tested a bit and PR solves quite some issues that are present on current model. Hyd reservoirs do not seem to rise in pressure anymore even when there's less pressure upstream than in reservoir.

src/systems/systems/src/pneumatic/mod.rs
Gurgel100: I changed the name of the variable

src/systems/systems/src/pneumatic/mod.rs
BlueberryKing: Yeah, in general, we don't want the volume (i.e. the physical volume, like the available space) of a container to change. The exception is the `VariableVolumeContainer `, which is used for hydraulic reservoirs, where the amount of space that is available to the bleed air, depends on the amount of hydraulic fluid present in the reservoir.

src/systems/systems/src/pneumatic/mod.rs
davidwalschots: As a side note I'd like to mention that comments in this PR aren't following the Rust convention. If you'd like to know more, please read [this chapter](https://doc.rust-lang.org/book/ch03-04-comments.html) of the book and [this section](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments) as well.

Gurgel100: > As a side note I'd like to mention that comments in this PR aren't following the Rust convention. If you'd like to know more, please read [this chapter](https://doc.rust-lang.org/book/ch03-04-comments.html) of the book and [this section](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments) as well.

Thank you for your comment. I didn't know there is a convention for comments in Rust. I will adopt them. But what I don't understand is why only line comments are convention but that's another topic.

davidwalschots:
@BlueberryKing has requested the author to make another change (through another communication channel) and will review this after that change has been made.

beheh:
@Gurgel100 Any objection if I move this PR back to draft until the suggestions have been addressed?

Gurgel100:
> @Gurgel100 Any objection if I move this PR back to draft until the suggestions have been addressed?

Sure. I have no problem with that.

Gurgel100:
I found a bug in the calculations of the mass flow which I fixed now. This also resolved the problem that prevented the mass to be stored and needed to be calculated.

crocket63:
.when? :D

src/systems/systems/src/pneumatic/mod.rs
BlueberryKing: Yes, this is true. I thought about this for a bit and I think I'm not too bothered by the duplication. I'd rather have a bit of duplication here and there than to abstract this away prematurely.

However, there are options I see to avoid it if we this becomes a problem in the future and we have more and more duplicate implementations of this.

For one, it might be possible to treat the entire environmental atmosphere as a `PneumaticContainer` and then reduce the `PneumaticExhaust` to a connector between any container and the atmosphere. In this case we only need one implementation of heat conduction on the connector. However, I'm not sure how feasible this is right now.

Secondly, we could move the computation to a macro or a top-level function. Whether this is how you should handle it this way in Rust, I don't know 😄

Gurgel100: Ok. Then I will move the code back to `PneumaticContainerConnector` and `PneumaticExhaust`.

BlueberryKing (approved):
LGTM!

mico975:

Discord          : mico#3145
Object of testing: #6810
Tier of Testing  : 1
Date             : 06/04/2022

Testing Process: 
Generally check the simvars (filter for PNEU_) and see if there is something unusual or hugely different from master. - **OK**
Also check that PNEU_ENG_{1,2}_PRECOOLER_OUTLET_PRESSURE converges to 46 psi. - **OK**
Play with the system. Try different outside temperatures and pressures in different flight phases and engine power. - **All OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
Nice work.

Media:       
image

.github/CHANGELOG.md
```diff
@@ -229,6 +229,7 @@
 1. [MCUD] Add LSK for vert rev at destination - @tracernz (Mike)
 1. [MODEL] Add detailed rivet mesh to allow for ALDB+COMP+NORM ribbons - @bouveng (Johan Bouveng)
 1. [SD] Rewrite Bleed page to React - @RichardPilbery (tricky_dicky#3571)
+1. [BLEED] Converted bleed system to you mass flow instead of volume flow - @Gurgel100 (Pascal)
```
crocket63, line 232: "To you mass flow"
 ?

Gurgel100: Oh lol. Should say "To use mass flow". Fixed.

John-UK:
Quality Assurance Report
Discord john P Maguire#9027
Object of testing: #6810
Tier of Testing: 2
Date: 07/04/2022

Testing Process:

Spawned C&D at EGLL with live weather OAT 10C. Set up with B-debug open for PNEU
pneu1


Flew a circuit on EGLL and observed the PNEUENG{1,2}_PRECOOLER_OUTLET_PRESSURE warmed up to 45 on start up and remained 45-47 in the air. Slightly higher on landing temporarily at 50C.

pneu2

Spawned for a second time with clear skies and 35C. Just one observation not related the PR. LP temperatures are much lower than OAT.

image

Same results in the higher temps outlet pressure 48 +-3
image

Testing Results
Passed. 

Conclusions 
Nothing out of the ordinary noted.

Gurgel100:
> Spawned for a second time with clear skies and 35C. Just one observation not related the PR. LP temperatures are much lower than OAT.
The reason for that is because the engine is not running and there is no heat exchange between the LP and HP stage and the outside world. But as soon as the engines are running it should increase the temperature. And thanks for your qa.

PR #6825, opened 2022-03-01
fix(hyd): fix cargo door unit test

crocket63:
Fixes #[issue_no]

## Summary of Changes

Cargo door opening unit test was sometimes failing due to too small margin in test timing between unlock and hydraulic control start.

## Additional context

## Testing instructions

No test needed as long as unit tests are GO.

### DIFF
--- a/src/systems/a320_systems/src/hydraulic/mod.rs
+++ b/src/systems/a320_systems/src/hydraulic/mod.rs
@@ -7055,9 +7055,9 @@ mod tests {
 
             let current_position_unlocked = test_bed.cargo_fwd_door_position();
 
-            test_bed = test_bed
-                .open_fwd_cargo_door()
-                .run_waiting_for(A320DoorController::DELAY_UNLOCK_TO_HYDRAULIC_CONTROL);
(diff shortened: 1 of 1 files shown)


crocket63:
> I am slightly confused by what `DELAY_UNLOCK_TO_HYDRAULIC_CONTROL` actually means. Apparently it doesn't mean the time until the door actually moves when commanded to open; even though the name suggests it does.
> 
> I'd like to note that there is already a second of time waited at line 7052. Might this test actually have identified something that is in fact a bug which occurs irregularly?

DELAY_UNLOCK_TO_HYDRAULIC_CONTROL Is the delay from unlock command to the start of hydraulic motion. 

So there's already a 1s delay after unlock until we capture a position for the test. That 1s could also be too small, because depending on how the door oscillates it could capture something that is not rest position.

My guess is that we might fall in a case where oscillations make the door go back under the captured position at the start of hydraulic motion. So by adding 1s we have a 2s margin on the expected delay.

I doubt that added margin could mask a bug unless you have some evil idea? :D

davidwalschots (approved):
> unless you have some evil idea? :D

😆

PR #6832, opened 2022-03-02
feat(hyd): Hydraulic Elevator, Rudder and Spoiler control

### LINKED ISSUE #180: flight controls should be locked until APU or engines started (Hydraulics simulation)

### LINKED ISSUE #6447: Hydraulic "droop" to elevators and ailerons when powered down
### Aircraft Version

Development

### Description

As with the real world aircraft, is it possible to have the slow hydraulic "droop" due to lack of pressure to the ailerons and elevators when parked up at a gate with engines off? 

### References (optional)

https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.reddit.com%2Fr%2Fwhatisthisthing%2Fcomments%2F4ahfme%2Fmultiple_aerial_things_on_the_end_of_the_plane%2F&psig=AOvVaw2_KRAJ6DUi52bu8WN91Yx3&ust=1640986465598000&source=images&cd=vfe&ved=0CAsQjRxqFwoTCIDugo69jPUCFQAAAAAdAAAAABAQ

_No response_

Larodian#0384

### LINKED ISSUE #6592: New spoiler animation on landing
### Aircraft Version

Experimental

### Description

When plane lands with spoilers armed their animation are flickering. I think this animation should be more smooth, than now

### References (optional)

_No response_

_No response_

MrFunny#8972

crocket63:
Fixes #6592
Fixes #6447 
Fixes #180

## Summary of Changes
Implements basic rudder, elevator and spoilers hydraulic servo control.
Trim mechanism is not implemented as part of this PR.
ELAC and FAC logics are placeholders

All surfaces react to wind forces.

### Elevator
- [x] Basic ELAC logic placeholder
- [x] Using real actuator characteristics
- [x] Dual actuator control for high pitch demands
- [x] Left/Right elevator separate instances
- [x] Flight control page update
- [x] Actuator centering mode
- [x] Random actuator sealing values for realistic asymmetric droop on power off

### Rudder
- [x] Basic Fac logic placeholder
-  Using real actuator characteristics / ADDRESSED IN ANOTHER PR
- [x] Triple actuator control
- [x] Flight control page update

### Spoilers
- [x] Each panel is modeled separately and connected on the correct circuit
- [x] React to aerodynamic forces
-  Using real actuator characteristics / ADDRESSED IN ANOTHER PR

Real position logic for each panel only depends on FBW demands and flight control computers state, so it will be updated later and not part of this PR.

*Known issue*: Animations on default plane can't be separated so as a workaround the plane spoiler group is moving as the mean of all simulated panels. It means that if only one panel can move, visually the whole spoiler group will just move 1/5.

## Additional context

## Testing instructions

1. No flight control move when no hydraulic pressure

2. All flight controls react to wind AND/OR gravity. 
Use your imagination using custom weather menu / shutting hydrualics off/on while surfaces in various positions
Use your imagination in flight shutting hydrualics off/on while surfaces in various positions  
Of course rudder should move into the wind!

3. Flight surfaces should only move when their associated hydraulic system is pressurised (flight controls ECAM should be accurate)

X. In any conditions, no stupid wobbling should appear on any surface

Known issues:
1. Spoiler animation only shows the mean position of al panels. So you can't see per panel hydraulic deactivation, but whole assembly will just rise lower
2. Leak measurement valves are coming in a different PR, so cargo door operation will pressurise controls


aguther (approved):

aguther:
Can also be tested on experimental version.

crocket63:
Dunno who clicked on "rudder real characteristic and spoiler real characteristic" but it's not me :D

Those two system still has totally random number

aguther:
Ok sorry that was me because I assumed recent updates was that.

So then I suggest to remove it and introduce it with a later PR.

mico975:

Discord          : mico#3145
Object of testing: #6832
Tier of Testing  : 1
Date             : 12/03/2022

Testing Process: 
No flight control move when no hydraulic pressure - **OK**

All flight controls react to wind AND/OR gravity. - **OK**

Flight surfaces should only move when their associated hydraulic system is pressurised (flight controls ECAM should be accurate) - **OK**, I understand that PTU should not affect this, i.e. green pressurized via PTU doesn't enable green part of controls.

In any conditions, no stupid wobbling should appear on any surface - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

donstim:
Just now saw this PR. Why are the spoilerons (roll spoiler function) being removed? Although I'm not so sure the aerodynamic effect is modeled (but it should be checked), this will remove the animation effect on the exterior model. Although it wasn't accurate in that it used all spoiler surfaces, removing it entirely is probably worse.

Do I understand this correctly that flight speedbrake deflection angle will be noticeably less in the external model animation? Do we really want to do this? At the least, the support/docs guys should be informed as there are likely to be people who notice this.

aguther:
@donstim 

Speaking about Spoilerons:
Back some time I've tested if the spoilerons have any effect and they do not have any effect. This was confirmed by Seb.

The spoilerons are still working as before, the difference is only that it was removed from the flight model and is now done in the C++ code. So no functionality is lost at all, a simple flight test shows that.

crocket63:
> Just now saw this PR. Why are the spoilerons (roll spoiler function) being removed? Although I'm not so sure the aerodynamic effect is modeled (but it should be checked), this removed the animation effect on the exterior model. Although it wasn't accurate in that it used all spoiler surfaces, removing it entirely is probably worse.
> 
> Do I understand this correctly that flight speedbrake deflection angle will be noticeably less in the external model animation? Do we really want to do this? At the least, the support/docs guys should be informed as there are likely to be people who notice this.

Only effect I can think of from previous behaviour is that, because I added the ground spoiler group functionality (spoiler 5 not moving unless for ground braking), it might affect the final animation position by 1/5 of its movement (because final animation is the mean of all panels, and ground spoiler one is staying retracted)

donstim:
> @donstim
> 
> Speaking about Spoilerons: Back some time I've tested if the spoilerons have any effect and they do not have any effect. This was confirmed by Seb. Spoilers do not create any roll moment on the flight model.
> 
> In terms of animation, the spoilerons are still working as before, the difference is only that it was removed from the flight model and is now done in the C++ code. So no functionality is lost at all, a simple flight test shows that.

Actually, I did a "simple flight test" before submitting the comment. What I didn't do was zoom in enough to see the small, nearly imperceptible spoiler deflection now associated with the roll spoilers. I assume this is related to the 1/5 spoiler movement issue. The only indication of the roll control spoilers simulation was the animation and now that has been diminished to the point of it effectively not being there. 

We have already received one comment from a user that it is missing. I expect many more from users who know that the A320 spoilers are used for roll control. I guess I just don't understand what is gained at this point by this PR when the animation loss is readily apparent. Why not wait until the negative aspects can be eliminated?

aguther:
I can confirm that at least for the demand generation I tested what demand was being generated and replicated this exactly as it was. It seems I wasn't checking enough the "end-result" by comparing values. I can make another check in flight, but there was definitively spoilerons active and it looked ok to me when I tested it.

IMHO I suggest to fix it on the Rust side by not applying the 1/5 logic until a more detailed (surface by surface) demand is generated which should fix the issue.

donstim:
> > Just now saw this PR. Why are the spoilerons (roll spoiler function) being removed? Although I'm not so sure the aerodynamic effect is modeled (but it should be checked), this removed the animation effect on the exterior model. Although it wasn't accurate in that it used all spoiler surfaces, removing it entirely is probably worse.
> > Do I understand this correctly that flight speedbrake deflection angle will be noticeably less in the external model animation? Do we really want to do this? At the least, the support/docs guys should be informed as there are likely to be people who notice this.
> 
> Only effect I can think of from previous behaviour is that, because I added the ground spoiler group functionality (spoiler 5 not moving unless for ground braking), it might affect the final animation position by 1/5 of its movement (because final animation is the mean of all panels, and ground spoiler one is staying retracted).
> 
> Someone was looking into a possible animation split, hope it will be possible before having the new model.

nvm. I think I misinterpreted your statement about how the spoiler animation is affected. Even though only panels 2,3 and 4 are used as speedbrakes, I really don't see any difference in their external view deflection angles from the pr-PR version. According to your explanation I expected to see the deflection angles nearly cut in half (actually 60%), but I'm not noticing that.

crocket63:
Did not really notice that issue, will have a check. Ground spoiler implementation would only remove 20% of final animation not more.

donstim:
> Did not really notice that issue, will have a check. Ground spoiler implementation would only remove 20% of final animation not more.

I'm agreeing that I did not notice an animation issue with in-flight speedbrake deployment. Maybe the 20% is there, but I didn't notice it. The reason I thought it would only be 60% of what it was before is that only 3 of the 5 panels are used for in-flight speedbrake deployment. I guess you haven't gotten that far yet?

crocket63:
Only ground spoiler group was added. I'll remove it to gain back those 20%.

Then it's either demand is not a lot at speed, or spoiler actuators do not have enough force to reach demand. That's a possibility as this was merged before I get real actuator data so they might be way too weak.

crocket63:
There was in fact ground spoiler 20% issue PLUS another bit missing (we command actuator position but surface position is different due to motion ratio). Will push a fix in a new PR

Issue #6484, opened 2022-01-05
A32NX_AUTOPILOT_ACTIVE resetting when switching AP


rthom91:
### Aircraft Version

Development

### Build info

```json
{
  "built": "2022-01-04T06:33:26+00:00",
  "ref": "master",
  "sha": "93b60b06f5db427fe1d9bf93531a513a3562845f",
  "actor": "tracernz",
  "event_name": "manual"
}
```

### Describe the bug

When switching between AP1 and AP2, the local var A32NX_AUTOPILOT_ACTIVE immediately resets to 0 and then back to 1 when switching from AP1 to AP2. It DOES NOT reset when switching from AP2 back to AP1. This results in a bad sync with YourControls and disconnects AP for all clients.

### Expected behavior

A32NX_AUTOPILOT_ACTIVE should remain as 1.

### Steps to reproduce

Open model behaviors list of local variables, search 'autopilot', and switch back and forth between AP1 and AP2.

rthom91:
Ignore.

PR #6860, opened 2022-03-06
feat: add session_id tag to sentry

ErickSharp:
## Summary of Changes
Adds a new Sentry tag named 'session_id' that is generated using nanoid so that users can show it to support when they encounter a bug. On flypadosv3, this session ID will appear when the flypad itself crashes, and also will be available in the About settings page.

## Additional context


src/instruments/src/EFB/index.tsx
```diff
@@ -77,4 +79,12 @@ const EFBLoad = () => {
 
 readSettingsFromPersistentStorage();
 
+const ALPHABET = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const SESSION_ID_LENGTH = 14;
+const nanoid = customAlphabet(ALPHABET, SESSION_ID_LENGTH);
+const generatedSessionID = nanoid();
+
+Sentry.setTag('session_id', generatedSessionID);
```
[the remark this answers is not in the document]
beheh: @Benjozork Are you suggesting we should hold off on proceeding with this PR until that has been addressed, or can this land regardless?

beheh:
I've added a tag to hold off on merging for the time being.

@ErickSharp Any objection to moving this back to draft until the feedback from Benjozork above has been addressed?

ErickSharp:
> I've added a tag to hold off on merging for the time being.
> 
> @ErickSharp Any objection to moving this PR back to draft until the feedback from Benjozork above has been addressed?

Yes sorry I forgot about this PR. I'll make the changes necessary soon :)

frankkopp (approved):

src/instruments/src/EFB/index.tsx
```diff
@@ -49,6 +53,8 @@ const EFBLoad = () => {
         }, 6000);
     }
 
+    useEffect(() => () => setSessionId(''), []);
```
beheh, line 56: Won't this always clear the `A32NX_SENTRY_SESSION_ID` persistent property when the EFB intiializes?

As the `generatedSessionID` below will only be set once on a JS load, before the EFB renders.

ErickSharp: > Won't this always clear the `A32NX_SENTRY_SESSION_ID` persistent property when the EFB intiializes?
> 
> 
> 
> As the `generatedSessionID` below will only be set once on a JS load, before the EFB renders.

Wouldn't the cleanup function for the useffect run only when the component gets unmounted?

beheh: Ah, you're right, I missed the double-arrow.

PR #6879, opened 2022-03-10
feat(pfd): AUTO BRAKE label

salvapatuel:
## Summary of Changes
The AIRBUS had a software update that shows the AUTOBREAK status on the PFD. This small change reads from the state and presents the label on the PFD only if there are no AUTO-THROTTLE warnings.

## Screenshots (if necessary)
Real Life: 
REAL - AutoBrakes on PFD

New Implementation: 
FBW - AutoBreak PFD

## Additional context
Only if there is no AUTOTHROTTLE warning that uses 2 rows

## Testing instructions
ON GROUND: Set the engines and set the autobreak
ON AIR: Select autobreak on the air, also switch the autothrottle and set the power outside the CLB radius to make sure it dissapear the label and returns when solved the CLB issue.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ## 0.8.0
 
+1. [PFD] Included Auto BRK messages in PFD as latest real life update - @brainshot (brainshot)
 1. [FBW] Improved speed dependent roll law model - @IbrahimK42 (IbrahimK42)
 1. [SOUND] Pack sounds use new APU model variables - @tracernz (Mike)
 1. [HYD] Hydraulic system updated. Multiple sections. Better regulation - @Crocket63 (crocket)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## 0.8.0
 
-1. [PFD] Included Auto BRK messages in PFD as latest real life update - @brainshot (brainshot)
+1. [PFD] Added Auto BRK messages to the FMA - @brainshot (brainshot)
 1. [FBW] Improved speed dependent roll law model - @IbrahimK42 (IbrahimK42)
 1. [SOUND] Pack sounds use new APU model variables - @tracernz (Mike)
 1. [HYD] Hydraulic system updated. Multiple sections. Better regulation - @Crocket63 (crocket)
--- a/src/instruments/src/PFD/FMA.tsx
+++ b/src/instruments/src/PFD/FMA.tsx
@@ -89,6 +89,7 @@ const A1A2Cell = () => {
     const AThrMode = getSimVar('L:A32NX_AUTOTHRUST_MODE', 'enum');
 
     let text: string | undefined;
+    let autobreakText: string | undefined;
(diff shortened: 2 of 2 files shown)


src/instruments/src/PFD/FMA.tsx
```diff
@@ -89,6 +89,7 @@ const A1A2Cell = () => {
     const AThrMode = getSimVar('L:A32NX_AUTOTHRUST_MODE', 'enum');
 
     let text: string | undefined;
+    let autobreakText: string | undefined;
```
beheh, line 92:
```suggestion
    let autobrakeText: string | undefined;
```
Auto Break != Auto Brake 😉

.github/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ## 0.8.0
 
+1. [PFD] Included Auto BRK messages in PFD as latest real life update - @brainshot (brainshot)
```
beheh, line 9:
```suggestion
1. [PFD] Added Auto BRK messages to the FMA - @brainshot (brainshot)
```
Slightly reworded.

beheh (changes requested):
Hi there, thanks for contributing!

Two small pieces of suggestion right away. I think additionally we'll need to look for some references to specify the exact conditions when this appears (e.g. flight phase), but other than that this looks good so far!

src/instruments/src/PFD/FMA.tsx
salvapatuel: Yes, sorry for the typo, do you change it or do you want me to change it?

salvapatuel:
> Hi there, thanks for contributing!
> 
> Two small pieces of suggestion right away. I think additionally we'll need to look for some references to specify the exact conditions when this appears (e.g. flight phase), but other than that this looks good so far!

Thanks for the comment. The conditions are matched to the AUTOBREAK logic to keep everything in order. This means, if the AUTOBREAK functionality is enabled then the label will show as this reads: A32NX_AUTOBRAKES_ARMED_MODE. The only exception as the real Airbus is if you have a warning message from the autothrottle (that usualy uses 2 rows), as is more important than the autobreak the label will not be shown, once the AUTOTHROTTLE warning is disabled it comes back again.

tracernz:
We need to have a wider discussion about S14 and beyond if we are to do this. The PFD should really be consistent rather than have parts from one version of the standard and parts from another version.

salvapatuel:
> 

Done

tracernz:
Check it after you land, while it's decelerating.

beheh:
@salvapatuel Thanks again for your contribution! I think your PR motivated #6919 which contains this change a few others. I'd love for you to take a look at it, and if you agree that it contains everything from here we could close this PR.

salvapatuel:
> @salvapatuel Thanks again for your contribution! I think your PR motivated #6919 which contains this change a few others. I'd love for you to take a look at it, and if you agree that it contains everything from here we could close this PR.

Thanks again for the development of the new features, I have been testing it today and works great. Please do close it. Thanks again.

PR #6913, opened 2022-03-13
feat(efcs): initial flight control computers implementation

lukecologne:
## Summary of Changes

This PR adds the computers of the Electronic Flight Control System (EFCS): ELACs, SECs and FCDCs.
The FACs are not part of the EFCS but the AFS (Autoflight system), and will be only partially modeled in this PR (some features may be included though, to replace old functionality).
The Computers realistically acquire data from the correct sources (ADIRS, RAs, SFCCs, LGCIUs etc.) and compute their logics and laws from this data.
They also realistically communicate with each other both via busses, and also discrete data.

The laws used in the computers are identical to the normal laws used before, and the alternate and direct laws have been added as a first version. They will be adjusted in the future to be more accurate/have better handling.
The laws are selected accurately via the computers built-in logic, which also performs other functions such as monitoring of the peripherals,(ADIRS, RA etc), and computation of miscellaneous logic, for example Aileron Droop, Ground spoiler and speedbrake functions.

The SECs now also command seperate spoiler extension commands for each spoiler panel, although the 3D model still isn't seperated in this regard, so the cockpit displays will accurately depict the spoiler extensions, but the model will only show the mean value of all spoilers.
The sim's spoiler position is now also computed from the actual spoiler positions.

The computers' outputs are transmitted to the hydraulics, with seperate commands for each computer/servo, so that failure cases are accurately modeled.
The communication with the other aircraft systems (displays for example) is handled via the FCDC, which acquires bus and discrete outputs of the other computers, and concentrates and outputs this data via it's bus and it's own discretes.

The sidestick priority logic is also implemented, it is computed by all the flight control computers (ELAC/SEC) independently, and it is consolidated and converted to output to the sidestick priority lights and it's bus by the FCDC. Aural warnings (DUAL INPUT, PRIORITY LEFT/RIGHT) are not yet implemented, as these are computed in the FWC.
Of course, without any proper cockpit sharing it will be useless, in particular, the left sidestick is assumed to be used at the moment (although this does not reflect in the animation). This will probably be configurable in the future.

Todo:
* Improve direct and alternate law: Direct law is not scaled down based on Flaps/Slats configuration, thus can be very sensitive at high speeds, Alternate law still uses TAS for the C* law, this is incorrect and will result in issues if no ADR is available.
* V_stall warn AoAs need to be tuned, currently they are just the same as alpha_prot.
* In some situations, AP might not be available in alternate law even though it should be. This is due to how the AP commands are currently executed, which requires the normal laws to be active (IRL this is different). Will take some time to fix this, not planned for this PR.
* The electrical bus connections are not quite correct yet, the ELAC 2 has rather complex logic for powersupply switching, which is not done yet. Not sure if that will make it in this PR.

## Additional context


crocket63:
Depending when this is merged, https://github.com/flybywiresim/a32nx/pull/6342 can provide hydraulic pressure switches (including hysteresis) directly.

donstim:
Adding this as a reminder. Need to be careful about how the overall spoiler deflection will be determined for injecting into MSFS. Currently, the lift/drag increments in MSFS are based on the max deflection angle in the flight_model.cfg file converted from degrees to radians (i.e., deg/57.3), then multiplied by the spoiler cl/cd increment in the flight model.cfg file. With the current (incorrect) 35 degree max deflection angle in the flight_model.cfg file, this means that the .cfg cl/cd increments are multiplied by 35/57.3, or 0.6108. Any changes to the max deflection angle, either by averaging all of the spoiler panels, or by correcting the in air or on ground max deflection angle needs to be accompanied by a corresponding change to the cl/cd increments in order to maintain the current flight performance/characteristics effects from spoiler deployment.

For example, for in-flight use of speedbrakes, the max deflection of spoilers 3 and 4 is 40 degrees, and the max deflection of spoiler 2 is 20 degrees. Spoilers one and 5 are not used at all. So, the average of the 5 spoiler max deflection angles would be 20 degrees. The flight_model cl/cd increments would need to be multiplied by 35/20 = 1.75 to retain the current cl/cd increments that are applied in the sim. 

Averaging the position of all 5 panels (per side) for determining the effective speedbrake angle doesn't seem to be correct since 2 of the panels are not being used at all as speedbrakes and one of the panels only extends half as much as the 2 panels that fully extend.

2hwk (approved):
Approved (for c++/react/ts) but still needs proper review of the fbw matlab and the (minor) system changes written in rust.

src/systems/a320_systems_wasm/src/spoilers.rs
```diff
@@ -91,5 +91,67 @@ pub(super) fn spoilers(builder: &mut MsfsAspectBuilder) -> Result<(), Box<dyn Er
...
+        |values| values,
+        Variable::named("HYD_SPOILER_4_LEFT_DEFLECTION"),
+    );
+    builder.map(
+        ExecuteOn::PostTick,
+        Variable::aspect("HYD_SPOIL_5_LEFT_DEFLECTION"),
+        |values| values,
+        Variable::named("HYD_SPOILER_5_LEFT_DEFLECTION"),
+    );
+
```
crocket63, line 155: As we now use directly the internal system values, seems like we don't need those .map at all now.
That'd would need a quick test to be sure though :)

crocket63 (dismissed):

crocket63:
Spoilers to check after my last 2 commits

src/instruments/src/PFD/shared/PFDSimvarPublisher.tsx
```diff
@@ -110,15 +102,40 @@ export interface PFDSimvars {
     autoBrakeDecel: boolean;
     fpaRaw: number;
     daRaw: number;
+    latAccRaw: number;
     ls1Button: boolean;
     ls2Button: boolean;
```
2hwk, line 107: Missing ``fcdcDiscreteWord1``

PR #6918, opened 2022-03-14
fix(APU): APU Bleed pressure not shown with ADIRI1 off

salvapatuel:
## Summary of Changes
The APU page is not showing the Bleed Air PSI when the ADIRI1 is not aligned which does not follow the right procedure as explained in the APU manual.

## Screenshots (if necessary)
APU page shows no pressure
BLEED ISSUE 2

Pneumatic bleed is working from the APU
BLEED ISSUE 1

Fixed Version with no ADIRI1:
image

## References
Manual shows no link to ADIRI1
image

## Testing instructions
Current issue: Start the APU and the APU bleed without aligning the ADIRI1
Correct state: Regardless of the ADIRI1 the bleed should be displayed.


lukecologne:
This change is incorrect. The bleed indication is indeed shown as unavailable if the ADR is not aligned, since only the total static pressure is delivered by the APU control box. To display the bleed pressure, the ambient static pressure is subtracted, which is delivered by the ADR. So, the APU bleed indication cannot be shown if the ADR is off.

tracernz:
The existing logic was almost correct, but it should look at the SSM for the corrected average static pressure for the ADR corresponding to the DMC drawing the page, rather than the knob position.

PR #6922, opened 2022-03-15
feat(cond): randomized starting positions for air conditioning selector knobs 

### LINKED ISSUE #3128: [REQUEST] More realistic starting positions for AIRCOND knobs

All three of the air conditioning knobs are always in the same starting position C&D (now 12 o'clock). That doesn't account for the small variations that happen from situation to another, and from pilot to pilot.

1- Instead of a fixed starting position for all three knobs, we could implement a **range** of positions, one of which gets selected when loading the plane. Either randomly, or based on distance from the equator/seasons or both, according to what IRL pilots find works best.

2- The 3 knobs shouldn't have the same exact starting value either, as if they were last turned by a robotic hand. So maybe a smaller range among the 3 knobs in the **same flight** as well.

**References**

According to @ImenesFBW in PR #3118 "Pilots have been giving feedback about how they rarely touch the air conditioning knobs and **usually** find them in the **10-12 o'clock** position in a cold&dark cockpit."

**Additional context**

3- If pilots reported that there is a 10% chance the knobs could be **outside** the 10-12 range for example, it would be nice to account for that as well.

rare-potato:
Fixes #3128

## Summary of Changes

On plane load the air conditioning selector knobs will be set to positions other than 12 o'clock based on a multimodal distribution centered on the 11 o'clock, 12 o'clock, and 1 o'clock notches, with more weight on the 11 and 12 o'clock notches.

## Additional context

Task List:
- [ ] Get feedback on most common positions outside of the 12 o'clock position
- [x] Program the multimodal distribution
- [ ] Determine if code should be implemented in a different place. 
- [ ] Make some temperature settings more common based on ambient air temperature
- [ ] Finish Pull Request formatting

Discord username (if different from GitHub): Rare Potato#9693

## Testing instructions

Load plane multiple times checking that the A/C Selector knobs change position based on the logged 
"WASM: A/C Knob Rounded Variable"


tracernz:
Take a look in https://github.com/flybywiresim/a32nx/tree/master/src/systems. The readme there explains some basics. Drill into the systems and a320_systems directories to find the code.

wpine215:
Will this be revisited anytime soon?

rare-potato:
Yeah, I'll go ahead and mark this closed since it's been so long since I last worked on it. I might revisit the issue at a later date with a new PR on a later fork if someone else doesn't pick it up before hand.

tracernz:
It appeared this was pretty close? Did it just need rebased, or still some implementation left to do?

rare-potato:
It mainly needed to be finished rebasing from my initial implementation. If I recall correctly I had started moving it to Rust but was having difficulty understanding the airframe initialization dataflow and never got around to finding someone to help me better understand it.

In other words, I wasn't able to figure out how exactly when/where the initialization of the airframe occurred to set the appropriate SIMVARS but wouldn't repeat itself.

PR #6937, opened 2022-03-18
fix: close old websocket before every connection attempt

Saschl:
Fixes #[issue_no]

## Summary of Changes

If a Websocket fails to connect it doesn't seem to close but stays open (visible in the debugger and in ram usage). If the MCDU Server is not used this leaks a Websocket connection every attempt. This slowly fills up the RAM and lead to increased CPU usage.

I removed the "readyState" condition on a reconnect so the old socket is closed and properly garbage collected.

## Screenshots (if necessary)

Before:
Lots of sockets...
image

After:
image
Just one WebSocket that also reports closed.
## Additional context


Benjozork (approved):

Saschl:
Merging this as it's been approved and a trivial fix that doesn't change the behaviour and functionality of the feature.

PR #6942, opened 2022-03-19
feat(efb): add active failures to comfort ui

Louis-Deez:
## Summary of Changes
Added active failures to comfort UI.

## Screenshots (if necessary)
image


Benjozork (approved):

ErickSharp:
To me, this wastes a great deal of space that makes the UI unnecessarily claustrophobic. I think if we are going to keep this we need some sort of horizontal scrolling.

FlightSimmer7700:
Just loaded into this to try a qa test it. Not sure why but whole planes systems don't work and clicking on the efb doesn't turn it on. Whole plane doesn't turn on basically.
Would also be nice if we were provided with a bit more precise information on what exactly it is there is to test and how it functions :)
Adding on to @ErickSharp's suggestion I think some sort of thin bar which comes up when there is a failure which you can then click on and open to show a list of the failures might be a better option?
image

2hwk:
Please solve merge conflicts

PR #6960, opened 2022-03-24
fix(hyd): fixed visual range of spoiler animations

crocket63:
Fixes #[issue_no]

## Summary of Changes

Having ground spoiler simulated in systems made animation losing 20% of range.

Further more we now ouput actuator position instead of  surface position. This is a workaround until we decide if command is in angle or in actuator position when control computers are implemented.

## Additional context

## Testing instructions

Test spoilers in various conditions.

Make sure that position of spoilers A32NX_HYD_SPOILERS_XXX_DEFLECTION is close to demand from FBW : A32NX_SPOILERS_XXX_DEFLECTION_DEMAND 


aguther (approved):

aguther:
Code looks ok to me. Functional wise it also seems to be ok for me.

@donstim Can you also check on experimental version to confirm?

crocket63:
Note this won't solve the demand value not changing when AP state changes though. Not sure if it's been like this forever or if it's something from the new spoileron small code added to fbw module?

aguther:
I just tested it in flight and there is definitively a difference between manual flight and with autopilot engaged (as it was in the past).

Looking at the spoiler demand code, it seems spoilers are limited from full deployment to half deployment when AP is engaged, but they do not extend when AP is disengaged. This seems very intentionally when you look at the code, otherwise it would be possible to implement with a simple limiter. I did a short search for the reference but I didn't find it yet.

I would consider this for the moment as "ok".

PR #6996, opened 2022-04-02
fix: SU9 issue with engine control

aguther:
## Summary of Changes
Fixes a weird issue with SU9 that causes the engine control to stop working. This happens when 100% is written to the sim throttle. In many cases this is when TOGA thrust is applied and the plane starts rolling and thrust limit is further increased.

## Testing instructions
- start on the runway in SU9
- take-off with TOGA power


tracernz (approved):

Pelice:
[Development and Experimental brach]
I think I've the same problem: after take off engines shut down and ECAM turn off. 
I've tried everything (cleaned community, changed settings in game and in efb following the recommended settings,.. ) but with developement and experimental branch I've got always this issue. 
I just discovered that with the stable branch, everything work fine.  
If from stable I switch to dev it doesn't work, then going back to stable and it works fine (without touching everything else).
Stable 0.8.1 - Dev 4bc7e01 - Exp 39fd7d4

PR #6997, opened 2022-04-02
fix: some small TCAS improvements

aguther:
## Summary of Changes
- use the actual radio altitude value
- reset all variables when TCAS is turned off to ensure clean state when you have been in a TCAS RA

## Testing instructions
- fly the plane and try to get into TCAS RA
- turn TCAS off -> result should be a clean state on the PFD

### DIFF
--- a/src/tcas/src/components/TcasComputer.ts
+++ b/src/tcas/src/components/TcasComputer.ts
@@ -467,7 +467,7 @@ export class TcasComputer implements TcasComponent {
             // information, we need to rely on the fallback method
             // this also leads to problems above 1750 ft (the threshold for ground detection), since the aircraft on ground are then shown again.
             // Currently just hide all above currently ground alt (of ppos) + 380, not ideal but works better than other solutions.
-            const groundAlt = this.planeAlt - this.radioAlt; // altitude of the terrain
+            const groundAlt = this.planeAlt - this.radioAlt.value; // altitude of the terrain
             const onGround = traffic.alt < (groundAlt + 360) || traffic.groundSpeed < 30;
             traffic.onGround = onGround;
             let isDisplayed = false;
--- a/src/tcas/src/components/TcasComputer.ts
+++ b/src/tcas/src/components/TcasComputer.ts
@@ -1138,6 +1138,15 @@ export class TcasComputer implements TcasComponent {
         this.updateInhibitions();
         this.updateStatusFaults();
         if (this.tcasMode.getVar() === TcasMode.STBY) {
+            this.advisoryState = TcasState.NONE;
+            this.tcasState.setVar(TcasState.NONE);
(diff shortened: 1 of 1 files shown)


src/tcas/src/components/TcasComputer.ts
```diff
@@ -467,7 +467,7 @@ export class TcasComputer implements TcasComponent {
             // information, we need to rely on the fallback method
             // this also leads to problems above 1750 ft (the threshold for ground detection), since the aircraft on ground are then shown again.
             // Currently just hide all above currently ground alt (of ppos) + 380, not ideal but works better than other solutions.
-            const groundAlt = this.planeAlt - this.radioAlt; // altitude of the terrain
+            const groundAlt = this.planeAlt - this.radioAlt.value; // altitude of the terrain
```
tracernz, line 470: Looks like this `updateTraffic` function runs even if the selected radioAlt state is failure warning... probably the TCAS should change state in line 294 if no radio altitude is available? https://github.com/flybywiresim/a32nx/blob/870c4a82217b312a139826edf49f61e8591aee0b/src/tcas/src/components/TcasComputer.ts#L294 I didn't find a ref for this case so far however...

aguther: True, I also guess the TCAS is inop with RA failure.

tracernz: You didn't actually make this any different though... just made it actually use the RA altitude... that's at least still a good improvement. 👍

tracernz (approved):

src/tcas/src/components/TcasComputer.ts
beheh: Yeah, this one can be merged regardless as it fixes a bug with functioning RAs (SSM=NO). TCAS INOP should already be properly working including ECAM on dual RA fault.

src/tcas/src/components/TcasComputer.ts
```diff
@@ -1138,6 +1138,15 @@ export class TcasComputer implements TcasComponent {
         this.updateInhibitions();
         this.updateStatusFaults();
         if (this.tcasMode.getVar() === TcasMode.STBY) {
+            this.advisoryState = TcasState.NONE;
+            this.tcasState.setVar(TcasState.NONE);
+            this.correctiveRa.setVar(false);
+            SimVar.SetSimVarValue('L:A32NX_TCAS_STATE', 'Enum', 0);
+            SimVar.SetSimVarValue('L:A32NX_TCAS_RA_CORRECTIVE', 'bool', 0);
```
beheh, line 1145: These two lines can be dropped, as they duplicates the lines above (`this.tcasState.setVar`/`this.correctRa.setVar`):
```suggestion
```

PR #7014, opened 2022-04-06
feat: use new steering axis for nosewheel

aguther:
## Summary of Changes
This PR switches the control of the nosewheel steering to the new axis for steering that was introduced in SU9.

**For that purpose this PR also updates the dev-env to use SDK 0.18.0.0.**

**NOT TO BE MERGED BEFORE SU9 RELEASE**

## Testing instructions
- test if airplane still works normal (FBW, AP, Systems)
- test if nosewheel steering is working with new axis

### DIFF
--- a/src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
+++ b/src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
@@ -42,7 +42,7 @@ pub(super) fn nose_wheel_steering(builder: &mut MsfsAspectBuilder) -> Result<(),
     // Lacking a better event to bind to, we've picked a mixture axis for setting the
     // tiller handle position.
     builder.event_to_variable(
-        "AXIS_MIXTURE4_SET",
+        "AXIS_STEERING_SET",
         EventToVariableMapping::EventData32kPosition,
         Variable::aspect("RAW_TILLER_HANDLE_POSITION"),
         |options| options.mask(),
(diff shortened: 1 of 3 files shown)


src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
```diff
@@ -42,7 +42,7 @@ pub(super) fn nose_wheel_steering(builder: &mut MsfsAspectBuilder) -> Result<(),
     // Lacking a better event to bind to, we've picked a mixture axis for setting the
     // tiller handle position.
```
pepperoni505, line 43: Should these comments be removed?

crocket63:
Can't test new beta, but is that new event just a working steering set event, or do this solve the issue of rudder affecting that steering? I think Boris told me it was the latter, so this pr might need a total removal of the rudder hack if it's the case.

src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
aguther: It seems that they just added a new and distinct axis, but the remaining behavior seems to be the same. Therefore we still need it.

davidwalschots: I agree with @pepperoni505. I think the comment should be removed, as we're no longer using the "mixture axis" when this change is incorporated.

```suggestion
```

aguther: Agree...done

crocket63 (approved):

davidwalschots (approved):

PR #7020, opened 2022-04-08
refactor(fmgc): message queue

derl30n:
## Summary of Changes

Made fmgc message queue a separate module/sub-system

## Additional context

Added separate message queue from https://github.com/flybywiresim/a32nx/pull/6076

## Testing instructions

Test mcdu interactions and messages. (no visible difference to master)


tracernz:
Does this mean we can have scratchpad/queue per subsystem, like the real one? Next steps after that are to keep the page you were on when switching back to a subsystem from MCDU MENU, and annunciator lights when a subsystem wants attention.

derl30n:
> Does this mean we can have scratchpad/queue per subsystem, like the real one? Next steps after that are to keep the page you were on when switching back to a subsystem from MCDU MENU, and annunciator lights when a subsystem wants attention.

To add to that, FMGC messages are not displayed in another subsystem, this is also important to note, I assume this is where different queues come into place.

Generally speaking there is a lot to be done, also the scratchpad needs a special state for a message like `please wait`, which are type I but prevent input into the scratchpad while the message is displayed (basically not clearable). While that may be easy to accomplish given the "standalone" scratchpad by adding the attribute "locked" to its class, testing will be needed.

Hopefully this can be accomplished in small prs step by step rather than in one big chungus xD

BlueberryKing (approved):
Very nice, quite a bit cleaner!

mico975:

Discord          : mico#3145
Object of testing: #7020
Tier of Testing  : 1
Date             : 15/04/2022

Testing Process: 
Tested MCDU interactions thoroughly and everything works as before.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7036, opened 2022-04-09
fix(bleed): Fix stack overflow when container contains NaNs

Gurgel100:
Fixes #[issue_no]

## Summary of Changes

This PR fixes a small issue introduced with #6810 where the systems wasm module would run into a stack overflow which leads to a sim crash. The stack overflow only occurs if there ever appears a NaN in a PneumaticContainer (which should never happen).

## Additional context

## Testing instructions

Really not needed because it just switches around the if condition.


Benjozork (approved):

NUTTTTTTTT:
Merging without QA as the change is minor and does not affect functionality. Code review completed.

PR #7045, opened 2022-04-12
fix: change sentry.io sample rate to 10%

Benjozork:
## Summary of Changes

* Changes sentry.io sample rate to 10% to avoid exceeding our quota. Default was 100%, which is way too much.

## Testing instructions

* Make sure screens still work, ig


ErickSharp (approved):
😏

PR #6671, opened 2022-02-08
fix(athr): fine tuning of athr concerning speed undershoot

IbrahimK42:
## Summary of Changes
This PR is a fine tuning of the ATHR normal mode concerning speed undershoot. The ATHR app mode is merged with the ATHR normal mode. The ATHR app mode is no more necessary. The engines shall behave faster in case of speed deceleration.

## Testing instructions

set clear weather:

1. do some speed acceleration and deceleration
1. the engines shall react faster than before. 
    maximum undershoot shall be -5 kn and maximum overshoot shall be 10 kn

## Media

You can see the behaviour here:

https://streamable.com/s8m6a2

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #7059, opened 2022-04-15
feat(engine): Realistic engine spool model

Taz5150:
Fixes #[issue_no]

## Summary of Changes

Updated engine model with a new and realistic spool characteristic:
- Concave engine spool characteristic at low N1s (instead of convex)
- Faster engine stabilization (time-constant) at commanded N1
- Smoother N1 behavior (lower spiking) between segments

## Screenshots (if necessary)

image
Actual engine model vs. real-life

image
New engine model vs. real-life

Note: both tests performed at the same airport and environmental conditions as its real-life counterpart

## References

Real life engine spool characteristic taken from: https://www.youtube.com/watch?v=xHxSU7PRujg

Druk Air A320-251N taking off from Paro, Bhutan. QNH 1025, UTC 2:05, Temperature -1 Celsius

## Additional context

## Testing instructions

To test against the YT video, start the plane at Paro (VQPR). Set environmental conditions to AGL: Temp = -1 Celsius and QNH = 1025 mb. In the aircraft, set Packs Off and QNH to 1025.


hotshotp (approved):
While I cannot review code, I tested the actual spool behavior and it LGTM, slow at the beginning and then faster later on

aguther:
Needs definitively testing of ATHR behavior in different situations, but especially in various approaches with different weights and wind.

donstim:
I have a question about the comparison you show. Your "real" N1 depiction stops at less than 80% N1. But the YouTube video you linked looked like it was using Flex 23C and a target N1 at zero airspeed of what looks like 89.3% with a mostly brakes on engine spool up. (I don't know how you are able to read those blurred numbers or how you are doing it at intervals of less than 1 second!:grin: )

In the A32NX, for a Flex of 23C, it shows a target N1 of 88.7% at the Paro conditions you specified. From the FCOM chart for a pressure altitude of 7000 feet (which is about what you get for QNH 1025 at the Paro elevation), it has 89.3% N1 for 0 degrees C TAT. So maybe the actual temp at that time was 0 degrees? But then, it is confusing to me that they got an N1 of 89.3% (the sane as the TOGA N1) for a Flex takeoff. (At that pressure altitude, it looks like Tref is no longer ISA+29C -- probably more like ISA+20C).

For QA testers trying to replicate the entire spin-up curve, the plateau at 50% is achieved by first pushing the thrust levers up to command 50%, pausing until the engines stabilize there, then setting the thrust levers to TOGA or FLEX. The engine spin up won't pause at 50% by itself.

Taz5150:
> Needs definitively testing of ATHR behavior in different situations, but especially in various approaches with different weights and wind.

Agreed, although I don’t expect changes in the overall performance behavior.

The new model has a different engine accel/ decel characteristic, specially in the low N1 regime (below 50%). This could affect minimally when making high N1 transitions, such as from Cruise to Idle while descending (from N1 in the 70s to 40s), but the difference should not be higher than a couple of seconds maximum worst case. Transitioning between lower ranges should not see any difference at all.

But yes, I agree this has to be tested 👍🏻

aguther:
The "issue" is, that to compensate for slow reaction in the lower N1 area we needed to increase the gain of the ATHR controller to get acceptable results. If the engines are now a bit faster, there is a risk of overreaction.

My first short tests looked good, so I put it on experimental version. I think it will be fine.

Taz5150:
Hi @donstim. You are absolutely right. Let me answer one by one to your questions/ doubts:

> @Taz5150, I have a question about the comparison you show. Your "real" N1 depiction stops at less than 80% N1. But the YouTube video you linked looked like it was using Flex 23C and a target N1 at zero airspeed of what looks like 89.3% with a mostly brakes on engine spool up. (I don't know how you are able to read those blurred numbers or how you are doing it at intervals of less than 1 second!😁 )

To get the N1 values as a function of time from the video, I use VLC. It allows me to zoom on the video and get time figures in milliseconds. Numbers are blurred, but can be read and with some "fine tuning" moving up and down on the video, I can get the values.

You are right: Flex is 23C and N1 limit is set at 89.3%. The reason why my "real" figure stops a little lower is because I could not get good readings on those, but I can confirm at some point in the video N1 goes as high as 90% during takeoff roll. I will try again and see if I can get at least a higher reading to complete the "real" characteristic.

> In the A32NX, for a Flex of 23C, it shows a target N1 of 88.7% at the Paro conditions you specified. **[Edit: Thought I could run the packs from the APU, but we apparently don't have that capability yet. With packs off, TOGA N1 at -1 deg C is 89.1%.]** From the FCOM chart for a pressure altitude of 7000 feet (which is about what you get for QNH 1025 at the Paro elevation), it has 89.3% N1 for 0 degrees C TAT. So maybe the actual temp at that time was 0 degrees? But then, it is confusing to me that they got an N1 of 89.3% (the same as the TOGA N1) for a Flex takeoff. (At that pressure altitude, it looks like Tref is no longer ISA+29C -- probably more like ISA+20C).

Regarding the temperature, yes it might be that the ambient temperature was actually 0C. What I had to do to get the temperature was a little complicated and I could be wrong by a degree or so. The video was uploaded on Jan 16th, 2022 and I knew it was a Druk Air flight taking off from Paro, at UTC 02:05 and QNH 1025. So basically I had to find on which day that aircraft was taking off from Paro at UTC 02:05 and double check that QNH was 1025. I found out that flight took off on Jan 10th, 2022. The METAR at VQPR UTC 02:00 was: 

**METAR VQPR 100200Z VRB02KT 9999 SKC M01/M01 Q1025=**

But, the METAR at 02:30 was:

**METAR VQPR 100230Z 01003KT 9999 NSC 00/M00 Q1025**

So it could be that the real ambient temperature was 0C at that moment.

> For QA testers trying to replicate the entire spin-up curve, the plateau at 50% is achieved by first pushing the thrust levers up to command 50%, pausing until the engines stabilize there, then setting the thrust levers to TOGA or FLEX. The engine spin up won't pause at 50% by itself.

Taz5150:
> The "issue" is, that to compensate for slow reaction in the lower N1 area we needed to increase the gain of the ATHR controller to get acceptable results. If the engines are now a bit faster, there is a risk of overreaction.
> 
> My first short tests looked good, so I put it on experimental version. I think it will be fine.

I see what you mean, and yes, it could mess up controller settings for ATHR in that case. Good to know some tests looked fine!!!

brewers10:

Discord          : brewers_10#5706
Object of testing: #7059 
Tier of Testing  : 1
Date             : 16/04/2022

Testing Process: 
Spawned in VQPR Runway 15. QNH 1025 and Temp -1. Flex 23oC. Flight deck replicated including heading + altitude.
Takeoff 1: Spooled engines up to 50%, sound matches and flex 23 was then set. Sound also matches.
Takeoff 2: Spooled engines up to 50%, sound once again matches and flex 23 was set. Sound matches once again.
Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
Very accurate.

mico975:

Discord          : mico#3145
Object of testing: #7059
Tier of Testing  : 1
Date             : 17/04/2022

Testing Process: 
Haven't done the check at VQPR as it is already done.

Did some flights and multiple (gusty) approaches with experimental version without major A/THR issues.

Negatives:       
Undershoot from idle can get to around -10kt which probably means that some A/THR gains will need to be tweaked - see #6671. Once it's above roughly 35% N1 it performs flawlessly.

Testing Results: 
Passed

Conclusions: 
LGTM

aguther:
We need to remember that even the real ATHR undershoots the target speed in those cases. What I’ve seen so far seems very close to what the real plane does (as far as I can tell).

donstim:
Checked it against this - https://github.com/flybywiresim/a32nx/issues/5262 and it still handles that closed issue well.

PR #7062, opened 2022-04-15
ci: recognise external downloads in installer

FoxtrotSierra6829:
## Summary of Changes

- add install.json to the distributed zip so the installer recognizes it as an installation

## Additional context

Discord username (if different from GitHub): Foxtrot Sierra#6420

## Testing instructions

- download PR artifact and install
- see the installer recognize it as updatable


aguther:
I've looked at the resulting zip and the install.json is added. What is a bit strange is that the structure of the zip is different than on other PRs. New PR like from master now have another zip inside and have a top folder included. This is missing here, so I wonder if the PR breaks this.

FoxtrotSierra6829:
> What is a bit strange is that the structure of the zip is different than on other PRs. New PR like from master now have another zip inside and have a top folder included. This is missing here, so I wonder if the PR breaks this.

Yes it is intentional to remove the double zipping, talked to QA about this

FoxtrotSierra6829:
@aguther fixed

aguther (approved):

PR #7068, opened 2022-04-16
feat: added config parameter to change key increment for throttles

aguther:
## Summary of Changes
This PR adds an option to the Throttle configuration that allows the change of the increment that is used when key events are pressed (normal and small).

The default increment is as before, and users should not see any difference. If a user wants to change that, he will be able to do so using the configuration file.

If requested, this can be added to the EFB later as an additional option in the Throttle calibration.

## Testing instructions
Test if the values can be changed in the `ThrottleConfiguration.ini` with the following entries:
```
[throttle_common]
; determines the increment for a key event (+/-):
key_increment_normal = 0.05
; determines the increment for a "small" key event (+/-):
key_increment_small = 0.025
```

Change the values in the config file and reload them (you can use the EFB button to load the settings from file for this).

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -238,6 +238,7 @@
 1. [PFD] Reimplement PFD with msfs-avionics framework for improved performance - @saschl (saschl#9432)
 1. [GPWS] Mute "100 above" callout during autoland - @patmack14 (Patrick Macken)
 1. [MCDU] Refactored mcdu key input events - @derl30n - (Leon)
+1. [AUTOTHRUST] Added configuration option to allow change of key increment size - @aguther (Andreas Guther)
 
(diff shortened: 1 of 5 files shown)


svengcz (approved):
LGTM

mico975:

Discord          : mico#3145
Object of testing: #7068
Tier of Testing  : 1
Date             : 17/04/2022

Testing Process: 
Test if the values can be changed in the ThrottleConfiguration.ini with the following entries - **OK**

Change the values in the config file and reload them (you can use the EFB button to load the settings from file for this). - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM, I agree with adding this to the throttle calibration page on the EFB.

PR #6969, opened 2022-03-25
refactor: W&B and boarding logic prep for EFB

sidnov:
Fixes #[issue_no]

## Summary of Changes

- [x] Basic optimization and rewrite for Boarding and OFP Data processing on the MCDU
- [x] Improved conversion when using LBS
- [x] Changed conversion factors overall (uniform across code, this is an experimental change, can be reverted if known to cause unnecessary effects)

## Additional context

Discord username (if different from GitHub): sidnov# 8337

## Testing instructions

1. Test the MCDU ATSU AOC W/B pages and make sure they work normally
2. Values in cargo hold now depict either Metric Tons or Metric Pounds depending on the unit selected on the EFB Settings for aircraft configuration
3. Test Simbrief loading using OFPs generated in KGs and LBS units
4. Test Manual loading with both KGs and LBS units selected.
5. Make sure the unit selected on the EFB->aircraft config is the same as the unit in Simbrief OFP generation(right hand sidebar)/airframe settings
6. Match payload and ZFW values after loading with that given in the Simbrief OFP

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #7078, opened 2022-04-17
fix: Payload weight conversion discrepancies

Taz5150:
Fixes ##6802

## Summary of Changes

Fixes discrepancies in payload (both Pax and Cargo) weight conversion (Lbs->Kgs and Kgs->Lbs) when comparing with SimConnect SimVars.

## Additional context

## Testing instructions

Go to W&B page in ATSU -> AOC Menu
Enter PAX number and compare to reported total payload stations on simconnect
Enter Cargo Tons and compare to reported total payload stations on simconnect

### DIFF
--- a/src/fadec/src/EngineControl.h
+++ b/src/fadec/src/EngineControl.h
@@ -60,8 +60,8 @@ class EngineControl {
   int engineImbalanced;
   double paramImbalance;
 
-  const double LBS_TO_KGS = 0.453592;
-  const double KGS_TO_LBS = 2.20462;
+  const double LBS_TO_KGS = 0.4535934000001;
+  const double KGS_TO_LBS = 2.2046176156881;
   const double FUEL_THRESHOLD = 661;  // lbs/sec
 
   bool isFlexActive = false;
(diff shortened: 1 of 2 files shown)


src/fadec/src/EngineControl.h
```diff
@@ -638,27 +638,28 @@ class EngineControl {
   /// </summary>
   void checkPayload() {
     double fuelWeightGallon = simVars->getFuelWeightGallon();
+	double paxWeightLbs = simVars->getPaxWeightKgs() * KGS_TO_LBS;                            // in LBS
```
tracernz, line 641: Indentation issue?

src/fadec/src/EngineControl.h
```diff
@@ -60,8 +60,8 @@ class EngineControl {
   int engineImbalanced;
   double paramImbalance;
 
-  const double LBS_TO_KGS = 0.453592;
-  const double KGS_TO_LBS = 2.20462;
+  const double LBS_TO_KGS = 0.4535934000001;
+  const double KGS_TO_LBS = 2.2046176156881;
```
tracernz, line 64: Can you elaborate on the source of these. Are they MSFS constants, or standard conversion factors (IIRC there is a discrepancy between those)?

donstim:
This works when the user is using kg as the weight unit for the airplane and simBrief OFP, but does not work if lbs is used. For example, I have a simBrief OFP with 159 passengers using a passenger weight of 176 lb/pax and 33 lb/pax of baggage. Payload should be 159 times (176+33) = 33,231. That is what simBrief shows, but the MCDU shows 32,600. Cargo weight in the MCDU is 2200 lb (rounded off), but it should be 5,247. With zero cargo, MCDU shows a payload of 29,445 vs what should be 27,984.

If MCDU thinks the passenger only payload is  29,445, but the passenger + bags payload is 32,600, then it must think the cargo payload is 3,155 even though it shows it as 2,200? If I "trick" the cargo loading by inserting a total passengers of 159, which zeros out the cargo, then select a cargo load, then select zero cargo, it then loads the correct cargo amount. But because the passenger payload is incorrect, the total payload is still incorrect.

Not connected with the PR, but it seems odd to me that if the user loads the payload manually, inserting a total passenger payload does not load any bags, but if a number is selected for the cargo hold, the weight of the passenger's bags are added to that value. That is sure to confuse people.

src/fadec/src/EngineControl.h
Taz5150: True. Updated already. 

Thank you!

src/fadec/src/EngineControl.h
Taz5150: I've set the conversion being used by the ATSU/ AOC code. Since I was not able to find the actual conversion factors in the code, I used SimvarWatcher to check PAYLOAD STATION simvars both in Lbs and Kgs.

Taz5150:
Thanks @donstim. I have some doubts regarding your questions, and will elaborate below.

On a general note, I did test both Kg and Lbs config settings in the EFB and seemed to work just fine. BUT I did not check SimBrief OFP integration. I suspect that, if the issue is with SimBrief integration, it had to be there before these changes.

> This works when the user is using kg as the weight unit for the airplane and simBrief OFP, but does not work if lbs is used. For example, I have a simBrief OFP with 159 passengers using a passenger weight of 176 lb/pax and 33 lb/pax of baggage. Payload should be 159 times (176+33) = 33,231. That is what simBrief shows, but the MCDU shows 32,600. Cargo weight in the MCDU is 2200 lb (rounded off), but it should be 5,247. With zero cargo, MCDU shows a payload of 29,445 vs what should be 27,984.
> 
I have doubts regarding the first sentence (comprehension from my side). Do you mean to say that the problem is ONLY with SimBrief OFP integration and at the same time, only when Pounds are being used? Am I right to believe that direct input on PAX and Cargo in the AOC works as expected both in Kgs and Lbs?

If this is the case, I'm almost certain this has nothing to do with the FADEC code. We should contact and work it out with the developer(s) that worked on the SimBrief integration. 

Elaborating a little more on this issue, when we input the PAX and Cargo manually in the AOC, a default PAX weight of 84 Kg and Cargo weight of 20 Kg is hardcoded and taken into account. The FADEC code reads those values from the LVars and makes the conversion from Kgs to Lbs. With zero cargo, MCDU shows a payload of 29,445 which is 159 x 185 lbs (the hardcoded PAX weight).

I believe that if you change the unit weight of PAX and Cargo in SimBrief, those LVars should be updated as well. If this is the case, my code should read the right values from them and there shouldn't be any problem.

> If the MCDU thinks the passenger only payload is 29,445, but the passenger + bags payload is 32,600, then it must think the cargo payload is 3,155 even though it shows it as 2,200? If I "trick" the cargo loading by inserting a total passengers of 159, which zeros out the cargo, then select a cargo load, then select zero cargo, it then loads the correct cargo amount. But because the passenger payload is incorrect, the total payload is still incorrect.
> 
> Not connected with the PR, but it seems odd to me that if the user loads the payload manually, inserting a total passenger payload does not load any bags, but if a number is selected for the cargo hold, the weight of the passenger's bags are added to that value. That is sure to confuse people.

sidnov:
> Not connected with the PR, but it seems odd to me that if the user loads the payload manually, inserting a total passenger payload does not load any bags, but if a number is selected for the cargo hold, the weight of the passenger's bags are added to that value. That is sure to confuse people.


This will be fixed with the refactor PR

tracernz:
Why is the payload code actually in the FADEC? That seems like a rather odd design choice that complicates this given that the other half of the code must be in JS.

Taz5150:
> Why is the payload code actually in the FADEC? That seems like a rather odd design choice that complicates this given that the other half of the code must be in JS.

The Payload code is not in the FADEC, only the UI override code (just because the Fuel override code was already part of the FADEC and within the Sim, Fuel and Payload share the same UI Menu). All the override function does in the FADEC is, depending on the Dev Mode LVar, allow for W&B UI tuning or not.

donstim:
Sounds like you guys may have it worked out (at least according to Discord), but I still have some questions/issues:
1. Where does the paxweightlbs come from? If I have MSFS and A32NX EFB units in lbs,, the per passenger weight should be 185 lbs. However, as noted in my previous comment, if I put in 159 passengers, I get 29445 lbs.  In your response, you said that 29445 = 185 (lbs/pax) times 159 pax, but on my calculator 185 times 159 is 29415. Seems like some conversion rounding is occurring. But for weights in lbs, each passenger should be 185 lbs, not 84 kg converted to lbs. If I put in 1 passenger, I get a payload of 185 lbs as expected, but if I put in102 passengers, I get 1852 lbs.

2. I am having trouble checking the cargo loading because of the issue I mentioned in my previous comment that looks to be fixed by the refactoring pr. For example, to the 10 passenger load from above, if I had 1 "unit" (1000 lbs) of cargo, it loads 1.2 thousand pounds. The one thousand I put in plus what looks like 200 kg? of bags at 20 kg/pax? The total payload is shown as 4497, which means the cargo load is 4497 - 1852 = 2645 lb even though the MCDU shows 1.2 "somethings." That 1.2 number is apparently in kg. I don't remember discussing what the bags/pax weight should be, but I think it should be an even 44 lb/pax just as the per pax weight should be an even 185 lb/pax (not an exact kg conversion) when using lbs as the weight unit.

3. I realize these comments are not directed solely at this pr, but rather the whole W/B integration, including simBrief integration. I hope they are helpful to you and Sid for making sure all this stuff works together. I think it is important for the payload UI blocker and MCDU payload constructor to be using consistent units and values, and that both kg and lbs be available as the primary unit depending on the EFB units option the user has selected. What I mean is, if the user selects lbs, the per passenger weight and per passenger baggage weight should be 185/44, but if they select kg, those weights should be 84/20. If they use simBrief integration, the weights should be those from simBrief. (I'm assuming simBrief passes along the weights in the units the user has selected in their simBrief profile for the airplane.)

Taz5150:
> Sounds like you guys may have it worked out (at least according to Discord), but I still have some questions/issues:
> 
> 1. Where does the paxweightlbs come from? If I have MSFS and A32NX EFB units in lbs,, the per passenger weight should be 185 lbs. However, as noted in my previous comment, if I put in 159 passengers, I get 29445 lbs.  In your response, you said that 29445 = 185 (lbs/pax) times 159 pax, but on my calculator 185 times 159 is 29415. Seems like some conversion rounding is occurring. But for weights in lbs, each passenger should be 185 lbs, not 84 kg converted to lbs. If I put in 1 passenger, I get a payload of 185 lbs as expected, but if I put in102 passengers, I get 1852 lbs.

I cannot answer on "where" does paxweightlbs come from, but I can ratify your analysis. It seems no matter what Units you set in the EFB (Kgs or Lbs), the LVars for unit weight per PAX and Bag are always shown in Kgs (84 Kgs + 20 Kgs). 

> 2. I am having trouble checking the cargo loading because of the issue I mentioned in my previous comment that looks to be fixed by the refactoring pr. For example, to the 10 passenger load from above, if I add 1 "unit" (1000 lbs) of cargo, it loads 1.2 thousand pounds. The one thousand I put in plus what looks like 200 kg? of bags at 20 kg/pax? The total payload is shown as 4497, which means the cargo load is 4497 - 1852 = 2645 lb even though the MCDU shows 1.2 "somethings." That 1.2 number is apparently in kg. I don't remember discussing what the bags/pax weight should be, but I think it should be an even 44 lb/pax just as the per pax weight should be an even 185 lb/pax (not an exact kg conversion) when using lbs as the weight unit.

You are also correct on this analysis. First, bags are hardcoded at 20 Kgs/pax and the LVar that contains this unit weight is always in Kgs as stated above. I concur that when you enter Cargo, Bags are also added as part of that Cargo. This is wrong; Bags should be treated as part of PAX weight, and not Cargo weight IMO. But nevertheless, if we were to consider Bags as Cargo, the MCDU (and thus the cargo variables) should be updated with Bag weight as soon as PAX enter the plane. This should be treated as an open issue.

Now, Cargo is always entered in metric Tons, no matter if in Kgs or Lbs. This I agree is also confusing when using Imperial Units and should be discussed and taken care of. Another issue I would open for this matter. 

> 3. I realize these comments are not directed solely at this pr, but rather the whole W/B integration, including simBrief integration. I hope they are helpful to you and Sid for making sure all this stuff works together. I think it is important for the payload UI blocker and MCDU payload constructor to be using consistent units and values, and that both kg and lbs be available as the primary unit depending on the EFB units option the user has selected. What I mean is, if the user selects lbs, the per passenger weight and per passenger baggage weight should be 185/44, but if they select kg, those weights should be 84/20. If they use simBrief integration, the weights should be those from simBrief. (I'm assuming simBrief passes along the weights in the units the user has selected in their simBrief profile for the airplane.)

I totally agree with these conclusions.

This PR fixed the conversion between Lbs and Kgs that was causing issues. But once I use the same conversion figures as the W&B code, the problem on the UI override code is gone. 

My opinion: if Q&A checks that the override conversion issue is solved by this PR, I would merge it. Then, once the new W&B code is ready for a new PR, I would suggest getting in touch with me so I can make pertinent changes to the override code and include it in that PR.

src/fadec/src/EngineControl.h
donstim: We should be using conversations consistent with what MSFS uses internally, regardless of how "correct" they are.

tracernz: It would be good to see a comment in the code noting where the numbers come from and why.

2hwk:
Closed, some elements already in #6969, the rest in a future PR

PR #7084, opened 2022-04-18
feat: add true heading and track

ErickSharp:
## Summary of Changes

This PR adds true heading and true track into the current ADIRS data.

## References

FBW-34-01

## Additional context

## Testing instructions

Make sure that the ND still accurately represents the heading and track on the display.

### DIFF
--- a/src/systems/systems/src/navigation/adirs.rs
+++ b/src/systems/systems/src/navigation/adirs.rs
@@ -1053,7 +1053,8 @@ impl InertialReference {
             SignStatus::NoComputedData
         };
 
-        self.true_heading.set_value(simulator_data.true_heading, ssm);
+        self.true_heading
+            .set_value(simulator_data.true_heading, ssm);
         self.heading.set_value(simulator_data.heading, ssm);
     }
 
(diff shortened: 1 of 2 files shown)


src/systems/systems/src/navigation/adirs.rs
```diff
@@ -1031,6 +1053,8 @@ impl InertialReference {
             SignStatus::NoComputedData
         };
 
+        self.true_heading
+            .set_value(simulator_data.true_heading, ssm);
```
tracernz, line 1057: The heading should become available prior to full allignment, whereas the ssm you're using here does not.

src/systems/systems/src/navigation/adirs.rs
```diff
@@ -1056,6 +1080,16 @@ impl InertialReference {
...
             ssm,
         );
+
+        self.true_track.set_value(
+            if ground_speed_above_minimum_threshold {
+                simulator_data.true_track
+            } else {
+                simulator_data.true_heading
+            },
+            ssm,
```
tracernz, line 1090: This should use the ssm of true heading as well below 50 knots.

tracernz (changes requested):

davidwalschots (approved):
🚀 Let's get this in!

tracernz (approved):
🚀

beheh (approved):
Looks good! It would be nice at some point to properly implement the refresh interval, but applies to all ADIRS values.

aguther:
Tested in sim and still working:
image

Looking at the code changes, there are no users of the new variables. This is to be expected.

PR #7086, opened 2022-04-19
build: Github workflows for nightly builds

frankkopp:
## Summary of Changes
This PR introduces a nightly build GitHub Action workflow and changes the workflow for master merges to not upload to CDN. 

Motivation:
One of flyByWire's biggest monthly  cost contributors is traffic cost from users downloading updates from the CDN.
To reduce these costs uploads of new changes (builds) will only happen once a day. This also will reduce the frequency users will be asked by the Installer to update the aircraft which currently can happen several times a day. 

## ToDo:

- [x] Nightly script
- [x] Master script change
- [ ] Installer version fetch must be changed - probably change to api necessary

## Additional context
This workflow makes a quick check to see if there have been any commits in the last 24h. 
If not it skips the rest of the workflow (build and upload to CDN).
If there have been changes the workflow is identical to the previous master build workflow. 

The master workflow will still build the artifact for download via github but will not do an upload to the CDN for updates in the Installer. 

## Testing instructions
Can't be tested by QA.
We need to have an eye on the workflow to make sure the nightly builds & uploads work well. 

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,13 +1,36 @@
-name: master
+name: nightly
 on:
-  push:
-    branches:
-      - master
+  schedule:
+    # https://crontab.guru/
+    # Every 24h at 12:00 am
+    - cron: '0 0 * * *'
+  workflow_dispatch:
 
 jobs:
+  # This checks if there has been any changes (commits) to master in the last 24 hours.
+  # If not the build and upload to CDN will be skipped.
+  check_date:
+    runs-on: ubuntu-latest
+    name: Check latest commit
+    outputs:
+      result: ${{ steps.should_run.outputs.result }}
+    steps:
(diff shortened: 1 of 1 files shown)


Benjozork:
@frankkopp This idea was originally meant to reduce the bandwidth used on BunnyCDN. Now that we have switched to R2 and our merges are larger, long-running PRs that don't occur multiple times a day, do you think this is still relevant?

frankkopp:
> @frankkopp This idea was originally meant to reduce the bandwidth used on BunnyCDN. Now that we have switched to R2 and our merges are larger, long-running PRs that don't occur multiple times a day, do you think this is still relevant?

I guess not. If there is no real benefit we should stick to the current process. I'll close this. We can rethink once A380X is out. But that might vhange a lot more anyway.

PR #7088, opened 2022-04-19
fix: fire handles can't be stowed

### LINKED ISSUE #4596: Fire handles can't be stowed
**Aircraft Version**

{
  "built": "2021-04-23T22:37:56+00:00",
  "ref": "experimental",
  "sha": "c43d0267084cae66e5cb3950c1c76e9b78bf43d0",
  "actor": "aguther",
  "event_name": "manual"
}

**Describe the bug**
Once a fire handle is pushed and released I can not stow it back by any means.

**To Reproduce**
Open the cover, push and relase the fire handle and try to reset it.

**Expected behavior**
The handle should be stowable by clicking on the released button.

**Actual behavior**
Once out it stays out.

**References**
N/A

**Additional context**
N/A

UNKNOWN

UNKOWN

juliansebline:
Fixes #4596

## Summary of Changes
After being pushed, the fire buttons couldn't be stowed. Now, the LVAR for a specific fire push button takes its opposite value every time there's a click on it. This couln't be done without changing a condition in Rust: at each frame it updates the LVAR with a variable but this variable was always true as soon as the button was pushed.

## Contact
Discord username (if different from GitHub): Julian Sebline#8476

## Testing instructions
1) Open the guard of a given fire push button
2) Click twice on the fire push button

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1025,3 +1025,4 @@
 1. [MISC] Standby Instrument stays ON if emergency power should be available, bug fixes - @2hwk (2Cas#1022 on discord)
 1. [CDU] Full +/- button functionality - @lhoenig (Lukas Hoenig)
 1. [DCDU] Fixed MSG- and MSG+ button labels - @tyler58546 (tyler58546)
+1. [OVHD] Fixed fire push button not being able to be stowed - Julian Sebline (Julian Sebline#8476 on Discord)
--- a/src/behavior/src/A32NX_Interior_Fire.xml
+++ b/src/behavior/src/A32NX_Interior_Fire.xml
@@ -78,7 +78,7 @@
             <UseTemplate Name = "ASOBO_GT_Interaction_LeftSingle_Leave_Code">
                 <LEFT_SINGLE_CODE>
                     (L:A32NX_FIRE_GUARD_#TYPE##ID#) 1 == if{
-                    1 (&gt;#TOGGLE_VAR#) #LEFT_SINGLE_CODE#
+                        (#TOGGLE_VAR#) ! (&gt;#TOGGLE_VAR#)
                     }
                 </LEFT_SINGLE_CODE>
                 <LEFT_LEAVE_CODE />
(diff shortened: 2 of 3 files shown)


aguther (approved):
I've fixed the rust unit test, so that it tests now that handle can be stowed after release.

LGTM, tested it in the sim and it works.

juliansebline:
Thank you.

Would you like me to update the unit test next time ?


De : Andreas ***@***.***>
Envoyé le :jeudi 21 avril 2022 20:30
À : ***@***.***>
Cc : Julian ***@***.***>; ***@***.***>
Objet :Re: [flybywiresim/a32nx] fix: fire handles can't be stowed (PR #7088)


@aguther approved this pull request.

I've fixed the rust unit test, so that it tests now that handle can be stowed after release.

LGTM, tested it in the sim and it works.

aguther:
Yeah, overall the PR should be ready overall. But it's a small thing and I've seen it by running on GitHub. So no problem at all!

aguther:
Thanks for your contribution, keep it coming :-)

davidwalschots:
I think this PR might have consequences beyond the code that was edited. For example, in ` engine_generators.rs` the engine generator determines whether or not it is active by checking `is_released`. I am not sure if pushing the fire handle back into their "non-released" position should make the engine generators function again? If it shouldn't, then this PR has broken that. Maybe @komp1821 can shed some light on that.

There are other usages as well, which might all have also worked under the assumption that `is_released` would remain `true` once `true`.

We might want to include a `was_released` function so that usages can themselves decide what type of information they require.

juliansebline:
Hello David,

I thought about this but I have decided to send this PR anyway because the engine would not stop anyway without my changes.


De : David ***@***.***>
Envoyé le :dimanche 24 avril 2022 23:30
À : ***@***.***>
Cc : Julian ***@***.***>; ***@***.***>
Objet :Re: [flybywiresim/a32nx] fix: fire handles can't be stowed (PR #7088)


I think this PR might have consequences beyond the code that was edited. For example, in engine_generators.rs the engine generator determines whether or not it is active by checking is_released. I am not sure if pushing the fire handle back into their "non-released" position should make the engine generators function again? If it shouldn't, then this PR has broken that. Maybe @komp1821<https://github.com/komp1821> can shed some light on that.

There are other usages as well, which might all have also worked under the assumption that is_released would remain true once true.

We might want to include a was_released function so that usages can themselves decide what type of information they require.

PR #7117, opened 2022-04-28
fix(fmgc): remove TAKEOFF to PREFLIGHT transition

beheh:
## Summary of Changes

This PR removes the transition from PREFLIGHT back to TAKEOFF phase. Previously we would return after you left TOGA while still on the ground, but we no longer believes this exists in reality.

## References

https://discord.com/channels/738864299392630914/747571985831297145/969214848758079598

## Additional context

## Testing instructions

1. Start a takeoff roll with Flex or TO/GA and on the PERF page observe TAKEOFF in green at the top, and most fields becoming non-editable
2. Abort the takeoff. Observe we remain in green TAKEOFF with most fields disabled.
3. Shut down the engines and observe the FMGC resetting (DONE phase)


tracernz (approved):
Verified with multiple sources.

aguther:
One observation from my side: even the power cycle of the whole plane is keeping it stuck in the TAKE OFF phase. I guess when both FMGC are cycled the flight phase should reset.

ShadowFrost3:
LGTM

aguther:
I missed updating the comment here. We had a quick discussion yesterday and it turned out that the FMGC flight phase resets to DONE when on ground for at least 30 seconds and both engines shutdown (N2 < 20%).

This function was already added to this PR but it needs a fix for the nose wheel compression in case of pushback (IRL the nose wheel is still compressed but in the sim, this is currently not the case and will assume the plane not on ground).

aguther:
I missed updating the comment here. We had a quick discussion yesterday and it turned out that the FMGC flight phase resets to DONE when on ground for at least 30 seconds and both engines shutdown (N2 < 20%).

This function was already added to this PR but it needs a fix for the nose wheel compression in case of pushback (IRL the nose wheel is still compressed but, in the sim, this is currently not the case and will assume the plane not on ground).

PR #7118, opened 2022-04-28
fix(hyd): improved actuator fluid return mechanism

crocket63:
Fixes #[issue_no]

## Summary of Changes

Actuators not in active control were still able to use fluid from hydraulic circuit. For example gravity extension was still consuming fluid from green system even after valve cut off closed.

The old workaround is removed so now only an active actuator will be able to use fluid from hydraulic input line. If actuator moves in a different mode, it will only return fluid to reservoir (or use fluid in return line depending on actuator direction).

## Additional context

## Testing instructions

Main impact that simple change can have is reservoir level inconsistencies.

Check all kind of actuator operations, including removing pressure while actuators are in use (for exemple plane pushing back while cargo door in operation until electric power is lost and yellow epump stops).

All reservoir levels should stay coherent with initial state.


BlueberryKing (approved):
Removing old workarounds is always great! Do you think you need to add test case for this?

crocket63:
> Removing old workarounds is always great! Do you think you need to add test case for this?

Workaround had also big drawbacks (taking fluid from the circuit even during gravity extension where valves are closed...) so now we should get a realistic return to reservoir in nominal cases.

Normally current test cases on reservoir levels should be good, and I ran locally the ignored test of resetting the emergency handle, and reservoir coherency.  I'd prefer to decide activating the emergency extension possibility to reset in a dedicated PR, where I'll de-ignore that test :)

mico975:

Discord          : mico#3145
Object of testing: #7118
Tier of Testing  : 1
Date             : 13/05/2022

Testing Process: 
Check all kind of actuator operations, including removing pressure while actuators are in use. - **OK**

Gear emergency extension should see fluid level going back to reservoir while gear goes down so it reaches a level consistent with "on ground gear down" fluid level - **OK**

All reservoir levels should stay coherent with initial state. - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7120, opened 2022-04-28
fix(adirs): various IRS data fixes

lukecologne:
## Summary of Changes

* Output correct units for Rotation rates and Body accelerations: Previously, RPM and percentage were used, instead of the correct units of deg/s and G.
* Use correct sign for Rotation rates and Attitude to conform to Aeronautical standards, in particular, flip the signs of Pitch, Bank, q, p, theta_dot, phi_dot.
* Add the gravitational vector to the body accelerations, since the sim does not include it in the body acceleration simvars.
* Change the G LOAD indication on the SD permanent data area to use IR data (Body Normal Acceleration):
   * It is now shown if the acceleration data is valid and above 1.4G or below 0.7G for more than 2s
   * It disappears 5 seconds after the acceleration returns to normal, or the data goes invalid.
   * If it is shown while the data is invalid (only during the 5s it remains displayed), the data is replaced with 'XX'

Since the Rotation rates and accelerations are not used yet, these changes will have no effect.
The Pitch and Roll data are used, and I have corrected all the usages I could find, although it should be confirmed that I actually found all usages.

## Additional context

## Testing instructions

* Make sure that the systems that use the IRS Pitch and Roll data continue to function normally, for example the PFD.
* Make sure the G LOAD indication works normally.

### DIFF
--- a/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
+++ b/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/EICAS/EICAS_Common.js
@@ -45,19 +45,24 @@ class EICASCommonDisplay extends Airliners.EICASTemplateElement {
         }
 
         const airDataReferenceSource = this.getStatusAirDataReferenceSource();
+        const inertialReferenceSource = this.getStatusInertialReferenceSource();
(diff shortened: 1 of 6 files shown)


davidwalschots (approved):

ShadowFrost3:


Object of testing: #7120
Tier of Testing:1
Date: 30/04/2022

Testing Process: Did a quick small flight. Monitored PFD and SD. 
G gauge works as expected
PFD - no known abnormalities 

I am not extensively familiar with how the PFD should look so it would be good to have a pilot verify or I can do so later. But nothing game breaking. No noticed differences other than airspeed arrow moving erractically but I am told I am either placeboing or its unrelated to this PR. 

Negatives: 
Testing Results: Passed

PR #7122, opened 2022-04-29
fix(fmgc): don't add legs to the empty segment

tracernz:
Fixes #[issue_no]

## Summary of Changes

There was a bug when building departures if the second waypoint was a "duplicate" of the first (also happens if the second leg is effectively null, e.g. a CA leg with the same altitude as the first CA leg), the departure segment would be deleted and the remaining legs would be inserted into the "EmptySegment" which is effectively a black hole for legs. This would only occur for poorly coded procedures, in this case the LPMA DEGU1E departure from runway 05 when using Navigraph/Jeppesen data (MSFS/Navblue and NavDataPro/LIDO databases have a better coding for this departure).

## Additional context

## Testing instructions

Try loading some departures and checking they appear okay.

### DIFF
--- a/src/fmgc/src/flightplanning/ManagedFlightPlan.ts
+++ b/src/fmgc/src/flightplanning/ManagedFlightPlan.ts
@@ -385,9 +385,19 @@ export class ManagedFlightPlan {
             this.reflowSegments();
             this.reflowDistances();
         } else {
-            let segment = segmentType !== undefined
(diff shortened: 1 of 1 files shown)


2hwk (approved):

Lucky38i (approved):
LGTM but I wonder if there's some way to tidy up the if else a bit. was trying for a guard clause but can't really fit it cleanly

PR #7128, opened 2022-04-30
fix(isis): force css animations

tracernz:
Fixes #[issue_no]

## Summary of Changes

Force css animations on in the ISIS.

## Additional context

## Testing instructions

Check that the selected bug in the bugs screen flashes with "menu animations" turned off in MSFS settings.

### DIFF
--- a/src/instruments/src/ISIS/ISISDisplayUnit.tsx
+++ b/src/instruments/src/ISIS/ISISDisplayUnit.tsx
@@ -37,6 +37,11 @@ export const ISISDisplayUnit: React.FC<ISISDisplayUnitProps> = ({ indicatedAirsp
                 setTimer(null);
             }
         }
+
+        // override MSFS menu animations setting for this instrument
+        if (!document.documentElement.classList.contains('animationsEnabled')) {
+            document.documentElement.classList.add('animationsEnabled');
+        }
     });
 
     useEffect(() => {


BlueberryKing (approved):
LGTM! Although I do wonder if we should make a change like this on a higher level. I think the PFD could also benefit from enabling animations.

I am not the biggest fan of forcing the animations to be enabled to be honest, so maybe we can talk about a few options.

tracernz:
> LGTM! Although I do wonder if we should make a change like this on a higher level. I think the PFD could also benefit from enabling animations.
 I did implement this in DisplayUnit which gets all the 6 display units... but not ISIS.

> I am not the biggest fan of forcing the animations to be enabled to be honest, so maybe we can talk about a few options.

Why? We 100% rely on animations working, so we can either enable them, or rely on a random MSFS setting with no clear linkage for the user to break our instruments.

BlueberryKing:
> Why? We 100% rely on animations working, so we can either enable them, or rely on a random MSFS setting with no clear linkage for the user to break our instruments. This has no effect outside our instruments.

Yes, but I assume the setting is there for a reason. Overriding such a setting always has a risk of breaking something with no way for users to avoid it.

tracernz:
I think they just didn’t consider that CSS animations could also be used in instruments. There’s a reason the setting is labelled “**menu** animations”.

PR #7148, opened 2022-05-02
feat: inverted nose wheel steering axis

aguther:
## Summary of Changes
This PR inverts the nosewheel steering axis. The implementation right now is as described in the sim dialog, but as several users have reported it, other planes including the default ones implement it in the opposite direction.

## Testing instructions
- test if the nose wheel steering is still working, in the inverted way
- axis mixture 4 is not changed

### DIFF
--- a/src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
+++ b/src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
@@ -49,7 +49,7 @@ pub(super) fn nose_wheel_steering(builder: &mut MsfsAspectBuilder) -> Result<(),
 
     builder.event_to_variable(
         "AXIS_STEERING_SET",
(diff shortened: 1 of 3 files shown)


src/systems/a320_systems_wasm/src/nose_wheel_steering.rs
```diff
@@ -101,6 +101,8 @@ pub(super) fn nose_wheel_steering(builder: &mut MsfsAspectBuilder) -> Result<(),
             let tiller_handle_position = values[2];
             let tiller_pedal_disconnect = to_bool(values[3]);
 
+            println!("WASM: raw tiller position: {:.2}", values[2]);
+
```
davidwalschots, line 105:
```suggestion
            println!("WASM: raw tiller position: {:.2}", values[2]);

```

davidwalschots (approved):

PR #7151, opened 2022-05-03
feat: disable all MSFS assistance settings

Saschl:
Fixes #[issue_no]

## Summary of Changes

MSFS AI assistances are not compatible with the addon and cause trouble (engine going to idle after takeoff, tiller not working) etc.). This forces all assistances to be off via an entry in the flight_model.cfg.

## Additional context

## Testing instructions

1. Spawn in the plane and make sure it behaves normally.
2. Go to the main menu, turn on all assistance settings and spawn in the plane again. Verify that it is not possible to change assistance settings (in the toolbar + in the settings menu).
3. Taxi, Takeoff and ensure the behavior has not changed even after enabling the settings in the menu before spawning.

### DIFF
--- a/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/flight_model.cfg
+++ b/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/flight_model.cfg
@@ -361,6 +361,7 @@ lift_coef_at_drag_zero_flaps = 0.10000
 [FLIGHT_TUNING]
 
 modern_fm_only = 1; 1 (true) forces use of modern flight model regardless of what user selected in MSFS options menu. 0 (false) allows use of user-selected flight model
+disable_assistances = 1; 1(true) disables all AI assistance settings as they are not compatible with the addon
 ;empty_cg_deviation_limit = 20 ; Maximum deviation of empty weight cg allowed in wegiht & balance UI menu (in feet)
 ;icing_scalar = 1 ; Scales effect of icing on lift and weight
 cruise_lift_scalar = 1


tracernz (approved):

2hwk:
LGTM 

Testing passed

FlightSimulator_Fi9nvWHJbW

image

aguther:
We have no control on this as we are using a built-in function of the sim that works like you see it now.

Tinh1000000:
> Disabling the assistance menu also prevent access to the automatic ATC answering feature which is quite handy for people flying with the default ATC. Is it possible to activate only this item as it should not cause any trouble to the FBW ?

> We have no control on this as we are using a built-in function of the sim that works like you see it now.

Since the authority to allow toggling the AI Radio Communications from the toolbar while toggling flight assistance options are disabled is on Asobo's part, I've raised the issue as a wishlist on the MSFS forums: https://forums.flightsimulator.com/t/allow-access-to-toggle-ai-radio-communications-atc-via-toolbar-when-disable-assistances-flag-in-flight-model-cfg-is-enabled/515313

carlosgaucho:
Have same issue - I like using auto ATC communications which now have disappeared. Seems to only affect FBW. Thanks @Tinh1000000 @aguther @lcanet for raising issue!

carlosgaucho:
Found out there is a workaround - what works for me is changing settings all of the Assistance settings to OFF before loading a flight. Get flight configured, programed, etc, lined up for takeoff - then I will hit ESC and go tot assistance and turn ON AI ATC radio communications - AI  will then takeover and keep in contact with ATC en rout

dannywooly:
This is massive disappointment, ATC assistance and Taxi assistance is something I use all the time, crazy its just been removed on a whim!

Saschl:
They are still available. The only thing that got disabled is the toolbar (only MSFS knows why it does disable the whole thing). Via the "normal" settings you can access AI ATC and Taxi Ribbons (as seen in the screenshot a few posts above and here: https://docs.flybywiresim.com/fbw-a32nx/settings/?h=assistance#deactivate-msfs-assistance-features). We know the toolbar is an annoyance and have created this: https://devsupport.flightsimulator.com/questions/7607/possibility-to-turn-off-assistance-functions.html and are constantly looking into options to have this more fine grained.

PR #7173, opened 2022-05-08
feat(lgciu/sd): add LGCIU bus output, modify WHEEL page for LGCIUs

lukecologne:
## Summary of Changes

This PR:
* Adds (part of) the LGCIU output bus, for a list of bits in the discrete words and their description, see the Lvar docs.
   * Discrete words 2 and 4 and some bits in discrete word 3 have not been added, since the functionality for them is not currently implemented in the LGCIU.
* Modifies the WHEEL SD page to use the LGCIU output bus, and improve the indications.
   * LGCIU failures should now be properly handled, and the "door partially opened" case has been added to the door indications. 
  
   

## Additional context

## Testing instructions

Check that the WHEEL page works correctly in all situations, including failures

Testing suggestion from Davy:
```
Use gravity gear extension on ground and in flight. If clicked and hold it should:

release doors on ground
release doors and gears in flight.
After use even with green system available, gear should be inoperative.

If clicked and held again, and if green system available it should:

close doors on ground
close doors and gears in flight. (Might need to recycle gear lever if LGCIU error)
```

### DIFF
--- a/src/instruments/src/SD/Pages/Wheel/Wheel.tsx
+++ b/src/instruments/src/SD/Pages/Wheel/Wheel.tsx
@@ -1,8 +1,10 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { render } from '@instruments/common/index';
 import classNames from 'classnames';
 import { SimVarProvider, useSimVar } from '@instruments/common/simVars';
 import { setIsEcamPage } from '@instruments/common/defaults';
+import { useArinc429Var } from '@instruments/common/arinc429';
+import { Arinc429Word } from '@shared/arinc429';
 import { PageTitle } from '../../Common/PageTitle';
 import './Wheel.scss';
 import { HydraulicsProvider, useHydraulics } from '../../Common/HydraulicsProvider';
@@ -25,6 +27,11 @@ export const WheelPage = () => {
 
     const maxTemperatureIndex = temperatures.reduce((maxIndex, element, index, array) => (element > array[maxIndex] ? index : maxIndex), 0);
 
+    const lgciu1DiscreteWord1 = useArinc429Var('L:A32NX_LGCIU_1_DISCRETE_WORD_1');
+    const lgciu2DiscreteWord1 = useArinc429Var('L:A32NX_LGCIU_2_DISCRETE_WORD_1');
+    const lgciu1DiscreteWord3 = useArinc429Var('L:A32NX_LGCIU_1_DISCRETE_WORD_3');
+    const lgciu2DiscreteWord3 = useArinc429Var('L:A32NX_LGCIU_2_DISCRETE_WORD_3');
+
     return (
         <EcamPage name="main-wheel">
             <PageTitle x={6} y={115} text="WHEEL" />
@@ -33,7 +40,14 @@ export const WheelPage = () => {
                 <Spoilers x={103} y={64} />
 
                 <NoseWheelSteering x={205} y={200} />
-                <LandingGearCtl x={255} y={263} />
+                <LandingGearCtl
+                    x={255}
+                    y={263}
+                    lgciu1DiscreteWord1={lgciu1DiscreteWord1}
+                    lgciu2DiscreteWord1={lgciu2DiscreteWord1}
+                    lgciu1DiscreteWord3={lgciu1DiscreteWord3}
+                    lgciu2DiscreteWord3={lgciu2DiscreteWord3}
+                />
                 <AntiSkid x={300} y={312} />
 
                 <NormalBraking x={220} y={333} />
@@ -42,7 +56,15 @@ export const WheelPage = () => {
 
             <AutoBrake x={300} y={460} />
 
-            <Gear x={18} y={210} location="left" />
+            <Gear
+                x={18}
+                y={210}
+                location="left"
+                lgciu1DiscreteWord1={lgciu1DiscreteWord1}
+                lgciu2DiscreteWord1={lgciu2DiscreteWord1}
+                lgciu1DiscreteWord3={lgciu1DiscreteWord3}
+                lgciu2DiscreteWord3={lgciu2DiscreteWord3}
+            />
             <Wheels
                 x={12}
                 y={318}
@@ -50,11 +72,27 @@ export const WheelPage = () => {
                 right={{ number: 2, temperature: temperatures[1], hottest: maxTemperatureIndex === 1 }}
             />
 
-            <Gear x={210} y={107} location="center" />
+            <Gear
+                x={210}
+                y={107}
+                location="center"
+                lgciu1DiscreteWord1={lgciu1DiscreteWord1}
+                lgciu2DiscreteWord1={lgciu2DiscreteWord1}
+                lgciu1DiscreteWord3={lgciu1DiscreteWord3}
+                lgciu2DiscreteWord3={lgciu2DiscreteWord3}
+            />
             <WheelArch x={298} y={370} type="bottom" />
             <WheelArch x={406} y={370} type="bottom" />
 
-            <Gear x={401} y={210} location="right" />
+            <Gear
+                x={401}
+                y={210}
+                location="right"
+                lgciu1DiscreteWord1={lgciu1DiscreteWord1}
+                lgciu2DiscreteWord1={lgciu2DiscreteWord1}
+                lgciu1DiscreteWord3={lgciu1DiscreteWord3}
+                lgciu2DiscreteWord3={lgciu2DiscreteWord3}
+            />
             <Wheels
                 x={436}
                 y={318}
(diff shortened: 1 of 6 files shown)


src/instruments/src/SD/Pages/Wheel/Wheel.tsx
```diff
@@ -178,114 +232,241 @@ type GearLocation = 'left' | 'center' | 'right';
...
+    let lgciu2GearNotUplocked: boolean;
+    let lgciu1GearDownlocked: boolean;
+    let lgciu2GearDownlocked: boolean;
+    let upLockFlagShown = lgciu1DataValid && lgciu2DataValid;
+    if (location === 'left') {
+        lgciu1GearNotUplocked = lgciu1DiscreteWord3.getBitValue(11);
+        lgciu2GearNotUplocked = lgciu2DiscreteWord3.getBitValue(11);
+        lgciu1GearDownlocked = lgciu1DiscreteWord1.getBitValue(23);
+        lgciu2GearDownlocked = lgciu2DiscreteWord1.getBitValue(23);
+        upLockFlagShown = upLockFlagShown && lgciu1DiscreteWord1.getBitValue(20) && lgciu1DiscreteWord1.getBitValue(20);
```
tracernz, line 366: Should be lgciu2 for the second one?

src/instruments/src/SD/Pages/Wheel/Wheel.tsx
```diff
@@ -178,114 +232,241 @@ type GearLocation = 'left' | 'center' | 'right';
...
+    } else if (location === 'right') {
+        lgciu1GearNotUplocked = lgciu1DiscreteWord3.getBitValue(12);
+        lgciu2GearNotUplocked = lgciu2DiscreteWord3.getBitValue(12);
+        lgciu1GearDownlocked = lgciu1DiscreteWord1.getBitValue(24);
+        lgciu2GearDownlocked = lgciu2DiscreteWord1.getBitValue(24);
+        upLockFlagShown = upLockFlagShown && lgciu1DiscreteWord1.getBitValue(21) && lgciu1DiscreteWord1.getBitValue(21);
```
tracernz, line 372: Should be lgciu2 for the second one?

src/instruments/src/SD/Pages/Wheel/Wheel.tsx
```diff
@@ -178,114 +232,241 @@ type GearLocation = 'left' | 'center' | 'right';
...
+    } else {
+        lgciu1GearNotUplocked = lgciu1DiscreteWord3.getBitValue(13);
+        lgciu2GearNotUplocked = lgciu2DiscreteWord3.getBitValue(13);
+        lgciu1GearDownlocked = lgciu1DiscreteWord1.getBitValue(25);
+        lgciu2GearDownlocked = lgciu2DiscreteWord1.getBitValue(25);
+        upLockFlagShown = upLockFlagShown && lgciu1DiscreteWord1.getBitValue(22) && lgciu1DiscreteWord1.getBitValue(22);
```
tracernz, line 378: Should be lgciu2 for the second one?

src/instruments/src/SD/Pages/Wheel/Wheel.tsx
lukecologne: Oh yeah, nice copy paste error. Will fix shortly

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -744,6 +756,131 @@ impl LandingGearControlInterfaceUnit {
...
+
+    fn discrete_word_1(&self) -> Arinc429Word<f32> {
+        if !self.is_powered {
+            Arinc429Word::new(0., SignStatus::FailureWarning)
+        } else {
+            let mut word = Arinc429Word::new(0., SignStatus::NormalOperation);
+            set_arinc429_bit(
+                &mut word,
+                11,
+                !self.sensor_inputs.left_gear_up_and_locked && !self.gear_handle_is_down(),
```
crocket63, line 768: I'm wondering if we should use sensor_inputs.left_gear_up_and_locked directly or use access methods that are also handling the value if failed/unpowered. Right now probably won't change anything as you just send invalid frame when unpowered, but something to remember to check when adding more things I believe.

crocket63 (changes requested):
Lgtm, however couldn't we add one or two tests to check per bit write/read in arinc429.rs ?

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -744,6 +756,131 @@ impl LandingGearControlInterfaceUnit {
     fn status(&self) -> LgciuStatus {
         self.status
     }
+
+    fn discrete_word_1(&self) -> Arinc429Word<f32> {
+        if !self.is_powered {
+            Arinc429Word::new(0., SignStatus::FailureWarning)
```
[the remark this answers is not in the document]
lukecologne: Yes, IRL a power failure would stop the transmission of the source computer, and the receiving computers would then register that and process that information accordingly. Some set the received label internally to Failure Warning, others do it differently. For now, we just set the SSM to FW at the source, that should be an alright approximation.

lukecologne:
I've added a bit-get method, and also added a test for both (similar to the existing one, testing the get and set at the same time. Hopefully it's adequate

crocket63 (approved):
Go go go

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -797,6 +898,10 @@ impl SimulationElement for LandingGearControlInterfaceUnit {
         );
 
         writer.write(&self.fault_ecam_id, self.status() != LgciuStatus::Ok);
+
+        writer.write(&self.discrete_word_1_id, f64::from(self.discrete_word_1()));
+        writer.write(&self.discrete_word_2_id, f64::from(self.discrete_word_2()));
+        writer.write(&self.discrete_word_3_id, f64::from(self.discrete_word_3()));
```
davidwalschots, line 904: **suggestion**: Add a `Writer` implementation for `Arinc429Word<f32>` which then calls the appropriate `From` implementation:

```rust
impl<T: Writer> Write<Arinc429Word<f32>> for T {
    fn convert(&mut self, value: Arinc429Word<f32>) -> f64 {
        value.into()
    }
}
```

```suggestion
        writer.write(&self.discrete_word_1_id, self.discrete_word_1());
        writer.write(&self.discrete_word_2_id, self.discrete_word_2());
        writer.write(&self.discrete_word_3_id, self.discrete_word_3());
```

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -744,6 +756,95 @@ impl LandingGearControlInterfaceUnit {
...
+            word.set_bit(20, false);
+            word.set_bit(21, false);
+            word.set_bit(22, false);
+            word.set_bit(23, self.sensor_inputs.downlock_state(GearWheel::LEFT));
+            word.set_bit(24, self.sensor_inputs.downlock_state(GearWheel::RIGHT));
+            word.set_bit(25, self.sensor_inputs.downlock_state(GearWheel::NOSE));
+            word.set_bit(26, self.sensor_inputs.left_gear_compressed(true));
+            word.set_bit(27, self.sensor_inputs.right_gear_compressed(true));
+            word.set_bit(28, self.sensor_inputs.nose_gear_compressed(true));
+            word.set_bit(29, self.gear_handle_is_down());
```
davidwalschots, line 802: **suggestion**: Consider creating a declarative macro to reduce the amount of ` word.set_bit` repetition. Not sure if there's some funky thing we can do here, might be worth looking into.

src/systems/systems/src/shared/arinc429.rs
```diff
@@ -40,6 +40,35 @@ impl<T: Copy> Arinc429Word<T> {
...
+impl Arinc429Word<f32> {
+    pub fn set_bit(&mut self, bit: u32, value: bool) {
+        self.value =
+            (((self.value as u32) & !(1 << (bit - 1))) | ((value as u32) << (bit - 1))) as f32;
+    }
+
+    pub fn get_bit(&self, bit: u32) -> bool {
+        ((self.value as u32 >> (bit - 1)) & 1) != 0
+    }
+}
```
davidwalschots, line 52: **thought**: I think we can store the value as `u32` instead of `f32`? I see a lot of conversion between these two numeric types, but am not really seeing a reason for doing so.

src/systems/systems/src/shared/arinc429.rs
```diff
@@ -40,6 +40,35 @@ impl<T: Copy> Arinc429Word<T> {
         self.ssm == SignStatus::NormalOperation
     }
 }
+impl Arinc429Word<f32> {
+    pub fn set_bit(&mut self, bit: u32, value: bool) {
```
davidwalschots, line 44: **suggestion**: I think we will not exceed 256 bits within set_bit. Therefore this can be a `u8`.
```suggestion
    pub fn set_bit(&mut self, bit: u8, value: bool) {
```

**suggestion**: Add `debug_assert!` to limit the valid usage to between 1 and 32 (or 0 and 31, depending on the standard).

src/systems/systems/src/shared/arinc429.rs
```diff
@@ -40,6 +40,35 @@ impl<T: Copy> Arinc429Word<T> {
...
+        self.value =
+            (((self.value as u32) & !(1 << (bit - 1))) | ((value as u32) << (bit - 1))) as f32;
+    }
+
+    pub fn get_bit(&self, bit: u32) -> bool {
+        ((self.value as u32 >> (bit - 1)) & 1) != 0
+    }
+}
+impl From<f64> for Arinc429Word<f32> {
```
davidwalschots, line 53: **thought**: This implementation was probably created in anticipation of future reading of the value from a simvar.  **suggestion**: Add a `Reader` implementation as well (next to the `Writer`).

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -744,6 +756,95 @@ impl LandingGearControlInterfaceUnit {
...
+
+            word
+        }
+    }
+
+    fn discrete_word_2(&self) -> Arinc429Word<f32> {
+        if !self.is_powered {
+            Arinc429Word::new(0., SignStatus::FailureWarning)
+        } else {
+            let mut word = Arinc429Word::new(0., SignStatus::NormalOperation);
```
davidwalschots, line 812: **suggestion**: Consider creating another constructor function which doesn't take a value, as it seems these usages always set the value to `0.`.

```rust
impl Arinc429Word<u32> {
    pub fn bits(ssm: SignStatus) -> Self {
        Self { value: 0., ssm }
    }
}
```

src/systems/systems/src/shared/arinc429.rs
lukecologne: Implemented the first suggestion, but why only `debug_assert`?

src/systems/systems/src/landing_gear/mod.rs
lukecologne: I think it's better to keep it more explicit, so you can directly see it initializes to 0 and don't have to look up what the constructor does. Also, IMO, it wouldn't really save much to write anyways

src/systems/systems/src/landing_gear/mod.rs
lukecologne: Not sure how to do that, any suggestions?

src/systems/systems/src/shared/arinc429.rs
lukecologne: Alright, added the `debug_assert` for the bit range that is possible with the ARINC 429 data word structure (bits 11 to 29).

2hwk:
Testing instructions could be a bit more detailed, it's hard to define what "behaving correctly" means

crocket63:
> Testing instructions could be a bit more detailed, it's hard to define what "behaving correctly" means, a list of test cases and expected outputs would be helpful

Can use emergency extension for a nice test case:

Use gravity gear extension on ground and in flight. If clicked and hold it should:

release doors on ground
release doors and gears in flight.
After use even with green system available, gear should be inoperative.

If clicked and held again, and if green system available it should:

close doors on ground
close doors and gears in flight. (Might need to recycle gear lever if LGCIU error)

src/systems/systems/src/landing_gear/mod.rs
```diff
@@ -744,6 +756,95 @@ impl LandingGearControlInterfaceUnit {
     fn status(&self) -> LgciuStatus {
         self.status
     }
+
+    fn discrete_word_1(&self) -> Arinc429Word<u32> {
```
beheh, line 760: Would you mind making these accessible to other systems (like the FWC in future?). You could either make the method public, or add it to a public trait and implement that (e.g. like `systems::shared::LgciuWeightOnWheels`);

lukecologne: I'll just make it public, I don't see a use of making it a trait, since it's not going to be different on different aircraft, or am I missing something?

PR #7176, opened 2022-05-09
feat: do no longer force disable assistance

aguther:
## Summary of Changes
This PR introduces the following changes:
- do no longer force disable assistance functions
- show the warning due to enabled assistance functions only once
- check for assistance not based on CAMERA STATE but based on A32NX_IS_READY variable
- add the A32NX_IS_READY variable to the FLT files
- add one additional datum related to AI controls to the FDR
- add nosewheel and tiller information to FDR

## Additional context
Based on user feedback and the fact that we now see mods to copy over the whole flight model to enable the assistance function again, we should no longer force disable it. After the next flight model changes the probability is quite high that users will overwrite it with an outdated version coming from a mod to enable assistance functions.

## Testing instructions
- assistance functions are not force disabled any more
- test if pop-up is shown once an assistance function is enabled (assisted landing, assisted take-off, AI auto-trim)


beheh (approved):

2hwk:
Notification will require a fix to deal with the issue of escaping to the menu before it is properly dismissed, I should be able to work on it over the weekend (or earlier)

PR #7180, opened 2022-05-10
perf: improved simulation variable hooks

Benjozork:
### Old hooks

The old SimVar hooks were based on a React context object provided by a `SimVarPovider`. This would centralize get/set calls into the context component, and allow employing a global cache. The client hooks would then call into that context, employing said global cache to access values.

On `set` operations, the global cache state would be mutated in identity, causing a re-render in every component using the hook. This was a tradeoff, effectively enabling dedupe-ing `SimVar.GetSimVarValuie` calls in the context component.

### New hooks

The new hooks simply employ two core mechanisms: a `useState` containing the current value, a `useRef` containing the latest timestamp of a get operation, and a `useUpdate` callback to compare the timer and get the value if necessary.

After performance testing, it is noticed that the deduep-ing tradeoff was not worth the resullting re-render penalty. CoherentGTUIThread times on idle on runway have went from 8-10ms to about 5-7, for instance.

## Summary of Changes

* This PR introduces experimental new simvar hooks, which are simpler and do not employ global caching. The main performance improvement is avoiding frequent re-renders, at the expense of no dedupeing simvar get calls.
* This changes the `maxStaleness` argument to `refreshInterval` - code will have to be checked for correcteness

* [x] Introduce new hooks
* [x] Port `useInteractionSimVar`
* [x] Adjust documentation
* [ ] Verify correctness of time-sensitive uses
* [x] Remove old hooks and provider

### DIFF
--- a/src/instruments/src/Common/simVars.tsx
+++ b/src/instruments/src/Common/simVars.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useContext, useEffect, useRef, useState } from 'react';
+import { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import { useInteractionEvents, useUpdate } from './hooks';
 
 /**
@@ -292,8 +292,29 @@ export const useSimVar = (
     maxStaleness = 0,
 ): [SimVarValue, (newValueOrSetter: SimVarValue | SimVarSetter
 ) => void] => {
-    const value = useSimVarValue(name, unit, maxStaleness);
-    const setter = useSimVarSetter(name, unit);
+    const lastUpdate = useRef(Date.now() - maxStaleness - 1);
+
+    const [value, setValue] = useState(() => SimVar.GetSimVarValue(name, unit));
+
+    const updateCallback = useCallback(() => {
+        const newValue = SimVar.GetSimVarValue(name, unit);
+
+        const delta = Date.now() - lastUpdate.current;
+
+        if (delta > maxStaleness) {
+            lastUpdate.current = Date.now();
+            setValue(newValue);
+        }
+    }, [name, unit, maxStaleness]);
+
+    useUpdate(updateCallback);
+
+    const setter = useCallback((value: any | SimVarSetter) => {
+        SimVar.SetSimVarValue(name, unit, typeof value === 'function' ? value(SimVar.GetSimVarValue(name, unit)) : value);
+
+        return value;
+    }, [name, unit]);
+
     return [value, setter];
 };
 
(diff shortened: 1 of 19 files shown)


src/instruments/src/Common/simVars.tsx
```diff
@@ -289,11 +286,32 @@ const SimVarProvider: React.FC = ({ children }) => {
...
 ): [SimVarValue, (newValueOrSetter: SimVarValue | SimVarSetter
 ) => void] => {
-    const value = useSimVarValue(name, unit, maxStaleness);
-    const setter = useSimVarSetter(name, unit);
+    const lastUpdate = useRef(Date.now() - refreshRate - 1);
+
+    const [value, setValue] = useState(() => SimVar.GetSimVarValue(name, unit));
+
+    const updateCallback = useCallback(() => {
+        const newValue = SimVar.GetSimVarValue(name, unit);
```
tracernz, line 297: Can’t this go inside the refreshRate check?

Benjozork: fixed

tracernz (approved):

svengcz:
The hotstart of ND and upper ECAM.
grafik

How to reproduce: Spawn on the runway with running engines.

src/instruments/src/DCDU/index.tsx
```diff
@@ -56,29 +59,44 @@ const DCDU: React.FC = () => {
...
         }
     };
-    const setStatus = (sender: string, message: string) => {
+    const setStatus = (sender: string, message: string, duration: number) => {
         const state = statusMessage;
         state.sender = sender;
         state.message = message;
-        state.remainingMilliseconds = 5000;
         setStatusMessage(state);
+        if (duration > 0) {
```
beheh, line 74: @svengcz How about instead of using the magic -1 you use `duration: Infinity` and check `if (duration.isFinite())` here? (because magic numbers are evil :-)

src/instruments/src/Common/simVars.tsx
```diff
@@ -431,19 +198,43 @@ export const useInteractionSimVar = (
...
-    const contextValue = useContext(context);
-    const value = useSimVarValue(name, unit, maxStaleness);
+    const lastUpdate = useRef(Date.now() - refreshInterval - 1);
+
+    const [stateValue, setStateValue] = useState(() => SimVar.GetSimVarValue(name, unit));
+
+    const updateCallback = useCallback(() => {
+        const delta = Date.now() - lastUpdate.current;
+
+        if (delta > refreshInterval) {
```
beheh, line 211: >=, no?

src/instruments/src/DCDU/index.tsx
svengcz: fixed

beheh (approved):

PR #7181, opened 2022-05-11
refactor(engine): remove static values for inner tank overflow check

Revyn112:
Fixes #[issue_no]

## Summary of Changes
This change removes the static values for the Inner Fuel Tank check and use the MSFS SimVar values instead.

## Additional context
The code is more dynamically when changes happend and so it is better for the general maintaining process. On the other hand, as the FBW A32NX gets integrated in other addons this change is very helpful for the developer to maintain they code.

## Testing instructions

1. Load standard appron configuration of A32NX (Cold & Dark).
2. Turn on the FlyPad (EFB).
3. Switch to the Fuel Page.
4. Refill all tanks completly.
4. Check if fuel gets transfered to center tank when inner tank limit is reached.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -248,6 +248,7 @@
 1. [MISC] Added custom Autobrake event for SimConnect - @aguther (Andreas Guther)
 1. [EWD] Fixed failures not re-triggering when resolved - @tricky_dicky (Richard Pilbery)
 1. [SOUND] Prevent autopilot disconnect from sounding on C+D spawn - @tricky_dicky (Richard Pilbery) and @saschl (saschl#9432)
+1. [ENGINE/FADEC] Remove static values for inner tank overflow check - @Revyn112 (Denis Freund)
 
 ## 0.7.0
 
--- a/src/fadec/src/EngineControl.h
+++ b/src/fadec/src/EngineControl.h
@@ -871,17 +871,21 @@ class EngineControl {
       fuelLeft = (fuelLeftPre - (fuelBurn1 * KGS_TO_LBS)) + xfrAuxLeft + (xfrCenter / 2);     // LBS
       fuelRight = (fuelRightPre - (fuelBurn2 * KGS_TO_LBS)) + xfrAuxRight + (xfrCenter / 2);  // LBS
 
+
+      double leftQuantity = simVars->getTankLeftQuantity() * fuelWeightGallon; // LBS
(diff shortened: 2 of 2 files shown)


Taz5150:
Hi @Revyn112,

You are prefixing this PR as a refactor, and I understand from your comment that was the intent. But you've changed the whole purpose of that function altogether; let me explain in more detail.

Previously, the function detected when there was an overflow in the left or right inner tanks, and compensated that overflow in the center tank. Right now, the code will not detect an overflow since you have dynamically changed the capacity of the tank itself (12,167.2 lbs) to the previous fuel quantity within the tank (by calling the SimVar). Although a simple test could work out just fine, you have to also test all other conditions like Sim being paused, MCDU refueling, etc.

It might be that with the latest changes in the code for the EFB and MCDU, this part of the code is not needed anymore. But since we are working on a full fledge fuel system (modern fuel system within MSFS), and the update on the fuel will also change end-to-end, I suggest not messing with this part of the code for the time being.

As I said, being the final objective to refactor the code, and due to the risks we take by changing those lines, I suggest not to approve this PR for the time being. But be sure that at some point those static values will disappear ... 

Thanks!

Revyn112:
@Taz5150 thank for your feedback. As i see I just used the wrong SimVar. My intention for this was the use of `FUEL TANK LEFT MAIN CAPACITY`, as this wouldn't be changed dynamic and is getting the value of the defined tank from the game itself

Anyways when there are changes incoming then i will have patience on this topic and will close this pull request for now.

Greetings

Revyn112:
@Taz5150 thank for your feedback. As i see I just used the wrong SimVar. My intention for this was the use of `FUEL TANK LEFT MAIN CAPACITY`, as this wouldn't be changed dynamic and is getting the value of the defined tank from the game itself

Anyways when there are changes incoming then i will have patience on this topic and will close this pull request for now.


So the origin of this pull request was that we recently changed this in the a3309x addon for better maintaining the behavior in the future: https://github.com/headwind-msfs/a330-900/blob/main/src-a339/fadec/src/EngineControl.h#L874

Greetings

Taz5150:
@Revyn112 Yes, I edited my original comment to let you know I understood what you were trying to do, but it was already too late.

I understand the need, since tank capacities between A32NX and A339x are different. But the problem we found with that SimVar was the unused fuel.

I will get back to you once we have the new fuel system implemented which will take over the whole fuel update function within the code.

Thank you for your comment!! Anything you might need from an engine/ fuel system perspective, please do let me know!

Cheers!

PR #7198, opened 2022-05-14
fix: Improve pitch trim & manual braking level

donstim:
## Summary of Changes

Changes:
1) Incorporates the (relatively) new elevator trim up and down limits, which are different for the A320. Previously, there was only one limit in MSFS for both up and down limits.
2) Reduces the pitch trim sensitivity at low speeds to reduce the possibility of premature liftoff or early nosewheel liftoff. Raises the pitch trim sensitivity at higher speeds to better match available pitch trim data at cruise conditions.
3) Reduces manual braking effectiveness to regain a match with stopping distance data. For some reason, the stopping distances have gotten shorter since this matching was first done.
## Additional context

Additional information is contained here: 
[Pitch Trim Adjusments.pdf](https://github.com/flybywiresim/a32nx/files/8692339/Pitch.Trim.Adjusments.pdf)
and here:
1) The pitch trim effectiveness changes only affects the pitch trim (THS) deflection angle for a given aerodynamic effect. It has no effect on the aerodynamics of the airplane in flight when autotrim is active in that autotrim will simply command the appropriate pitch trim deflection (unless it causes the airplane to need more pitch trim deflection than the deflection angle limit). It does have an effect on the ground during takeoff where the pitch trim is set manually. A given amount of pitch trim will have less aerodynamic effect with this mod.

2)The pitch trim change is done though the elevator trim elasticity table. It could also have been done through the elevator trim effectiveness scalar, but changing this scalar would change the pitch trim level in all conditions. Using the elasticity scalar allows reducing the effectiveness at low speeds, but increasing it at high speeds to match the available data. (Note: The elasticity table is actually a function of dynamic pressure, which is correlated with speed, rather than being a function of speed itself.)

3) Because approach speeds overlap with takeoff speeds, the pitch trim deflection angles for approach will be increased. This doesn't affect the approach flight characteristics. It just means that if you look at the pitch trim deflection on approach, it is probably higher then it would be on the real airplane (although we do not have real airplane data for that yet). 

## Testing instructions
Perform a number of takeoffs with pitch trim set (a) at the proper setting for the takeoff CG position and (b) at the limits of the green band per the FBW checklist pitch trim vs takeoff CG chart (https://github.com/flybywiresim/manuals/blob/master/pdf/A32NX%20Documentation/FBW%20A32NX%20Checklist.pdf). 

Takeoffs at the proper THS position should not encounter premature liftoff or early rotation. For takeoffs at the nose up green band limit, some early nosewheel liftoff tendency is acceptable if it can be easily maintained on the runway with forward stick pressure. For takeoffs anywhere in the green band, use of the same stick deflection for rotation should result in about the same flyaway characteristics. For example, when THS is set to the nose down green band limit (2.5 degrees nose down), it should not be unduly difficult to rotate the airplane -- it should respond to the stick input pretty much the same any other THS setting within the green band.

Accomplish a circuit to determine if there are any changes in flying characteristics caused by this mod.


2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #7198
Tier of Testing  : 1
Date             : 15/05/2022

Testing Process: 
Performed takeoffs with pitch trim set in various positions including full up/down. No premature rotation in green band up position and rotation feels similar regardless of set trim.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7229, opened 2022-05-22
feat: improving flypadOS 3 Ground Service page

frankkopp:
## Summary of Changes

Improvement of flyPadOS 3 Ground Service page:

- [x] Ensure service are not available when pushback tug is attached, aircraft is moving or in the air
- [x] Change button handling to a more readable and maintainable code base (no more arrays)
- [x] Add Cargo Door
- [x] Add/improve state handling for buttons, services and doors
- [x] Add TUG message
- [x] Use redux to keep state of service buttons which do not have a simvar (change existing redux for buttons)
- [x] Reduce code redundancy while still maintain readability/maintainability 
- [x] Testing

## Screenshots (if necessary)
image
image

## Testing instructions

Testing should focus on:
- extensively use ground services
- making sure buttons states are staying in sync with sim's services 
  - this is challenging as the services cannot be queried from the aircraft and must be maintained manually. As the sim influences them independently a lot can go wrong. There are edge cases where it is still possible to get out of sync - but in "normal" situations this should not happen
- Test in "normal" gate situations (services available) - also test with no jetway but stairs
- Test at places where no services or not all services are available (e.g. KFBW or so)
- Test edge cases like spamming buttons, hectic clicking, changing EFB pages during service activations, etc.
- Test for incompatibilities with in-sim ATC or other pushback tools if you have them available. 

It will not be 100% perfect in all situations but already much better as before. I will continue to work on a solution which requires changing the vehicles to signal the aircraft if they are connected or not. This will be a future PR. 


src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
```diff
@@ -38,159 +43,524 @@ const ServiceButtonWrapper: FC<ServiceButtonWrapperProps> = ({ children, classNa
...
+            if (cabinDoorOpen === 1) {
+                SimVar.SetSimVarValue('K:TOGGLE_AIRCRAFT_EXIT', 'enum', 1);
+            }
+            if (aftDoorOpen === 1) {
+                SimVar.SetSimVarValue('K:TOGGLE_AIRCRAFT_EXIT', 'enum', 4);
+            }
+            if (cargoDoorOpen === 1) {
+                SimVar.SetSimVarValue('K:TOGGLE_AIRCRAFT_EXIT', 'enum', 6);
+            }
+        } else if (cateringButtonState === ServiceButtonState.DISABLED
```
ErickSharp, line 453: To reduce a lot of this repetition and if this is something you prefer, you can do something like this: 

```typescript
[cateringButtonState, jetWayButtonState, fuelTruckButtonState, gpuButtonState, cargoDoorButtonState, baggageButtonState, aftDoorButtonState, cateringButtonState].every((buttonState) => buttonState === ServiceButtonState.DISABLED);
```

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
```diff
@@ -38,159 +43,524 @@ const ServiceButtonWrapper: FC<ServiceButtonWrapperProps> = ({ children, classNa
...
+    const cargoDoorButtonState = useAppSelector((state) => state.groundServicePage.cargoDoorButtonState);
+    const baggageButtonState = useAppSelector((state) => state.groundServicePage.baggageButtonState);
+    const aftDoorButtonState = useAppSelector((state) => state.groundServicePage.aftDoorButtonState);
+    const cateringButtonState = useAppSelector((state) => state.groundServicePage.cateringButtonState);
 
-    useEffect(() => {
-        setWheelChocksVisible(wheelChocksEnabled && isGroundEquipmentVisible);
-        setConesVisible(conesEnabled && isGroundEquipmentVisible);
-    }, [isGroundEquipmentVisible, wheelChocksEnabled, conesEnabled]);
+    // Required so these can be used inside the useTimeout callback
```
ErickSharp, line 150: Can you explain why you have had to wrap these states in refs? What issues in particular have you encountered in trying to use the states directly?

ErickSharp (changes requested):
Your solution manages complexity very well. I am so proud of you! Great work!!!!!!!

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
```diff
@@ -38,159 +43,524 @@ const ServiceButtonWrapper: FC<ServiceButtonWrapperProps> = ({ children, classNa
...
-                    dispatch(removeActiveButton(index));
-                }
-                if (disabledButton) {
-                    const disabledIndex = disabledButtons.indexOf(disabledButton);
-                    dispatch(removeDisabledButton(disabledIndex));
+        // Door Button: enable door button after a timeout if it was disabled
+        if (doorButtonState === ServiceButtonState.DISABLED) {
+            setTimeout(() => {
+                // service button could have been pressed again in the meantime
+                if (serviceButtonStateRef.current < ServiceButtonState.CALLED) {
```
ErickSharp, line 206: I would avoid doing these types of comparisons with an enum as it isn't very intuitive, is error-prone if the order of enum members is ever changed, and not very explicit. I would change this to more explicitly define the condition for the behavior specified.

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
```diff
@@ -38,159 +43,524 @@ const ServiceButtonWrapper: FC<ServiceButtonWrapperProps> = ({ children, classNa
...
+    'opacity-20 pointer-events-none',
+    // INACTIVE
+    'hover:bg-theme-highlight text-theme-text hover:text-theme-secondary transition duration-200 disabled:bg-grey-600',
+    // CALLED
+    'text-white bg-amber-600 border-amber-600 hover:bg-amber-400',
+    // ACTIVE
+    'text-white bg-green-700 border-green-700 hover:bg-green-500 hover:text-theme-secondary',
+    // RELEASED
+    'text-white bg-amber-600 border-amber-600 pointer-events-none',
+];
```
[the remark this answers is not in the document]
frankkopp: Good one of course....will change.

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
frankkopp: They simply had the old state when the timeout was called - tried all the methods I knew - added a parameter array, adding parameter, etc. - the useTimeout used the variable state when defining the timeout instead the state of the variables when executing it. The refs where the easiest and most reliable way to get the current value.

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
frankkopp: Why would anybody every change the order of an enum in a language where the the order mattes? 
This is a very common practice in C/C++ to have serial/ordered enums used like this by taking advantage of the fact that enums are in fact just automatically assigned incremental const integers. With this it is much easier to write these if-statements otherwise I would have to always compare all states I want to cover in the if. It is less code and I think it is readable if you are used to it. 
I think in a language where you use .map or closures so often using enums like this is surely one of the more readable parts :D. Just look how complicated the previous code for this page was - unreadable and unmaintainable. 
But lets discuss on Discord :)

src/instruments/src/EFB/Ground/Pages/ServicesPage.tsx
frankkopp: @ErickSharp you want to look more into this or is this ok as it is?

frankkopp: @ErickSharp bump 😀

ErickSharp: @frankkopp im very sorry! I haven’t been at my computer for a while. Yes, this is fine with me 🙂

mico975:

Discord          : mico#3145
Object of testing: #7229
Tier of Testing  : 1
Date             : 12/06/2022

Testing Process: 
Make sure buttons states are staying in sync with sim's services - **OK**
Test in "normal" gate situations (services available) - also test with no jetway but stairs - **OK**
Test at places where no services or not all services are available (e.g. KFBW or so) - **OK**
Test edge cases like spamming buttons, hectic clicking, changing EFB pages during service activations, etc. - **OK**
Test for incompatibilities with in-sim ATC or other pushback tools if you have them available. - **OK**, tested with toolbar pushback, no incompatibilities found. In sim ATC not tested (yet).

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

Awemeter:

Discord: Awemeter#3110
Object of testing: #7229 
Tier of Testing: 1
Date: 12/6/2022

Testing Process: 
Spawned into KFBW 
Tested all buttons
Spawned into KDFW due to the fact that KFBW has no catering truck and baggage carts and retested

Negatives:       

- If you start opening a door and then request the service that corresponds with that door but the service is not avail the door button becomes unusable (greyed out).

- You can request services even if they are unavailable.

- After you deactivate a service and it has fully been retracted/gone away, you cannot click the button again for 5ish seconds.

- The wheel chocks and safety cones do not disappear if you have the beacon light switch on but the light itself is not on due to the plane not having enough power.


Testing Results: 
Not Passed

Conclusions: 
Other than the issues I found I do think there should be a button to remove chocks and cones instead of it being linked to the beacon light.

frankkopp:
> * If you start opening a door and then request the service that corresponds with that door but the service is not avail the door button becomes unusable (greyed out).

will check again - thought I had covered this

> * You can request services even if they are unavailable.

Yep - no way to prevent this for now as the sim does not have an API to query available services from within the aircraft.
Only possible by changing the default vehicles/services (I did a prototype but that has some other consequences). 

> * After you deactivate a service and it has fully been retracted/gone away, you cannot click the button again for 5ish seconds.

That is as intended - there is a 5sec delay to avoid getting out of sync. 

> * The wheel chocks and safety cones do not disappear if you have the beacon light switch on but the light itself is not on due to the plane not having enough power.

Not part of this PR but I will look into this.

frankkopp:
> > * If you start opening a door and then request the service that corresponds with that door but the service is not avail the door button becomes unusable (greyed out).
> 
> will check again - thought I had covered this

Works - @Awemeter could you doublecheck? 
https://user-images.githubusercontent.com/16833201/173237395-ea215f22-2a05-45ad-8bd4-b303a4214ade.mp4

> > * You can request services even if they are unavailable.

> Yep - no way to prevent this for now as the sim does not have an API to query available services from within the aircraft. Only possible by changing the default vehicles/services (I did a prototype but that has some other consequences).

But it should work as intended: 
https://user-images.githubusercontent.com/16833201/173237505-57bf0241-26fe-41e0-8db9-88152e7a43a2.mp4

> > * After you deactivate a service and it has fully been retracted/gone away, you cannot click the button again for 5ish seconds.
> 
> That is as intended - there is a 5sec delay to avoid getting out of sync.
 
> > * The wheel chocks and safety cones do not disappear if you have the beacon light switch on but the light itself is not on due to the plane not having enough power.
> 
> Not part of this PR but I will look into this.

Checked this - AC power need to be available (Ext. or APU/ENG) - then the Beacon switch makes the Chocks/Cones disappear.
So works as intended.

PR #7238, opened 2022-05-25
perf: reduce fuel persistency update rate and enable clang optimizations

Saschl:
Fixes #[issue_no]

## Summary of Changes

I noticed that the file for fuel persistency was written every update cycle while being on ground which showed quite a high runtime in the WASM debug window. I've added a timer to only update it every second which should be sufficient IMO.

Furthermore I've enabled clang optimization which also showed a slight improvement in runtime. Note that I used O3 which might be aggressive but I've done a few flights over the last week and didn't notice any erratic behaviour.

### Before

fadec_before
fbw_before

### After
fadec_after
fbw_after

## Additional context

## Testing instructions

1. Spawn C&D and verify the fuel state is saved correctly every second (check the Modified timestamp of the file in the "work/AircraftStates" for the selected livery.
2. Take off and verify the fuel state is not updated (as it was before this change)
3. Observe system behaviour (FBW, AP and Engines) for any possible strange behaviour (very high or low numbers, unusual behaviour)

### DIFF
--- a/src/fadec/src/EngineControl.h
+++ b/src/fadec/src/EngineControl.h
@@ -20,11 +20,11 @@
 
 /* Values in gallons */
 struct Configuration {
-    double fuelCenter = 0;
-    double fuelLeft = 400;
-    double fuelRight = fuelLeft;
-    double fuelLeftAux = 228;
-    double fuelRightAux = fuelLeftAux;
+  double fuelCenter = 0;
+  double fuelLeft = 400;
+  double fuelRight = fuelLeft;
+  double fuelLeftAux = 228;
+  double fuelRightAux = fuelLeftAux;
 };
 
 class EngineControl {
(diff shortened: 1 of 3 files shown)


src/fadec/src/EngineControl.h
```diff
@@ -20,11 +20,11 @@
 
 /* Values in gallons */
 struct Configuration {
-    double fuelCenter = 0;
```
Saschl, line 23: Looks like VSCode applied formatting in some places.. Maybe makes sense at some places but I can revert that if needed.

Saschl:
CC @Taz5150

Taz5150:
> CC @Taz5150

Sorry, did not see this message. Yes please, feel free to make pertinent changes.

Saschl:
Done @2hwk

2hwk (approved):

2hwk:
Tests passed:
- [x] Spawn C&D and verify the fuel state is saved correctly every second (check the Modified timestamp of the file in the "work/AircraftStates" for the selected livery.
- [x] Take off and verify the fuel state is not updated (as it was before this change)
- [x] Observe system behaviour (FBW, AP and Engines) for any possible strange behaviour (very high or low numbers, unusual behaviour)

PR #7291, opened 2022-06-10
fix: update gross weight logic across FM and monitoring systems

### LINKED ISSUE #6584: Handling of ice accretion weight
### Aircraft Version

Experimental

### Build info

```json
{
  "built": "2022-01-19T05:55:43+00:00",
  "ref": "experimental",
  "sha": "44e9ed3ea1d214791968f7e90d175e5007910ce4",
  "actor": "aguther",
  "event_name": "manual"
}
```

### Describe the bug

As part of the MSFS simulation of icing, Ice accreting on the airplane can add significant weight, much more than would occur IRL. When deicing occurs, the airplane weight decreases. The airplane's weight can be observed in two places in the A32NX: 1) on the lower ECAM screen (after engine start); and 2) on the MCDU Fuel Prediction page. The airplane's weight is also used in calculating the takeoff speeds shown on the MCDU PERF TAKE OFF RWY page and the VLS and approach speed shown on the MCDU PERF APPR. The ZFW shown on the MCDU Fuel prediction page is also affected; the ZFW will change if the LSK is pressed at different times while icing or de-icing is occuring.

The gross weight shown on the MCDU fuel prediction page includes the weight of the ice. The gross weight shown on the lower ECAM page does not include the weight of the ice, so there is an inconsistency in the weights shown in these two places.  See the following screenshot:
Screenshot (1491)_LI

Additionally, Vapp and VLS shown on the MCDU PERF APPR page appear to be based on the same gross weight shown on the lower ECAM and will also not reflect the weight of any ice accreted. In contrast, the takeoff speeds on the MCDU PERF TAKE OFF RWY page do appear to reflect the weight of any ice accreted.

The effect of the increased weight is that the stall speeds will increase. In MSFS, the weight can increase by up to 2300 kg, which leads to about a 2-3 knot increase in VLS for flaps full, for example. Since this is within the 5 knot minimum additive to VLS to determine Vapp, it should not cause any significant issues on approach and landing. It is also within the margins already applied to the takeoff speeds. Since the low speed/high angle-of-attack envelope protection system is based on AOA rather than the airplane weight value, any weight error resulting from ice accretion should not matter there either.

### Expected behavior

The airplane weights displayed on the lower ECAM and on the MCDU Fuel Prediction page should be the same. They should not include the weight of ice accreted on the airplane as they would not include this IRL. The weights shown on these pages are not sensed directly by the airplane.

Similarly, neither the takeoff nor the VLS and approach speeds shown on the MCDU Perf pages should include the weight of the ice. They should be derived using the input payload added to the OEW and the weight of the current fuel load.

### Steps to reproduce

Covered in the description

### References (optional)

_No response_

_No response_

donbikes#4084

sidnov:
Fixes #6584 

## Summary of Changes

- The code for gross weight in the MCDU Fuel Prediction/INIT B page considered the ``TOTAL WEIGHT`` simvar which also included weight from accumulated ice on the airframe.

- This PR changes the source to make it more specific by including only payload, fuel and empty weight in GW calculations.

Source of GW information for the following items has been modified according to references 
- [x] ECAM GW indication 
- [x] VLS, GD, F, S speeds
- [x] V1, VR, V2 speed calculation (fictional, QoL feature)
- [x] MCDU Predictions (INIT B, FUEL PRED)

- Removed Simplane usages for better alternatives such as ``A32NX_ENGINE_FF``.
- Also fixed an error in CG input in Fuel Pred/INIT B I introduced in https://github.com/flybywiresim/a32nx/pull/5895/
- Add required type II messages

### To-Do
Once FAC is implemented, replace ```TOTAL WEIGHT``` to compare FAC-GW and FM-GW

## Screenshots (if necessary)

image

## References

image
image
image
image
image

## Additional context

Discord username (if different from GitHub): sidnov# 8337

## Testing instructions

Check for GW values mismatch on the lower ECAM vs MCDU fuel pred page in flight. The difference should be nil/negligible since the update rates and rounding applied can vary. Repeat test for both KGs and LBS.

Test in icy conditions to find any drastic changes in GW in both lower ECAM and MCDU.


frankkopp (approved):

flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -4801,7 +4801,10 @@ class FMCMainDisplay extends BaseAirliners {
      */
     //TODO: Can this be util?
     getGW() {
-        return (SimVar.GetSimVarValue("TOTAL WEIGHT", "Pounds") * 0.4535934) / 1000;
+        const fuelWeight = SimVar.GetSimVarValue("FUEL TOTAL QUANTITY WEIGHT", getUserUnit());
+        const emptyWeight = SimVar.GetSimVarValue("EMPTY WEIGHT", getUserUnit());
+        const payloadWeight = getTotalPayload();
+        return (emptyWeight + fuelWeight + payloadWeight) / 1000;
```
donstim, line 4807: GW should be set equal to the ZFW input + block fuel input by the pilot on the INIT A page, regardless of whether the ZFW is typed in by the pilot, or input by double-clicking the LSK key to automatically load the current ZFW. It should not be based on independently adding the total payload to the empty weight. If the pilot happens to insert the incorrect ZFW and/or block fuel, the weight shown on the ECAM and the weight used for flight path predictions and CDU/PFD display of operating speeds S, F, green dot, and VLS  should be based on these incorrect weights (even though the airplane's flight characteristics/performance will reflect the actual gross weight). Until the ZFW and block fuel are entered, there is no GW value for displaying in the ECAM or to use for flight path or speed predictions.

Once the first engine has started, the fuel weight is then taken from the fuel quantity indicating system instead of the block fuel entered by the pilot. So if the block fuel entered into the MCDU by the pilot is incorrect, the gross weight will be incorrect until the first engine is started.

sidnov: Thanks for the detailed refs, Don. I was looking exactly for this. Will make the changes to the logic

donstim: Also, although this is getting beyond your initial intention, if you want to add another weight initialization related thing, if the ZFW, ZFWCG, or Block fuel inputs are blank after the first engine is started, the MCDU scratchpad should display the message "INITIALIZE WEIGHT/CG."

donstim: Here are all the references:
Screenshot (2110)
Screenshot (2111)
Screenshot (2112)
Screenshot (2113)
Screenshot (2114)
Screenshot (2115)_LI
Screenshot (2116)

sidnov: @donstim @tracernz 
this one should now be ready for a re-review

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXSystemMessages.js
```diff
@@ -61,6 +61,7 @@ const NXSystemMessages = {
     cancelAtisUpdate:       new TypeIMessage("CANCEL UPDATE BEFORE"),
     checkMinDestFob:        new TypeIIMessage("CHECK MIN DEST FOB"),
     checkToData:            new TypeIIMessage("CHECK TAKE OFF DATA", true),
+    checkWeight:            new TypeIIMessage("CHECK WEIGHT", true),
```
tracernz, line 64: This message relates to the secondary flight plan. Are you sure it's the one you want?

donstim:
Only had time for a very short test so far. I loaded in cold and dark, boarded passengers and left the default fuel level. Without entering the ZFW/ZFWCG or Block Fuel in the INIT FUEL PRED page, I went to the PERF TAKEOFF page and saw that F, S, and O speeds were provided. These should not be there until the weights (ZFW and Block Fuel) are input. Same thing with all the speeds on the PERF APPR page. An EFOB is shown on the PERF CLB, CRZ, and DES pages. I'm not sure about that one, but I don't see where that information would come from if the weights are not initialized.

mico975:

Discord          : mico#3145
Object of testing: #7291
Tier of Testing  : 1
Date             : 07/07/2022

Testing Process: 
Check for GW values mismatch on the lower ECAM vs MCDU fuel pred page in flight. The difference should be nil/negligible since the update rates and rounding applied can vary. Repeat test for both KGs and LBS. - **OK**

Test in icy conditions to find any drastic changes in GW in both lower ECAM and MCDU. - **OK**

N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7297, opened 2022-06-13
fix: use visualframe to avoid msfs bug

Saschl:
Fixes #[issue_no]

## Summary of Changes

According to 
https://devsupport.flightsimulator.com/questions/8769/requestdataonsimobject-with-sim-frame-sometimes-st.html

the period SimFrame for receiving simconnect data might be buggy and stop updating. This might explain the issue with deltatime that sometimes appears and causes erratic behaviour. VisualFrame is potentially not affected and this is what this PR attempts to try out.

## Additional context

This should probably tested on exp as there the issue is most present.

## Testing instructions

1. Fly and observe all systems for unwanted behaviour (hydraulic pressure for example). And of course: no hard banking and the plane should be controllable all the time. 

### DIFF
--- a/src/systems/systems_wasm/src/lib.rs
+++ b/src/systems/systems_wasm/src/lib.rs
@@ -545,7 +545,7 @@ impl Time {
         sim_connect.request_data_on_sim_object::<SimulationTime>(
             SimulationTime::REQUEST_ID,
             SIMCONNECT_OBJECT_ID_USER,
-            Period::SimFrame,
+            Period::VisualFrame,
         )?;
 
         Ok(Self {


2hwk (approved):

John-UK:
Same fix Fenix will be applying.

2hwk:
LGTM

rockpapst1:
Short Flight from KLAS to KLAX everything went very Smoth. 
the  VNAV didn't worked - No TOD Where shown but I think thats not the point of this PR. 
So Thumbs up it worked good!

PR #7314, opened 2022-07-01
feat(hyd): hydraulic actuator locking function

crocket63:
Fixes #[issue_no]

## Summary of Changes

This PR adds a mechanism to be able to limit actuator speed in extension or retraction (up to a zero speed).
It can be used to limit the movement of any actuator such as in a failure case.

This will also implement the downlocking function of spoiler actuators, preventing them to go up when no pressure is available.

## Additional context

## Testing instructions

Check that any unpressurised spoiler panel can go down but never up.


src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -799,15 +867,69 @@ impl<const N: usize> HydraulicLinearActuatorAssembly<N> {
...
+    fn update_soft_lock_mechanism(&mut self, assembly_controllers: &[impl HydraulicLocking]) {
+        let mut no_lock = true;
+
+        let mut min_velocity = AngularVelocity::new::<radian_per_second>(-10000.);
+        let mut max_velocity = AngularVelocity::new::<radian_per_second>(10000.);
+
+        // Min velocity will be the max one amongst all the limit velocities requested
+        // Max velocity will be the min one amongst all the limit velocities requested
+        // This way we use the more restrictive speed limitation over all the controller demands
+        for (idx, controller) in assembly_controllers.iter().enumerate() {
```
BlueberryKing, line 901: I am not sure this comment is quite accurate? It seems you `break` out of the for loop as soon as you find the first controller or actuator that should soft lock.

BlueberryKing (changes requested):
Nice work!

I think there's something you could consider changing here. What I don't really like is that the `soft_lock_velocity` property is often meaningless when the actuator is not soft-locked. To fix this, we could introduce an enum:
```rust
enum SoftLockMode {
    None,
    Locked(AngularVelocity, AngularVelocity), // Min and max angular velocity when the body is soft-locked
}
```
Then, `CoreHydraulicForce` would only have a single property `soft_lock_mode: SoftLockMode`.

To implement the `HydraulicLocking` trait, you could use the following:
```rust
impl HydraulicLocking for CoreHydraulicForce {
    fn should_soft_lock(&self) -> bool {
        self.soft_lock_mode === SoftLockMode::Locked
    }

    fn soft_lock_velocity(&self) -> (AngularVelocity, AngularVelocity) {
        match self.soft_lock_mode {
            SoftLockMode::None => (AngularVelocity::default(), AngularVelocity::default()),
            SoftLockMode::Locked(min, max) => (min, max),
        }
    }
}
```
This is only a suggestion, let me know if you think this is not appropriate to use in this case. 😄

crocket63:
> Nice work!
> 
> I think there's something you could consider changing here. What I don't really like is that the `soft_lock_velocity` property is often meaningless when the actuator is not soft-locked. To fix this, we could introduce an enum:
> 
> ```rust
> enum SoftLockMode {
>     None,
>     Locked(AngularVelocity, AngularVelocity), // Min and max angular velocity when the body is soft-locked
> }
> ```
> 
> Then, `CoreHydraulicForce` would only have a single property `soft_lock_mode: SoftLockMode`.
> 
> To implement the `HydraulicLocking` trait, you could use the following:
> 
> ```rust
> impl HydraulicLocking for CoreHydraulicForce {
>     fn should_soft_lock(&self) -> bool {
>         self.soft_lock_mode === SoftLockMode::Locked
>     }
> 
>     fn soft_lock_velocity(&self) -> (AngularVelocity, AngularVelocity) {
>         match self.soft_lock_mode {
>             SoftLockMode::None => (AngularVelocity::default(), AngularVelocity::default()),
>             SoftLockMode::Locked(min, max) => (min, max),
>         }
>     }
> }
> ```
> 
> This is only a suggestion, let me know if you think this is not appropriate to use in this case. 😄

Idea looks good, now need to try :)
What the hell is === ?? :D

BlueberryKing:
> What the hell is === ?? :D

Oh no, JavaScript has cursed me 😄 .

crocket63:
> > What the hell is === ?? :D
> 
> Oh no, JavaScript has cursed me 😄 .

Problem then is when you want to set it to soft locked.... You need to know the velocities and can't just turn it true:
if self.locks_position_in_closed_mode {
            self.soft_lock = SoftLockMode::Locked((), ());
        }

BlueberryKing:
> > > What the hell is === ?? :D
> > 
> > 
> > Oh no, JavaScript has cursed me 😄 .
> 
> Problem then is when you want to set it to soft locked.... You need to know the velocities and can't just turn it true: if self.locks_position_in_closed_mode { self.soft_lock = SoftLockMode::Locked((), ()); }

Ah, yeah, that's an issue, forget what I said!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Comment is accurate. But the break has nothing to do in here lol

BlueberryKing (approved):
LGTM!

2hwk:
This may have conflicts with https://github.com/flybywiresim/a32nx/pull/6913

crocket63:
> This may have conflicts with #6913

There's no functional conflict, will just have to merge carefully. If we can test this one on master context, then I'll do the merge on #6913

mico975:

Discord          : mico#3145
Object of testing: #7314
Tier of Testing  : 1
Date             : 23/08/2022

Testing Process: 
Check that any unpressurised spoiler panel can go down but never up. - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

dragold:
Category: QA Tester
Name: Mr.Blank#6197
Date of testing MM/DD/YYYY:08/26/2022

PR Tested: 7314
Tier of testing:1

Testing technique:  Brute force tryouts
Test Instructions:
Check that any unpressurized spoiler panel can go down but never up.

Methodology:
All as per Testing instructions
Off Normal input test  
Negatives: NIL
Testing Results: PASS 
Conclusions: results were of normal outputs

PR #7332, opened 2022-07-14
fix(fmgc): string stars with non-runway-specific approaches

tracernz:
Fixes #[issue_no]

## Summary of Changes

Couldn't string a STAR with a non-runway-specific approach e.g. ELRU2A to RNAV F at NZQN. This removes the STAR filtering for non-runway-specific approaches in the MCDU page, and allows stringing runway legs in the case where there is only one runway in the STAR.

IRL behaviour is unclear at this stage, and in trying to get a reference for this particular approach I found the IRL database for a particular airline actually doesn't have these approaches in it. We can revisit in future when we can get better references. I think there are three options I can foresee:
- these approaches are never in the IRL database, in which case we should filter them out entirely,
- these approaches are processed into multiple runway-specific approaches by the FMS database tools,
- works as-is here

## Additional context

## Testing instructions

Check that NZQN RNAV F or G can be strung with ELRUV2A. Check that regular approach/STAR combos you're familiar with work okay.


2hwk (approved):

2hwk:
Suggested Routes:
KLAS/26R RADYR2 BLAQQ DCT SHTNR ANJLL4 KLAX/24R
KFLL/28R REGAE1 PERMT AR16 ILM SBY J79 VILLS DPK DPK3 KBDL/24
ESGG/21 TOPLA3J TOPLA DCT PENOR N133 DOMAG T240 OGDAV OGDAV4T EPGD/29
YPAD/05 PANKI3 PANKI H247 CULIN Y59 RIVET RIVET3 YSSY (No Arrival Runway Selected)
SBGL/15 TIVRO1A TIVRO DCT IBDA1A SBSP/35L

PR #7344, opened 2022-07-17
chore: bump @flybywiresim/eslint-config to 0.2.2 (latest)

ErickSharp:
## Summary of Changes

Bumps @flybywiresim/eslint-config to v0.2.2 and makes the necessary changes to eliminate lint errors introduced by the upgrade.

## Additional context


Benjozork:
During fixing some issues we've encountered a lot of warnings generated by the exhaustive hook dependencies rules.

It is right now a warning since there are a lot of instances of this rule triggering, some of them can break behaviour if not fixed with extreme care.

A lot of the instances include using effects as state machines on one state variable. This would no doubt be better with msfs-avionics.

ErickSharp:
> During fixing some issues we've encountered a lot of warnings generated by the exhaustive hook dependencies rules.
> 
> It is right now a warning since there are a lot of instances of this rule triggering, some of them can break behaviour if not fixed with extreme care.
> 
> A lot of the instances include using effects as state machines on one state variable. This would no doubt be better with msfs-avionics.

I think this warning can be overly aggressive in pushing the programmer to include items that theoretically could change but any sensible programmer knows they wouldn’t i.e Redux’s Dispatch hook, useState setters, or at least wouldn’t matter even if they did. How do you think we should proceed from here? Should we disable the warning across the board or allow it to remain just as a warning?

Benjozork:
I'm going to do this from the ground up using a new lint config based on prettier, and without the react hooks rules because those always end up stalling those PRs.

Introduction of prettier is being trialed in the SimBridge repo, we will see how to proceed once that pans out.

PR #7177, opened 2022-05-09
fix(cond): pack pushbutton fault light

mjuhe:
## Summary of Changes

The introduction of #6220 removed the pack pushbutton FAULT light illumination when the position of the valve disagreed with the commanded position. This PR re-introduces that functionality (to the best of my knowledge).

Now the FAULT light will illuminate when the position of the pack flow valve disagrees with the commanded position. This is mainly seen on the ground when there is no bleed air available and the pack pushbutton is pushed in (ON position, white light extinguished). In the future when failures are included for this system there will be other scenarios where the real position and the commanded positions of the valve will disagree.

## Screenshots (if necessary)

image

## References

See issue #1088

See conversation on Discord #ata-21-pressurisation

Also FCOM extract for reference:

image

## Additional context

No ECAM warnings have been included in this PR. For future work when the new FWC is implemented.

Discord username (if different from GitHub): Commander Ga#1548

## Testing instructions

1. Start cold and dark and power up the aircraft.
2. Ensure the pack pushbuttons are in the ON position (no white lights).
3. Check that the FAULT light illuminates with engines off and APU off.
4. When APU bleed is switched on, the FAULT light extinguishes.
5. Even after APU or engines are on, removing the bleed from the packs (ENG 1/2 BLEED or APU BLEED pb are off) illuminates the FAULT light.
6. Ensure the COND and BLEED systems behave normally.

Please note: on the ground without bleed the valve remains closed but commanded position is open (that's why the FAULT light illuminates). If you switch the engine start mode to IGN, or press the DITCHING pushbutton on the overhead, the commanded position will be closed (as per the FCOM above), and the FAULT light will extinguish, even without bleed.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #7355, opened 2022-07-23
refactor(cond): merge air cond and pneumatic pack flow valve

mjuhe:
## Summary of Changes

PR #6220 introduced the first building blocks of the air conditioning system. One of those blocks was the pack flow valves. However, the pack flow valves had already been modelled within the pneumatic system in a more realistic way. This PR removes the temporary valves in the air conditioning section of the code as well as the temporary pack flow valve controller from the pneumatic section. Now the pack flow valves live within the pneumatic system, and the pack flow valve controller within the ACSC.

Visual summary: 
image

Todos:

- [x] Change ACSC PackFlowController to be an array of two controllers, one per pack
- [x] Get open amount of Pneumatic pack_flow_valve to ACSC PackFlowController
- [x] Merge Pneumatic PackFlowValveController's PID into ACSC PackFlowController
- [x] Remove unused code in pneumatic
- [x] Add code from AirConditioning PackFlowValve into ACSC PackFlowController
- [x] Remove AirConditioning PackFlowValve
- [x] Separate A32NX_COND_PACK_FLOW LVar into two, one per pack
- [x] Use pack_flow from Pneumatic valve for calculations instead of demand flow from ACSC PackFlowController

## Additional context

I decided to keep this PR simple by not adding any new functionality. Hence the future work is still the same as it was:
- Integrate with electrical system
- Add delay in Pack FAULT light as per real life
- Add faults
- Integrate with pressurization system

## Testing instructions

No new functionality has been added to this PR. Ensure nothing has been broken when playing with the packs and the air cond system. You can use PR #5384, #6220 and #7177 as reference for testing instructions.


src/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -533,27 +538,48 @@ impl<const ZONES: usize> SimulationElement for ZoneController<ZONES> {
...
+    fn target_open_amount(&self) -> Ratio {
+        self.target_open_amount
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum PackId {
+    Pack1 = 0,
+    Pack2 = 1,
+}
```
davidwalschots, line 558: **thought**: I'm not sure if the usage of this enum is idiomatic. I would consider introducing a "new type" as an alternative, as this saves you from some of the `as` casting hassle:

```rust
pub struct Pack(usize);

impl From<usize> for Pack {
    fn from(value: usize) -> Self {
        Pack(value)
    }
}
```

Note that `#[repr(usize)]` is also helpful when you want to represent an enum with a numeric type greater than the minimum necessary size. Any `#[repr(u*)]` or `#[repr(i*)]` is allowed.

mjuhe: Thanks for the review as always @davidwalschots! I've changed this now and followed your suggestion. I do like the idea but the downside is that you could essentially create a pack with any `id` you wanted, so I've added an `if` statement to prevent pack numbers different than 1 or 2 from being created (the A380 also has 2 packs). 

I've also added a `From` implementation for usize, as the pack number is always 1 higher than the pointer.

davidwalschots: Looks good! Regarding the `From<Pack> for usize` implementation, the `number-1` might cause confusion, as it makes `From` specific to one use case (index-based access). An alternative could be adding a function `to_index(&self) -> usize`, to ensure it's 100% clear there's a conversion to index happening. This would result in e.g.:
```rust
self.pack_flow_controller[pack_id.to_index()]
```

mjuhe: Good idea, I've changed it now as suggested.

2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #7355
Tier of Testing  : 2
Date             : 08/08/2022

Testing Process: 
No new functionality has been added to this PR. Ensure nothing has been broken when playing with the packs and the air cond system. - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7357, opened 2022-07-24
fix: update spoiler drag in fmcg flight guidance model

donstim:
## Summary of Changes

Asobo recently clarified the relationship between the spoiler delta cd increment in the flight_model.cfg file and what is applied in MSFS - https://devsupport.flightsimulator.com/questions/9635/are-air-and-ground-spoiler-aerodynamics-implemente.html. The spoiler cd increment in the flight_model.cfg file is multiplied by the spoiler deflection angle in radians to determine the delta cd increment used in the sim. 
Screenshot (2210)

The current flight model uses a max spoiler deflection angle of 35 degrees. In terms of radians, this is 35/57.3 = 0.6108. Since the delta cd under autopilot control for this spoiler deflection in the cfg file is 0.01, this means the delta cd being applied in the sim is 0.6108 * 0.01 = 0.006108, rounded to 0.00611. (See screenshot of drag debug window below for verification.)

## Screenshots (if necessary)

Screenshot (2211)
## Additional context

Since there have not been any negative pilot comments about the accuracy of the spoiler lift/drag characteristics, no changes in speedbrake lift/drag are proposed or planned relative to the Asobo MSFS SDK update. The current spoiler lift/drag vales in the flight_model.cfg file were determined empirically, based on the resulting vertical speed and deceleration effects, rather than on a specific known lift/drag increment. So, the effect of the SDK update is just a realization that the spoiler lift/drag increments being applied in MSFS are simply not the same as the values in the flight_model.cfg file.

There are plans to fix the max spoiler deployment angle, both for speedbrake use in the air, and for SU 10, spoiler use in the air. The max deflection angle for in air use should be 40 degrees instead of the 35 degrees used currently. For on ground use, it should be 50 degrees. In both cases, the delta lift/drag increments in the flight_model.cfg file need to be adjusted to provide the same MSFS applied lift/drag increment that would be applied with the current max deflection angle of 35 degrees (unless there is a reason to change the lift/drag increments applied in MSFS).

## Testing instructions
 Shouldn't be any QA testing necessary. Just needs acknowledgement from @BlueberryKing that this is understood and acceptable from a VNAV accuracy standpoint.


BlueberryKing:
Thank you for your detailed work as always! A small thing that is not quite clear from the existing code, but would you mind using the drag value for the speedbrakes at half deflection? Whenever VNAV runs a prediction with "speedbrakes extended, " it assumes they are at half deflection as the AP will only allow half deflection.

donstim:
> Thank you for your detailed work as always! A small thing that is not quite clear from the existing code, but would you mind using the drag value for the speedbrakes at half deflection? Whenever VNAV runs a prediction with "speedbrakes extended", it assumes they are at half deflection as the AP will only allow half deflection.

That's what I used. A delta cd of 0.02 is what is in the flight_model.cfg file, so 0.01 is what is used with AP on.

donstim:
Oops!

BlueberryKing:
> That's what I used. A delta cd of 0.02 is what is in the flight_model.cfg file, so 0.01 is what is used with AP on.

Perfect!

BlueberryKing (approved):

PR #7416, opened 2022-08-16
feat(sd): HYD page improvements

lukecologne:
## Summary of Changes

* Visual improvements of the HYD page
* General refactor of the structure of the page
* Fix the systems wasm, so that the pump section pressure switch simvars are correctly updated
  * Also change the hysteresis values of the blue Epump pressure switch, so that it uses the same values as the system pressure switches
* Change some indications to use low pressure switches, instead of the numerical pressure
  * This fixes the indications during cargo door operations, the "arrow" and the system title at the top should stay amber
* Add the LO AIR PRESS indication, from the low air pressure switches
* Change display conditions for the blue elec pump LO indicaiton, so that it only shows when the blue pump section has abnormal low pressure (and not when the pump is off due to the automatic logic) 
* Complete overhaul of the PTU activation indication logc. Should be less prone to flickering now

## Screenshots (if necessary)

| Before | After |
| ----------- | ----------- |
|engines_off_before | engines_off_after |
|engines_running_before | engines_running_after|
|left_fire_valve_before |  left_fire_valve_after |
| yellow_elec_before |  yellow_elec_after |

Cargo door operation:
| Before | After |
| -------|-------|
cargo_door_before | cargo_door_after |

New LO AIR PRESS indication: 
lo_air_press_after

Video demonstrating PTU display logic and other indications:
| Before | After |
|--------|-------|
| https://streamable.com/wcad4h | https://streamable.com/mlwks0 |

## Additional context

## Testing instructions

Check that the indications on the HYD SD page correspond to the actual hydraulic configuration, in as many (hydraulic) situations as possible.
Take special care to verify that the PTU indication is correct, since this is the area that received the biggest logic change.


crocket63 (approved):
Looks good to me, cannot be too sure about the react part though :p

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+            setTransferColor(TransferColor.Amber);
+        } else {
+            setTransferColor(TransferColor.Green);
+        }
+    }, [ptuControlValveOff]);
+
+    const triangleFill = transferState !== TransferState.None ? 1 : 0;
+    const triangle1Orintation = transferState !== TransferState.GreenToYellow ? -90 : 90;
+    const triangle2Orintation = transferState !== TransferState.GreenToYellow ? -90 : 90;
+    const triangle3Orintation = transferState !== TransferState.YellowToGreen ? 90 : -90;
```
BlueberryKing, line 433:
```suggestion
    const triangle1Orientation = transferState !== TransferState.GreenToYellow ? -90 : 90;
    const triangle2Orientation = transferState !== TransferState.GreenToYellow ? -90 : 90;
    const triangle3Orientation = transferState !== TransferState.YellowToGreen ? 90 : -90;
```

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+
+    console.log(transferState);
```
BlueberryKing, line 435: I don't think this should be here.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+    const litersPerGallon = 3.79;
+
+    return lowerPressureSystemPumpLowPress
+        && (transferDirection === TransferState.GreenToYellow && !yellowElecPumpOn || transferDirection === TransferState.YellowToGreen)
+        && ((highPressureSystemPressure > 1450 && lowerPressureSystemQuantity * litersPerGallon < 2.5)
+        || (lowerPressureSystemPressure > 1500 && lowerPressureSystemQuantity * litersPerGallon > 2.5));
+};
+
+const PTU = ({ x, y, yellowPressure, greenPressure, yellowPumpLowPressure, greenPumpLowPressure, yellowQuantity, greenQuantity, ptuControlValveOff, yellowElecPumpOn } : PTUProps) => {
+    const [transferColor, setTransferColor] = useState(TransferColor.Green);
```
BlueberryKing, line 371: Why does this have to be stored in state? It looks like it only depends on `ptuControlValveOff`.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+        <SvgGroup x={x} y={y}>
+            <Triangle x={0} y={0} colour={!systemPressurisedSwitch ? 'Amber' : 'Green'} fill={0} orientation={0} />
+            <text className={`Huge Center${!systemPressurisedSwitch ? ' Amber' : ''}`} x={hydTitleXPos} y={50}>{title}</text>
+            <text className={`Huge Center ${pressureNearest50 <= lowPressure ? 'Amber' : 'Green'}`} x={1} y={92}>{pressureNearest50}</text>
+
+            <line className={pressureNearest50 <= lowPressure ? 'AmberLine' : 'GreenLine'} x1={0} y1={title === 'BLUE' ? 217 : 151} x2={0} y2={103} />
 
             <HydEngPump
                 system={title}
                 pumpOn={pumpPBStatus}
```
BlueberryKing, line 171: Do we actually want to pass along the PB state or the pump state? The variable names are somewhat ambiguous. It looks like we are using `bluePumpPBStatus && bluePumpActive`.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+        <line className={fireValve ? 'GreenLine' : 'Hide'} x1={0} y1={42} x2={x} y2={0} />
+        <line className={fireValve ? 'Hide' : 'AmberLine'} x1={-21} y1={21} x2={21} y2={21} />
+    </SvgGroup>
+);
+
+const levels = [
+    { system: 'GREEN', max: 14.5, low: 3.5, norm: 2.6 },
+    { system: 'BLUE', max: 6.5, low: 2.4, norm: 1.6 },
+    { system: 'YELLOW', max: 12.5, low: 3.5, norm: 2.6 },
+];
```
BlueberryKing, line 234: How about using something like this? We can avoid the `.filter` down the line.
```suggestion
const levels = {
    'GREEN': { max: 14.5, low: 3.5, norm: 2.6 },
    'BLUE': { max: 6.5, low: 2.4, norm: 1.6 },
    'YELLOW': { max: 12.5, low: 3.5, norm: 2.6 },
};
```

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
```diff
@@ -209,243 +123,329 @@ const RAT = ({ x, y }: RATProps) => {
...
+    highPressureSystemPressure: number,
+    lowerPressureSystemPumpLowPress: boolean,
+    yellowElecPumpOn: boolean,
+    transferDirection: TransferState,
+) => {
+    if (transferDirection === TransferState.None) {
+        return false;
+    }
+
+    const litersPerGallon = 3.79;
```
BlueberryKing, line 362: Maybe we can make this a constant in this file? I see 3.79 appearing as a magic number later.

BlueberryKing (commented):
Nice work! There's just a few things I noticed.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
lukecologne: You're right. I originallly thought that it would need to persistent too, but it turned out it doesn't and I forgot to remove this.

src/instruments/src/SD/Pages/Hyd/Hyd.tsx
lukecologne: Yes this is somewhat unclear. The blue pump is a special case though, since the LO indication for it also depends on if the pump is actually running. I'll rename the parameters to make it clearer

lukecologne:
Thanks for the feedback, I've tried to rename the pump PB vars to make them more clear, I'd appreciate if you could check if they are better now.

BlueberryKing (approved):
Lovely!

mico975:

Discord          : mico#3145
Object of testing: #7416
Tier of Testing  : 1
Date             : 31/08/2022

Testing Process: 
Check that the indications on the HYD SD page correspond to the actual hydraulic configuration, in as many (hydraulic) situations as possible. - **OK**
Take special care to verify that the PTU indication is correct, since this is the area that received the biggest logic change. - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7417, opened 2022-08-18
fix(adirs): correct local gravity vector calc for IR body accelerations

lukecologne:
## Summary of Changes

As the title says, it changes a `cos` to a `sin` in the IR local gravity calculation, so it correctly works as the euler angle transformation.

The affected value is not (yet) used in master and exp, so should have no effect at the moment

## References

https://de.wikipedia.org/wiki/Eulersche_Winkel#Anwendungsbeispiel

## Additional context

### DIFF
--- a/src/systems/systems/src/navigation/adirs.rs
+++ b/src/systems/systems/src/navigation/adirs.rs
@@ -1129,7 +1129,7 @@ impl InertialReference {
         // based on Theta and Phi.
         let g = Acceleration::new::<meter_per_second_squared>(9.81);
         self.body_longitudinal_acc
-            .set_value(context.long_accel() / g - pitch.cos(), ssm);
+            .set_value(context.long_accel() / g - pitch.sin(), ssm);
         self.body_lateral_acc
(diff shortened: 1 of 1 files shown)


src/systems/systems/src/navigation/adirs.rs
```diff
@@ -1129,7 +1129,7 @@ impl InertialReference {
         // based on Theta and Phi.
         let g = Acceleration::new::<meter_per_second_squared>(9.81);
```
crocket63, line 1130: You can check UpdateContext::acceleration_plane_reference_filtered_ms2_vector, I think it might do what you want already.

lukecologne: Unfortunately, filtering the accelerations too much would affect the guidance and flight controls, and it seems like the filter is pretty strong. I'd need the unfiltered accelerations in g for this

beheh:
Is this PR still relevant?

lukecologne:
Superseded by #8557

PR #7442, opened 2022-08-27
refactor(efb): store pax seating in LVars (vs local storage)

2hwk:
Fixes #[issue_no]

## Summary of Changes

This is part 1 of a multi-part refactor, the next will be to move the boarding/cargo loading and synchronization of payload weights (the backend) into WASM (rust). 

The first stage of this refactor is to move the bitwise flags that control passenger seating from localStorage (Which is slow and can only be accessed via coherentJS) into LocalVars instead, as well as to deprecate the old ``TOTAL_ROWS_{}_{}`` Local:Vars which are redundant with this bitpacked pax representation.

This has several benefits:

* Performance is much faster than using LocalStorage, where before frametimes could spike by 20ms or more with just 4 pax stations when trying to store several pax flag vars, now the impact is barely noticable to the end user (~1-2ms, which can be further improved). 

* Enables state saving/loading of seating by MSFS's default save file system

* Allows manipulation of passenger seating in WASM (rust/C++), and allows other systems to be able to read current passenger location on a per passenger level of granularity (vs per section/payload station as before)

* External programs can now set a seating layout by first setting the relevant ``PAX_FLAG_{}_DESIRED`` Local:Var with a bitpacked 64-bit long number, i.e. 7, or ``1111111`` would set the first 7 seats in a passenger section (station) as seated, then start the boarding by setting the ``BOARDING_STARTED_BY_USR`` Local:Var as ``true``. This will start the boarding process as though it was triggered by the EFB. 

### Changes

* Fix bug where fuel page was not being properly overriden by the EFB payload page
* Adds seat type for widebody aircraft.
* Improves compatibility with non-A320 aircraft
* Allows external apps and WASM to manipulate passenger seating
* Allows passenger seating to be saved/loaded from .FLT files

## Screenshots (if necessary)

No visual changes

## Additional context

## Testing instructions

Payload page should work as before

### New Payload Backend

1. Go to Ground > Services page and check that all elements are correctly rendered
2. Go to Payload page, 
3. Adjust seating by clicking on the seating layout. This should add new seats and increase the table value for planned passengers.
4. Add baggage by clicking on the 4 sliders, this should affect the desired cargo value in the table.
5. Change per passenger and per bag weight from 84 and 20 to higher/lower values (or leave at default)
6. Input desired passenger and desired cargo in the table
7. Check that ZFW calculation is accurate (Empty Weight is 42500KG, add passenger, passenger bags and cargo weight to this value and this should match desired ZFW)
8. Board passengers with desired time setting (Real/Fast/Instant). Should be able to change speed while passengers are boarding and have it transition between boarding times seamlessly.
9. Deboard passengers (either by inputting in the table, or clicking on the seating layout).
10. Go to the main EFB dashboard by clicking the FBW Logo in the top left, import a simbrief OFP
11. Go back to the payload page and click the cloud icon next to the table to load the OFP. Check that all relevant table values are correctly imported.
12. Load passengers/cargo again
13. Fill out INIT-A and INIT-B in MCDU (Init fuel pred). Check that MLDW indicator now shows on the performance envelope chart.
14. Proceed with normal flight operations. Check that when in air, MTOW envelope is replaced by the FLIGHT envelope.
15. Check that visuals in all themes (dark/light/blue) are correct and readable, with both bright daytime and at night (in sim time)
16. Press COND page and check that air conditioning is still functioning as expected when the air conditioning knobs are adjusted. 
17. Create a new save file with the new seating loadout, then load it. The seating layout should show when you go into the EFB's payload page.

### Check GSX Integration still works

There's a few things to be aware of when using GSX Sync. 
* There's a new page in the EFB settings labelled `3rd Party Options`. Inside there are two options to give you the freedom on what to sync with GSX. This activates the synchronisation. Without it GSX won't sync as expected. It's important to note various items and fields get disabled in the payload and fuel page with these options enabled.
* De-fueling Is NOT supported, GSX only checks if the fuel has changed to a number higher than the current value. If you call the refueling truck and attempt to defuel, the aircraft will defuel but GSX will nag you to change fuel and not actually do anything, constantly nagging you.
* Fueling is managed by A32NX, GSX simply acts as a visual indicator here which allows the user to still use Instant/Fast/Real options in the EFB. However, Fuel cannot be started until the fueling truck has connected the hose as shown by a grey start button. The fueling truck will remain and count for the duration of the fueling period set by A32NX, you will of course get a bill as usual.
* Boarding is managed by GSX (in a sense), this means you cannot adjust the boarding times in the EFB, you can only adjust boarding time, through GSX's own settings. If one physical passengers boards the aircraft through GSX, then A32NX's Boarding backend is made aware and acts accordingly. Passenger seats are still loaded in the same way they would without GSX Sync.
* The same is regarded with Cargo/Baggage, however due to limitations with the simvars available from GSX, all stations are loaded at the same time. So if GSX says baggage is 33% loaded, all stations will be loaded to 33% of their given weight target. This removed a bit of the randomness to loading but the limitations is out of my hands.

An aircraft profile is provided, you don't need to do anything GSX picks up on this

- [ ] Ensure simbrief loading still performs as before in EFB
- [ ] Fueling
  - [ ] Test GSX properly plays the re-fueling animation when you begin refueling with the EFB
  - [ ] Test the refueling animation correctly stops when the EFB stops refueling the aircraft
  - [ ] Test the refueling animation is instant when you're using Instant refueling time
  - [ ] Ensure you can still refuel in-flight with GSX Fuel Sync enabled
- [ ] Payload
  - [ ] Set in the EFB how much passengers you want as-well as cargo, to notify GSX of these details. 
  - [ ] Ensure passenger current in EFB, matches GSX tooltip
  - [ ] Confirm as GSX baggage tooltip percentage increase, current cargo weight increase in EFB
  - [ ] Ensure baggage weight decreases when deboarding
  - [ ] Ensure passenger amount decreases when deboarding
  - [ ] Ensure passenger/cargo amount in EFB cannot be modified after boarding is requested and only appears again after deboarding is completed 


src/instruments/src/EFB/Ground/Pages/Payload/Payload.tsx
```diff
@@ -23,6 +23,40 @@ import { useAppSelector } from '../../../Store/store';
...
+    const setDesiredFlags: ((newValueOrSetter: any) => void)[] = [];
+
+    const cargo: number[] = [];
+    const cargoDesired: number[] = [];
+    const setCargoDesired: ((newValueOrSetter: any) => void)[] = [];
+
+    Loadsheet.seatMap.forEach((station) => {
+        const [stationPax] = useSimVar(`L:${station.simVar}`, 'Number', 300);
+        pax.push(stationPax);
+        const [stationPaxDesired, setStationPaxDesired] = useSimVar(`L:${station.simVar}_DESIRED`, 'Number', 300);
```
beheh, line 41: Hooks should not be called in control structures, like loops. This can break for various reasons and cause hard-to-debug bugs in future. See the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).

src/instruments/src/EFB/Ground/Pages/Payload/Payload.tsx
```diff
@@ -23,6 +23,40 @@ import { useAppSelector } from '../../../Store/store';
...
+        paxDesired.push(stationPaxDesired);
+        setPaxDesired.push(setStationPaxDesired);
+        const [pFlg, setPFlg] = useBitFlags(station.bitFlags);
```
beheh, line 44: Breaks the rules of hooks.

src/instruments/src/EFB/Ground/Pages/Payload/Payload.tsx
```diff
@@ -23,6 +23,40 @@ import { useAppSelector } from '../../../Store/store';
...
+        activeFlags.push(pFlg);
+        setActiveFlags.push(setPFlg);
+        const [pFlgDesired, setPFlgDesired] = useBitFlags(`${station.bitFlags}_DESIRED`);
```
beheh, line 47: Breaks the rules of hooks.

src/instruments/src/EFB/Ground/Pages/Payload/Payload.tsx
```diff
@@ -23,6 +23,40 @@ import { useAppSelector } from '../../../Store/store';
...
+        desiredFlags.push(pFlgDesired);
+        setDesiredFlags.push(setPFlgDesired);
+    });
+
+    Loadsheet.cargoMap.forEach((station) => {
+        const [stationWeight] = useSimVar(`L:${station.simVar}`, 'Number', 300);
```
beheh, line 53: Breaks the rules of hooks.

src/instruments/src/EFB/Ground/Pages/Payload/Payload.tsx
```diff
@@ -23,6 +23,40 @@ import { useAppSelector } from '../../../Store/store';
...
+        cargo.push(stationWeight);
+        const [stationWeightDesired, setStationWeightDesired] = useSimVar(`L:${station.simVar}_DESIRED`, 'Number', 300);
```
beheh, line 55: Breaks the rules of hooks.

beheh (dismissed):

2hwk:
Rewriting to use a custom hook, converting to draft for now

2hwk:
Issues with this new implementation using a custom hook:

- Significant delay between clicking and the UI elements updating
- Seat map can glitch out if the user clicks on the UI too quickly

I've already tried to make this work without breaking the rules of hooks, and just due to the low performance of this approach (having to loop through ALL the simvars/localstorage values is very slow, vs creating a dynamic number of hooks that will handle each station seperately), and the fact that if the initial loop does not change (as is the case here), the order that the hooks are created does not matter, I think it is better to revert to the original implementation.

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
```diff
@@ -10,10 +12,12 @@ function airplaneCanBoard() {
 }
 
 function setDefaultWeights(simbriefPaxWeight, simbriefBagWeight) {
-    const perPaxWeight = (simbriefPaxWeight === 0) ? Math.round(NXUnits.kgToUser(84)) : simbriefPaxWeight;
-    const perBagWeight = (simbriefBagWeight === 0) ? Math.round(NXUnits.kgToUser(20)) : simbriefBagWeight;
+    const perPaxWeight = (simbriefPaxWeight === undefined) ? Math.round(NXUnits.kgToUser(84)) : simbriefPaxWeight;
+    const perBagWeight = (simbriefBagWeight === undefined) ? Math.round(NXUnits.kgToUser(20)) : simbriefBagWeight;
     const conversionFactor = (getUserUnit() == "Kilograms") ? 0.4535934 : 1;
+    console.log("Setting A32NX_WB_PER_PAX_WEIGHT to ", perPaxWeight);
```
svengcz, line 18: Do we need this logging?

flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Core/A32NX_Boarding.js
```diff
@@ -10,10 +12,12 @@ function airplaneCanBoard() {
...
     SimVar.SetSimVarValue("L:A32NX_WB_PER_PAX_WEIGHT", "Number", parseInt(perPaxWeight));
+    console.log("Setting A32NX_WB_PER_BAG_WEIGHT to ", perBagWeight);
```
svengcz, line 20: Same here

svengcz (changes requested):

donstim:
FYI - https://github.com/flybywiresim/a32nx/pull/7691

docs/a320-simvars.md
```diff
@@ -1400,23 +1400,28 @@
...
+- A32NX_BOARDING_STARTED_BY_USR
+    - Bool
+    - Indicates current pax/cargo loading state
 
-- A32NX_PAX_TOTAL_ROWS_{rows}_DESIRED
-    - Number
-    - Indicates the target number of pax in the selected rows
+- A32NX_PAX_FLAGS_{station}
+    - Bitwise
+    - Indicates the current pax in the selected rows (max 53 bits)
```
tracernz, line 1409: You can't safely write 53-bits to a simvar from JS. A single precision float is used in the backend somewhere.

src/shared/src/bitFlags.ts
```diff
@@ -34,7 +44,113 @@ export class BitFlags {
...
+                if ((1 << i & flag) === (1 << i)) {
+                    total++;
+                }
+            }
+        });
+        return total;
+    }
+}
+
+export class SeatFlags extends BitFlags {
```
tracernz, line 82: There should not be any payload or other system specific logic in a generic utility class file.

2hwk:
Hold for https://github.com/flybywiresim/a32nx/pull/7691 CG changes

src/shared/src/bitFlags.ts
2hwk: This is not system specific to only weight/balance as other systems that need to read passenger seating information will use this utility class.

tracernz: Would be better to create a file for the loading specific stuff rather than stuffing it in bitFlags.ts.

docs/a320-simvars.md
2hwk: resolved on discord, turns out the limit is indeed 53-bits :P

2hwk:
Closing for now as this will be delayed until after #7695 has been merged

PR #7446, opened 2022-08-27
fix(elec): prevent power loss on runway spawn

beheh:
## Summary of Changes

Currently the aircraft briefly loses power after clicking "Ready to Play" when spawning on the runway. This is because the `runway.FLT` file forces the N2s to 55%, which is below our generator threshold of 58%. By setting it to 68% instead, which is around where the N2 stabilizes naturally, we avoid this.

This PR is an important change for the FWCs at #4872, because otherwise the FWCs will lose power on spawn in and take almost a minute to startup again.

## Additional context

## Testing instructions

This test only affects the "Start on Runway" situation. Select a runway for departure on the world map and start the flight.
Spawn in on the runway and click "Ready to Fly". While some screens may turn off, make sure the aircraft doesn't experience a total power loss like it did previously.


davidwalschots (approved):
5 out of 5 ⭐

marcman86:
Quality Assurance Report

Discord: marcman86#4907
Object of testing: #7446
Tier of Testing: 1
Date: 28/08/2022

Testing Process:
- Spawned on runway EDDP
- Clicked Ready to fly > no power loss
- Menue > Restart
- Clicked Ready to fly > power loss, game freeze for like 5s, then powers up again

Negatives: Does not work on restart.

Testing Results: Not passed
Conclusions: -

Hope this helps :-)

PR #7476, opened 2022-09-06
feat: Incorporate SU10 flight model crosswind parameters

donstim:
## Summary of Changes
MSFS SU10 adds the following crosswind parameters to the [FLIGHT TUNING] section of the flight_model.cfg file: 

`ground_crosswind_effect_zero_speed` | This parameter represents the world speed (in ft per second) at which 0% of the crosswind effect is applied to the aircraft. This parameter will work in two different ways: With the ground rudder assistance enabled, at the given speed and below, the lateral (x) component of the wind is set to zero. With the ground rudder assistance disabled, crosswind is completely cancelled out below ground_crosswind_effect_zero_speed ft per seconds of IAS, and it is gradually blended in up to 100% at ground_crosswind_effect_max_speed ft per seconds of IAS.Note that this value can be set to -1000 to have a 100% realistic simulation where the crosswind is never cancelled out. Default value is 5.

`ground_crosswind_effect_max_speed` | This parameter represents the world speed (in ft per second) at which 100% of the crosswind effect is applied to the aircraft.Note that this value can be set to -1000 to have a 100% realistic simulation where the crosswind is never cancelled out. Default value is 80. 

`ground_high_speed_steeringwheel_static_friction_scalar` | At high speeds, tires are rolling and - depending on their shape and width and how much they are inflated - they will more or less resist rotation or sideways motion. This parameter allows you to define how much a moveable wheel resists static friction which goes sideways or resists rotation around the vertical axis. Essentially, it allows you to control how much the aircraft will move into the crosswind when rolling at higher speeds, and reducing the scalar will reduce the friction, so the aircraft is more likely to slide. Note that only the lateral forces are impacted by this value, so rolling friction and braking when your aircraft is rolling straight will not be influenced. Default value is 1. 

`ground_high_speed_otherwheel_static_friction_scalar` | At high speeds, tires are rolling and - depending on their shape and width and how much they are inflated - they will more or less resist rotation or sideways motion. This parameter allows you to define how much a non-moveable wheel resists static friction which goes sideways or resists rotation around the vertical axis. Essentially, it allows you to control how much the aircraft will move into the crosswind when rolling at higher speeds, and reducing the scalar will reduce the friction, so the aircraft is more likely to slide. Note that only the lateral forces are impacted by this value, so rolling friction and braking when your aircraft is rolling straight will not be influenced. Default value is 1.

The crosswind speed parameters have been set to -1000 so that any crosswind is never cancelled out. The sliding friction scalars have been kept at the default value of 1. (At a maximum crosswind value of <40 knots, there should not be any sliding from the A320 tires.)

## Additional context

## Testing instructions

Test takeoffs and landings using beta SU10 in various crosswind conditions up to 35 knots for takeoff and 38 knots for landing. The only noticeable differences should be that: 1) rudder will need to be applied sooner during takeoff (like from brake release) and held longer for the landing rollout, and 2) it should be easier to hold center line without oscillations due to the crosswind suddenly coming in during the takeoff roll or being cancelled on the landing rollout.

More explanation:  Without the PR (and pre-SU10), all crosswind is cancelled below a certain speed and gradually applied as  you speed up during takeoff, or removed as you slow down during landing.  This can cause problems when the crosswind kind of suddenly appears during the takeoff roll, or disappears during the landing roll -- leading to over or under-controlling and sort of "weaving" (oscillating) your way down the runway. By always having the crosswind present, regardless of speed, it should allow a more natural, realistic application of rudder control during both takeoff and landing. You will, however, still have to deal with the variation of rudder aerodynamic control effectiveness with airspeed -- the rudder control will be more effective as you speed up. You can still over or under-control and oscillate about the runway centerline, but this should improve with proficiency and not be due to suddenly having the crosswind appear or disappear. (There can also still be gusts that will change the crosswinds!)


tracernz (approved):

Letreivchen:

Discord          : Letreiv#9404
Object of testing:  #7476 
Tier of Testing  : 1
Date             : 21/09/2022

Testing Process: 
MSFS SU10 Beta
Several circuits at EDLP in various crosswind conditions, ranging from 0 to 38 knots and gusts up to 45 knots, config 3 and FULL
Checked that:

- rudder will need to be applied sooner during takeoff (like from brake release) and held longer for the landing rollout
- a more natural, realistic application of rudder control during both takeoff and landing can be felt
- the rudder control is more effective as you speed up

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
This PR fixes the sudden appearance of crosswind at a certain speed which makes the rudder control more natural and prevents uncontrolled oscillation from the centerline during takeoff and landing.
Gusts still can be felt.

Media:       
N/A

PR #7511, opened 2022-09-25
refactor: aircraft presets reordering based on pilot feedback and sop

frankkopp:
## Summary of Changes
Changes order of procedure steps based on pilot feedback and aligned with SOP.
Also adds some additional steps (CVR Test, T.O Config, Cabin Ready)
Some C++ clangd-tidy cleanup

## Testing instructions
- Test that Presets still work in general and that it is possible to switch from one state to the other without issues. 
- Order of steps should roughly be as in the SOP (but not identical as presets are adapted to sim usage)


2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #7511
Tier of Testing  : 1
Date             : 17/10/2022

Testing Process: 
Test that Presets still work in general and that it is possible to switch from one state to the other without issues. - **OK**
Order of steps should roughly be as in the SOP (but not identical as presets are adapted to sim usage) - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7512, opened 2022-09-25
feat(fac): Move speedscale computation to the FACs

lukecologne:
Fixes #[issue_no]

## Summary of Changes

This PR adds on the functionality of the FAC introduced in #6913, by moving the rest of the speedscale computation to the FAC.
This includes:

- [x] Vmax, Vfenext
- [x] Vctrend
- [x] V3 (F Speed)
- [x] V4 (S Speed)
- [x] Vman (Green Dot Speed)
- [x] Vls

(Vls/Vs/S/F is partly done, first iteration is pretty much a direct copy of the previous calculation from JS)

It also includes some improvements to the speeds:
- [x] Vmax now also takes Vle into account when the gear is extended
- [ ] Vls now uses actual flaps/slats position, so even in abnormal flaps/slats configs, Vls should be correct. This also means that it no longer requires filtering, as it moves linearly interpolated with the flaps setting
- [x] Vls is now also correct for speedbrake position
- [x] Vls is now properly corrected for 0.3g margin to buffet onset, instead of just adding a fixed amount of speed above 20000ft
- [x] A hysteresis is now applied to VCtrend, i.e. the arrow appears at 2kts, and disappears at 1kts. This was lost in the avionics-framework port of the PFD

## Additional context

## Testing instructions

Verify that the values computed by the FAC on the speedscale (i.e. the values mentioned above) are sensible, and match with the old implementation (the values of the old compuation can still be reviewed in the Lvar debug window, under A32NX_SPEEDS). And exception to that is Vls above 20000ft, which should be higher than the preivous values. Also make sure the AP/FD correctly use the Vls values.


crocket63 (approved):
Did not check simulink models generated code.
Not an expert of that function, but could not spot obvious copy paste errors :)

src/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -468,47 +440,47 @@ class SpeedTrendArrow extends DisplayComponent<{ airspeed: Subscribable<number>,
...
-                const offset = -targetSpeed * DistanceSpacing / ValueSpacing;
-                const neutralPos = 80.823;
-                if (sign > 0) {
-                    pathString = `m15.455 ${neutralPos + offset} l -1.2531 2.4607 M15.455 ${neutralPos + offset} l 1.2531 2.4607`;
-                } else {
-                    pathString = `m15.455 ${neutralPos + offset} l 1.2531 -2.4607 M15.455 ${neutralPos + offset} l -1.2531 -2.4607`;
-                }
-
-                this.offset.set(`m15.455 80.823v${offset.toFixed(10)}`);
+        sub.on('vCTrend').handle((word) => {
```
Saschl, line 480: Maybe `withArinc429Precision` would  be appropriate here?

src/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -523,39 +495,35 @@ class SpeedTrendArrow extends DisplayComponent<{ airspeed: Subscribable<number>,
...
-        const seconds = deltaTime / 1000;
-        const vls = SmoothSin(this.vls, vlsDestination, 0.5, seconds);
-        this.previousTime = currentTime;
-        return vls;
-    };
-
     private setVlsPath() {
-        const normalLawActive = this.fcdc1DiscreteWord1.getBitValueOr(11, false) || this.fcdc2DiscreteWord1.getBitValueOr(11, false);
+        if (this.vls.isNormalOperation()) {
+            this.refElement.instance.style.visibility = 'visible';
```
Saschl, line 516: I believe it is more efficient to use as subject (e.g. `Subject.create('visible')`) and use that in [Line 566](https://github.com/flybywiresim/a32nx/pull/7512/files#diff-258e500ddb10dce616498c2ec1b32b7637850e32a26739a176761f7dbe9f794bR566) as a visibility toggle. This way the DOM only gets modified when there are actual changes. I know it's been done the way you did in in many places, but I didn't know that the subject way is better back then. You can also keep it for now as I planned to make another pass with the PFD to change it in the other places as well.

src/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -578,8 +546,8 @@ class VLsBar extends DisplayComponent<{ bus: EventBus }> {
             this.setVlsPath();
         });
 
-        sub.on('vls').handle((vls) => {
-            this.vls = this.smoothSpeeds(vls);
+        sub.on('vLs').handle((vls) => {
```
Saschl, line 549: `withArinc429Precision` maybe? Not mandatory IMO, but if it won't hurt, I'd add it :)

Saschl (commented):
Reviewed the PFD part. Just some optional comments about optimization but approved from my side!

lukecologne:
Thanks for the review, I decided to fix the subject stuff, no need to make your job harder later

donstim:
Here's one that looks a little off to me. Just after takeoff at 60.3T, CONF 1+F, S speed should be 186 KCAS according to QRH and LVAR, but looks to be about 183 or 184 on the PFD if I'm not mistaken.
Screenshot (2417)

donstim:
For what it's worth, I checked the green dot, F, and S, and CONF 3 and FULL VLS speeds on approach at 45T up to 80T by 5T. All were spot on.

PR #7516, opened 2022-09-25
feat(efcs): move elevator elasticity table to laws

lukecologne:
## Summary of Changes

This PR removes the elevator elasticity table from the flight model definition file. They are now included as gains in the control laws,
so no change in the feeling should be encountered, as the net output is still the same.
The only difference is in the simulated hydraulics deflection of the surfaces (and the reported deflections from the hydraulics), as they did not account for the scaling the elasticity tables introduced after hydraulics. This caused the surfaces simulated by hydraulics to be deflected too much, which can be also seen on the F/CTL page, as the simulated hydraulics deflection is also reported by the F/CTL computers. This deflection should now be more sensible (see videos).

## Screenshots (if necessary)

Before (note the very large deflections, even at high speeds):
https://streamable.com/jat8gc
After (deflections are now more sensible):
https://streamable.com/pl6kgf
This also allows for some fun stuff in direct law, as with the new, higher maximum deflections, we can pull some serious g's:
https://streamable.com/0qtsvt

## Additional context


tracernz (approved):

tracernz:
Ah, there is now a failed test relating to aileron position.

donstim:
One potential issue is that I was going to try to see if this table could be used to adjust the pitch sensitivity of the airplane, especially at low speed (low dynamic pressure in this table). My previous attempt at adjusting pitch sensitivity using the slope of the elevator lift coefficient slope or the elevator effectiveness scalar did not work due to other unintended consequences of changing those parameters.

Having said that, I think I also tried this before and was unsuccessful. However, I would like to still have the opportunity to test it again. Do I understand you are simply moving the current values of this table into the control law gains, and I would still have the opportunity to use such an elasticity table in my own tests, but with the understanding that the current gains are already hardcoded in by this PR (if merged) and would need to be accommodated for?

lukecologne:
Yeah you can continue to use the table, the problem with the previous usage as a control law gain (which it effectively was) is that it limited the maximum deflection especially at high speeds, and that it was incompatible with the current hydraulics implementation. If you use it as a sort of "fine control" for the aerodynamics, which is what you are trying to do as far as I understand it, it should be fine

Saschl:
Quality Assurance Report

Discord : saschl
object of testing: https://github.com/flybywiresim/a32nx/pull/7516
Tier of Testing: 1
Date: 28/09/2022

Testing Process:


Replicated behaviour shown in the description and the results were consistent with the explanation and videos.

No negative/weird behaviour noticed.


Negatives:
N/A

Testing Results:
Passed ;)

donstim:
This should have been vetted more carefully. Where is the IRL data this is based on? What are "sensible" values for elevator deflection at different speeds? Other than due to envelope protection, why would the elevator deflection be limited in normal law? The control inputs being made in the videos are not shown, nor is the dynamic pressure, nor the actual elevator deflection values (simvars), nor does it seem to be considered that the maximum elevator deflections in the flight model being  used are not the maximum elevator deflection values for the real airplane.

The "serious" g-values that can be attained in direct law are not shown in the videos, but they look to be completely unrealistic. According to a pilot I worked with on the new flight model, you might get to 3-3.5 g with full back stick at high speed. That is where a corrected elevator elasticity table comes into play.

This PR threw the baby out with the bath water rather than fixing any issues.

lukecologne:
Well, as you know, in direct law, the control surface deflection per sidestick deflection is scaled down as a function of the current flaps/slats configuration, which would of course limit the maximum load factor. Note that this can **not** be achieved by the elasticity table, as 

1. It scales with speed, which is not correct
2. It would also affect the C* law, and 
3. It doesn't work well with the hydraulics as we established.

Now, it should be obvious from the video that I did not yet implement this, neither in roll nor in pitch direct law, as I did not have the time to do so. In my opinion this is not that big of a problem, since abnormal operations are not guaranteed to be 100% accurate, as the sign on the EFB says.

donstim:
According to the FCOM, in direct law, pitch is a direct stick-to-elevator relationship. It's in roll that system gains correspond to slat/flap configuration. I agree that such a gain, where it exists, should be a property of the flight control laws, not the airplane flight model. However, when control laws are not active (when mechanical law is in effect, or the current A32NX direct law), the flight model should ideally address this. I maintain that the airplane is not a fighter jet and should not be able to maneuver like in your video (even if the structure does not fail). The updated flight model can and does address this very well by use of the elevator elasticity table. (Too bad you'll never see it.)

You seem to forget that this is MSFS. Due to the way in which MSFS models the pitch control, the elevator elasticity table is a very useful, maybe even necessary, tool. Other MSFS airplanes also use it, including two very well known payware airplanes, one of which is a fly-by-wire airplane like ours. 

From my viewpoint, which may not be the same as yours, accurately simulating the flight characteristics and performance of the airplane is of primary importance. The hydraulics, gauges, control laws, etc., which can be programmed outside of MSFS, should be adjusted to work with the MSFS-compliant flight model, not the other way around.

lukecologne:
> However, when control laws are not active (when mechanical law is in effect, or the current A32NX direct law), the flight model should ideally address this

In mechanical control, only the THS can be controlled. The elevators will be automatically in centering mode if hydraulic power is available. Not sure what you mean with "current A32NX direct law", direct law is very much controlled by the flight control computers, since #6913. Now, for your FFS reference, here we go:
* With current master (i.e. with this PR):
grafik
* FFS:
grafik
Unfortunately I can't post the full video here on github.
It looks to me like the deflection in the FFS is even lower than in MSFS? I don't know what the CG position was in the FFS, but in MSFS it was 25% as far as I can remember.

donstim:
Yes, I was incorrect in calling it mechanical law or equating that with A32NX direct law.  The point of my comments remain unchanged.

You can place the video on another site and provide link in github. There certainly are some funky things going on with the elevator deflections displayed on the ECAM. They certainly aren't the same deflections as shown by the MSFS simvars. Seems somewhat arbitrary to me and that they could be changed to match what is expected. The important thing is, does the airplane behave as it should? If you instead start playing around with the flight model to obtain the desired deflections, you run the risk of messing with other fight characteristics that are also affected by that flight model variable or variables. Then it can be a goose chase changing other flight model parameters to resolve those issues as there is a lot of integration within the flight model between parameters.

lukecologne:
> You can place the video on another site and provide link in github

That's not what I meant. I can't post it publicly.

The deflection shown on the F/CTL SD page is the actual elevator position provided by hydraulics, which is the same thing that is finally written to the MSFS `ELEVATOR POSITION` simvar. That being said, there seems to be a bug in the hydraulics code converting from internal deflection to MSFS deflection, causing the nose down deflection to be incorrectly scaled. Nose up should still be correct though. Not sure what exactly you mean with the displayed deflections not being the same as the MSFS deflections. The only difference should be (excluding the aforementioned bug), if the elasticity table is not equal to 1, as then MSFS will scale down the position written to `ELEVATOR POSITION`.

donstim:
It looks to me like it might be scaled based on a hardcoded max deflection value rather than using what it is in the flight_model.cfg file. If so, you should see that when you changed the deflection values to the 30 up and 17 down.  For an example, I have 25 up and 17 down in the draft revised flight model (which are used in an integrated fashion with other parameters to get the desired pitch sensitivity and overall control authority at high and low speeds). For 10 degrees of up elevator deflection, which is 40% of the available elevator, the ECAM shows over 60%. 10 degrees out of 25 is 40%, while 10 degrees out of 16 (the value in the current flight model) is 62.5%.

Since you replied before I finished my edits of my above comments, I'll put them here as well:

Interesting that you tried changing the max deflection value when you told me this causes problems with the hydraulics simulation. Have you checked what increasing the max elevator deflection does to pitch sensitivity (which is a major issue with the current flight model)? Does this also improve things at low speed, or does it make this comparison (if you have it) worse?

Okay, I will try using this pr with the draft new flight model and try to revise the new elasticity table if necessary. It only reduces elevator deflection for dynamic pressures of 300 and above -- 0.75 at 300 and 0.60 at 400. But I expect your hardcoding of the old values into normal law will create issues with either pitch sensitivity or lift/drag/pitch values relative to this flight model that is based on a 1.0 elevator elasticity at dynamic pressures less than 300. If it doesn't, I'll proceed with updating the flight model. If it does, then you are now responsible for the flight model.

PR #7538, opened 2022-10-06
fix: repair fdr data for support analysis 

frankkopp:
## Summary of Changes
Unfortunately #6913 removed many vital data points in the fdr files without adequate replacement. This renders FDR files useles for support purposes (AP Disconnect Analysis, COntroller Input Analysis, High AOA Analysis, etc.). 

This PR attempts to bring these data points back into the fdr files. 

## Testing instructions
Testing the FDR with this PR only is not possible. It also requires the coversion tool and graph tool. 

But it shoul be test against general issues - basically anything should work as before. Probably best to simply be used in Exp. 

### DIFF
--- a/src/fbw/src/FlyByWireInterface.h
+++ b/src/fbw/src/FlyByWireInterface.h
@@ -94,6 +94,25 @@ class FlyByWireInterface {
   bool last_ls1_active = false;
   bool last_ls2_active = false;
 
+  // Fix missing data for FDR Analysis
+  // controller input data
+  double inputElevator = 0.0;
+  double inputAileron = 0.0;
+  double inputRudder = 0.0;
+  // additional
+  double simulation_rate = 0.0;
+  // ambient data
+  double ice_structure_percent = 0.0;
+  double ambient_pressure_mbar = 0.0;
+  double ambient_wind_velocity_kn = 0.0;
+  double ambient_wind_direction_deg = 0.0;
+  double total_air_temperature_celsius = 0.0;
+  // aoa - these are not correct yet
+  double alpha_floor_command = 0.0;
+  double protection_ap_disc = 0.0;
+  double v_alpha_prot_kn = 0.0;
+  double v_alpha_max_kn = 0.0;
+
   FlightDataRecorder flightDataRecorder;
 
   SimConnectInterface simConnectInterface;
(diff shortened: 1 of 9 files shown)


src/fbw/src/FlyByWireInterface.h
```diff
@@ -94,6 +97,26 @@ class FlyByWireInterface {
...
+  double ambient_wind_velocity_kn = 0.0;
+  double ambient_wind_direction_deg = 0.0;
+  double total_air_temperature_celsius = 0.0;
+  // aoa
+  double alpha_floor_condition = 0.0;
+  // these are not correct yet
+  double protection_ap_disc = 0.0;
+  double v_alpha_prot_kn = 0.0;
+  double v_alpha_max_kn = 0.0;
+
```
lukecologne, line 119: Seems like these fields are not used and are duplicated from [AdditionalData.h](https://github.com/flybywiresim/a32nx/pull/7538/files#diff-6a74ba4307cc2097b8e85392beee956d961c74424e0d2192497afbc6c09bf88f)

src/fbw/src/failures/FailuresConsumer.cpp
```diff
@@ -50,3 +50,12 @@ bool FailuresConsumer::setIfFound(double identifier, bool value) {
     return true;
   }
 }
+
+bool FailuresConsumer::isAnyActive() {
+  if (std::any_of(activeFailures.begin(), activeFailures.end(), [](auto pair) {
```
lukecologne, line 55: Vscode tells me that `std::any_of` needs to have `algorithm` included? Not sure why it compiles without it

lukecologne (changes requested):

src/fbw/src/failures/FailuresConsumer.cpp
frankkopp: My IDE shows that it seems to be included down the line somewhere - but maybe better to add it explicitly.

src/fbw/src/FlyByWireInterface.h
frankkopp: yea - my fault - interestingly these inits seem to not even be necessary. 
Will fix

lukecologne (approved):
Looks good!

PR #7539, opened 2022-10-08
fix: avoid flyPad crash when SimBrief cruise profile is other than "CI"

frankkopp:
## Summary of Changes
When a user choses anything other than "CI" on the SimBrief Dispatch Options page the flyPad crashed. [0]
Root cause is that the SimBrief API returns an empty object instead of the expected string. [1][2]

This PR adds a check if the returned `general.costindex` is a string and if not it returns a string containing the cruise profile text which indicated what has been chosen in the SimBrief context [3]. 
This could of course be set to a default CI value - currently unclear what would be better. 

## Screenshots (if necessary)
[0]
image

[1] Correct with "CI"
image

[2] Incorrect with "M78"
image

[3] Using the cruise_profile text:
image

## Testing instructions

- Check if SimBrief imports to flyPad and MCDU work even when Cruise Profile is set to something other than "CI".
- Check with "CI" to test for any regressions. 
- MCDU will show a "NOT ALLOWED" and the Cost Index with orange boxes but should not crash or show any other issues. 
- flyPad Dashboard should show "n/a" for Cost Index. 
- Check anything else that uses SimBrief and/or CI

### DIFF
--- a/src/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
+++ b/src/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
@@ -47,7 +47,7 @@ const NoSimBriefDataOverlay = ({ simbriefDataLoaded, simbriefDataPending, fetchD
                             <button
                                 type="button"
                                 onClick={fetchData}
-                                className="flex justify-center items-center p-2 space-x-4 w-full rounded-md border-2 transition duration-100 text-theme-body hover:text-theme-highlight bg-theme-highlight hover:bg-theme-body border-theme-highlight"
+                                className="flex justify-center items-center p-2 space-x-4 w-full text-theme-body hover:text-theme-highlight bg-theme-highlight hover:bg-theme-body rounded-md border-2 border-theme-highlight transition duration-100"
                             >
                                 <CloudArrowDown size={26} />
                                 <p className="text-current">{t('Dashboard.YourFlight.ImportSimBriefData')}</p>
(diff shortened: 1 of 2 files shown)


tracernz:
Should be default cost index of 0 when it’s not included in the uplink?

frankkopp:
> Should be default cost index of 0 when it’s not included in the uplink?

Definitely an option. But the way I have done it it shows the SimBrief setting clearly in the flyPad dashboard. 
0 would be more confusing to users if they chose "M78" or "LCR". 
Internally 0 will be used anyway - that is in the MCDU Perf pages 0 is shown. 

Wdyt?

tracernz:
> Definitely an option. But the way I have done it it shows the SimBrief setting clearly in the flyPad dashboard. 0 would be more confusing to users if they chose "M78" or "LRC". Internally 0 will be used anyway - that is in the MCDU Perf pages 0 is shown.
> 
> Wdyt?

SGTM

frankkopp:
> > Definitely an option. But the way I have done it it shows the SimBrief setting clearly in the flyPad dashboard. 0 would be more confusing to users if they chose "M78" or "LRC". Internally 0 will be used anyway - that is in the MCDU Perf pages 0 is shown.
> 
> > 
> 
> > Wdyt?
> 
> 
> 
> SGTM

Navigraph said that the empty object was indeed a bug on their side. They changed it to s string now. I might close this PR accordingly after some additional testing. 
@tracernz or do you think we should keep it as a precaution? 
I wouldn't think so as we wouldn't have this on other fields either.

frankkopp:
Will close as Navigraph fixed this on their side. Users need to regenerate their OFPs for the fix to work.

PR #7542, opened 2022-10-08
fix(engine): Engine primary parameters fine tuning – Part 1 of 2

Taz5150:
## Summary of Changes

Engine primary parameters (N1, N2, EGT and Fuel Flow) were originally modeled using real Leap-1A26 performance data. 

-	**EGT** and **Fuel Flow** have been included in the WASM code as polynomials after an optimization process performed with ML algorithms.

-	**N1** and **N2** (with respect to thrust) have been modeled using legacy tables within the _engines.cfg_ file, mostly based on bi-linear interpolations.

The errors of these variables with respect to published figures at ISA conditions (FCOM) are more than reasonable (maximum absolute errors of around 5%). This PR solves non-ISA situations and brings further down the maximum error.

This is a two-part PR: in **Part 1** (this PR), we are fine-tuning **EGT** and **Fuel Flow**. **Part 2** will focus on N1 and N2.

## Screenshots (if necessary)

- **EGT % Error wrt. Real Airbus Data** - Actual Model (Orange) vs. New Model (Blue)
   Maximum % Error Improvement: from **3.45%** to **2.11%**

image

- **Fuel Flow % Error wrt. Real Airbus Data** - Actual Model (Orange) vs. New Model (Blue)
   Maximum % Error Improvement: from **5.30%** to **2.72%**

image

## References

- For Fuel Flow figures, the main reference is the FCOM performance tables.

image

- For EGT, the reference is taken from Airbus PEP simulation output.

## Additional context

Since both EGT and Fuel Flow are a direct functions of  CN1 (among other parameters), results depend on the performance of N1 - so if N1 is a little off, EGT and Fuel Flow will also be a little off. Polynomials have been thus tuned to FCOM and Airbus N1 values.

Discord username (if different from GitHub): TazX[Z+2]#0405

## Testing instructions

Due to the dependence on N1, and since N1 is not fine-tuned yet, it will be hard to make a sound decision whether the new model is better or not than the old one.

That being said, the idea is to test FCOM performance points and check Fuel Flow is within specs. Also, it would be good to make a reasonable long trip and check that fuel consumption matches fuel flow.


mico975:

Discord          : mico#3145
Object of testing: #7542
Tier of Testing  : 1
Date             : 17/10/2022

Testing Process: 
That being said, the idea is to test FCOM performance points and check Fuel Flow is within specs. - **OK**
Also, it would be good to make a reasonable long trip and check that fuel consumption matches fuel flow. - **OK**, 4 hour flight, fuel consumption as expected.

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM, no reports on any exp issues so imo good to go.

PR #7554, opened 2022-10-16
feat(hyd):  Electro Hydrostatic Actuator (E(B)HA)

crocket63:
Fixes #[issue_no]

## Summary of Changes

Some bigger planes are using Electro Hydrostatic Actuators (EHA).

This PR adds an electrical backup option for linear actuators.

- [x] Electrical pump speed
- [x] Electrical consumption
- [x] Simple accumulator model
- [x] Drivable refill mechanism from aircraft hydraulic circuit
- [x] EHA / EBHA option (EBHA can use either electrical or aircraft hydraulic power)

## Additional context

## Testing instructions

Not implemented yet in an aircraft so unit test are enough for now


src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -53,6 +56,268 @@ pub enum LinearActuatorMode {
...
+
+        self.update_power_consumed(actuator_flow, actuator_pressure);
+        self.speed.update(context.delta(), new_speed);
+    }
+
+    fn update_power_consumed(&mut self, actuator_flow: VolumeRate, actuator_pressure: Pressure) {
+        // Horsepower = Pressure (PSIG) × Flow (GPM)/ 1714
+        let current_fluid_power = Power::new::<horsepower>(
+            actuator_pressure.get::<psi>() * actuator_flow.get::<gallon_per_minute>() / 1714.,
+        );
```
BlueberryKing, line 118: Using SI units here should let you get rid of the magic number `1714`.
```suggestion
        let current_fluid_power = Power::new::<watt>(
            actuator_pressure.get::<pascal>() * actuator_flow.get::<cubic_meter_per_second>(),
        );
```

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -53,6 +56,268 @@ pub enum LinearActuatorMode {
...
+
+    fn update(
+        &mut self,
+        context: &UpdateContext,
+        input_pressure: Pressure,
+        controller: &impl ElectroHydrostaticPowered,
+    ) {
+        let mut new_pressure =
+            if input_pressure > self.pressure.output() && controller.should_open_refill_valve() {
+                Pressure::new::<psi>(1800.)
```
BlueberryKing, line 188: What does the `1800` mean here? This should probably be extracted into a constant.

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -53,6 +56,268 @@ pub enum LinearActuatorMode {
...
+
+    fn should_activate_electrical_mode(&self) -> bool {
+        false
+    }
+}
+
+#[derive(PartialEq, Copy, Clone)]
+pub enum ElectroHydrostaticActuatorType {
+    EHA,  // Can only run on electric mode or is damping
+    EBHA, // Can run in electric or from aircraft hydraulic pressure
```
BlueberryKing, line 249: The comments are excellent, so please leave them in, but I suggest we use the full names for these, rather than the abbreviations.

BlueberryKing (changes requested):
Looks good!

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: I don't know :D

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Ends up with a complicated enigmatic name :D

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -53,6 +56,267 @@ pub enum LinearActuatorMode {
...
+
+    fn should_activate_electrical_mode(&self) -> bool {
+        false
+    }
+}
+
+#[derive(PartialEq, Copy, Clone)]
+pub enum ElectroHydrostaticActuatorType {
+    ElectroHydrostaticActuator, // Can only run on electric mode or is damping
+    ElectroBackupHydrostaticActuator, // Can run either in electric backup mode or from aircraft hydraulic pressure
```
lukecologne, line 248: EBHA actually stands for "Electrical Backup Hydraulic Actuator", not hydrostatic

crocket63: And they just add (B) to describe it doh :p

lukecologne: Well, you know how it is with airbus and abbreviations

src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -53,6 +56,271 @@ pub enum LinearActuatorMode {
...
+    fn is_active(&self) -> bool {
+        self.should_activate_electrical_mode && self.is_powered
+    }
+
+    fn max_available_pressure(&self, accumulator_pressure: Pressure) -> Pressure {
+        if self.is_active() && accumulator_pressure.get::<psi>() > 100. {
+            Pressure::new::<psi>(Self::NOMINAL_MAX_PRESSURE_PSI)
+        } else {
+            Pressure::default()
+        }
```
BlueberryKing, line 136: Hmm, does it make sense to use a fixed threshold cutoff here? With this, it looks like the backup would be immediately available as long as there's any accumulator pressure.

Don't get me wrong, you probably still get some interesting dynamics, but I wonder if plan on going even more in-depth with this in the future 😄.

crocket63: Well we could map max pressure on accumulator pressure, but keep in mind accumulator in those actuators is only part of the low pressure side. So it is in fact the equivalent of the air pressurisation of our hyd reservoirs. So a low pressure here is in fact a fluid level indication basically. So we are already on a super edge case where the actuator would probably declare a fault and do not work at all anyway..... We can still add a map at some point if we feel there's a failure scenario for it to be noticeable

BlueberryKing (approved):
Good to go!

PR #7563, opened 2022-10-21
refactor: FADEC code for A32NX & A380X options

Taz5150:
## Summary of Changes

This PR introduces a preliminary FADEC template for the A380X aircraft. The code has been adapted for a 4 engine configuration, but still based on Leap-1A26 characteristics.

The path structure within the source has also been adapted, and instructions on how to build for the A380X option has been placed in the Readme.md at ./src/fadec/

## Additional context


src/fadec/a380_fadec/src/FadecGauge.cpp
```diff
@@ -0,0 +1,32 @@
+#include "FadecGauge.h"
+
+FadecGauge FADEC_GAUGE;
+
+__attribute__((export_name("FadecGauge_gauge_callback"))) extern "C" bool FadecGauge_gauge_callback(FsContext ctx,
```
Saschl, line 5: Instead of `__attribute__((export_name("FadecGauge_gauge_callback"))) ` you could also just use `MSFS_CALLBACK`. Nothing huge, just looks a bit better :D

src/fadec/README.md
```diff
@@ -0,0 +1,9 @@
+# Engine & FADEC - A32NX/ A380X Build Options
+
+By default, the FADEC code will be built for the A32NX aircraft. To build for the A380X, you will have to edit the build.sh file accordingly:
+
+A380X Option:
+Line 43:   "${DIR}/a380_fadec/src/FadecGauge.cpp"
```
Saschl, line 6: Maybe it would be useful to parameterize this in the build script, so they can be both built by CI in the future?

Taz5150: Agreed. But for the time being, we have to keep it this way. Once we have an a380 SimObjects source within the same repo, we can build it accordingly.

src/fadec/a380_fadec/src/FadecGauge.cpp
Taz5150: It's been a while, but I recall Andreas giving me the tip to change it long time ago. I recall it made sense, but to be honest cannot tell you why :)

.github/CHANGELOG.md
```diff
@@ -87,6 +87,7 @@
 1. [LIGHTS] Increased max brightness of cockpit displays - @frankkopp (Frank Kopp)
 1. [EFB] Improved Aircraft Presets procedures - @frankkopp (Frank Kopp)
 1. [FMGC] Make DES SPD LIM available prior to descent - @tracernz (Mike)
+1. [ENGINE] Refactor FADEC code for A32NX & A380X - @Taz5150 (TazX [Z+2]#0405)
```
tracernz, line 90: 0.9 has been and gone. Need to target 0.10. But... does this need a changelog? If it doesn't change anything user-facing then I'd say no.

tracernz (changes requested):

.github/CHANGELOG.md
Taz5150: Yes, this will be changed to 0.10 but should be there since it possibilitates the A380

Saschl (approved):

.github/CHANGELOG.md
tracernz: If it doesn't make a user visible change there's no point adding noise to the changelog. We have git history for code changes.

Taz5150: Understood and changed.

Thank you!

tracernz (approved):

PR #7574, opened 2022-10-28
Memory leak fixes & using namespace std

MarcDirven:
Fixes [#[7572]](https://github.com/flybywiresim/a32nx/issues/7572)

## Summary of Changes

- Removes unwanted namespace pollution from `using namespace std` in header files
- Removes `using namespace std` in source files
- Fixes multiple memory leaks
- Uses static initialization on the stack instead of static initialization on the heap using `new`

## References

See issue.

## Additional context

I'd suggest also using `clang`'s C++20, instead of C++14 is now used afaik. I tried using [fold expressions](https://en.cppreference.com/w/cpp/language/fold) but got a compiler error.

See issue.

## Testing instructions

Not sure, I saw the code and I improved it. Functionality wise, nothing should've changed although the following files have been changed:
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/FlyPadBackend.h#L27
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/Aircraft/AircraftPreset.h#L20
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/Pushback/Pushback.h#L20
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/Aircraft/AircraftProcedures.h#L7
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fadec/src/FadecGauge.h#L32
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fbw/src/FlightDataRecorder.cpp#L14
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fbw/src/FlyByWireInterface.cpp#L11
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fbw/src/InterpolatingLookupTable.cpp#L3
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fbw/src/ThrottleAxisMapping.cpp#L6
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fbw/src/interface/SimConnectInterface.cpp#L7
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fdr2csv/src/FlightDataRecorderConverter.cpp#L6
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fdr2csv/src/main.cpp#L14
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/fdr2csv/src/main.cpp#L69
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/Aircraft/AircraftProcedures.h#L32 and
- https://github.com/flybywiresim/a32nx/blob/660b3bbc7e68c1a22a74bf1f9b1210b00895beec/src/flypad-backend/src/Aircraft/AircraftProcedures.h#L227

### DIFF
--- a/src/flypad-backend/src/Aircraft/AircraftPreset.cpp
+++ b/src/flypad-backend/src/Aircraft/AircraftPreset.cpp
@@ -38,7 +38,7 @@ void AircraftPreset::onUpdate(double deltaTime) {
     // needs to be initialized
     if (!loadingIsActive) {
       // check if procedure ID exists
-      vector<ProcedureStep*>* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
+      const auto* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
       if (requestedProcedure == nullptr) {
         std::cout << "FLYPAD_BACKEND: Preset " << loadAircraftPresetRequest << " not found!"
                   << std::endl;
--- a/src/flypad-backend/src/Aircraft/AircraftPreset.h
+++ b/src/flypad-backend/src/Aircraft/AircraftPreset.h
@@ -17,8 +17,6 @@
 #include "AircraftProcedures.h"
 #include "../Units.h"
 
-using namespace std;
-
 /**
  * Class for handling aircraft presets.
  */
@@ -42,7 +40,7 @@ private:
   // current procedure ID
   int64_t currentProcedureID = 0;
   // current procedure
-  vector<struct ProcedureStep*>* currentProcedure = nullptr;
+  const std::vector<const ProcedureStep*>* currentProcedure = nullptr;
   // flag to signal that a loading process is ongoing
   bool loadingIsActive = false;
   // in ms
--- a/src/flypad-backend/src/Aircraft/AircraftProcedures.h
+++ b/src/flypad-backend/src/Aircraft/AircraftProcedures.h
@@ -3,8 +3,7 @@
 #include <map>
 #include <string>
 #include <vector>
-
-using namespace std;
+#include <array>
 
 struct ProcedureStep {
   std::string description;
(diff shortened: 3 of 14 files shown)


src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -29,272 +28,264 @@ struct ProcedureStep {
...
-  vector<ProcedureStep*>* readyForTakeoff = new vector<ProcedureStep*>;
-  // just to get a list of IDs
-  vector<ProcedureStep*>* all = new vector<ProcedureStep*>;
+  std::vector<const ProcedureStep*> coldAndDark;
+  std::vector<const ProcedureStep*> powered;
+  std::vector<const ProcedureStep*> readyForPushback;
+  std::vector<const ProcedureStep*> readyForTaxi;
+  std::vector<const ProcedureStep*> readyForTakeoff;
+
+  void printProcedure(const std::vector<const ProcedureStep*>& procedures) {
```
Gurgel100, line 232: Why does this need to be a member function? I would make it an static member function.
Also I think this function has the wrong signature: Further down the function is called with a `const vector<const ProcedureStep>&` but the signature requires a `std::vector<const ProcedureStep*>&`

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -29,272 +28,264 @@ struct ProcedureStep {
...
+    std::for_each(begin(procedures), end(procedures), [](const auto* procedure) {
+        std::cout << procedure->id << " = " << procedure->description << std::endl;
+    });
+  }
+
+  void doInsertProcedures(std::vector<const ProcedureStep*>& dest, const std::vector<ProcedureStep>& src) {
```
Gurgel100, line 238: Same here: Why not static member function?

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -29,272 +28,264 @@ struct ProcedureStep {
...
+    std::transform(begin(src), end(src), back_inserter(dest), [](const auto& procedure) {
+      return &procedure;
+    });
+  }
+
+    void insertProcedures(std::vector<const ProcedureStep*>& dest, const std::vector<ProcedureStep>& a, const std::vector<ProcedureStep>& b,
```
Gurgel100, line 244: Same here: Why not static member function? And also this function would be more general if it would accept an unspecified number of arguments instead of a fixed number of vectors (lookup variadic functions or you can use std::initializer_list because all have the same type).

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -29,272 +28,264 @@ struct ProcedureStep {
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 // @formatter:off
-static vector<ProcedureStep*>* POWERED_CONFIG_ON = new vector<ProcedureStep*>{
+static const std::vector<ProcedureStep> POWERED_CONFIG_ON = {
```
Gurgel100, line 31: An even better solution would be to use `std::array` instead of `std::vector` also because you already added the include but never used it. Also it would be better to move the declaration inside of the `AircraftProcedures` class to make it "private"

src/flypad-backend/src/Aircraft/AircraftPreset.cpp
```diff
@@ -38,7 +38,7 @@ void AircraftPreset::onUpdate(double deltaTime) {
     // needs to be initialized
     if (!loadingIsActive) {
       // check if procedure ID exists
-      vector<ProcedureStep*>* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
+      const auto* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
```
Gurgel100, line 41: This variable is later assigned to the member `currentProcedure` which would remove constness so you also need to change the type of `currentProcedure`.

Gurgel100 (changes requested):
I like these changes. Just some small improvements. Did you compile? Because I think the current changes will not compile.

src/flypad-backend/src/Aircraft/AircraftPreset.cpp
MarcDirven: That is already the case otherwise it wouldn't compile. See [your comment here](https://github.com/flybywiresim/a32nx/pull/7574#discussion_r1008314075)

MarcDirven:
Gonna take a look at the feedback tomorrow!

src/flypad-backend/src/Aircraft/AircraftPreset.cpp
Gurgel100: Yes you're right.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
MarcDirven: Initialiazer list would be possible, however I think it's cleaner to use variadic template arguments. I've added a new push, let me know what you think.

MarcDirven:
Alright, I got the chance actually to process the feedback now. I've also added `-std=c++17` for the clang compiler, in order to use structured bindings.

Gurgel100 (changes requested):

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
```diff
@@ -0,0 +1,29 @@
+#include "AircraftProcedures.h"
+
+#include <iostream>
+#include <algorithm>
```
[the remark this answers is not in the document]
tracernz: Why? Only need the definitions in the public interface if it makes use of them.

src/fdr2csv/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(fdr2csv LANGUAGES C CXX)
 
-set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
MarcDirven: It's added at the bottom. Globally setting the Cmake variables isn't always considered good practice.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -15,244 +15,238 @@ struct ProcedureStep {
...
 
-    void insertProcedures(std::vector<const ProcedureStep*>& dest, const std::vector<ProcedureStep>& a, const std::vector<ProcedureStep>& b,
-                          const std::vector<ProcedureStep>& c, const std::vector<ProcedureStep>& d) {
-    dest.reserve(a.size() + b.size() + c.size() + d.size());
-    for (const auto& p : {a, b, c, d}) {
-      doInsertProcedures(dest, p);
-    }
+  template<std::size_t I, class... Arrays>
+  static constexpr void insertProcedures(std::array<const ProcedureStep*, I>& dest, const Arrays&... procedures) {
+    (doInsertProcedures(dest, procedures), ...);
```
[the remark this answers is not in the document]
MarcDirven: Sharp eyes, totally didn't see that

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -15,244 +15,238 @@ struct ProcedureStep {
...
   }
 
 public:
-  AircraftProcedures() {
+  constexpr AircraftProcedures() {
```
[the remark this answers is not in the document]
MarcDirven: It can be constexpr if debug isn't defined, so I'll just add that

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace {
+  void printProcedure(const std::vector<const ProcedureStep*>& procedures) {
+    for (const auto& p : procedures) {
+      std::cout << p->id << " = " << p->description << std::endl;
+    }
+  }
+}
+
+std::pair<const ProcedureStep*, const ProcedureStep*> AircraftProcedures::getProcedure(int64_t pID) const {
```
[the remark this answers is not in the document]
MarcDirven: It wouldn't make a lot of sense to make these const as the context wherein it is called will just be making a copy of these pointers.

src/flypad-backend/src/Aircraft/AircraftPreset.h
```diff
@@ -35,12 +35,12 @@ class AircraftPreset {
...
   // Procedures
-  AircraftProcedures procedures{};
+  AircraftProcedures procedures;
 
   // current procedure ID
   int64_t currentProcedureID = 0;
-  // current procedure
-  const std::vector<const ProcedureStep*>* currentProcedure = nullptr;
+  // current procedure, containing a pointer to the begin of the array and to the end of the array
+  std::pair<const ProcedureStep*, const ProcedureStep*> currentProcedure = std::make_pair(nullptr, nullptr);
```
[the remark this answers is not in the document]
MarcDirven: This isn't possible because then currentProcedure can't be assigned due to its constness.

src/fdr2csv/CMakeLists.txt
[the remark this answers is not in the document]
Gurgel100: Ok I didn't know that.

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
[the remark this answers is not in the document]
Gurgel100: Sure but as far as I see the algorithm used is `std::transform` which is called in the header file.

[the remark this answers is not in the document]
MarcDirven: It wasn't at first, but now it is since the function that uses it has become a template function.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -3,7 +3,10 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <algorithm>
```
Gurgel100, line 6: Can you cleanup the includes here? For example map is not used, etc.

MarcDirven: AFAIK map was already included, but I'll take a look

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -233,21 +236,34 @@ class AircraftProcedures {
   std::array<const ProcedureStep*, RDY_FOR_TAKEOFF_SIZE> readyForTakeoff;
 
   template<std::size_t N1, std::size_t N2>
-  static constexpr void doInsertProcedures(std::array<const ProcedureStep*, N1>& dest, const std::array<ProcedureStep, N2>& src) {
-    std::transform(begin(src), end(src), begin(dest), [](const auto& procedure) {
+  static constexpr void doInsertProcedures(std::array<const ProcedureStep*, N1>& dest, const std::array<ProcedureStep, N2>& src, const std::size_t offset) {
+    std::transform(begin(src), end(src), begin(dest) + offset * N2, [](const auto& procedure) {
```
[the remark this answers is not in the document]
MarcDirven: You want the behaviour to be like this right? https://godbolt.org/z/PKqEGGcfb If yes, then this implementation is correct.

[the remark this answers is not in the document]
MarcDirven: Don't be sorry, I'll fix it, thought this was the right implementation.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
Gurgel100: Yes. But if you already are changing the codestyle I think a little cleanup doesn't hurt

src/flypad-backend/src/Aircraft/AircraftProcedures.h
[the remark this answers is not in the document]
MarcDirven: Fixed it, here's my test: https://godbolt.org/z/Ehrqjfo85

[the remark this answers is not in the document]
MarcDirven: Don't be sorry, I'll fix it. Just made some commits but those aren't done yet.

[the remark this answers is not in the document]
MarcDirven: Just pushed it, [tested it here](https://godbolt.org/z/86dPaednc) this is the intended behaviour right?

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -237,29 +238,47 @@ class AircraftProcedures {
...
+  template<std::size_t... I>
+  static constexpr void doInsertProcedures(auto& dest, const auto& sources, const auto& offsets, std::index_sequence<I...>) {
+    (doInsertProcedures(dest, std::get<I>(sources), offsets[I]), ...);
+  }
+
+  template<class Tuple, std::size_t... I>
+  static constexpr std::array<std::size_t, sizeof...(I) + 1> getOffsets(const Tuple& procedures, std::index_sequence<I...>) {
+    std::array sizes = {(std::size_t)0, std::get<I>(procedures).size()...};
+    std::array<size_t, sizeof...(I) + 1> offsets{};
+    std::inclusive_scan(begin(sizes), end(sizes), begin(offsets));
```
Gurgel100, line 260: What you implement here is essentially `std::exclusive_scan`:
```suggestion
static constexpr std::array<std::size_t, sizeof...(I)> getOffsets(const Tuple& procedures, std::index_sequence<I...>) {
    std::array sizes = {std::get<I>(procedures).size()...};
    std::array<size_t, sizeof...(I)> offsets{};
    std::exclusive_scan(begin(sizes), end(sizes), begin(offsets), 0);
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -237,29 +238,47 @@ class AircraftProcedures {
...
   }
 
   template<std::size_t I, class... Arrays>
   static constexpr void insertProcedures(std::array<const ProcedureStep*, I>& dest, const Arrays&... procedures) {
     constexpr std::size_t accumulatedSize = (... + std::tuple_size<Arrays>::value);
     static_assert(I == accumulatedSize, "Accumulated array sizes do not match. Append more arrays to match size of dest.");
-    doInsertProcedures(dest, std::make_tuple(procedures...), std::make_index_sequence<sizeof...(Arrays)>());
+
+    const auto tupleProcedures = std::make_tuple(procedures...);
+    auto offsets = getOffsets(tupleProcedures, std::make_index_sequence<sizeof...(Arrays) - 1>());
```
Gurgel100, line 270:
```suggestion
    auto offsets = getOffsets(tupleProcedures, std::make_index_sequence<sizeof...(Arrays)>());
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -237,29 +238,47 @@ class AircraftProcedures {
...
-    std::transform(begin(src), end(src), begin(dest) + offset * N2, [](const auto& procedure) {
+    std::transform(begin(src), end(src), begin(dest) + offset, [](const auto& procedure) {
       return &procedure;
     });
   }
 
   template<class Tuple, std::size_t S, std::size_t... I>
   static constexpr void doInsertProcedures(std::array<const ProcedureStep*, S>& dest, const Tuple& procedures, std::index_sequence<I...>) {
-    (doInsertProcedures(dest, std::get<I>(procedures), I), ...);
+    (doInsertProcedures(dest, std::get<I>(procedures), (I == 0 ? 0 : I - 1) * std::get<(I == 0 ? 0 : I - 1)>(procedures).size()), ...);
```
Gurgel100, line 248: Also this overload is never called and I think has not the correct semantics so I suggest to remove it

Gurgel100 (changes requested):
It's getting there

src/flypad-backend/src/Aircraft/AircraftProcedures.h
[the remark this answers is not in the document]
Gurgel100: Seems rights yes.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
MarcDirven: Oh whoops that's still my little sandbox environment

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -1,13 +1,16 @@
...
+#ifdef DEBUG
+#include <iostream>
+#endif
+#include <array>
+#include <span>
+#include <numeric>
 
 struct ProcedureStep {
-  std::string description;
+  const char* description;
```
frankkopp, line 13: If you are trying to make things more cpp-like I do not understand this change - I'd prefer to keep the std::string
That counts for all your changes from std:string to char*

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -16,285 +19,284 @@ struct ProcedureStep {
...
+  ProcedureStep{"BAT1 On",                  1010, false, 1000, "(L:A32NX_OVHD_ELEC_BAT_1_PB_IS_AUTO)",                 "1 (>L:A32NX_OVHD_ELEC_BAT_1_PB_IS_AUTO)"},
+  ProcedureStep{"BAT2 On",                  1020, false, 3000, "(L:A32NX_OVHD_ELEC_BAT_2_PB_IS_AUTO)",                 "1 (>L:A32NX_OVHD_ELEC_BAT_2_PB_IS_AUTO)"},
 
-  new ProcedureStep {"EXT PWR On",               1030, false, 3000, "(L:A32NX_OVHD_APU_MASTER_SW_PB_IS_ON) "
+  ProcedureStep{"EXT PWR On",               1030, false, 3000, "(L:A32NX_OVHD_APU_MASTER_SW_PB_IS_ON) "
                                                                     "(L:A32NX_OVHD_APU_MASTER_SW_PB_IS_ON) && "
                                                                     "(L:A32NX_ENGINE_STATE:1) 1 == || "
                                                                     "(L:A32NX_ENGINE_STATE:2) 1 == || "
                                                                     "(A:EXTERNAL POWER ON:1, BOOL) ||",                     "(A:EXTERNAL POWER ON:1, BOOL) ! if{ 1 (>K:TOGGLE_EXTERNAL_POWER) }"},
```
frankkopp, line 43: Please fix the formatting of this file - the line identation and spacing was aligned for a reason

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -16,285 +19,284 @@ struct ProcedureStep {
...
+  template<std::size_t... I>
+  static constexpr void doInsertProcedures(auto& dest, const auto& sources, const auto& offsets, std::index_sequence<I...>) {
+    (doInsertProcedures(dest, std::get<I>(sources), offsets[I]), ...);
+  }
 
-    readyForPushback->insert(readyForPushback->end(), TAKEOFF_CONFIG_OFF->begin(), TAKEOFF_CONFIG_OFF->end());
-    readyForPushback->insert(readyForPushback->end(), TAXI_CONFIG_OFF->begin(), TAXI_CONFIG_OFF->end());
-    readyForPushback->insert(readyForPushback->end(), POWERED_CONFIG_ON->begin(), POWERED_CONFIG_ON->end());
-    readyForPushback->insert(readyForPushback->end(), PUSHBACK_CONFIG_ON->begin(), PUSHBACK_CONFIG_ON->end());
+  template<class Tuple, std::size_t... I>
```
frankkopp, line 251: Please revert these - these are much more than simple improvements. Especially these templates are much less readable. Not everything needs to be complicated sophisticated cpp code for no good reason.

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
```diff
@@ -0,0 +1,19 @@
+#include "AircraftProcedures.h"
+
+
+std::span<const ProcedureStep*> AircraftProcedures::getProcedure(int64_t pID) const {
```
frankkopp, line 4: not sure adding a class for the procedures is an improvement

frankkopp (changes requested):
Concerning flypad-backend:
Some changes - esp. namespace in header file fix - are good. 
Most changes are unnecessary and no real improvement. Not sure we should change other people's code just because of personal preferences. 
Especially as the current code is tested and running well we should not risk regression without a very good reason. And making code more complicated (e.g. adding templates) usually doesn't improve the code a lot. I deliberately did not use them. 
Please consider removing these changes and only do improvements which are indeed objectively better (e.g. namespace). You could create a new PR with these other changes and/or let's discuss on Discord (Cdr_Maverick#6475) - I'm always happy to improve my code but it actually should be an improvement and it should have a good reason.

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
MarcDirven: You can't return an `std::array`, the only way to return a `std::array` is to return a `std::span`, or a `std::pair` containing the beginning of the array and the ending of the array (or the size). Returning a `std::span` is more readable than returning a `std::pair` thus `std::span` it is. Also, it's part of the standard library so technically, I'm not adding anything.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
MarcDirven: This way you can make every `std::array` `constexpr`. AFAIK `std::string` is `constexpr` in C++20, but at the time of development, I was still using C++17.

MarcDirven:
It's not clear to me what you would like to see differently, concretly. I've sent you a friend request on Discord.

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
frankkopp: Bad comment from my side - I meant adding a cpp file just for this

frankkopp: Bad comment from my side - I meant adding a cpp file just for this

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -16,285 +19,284 @@ struct ProcedureStep {
...
-
-    std::for_each(all->begin(), all->end(), [&](ProcedureStep* item) {
-      std::cout << item->id << " = " << item->description << std::endl;
+  static constexpr std::size_t COLD_DARK_SIZE = TAKEOFF_CONFIG_OFF.size() + TAXI_CONFIG_OFF.size() + PUSHBACK_CONFIG_OFF.size() + POWERED_CONFIG_OFF.size();
+  static constexpr std::size_t POWERED_SIZE = TAKEOFF_CONFIG_OFF.size() + TAXI_CONFIG_OFF.size() + PUSHBACK_CONFIG_OFF.size() + POWERED_CONFIG_ON.size();
+  static constexpr std::size_t RDY_FOR_PUSHBACK_SIZE = TAKEOFF_CONFIG_OFF.size() + TAXI_CONFIG_OFF.size() + POWERED_CONFIG_ON.size() + PUSHBACK_CONFIG_ON.size();
+  static constexpr std::size_t RDY_FOR_TAXI_SIZE = TAKEOFF_CONFIG_OFF.size() + POWERED_CONFIG_ON.size() + PUSHBACK_CONFIG_ON.size() + TAXI_CONFIG_ON.size();
+  static constexpr std::size_t RDY_FOR_TAKEOFF_SIZE = POWERED_CONFIG_ON.size() + PUSHBACK_CONFIG_ON.size() + TAXI_CONFIG_ON.size() + TAKEOFF_CONFIG_ON.size();
+
+  std::array<const ProcedureStep*, COLD_DARK_SIZE> coldAndDark;
```
frankkopp, line 233: Rather keep these a vector - it is planned to maybe move the procedure config to a file which is read at runtime.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -16,285 +19,284 @@ struct ProcedureStep {
...
+  static constexpr void doInsertProcedures(auto& dest, const auto& sources, const auto& offsets, std::index_sequence<I...>) {
+    (doInsertProcedures(dest, std::get<I>(sources), offsets[I]), ...);
+  }
 
-    readyForPushback->insert(readyForPushback->end(), TAKEOFF_CONFIG_OFF->begin(), TAKEOFF_CONFIG_OFF->end());
-    readyForPushback->insert(readyForPushback->end(), TAXI_CONFIG_OFF->begin(), TAXI_CONFIG_OFF->end());
-    readyForPushback->insert(readyForPushback->end(), POWERED_CONFIG_ON->begin(), POWERED_CONFIG_ON->end());
-    readyForPushback->insert(readyForPushback->end(), PUSHBACK_CONFIG_ON->begin(), PUSHBACK_CONFIG_ON->end());
+  template<class Tuple, std::size_t... I>
+  static constexpr std::array<std::size_t, sizeof...(I)> getOffsets(const Tuple& procedures, std::index_sequence<I...>) {
```
frankkopp, line 252: It is really not required to have all constexpr - there is nearly no runtime cost and it is not worth to create such complicated code just to be constexpr.

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -16,285 +19,284 @@ struct ProcedureStep {
...
+    std::exclusive_scan(begin(sizes), end(sizes), begin(offsets), 0);
+    return offsets;
+  }
 
-    readyForTaxi->insert(readyForTaxi->end(), TAKEOFF_CONFIG_OFF->begin(), TAKEOFF_CONFIG_OFF->end());
-    readyForTaxi->insert(readyForTaxi->end(), POWERED_CONFIG_ON->begin(), POWERED_CONFIG_ON->end());
-    readyForTaxi->insert(readyForTaxi->end(), PUSHBACK_CONFIG_ON->begin(), PUSHBACK_CONFIG_ON->end());
-    readyForTaxi->insert(readyForTaxi->end(), TAXI_CONFIG_ON->begin(), TAXI_CONFIG_ON->end());
+  template<std::size_t I, class... Arrays>
+  static constexpr void insertProcedures(std::array<const ProcedureStep*, I>& dest, const Arrays&... procedures) {
```
frankkopp, line 260: Same here

frankkopp:
> It's not clear to me what you would like to see differently, concretely. I've sent you a friend request on Discord. Also, regression shouldn't be an argument to prevent code from optimized/modernized since you'd normally have tests for that. I agree though having heavily templated code isn't always as readable.

I agree with the namespace and removing "new" changes. 
Also using array for the procedure is somewhat ok - although not a huge improvement. 

But please remove all attempts to make stuff constexpr by adding complicated templates and char* for a class the is instantiated once when loading the aircraft. It is just not worth it.

Look at this code just to save on one single instantiation of a rather simple class?
```
 template<std::size_t N1, std::size_t N2>
  static constexpr void doInsertProcedures(std::array<const ProcedureStep*, N1>& dest, const std::array<ProcedureStep, N2>& src, const std::size_t offset) {
    std::transform(begin(src), end(src), begin(dest) + offset, [](const auto& procedure) {
      return &procedure;
    });
  }

  template<std::size_t... I>
  static constexpr void doInsertProcedures(auto& dest, const auto& sources, const auto& offsets, std::index_sequence<I...>) {
    (doInsertProcedures(dest, std::get<I>(sources), offsets[I]), ...);
  }

  template<class Tuple, std::size_t... I>
  static constexpr std::array<std::size_t, sizeof...(I)> getOffsets(const Tuple& procedures, std::index_sequence<I...>) {
    std::array sizes = {std::get<I>(procedures).size()...};
    std::array<size_t, sizeof...(I)> offsets{};
    std::exclusive_scan(begin(sizes), end(sizes), begin(offsets), 0);
    return offsets;
  }

  template<std::size_t I, class... Arrays>
  static constexpr void insertProcedures(std::array<const ProcedureStep*, I>& dest, const Arrays&... procedures) {
    constexpr std::size_t accumulatedSize = (... + std::tuple_size<Arrays>::value);
    static_assert(I == accumulatedSize, "Accumulated array sizes do not match. Append more arrays to match size of dest.");

    const auto tupleProcedures = std::make_tuple(procedures...);
    auto offsets = getOffsets(tupleProcedures, std::make_index_sequence<sizeof...(Arrays)>());
    doInsertProcedures(dest, tupleProcedures, offsets, std::make_index_sequence<sizeof...(Arrays)>());
  }

#ifdef DEBUG
  template<std::size_t S>
  void printProcedure(const std::array<ProcedureStep, S>& procedures) {
    for (const auto& p : procedures) {
      std::cout << p.id << " = " << p.description << std::endl;
    }
  }
#endif
```

MarcDirven:
Hey, I just saw your comments, will be working on it somewhere this week.

src/flypad-backend/src/Aircraft/AircraftProcedures.cpp
MarcDirven: Ah check, yes agreed. I sort of forgot about because during development, there used to be more functions in the cpp/h files which later on have been (re)moved

src/flypad-backend/src/Aircraft/AircraftProcedures.h
MarcDirven: That's pretty odd, I think my `clang-format` messed things up here.

frankkopp: You need to exclude this file and format manually if you are not using an editor which respects `// @formatter:off` (IDEA CLION does).
And the no format parts need to be done manually anyway. 
It is somewhat a compromise of readability and not having too long lines.

src/flypad-backend/src/Aircraft/AircraftPreset.cpp
```diff
@@ -38,8 +38,8 @@ void AircraftPreset::onUpdate(double deltaTime) {
     // needs to be initialized
     if (!loadingIsActive) {
       // check if procedure ID exists
-      vector<ProcedureStep*>* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
-      if (requestedProcedure == nullptr) {
+      const std::vector<const ProcedureStep*>* requestedProcedure = procedures.getProcedure(loadAircraftPresetRequest);
+      if (requestedProcedure->empty()) {
```
frankkopp, line 42: This would create a runtime error if requestedProcedure is nullptr.
getProcedure will return a nullptr when loadAircraftPresetRequest != 1|2|3|4|5

```suggestion
      if (requestedProcedure == nullptr) {
```

src/flypad-backend/src/Aircraft/AircraftPreset.h
```diff
@@ -37,12 +35,12 @@ class AircraftPreset {
...
   // Procedures
-  AircraftProcedures procedures{};
+  AircraftProcedures procedures;
 
   // current procedure ID
   int64_t currentProcedureID = 0;
-  // current procedure
-  vector<struct ProcedureStep*>* currentProcedure = nullptr;
+  // current procedure, containing a pointer to the begin of the array and to the end of the array
+  const std::vector<const ProcedureStep*>* currentProcedure = nullptr;
```
frankkopp, line 43:
```suggestion
  // current procedure
  const std::vector<const ProcedureStep*>* currentProcedure = nullptr;
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -1,10 +1,12 @@
...
-#include <string>
 #include <vector>
-
-using namespace std;
+#include <algorithm>
+#ifdef DEBUG
+#include <iostream>
+#endif
+#include <array>
+#include <string>
```
frankkopp, line 9:
```suggestion
#include <string>
#ifdef DEBUG
#include <iostream>
#endif
```

Remove array - not needed any more, right?

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
-  // as these are only on Exp this first iteration uses A32NX_FWC_FLIGHT_PHASE which also work on
-  // master and is equivalent for this specific purpose. Will be changed when the FWC is on master.
-  new ProcedureStep {"Waiting on FWC Initialization", 1065, true,  5000, "",                                                "(L:A32NX_FWC_FLIGHT_PHASE)"},
-  new ProcedureStep {"Waiting...",                    9999, false, 5000, "(L:A32NX_AIRCRAFT_PRESET_FWC_INIT_DONE)",         "1 (>L:A32NX_AIRCRAFT_PRESET_FWC_INIT_DONE)"},
-
-  // APU fire test
-  new ProcedureStep {"APU Fire Test On",         1035, false, 2000, "(L:A32NX_AIRCRAFT_PRESET_FIRE_TEST_APU_DONE)",         "1 (>L:A32NX_FIRE_TEST_APU)"},
-  new ProcedureStep {"APU Fire Test Off",        1036, false, 2000, "(L:A32NX_AIRCRAFT_PRESET_FIRE_TEST_APU_DONE)",         "0 (>L:A32NX_FIRE_TEST_APU) 1 (>L:A32NX_AIRCRAFT_PRESET_FIRE_TEST_APU_DONE)"},
+class AircraftProcedures {
+  // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
frankkopp, line 28:
```suggestion

  // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
+  // Please remember to also update the EFB Presets page for the step description
+  // if you make any changes to this list.
+  // src/instruments/src/EFB/Presets/Widgets/AircraftPresets.tsx
+  // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+  static const inline std::vector POWERED_CONFIG_ON {
```
frankkopp, line 33:
```suggestion

  static const inline std::vector POWERED_CONFIG_ON {
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
+  std::vector<const ProcedureStep*> readyForPushback;
+  std::vector<const ProcedureStep*> readyForTaxi;
+  std::vector<const ProcedureStep*> readyForTakeoff;
+
+  static void insert(std::vector<const ProcedureStep*>& dest, const std::vector<ProcedureStep>& src) {
+    std::transform(begin(src), end(src), back_inserter(dest), [](const auto& procedure) {
+      return &procedure;
+    });
+  }
+
```
frankkopp, line 238:
```suggestion
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
-// @formatter:on
 
-class AircraftProcedures {
-  vector<ProcedureStep*>* coldAndDark = new vector<ProcedureStep*>;
-  vector<ProcedureStep*>* powered = new vector<ProcedureStep*>;
-  vector<ProcedureStep*>* readyForPushback = new vector<ProcedureStep*>;
-  vector<ProcedureStep*>* readyForTaxi = new vector<ProcedureStep*>;
-  vector<ProcedureStep*>* readyForTakeoff = new vector<ProcedureStep*>;
-  // just to get a list of IDs
-  vector<ProcedureStep*>* all = new vector<ProcedureStep*>;
```
frankkopp, line 226:
```suggestion
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
+
+
+#ifdef DEBUG
+  template<std::size_t S>
+  void printProcedure(const std::vector<ProcedureStep, S>& procedures) {
+    for (const auto& p : procedures) {
+      std::cout << p.id << " = " << p.description << std::endl;
+    }
+  }
+#endif
```
frankkopp, line 247:
```suggestion
#endif

```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
 public:
   AircraftProcedures() {
-    // Map the procedure groups
-
 #ifdef DEBUG
```
frankkopp, line 250:
```suggestion

#ifdef DEBUG
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
-    all->insert(all->end(), TAKEOFF_CONFIG_OFF->begin(), TAKEOFF_CONFIG_OFF->end());
-    all->insert(all->end(), TAXI_CONFIG_OFF->begin(), TAXI_CONFIG_OFF->end());
-    all->insert(all->end(), PUSHBACK_CONFIG_OFF->begin(), PUSHBACK_CONFIG_OFF->end());
-    all->insert(all->end(), POWERED_CONFIG_OFF->begin(), POWERED_CONFIG_OFF->end());
-
-    std::for_each(all->begin(), all->end(), [&](ProcedureStep* item) {
-      std::cout << item->id << " = " << item->description << std::endl;
-    });
+    // Map the procedure groups
+    // P{rint to console to add them to the EFB code to display the current step.
```
frankkopp, line 252:
```suggestion
    // Print to console to add them to the EFB code to display the current step.
```

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
+    printProcedure(POWERED_CONFIG_ON);
+    printProcedure(PUSHBACK_CONFIG_ON);
+    printProcedure(TAXI_CONFIG_ON);
+    printProcedure(TAKEOFF_CONFIG_ON);
+    printProcedure(TAKEOFF_CONFIG_OFF);
+    printProcedure(TAXI_CONFIG_OFF);
+    printProcedure(PUSHBACK_CONFIG_OFF);
+    printProcedure(POWERED_CONFIG_OFF);
 #endif
```
frankkopp, line 261:
```suggestion
#endif

```

frankkopp (changes requested):
Thanks for the changes. 
One potential run time error and some small things like formatting, typos.
Once fixed I'm good with these changes!

src/flypad-backend/src/Aircraft/AircraftProcedures.h
```diff
@@ -22,279 +24,282 @@ struct ProcedureStep {
...
+  }
+
+
+#ifdef DEBUG
+  template<std::size_t S>
+  void printProcedure(const std::vector<ProcedureStep, S>& procedures) {
+    for (const auto& p : procedures) {
+      std::cout << p.id << " = " << p.description << std::endl;
+    }
+  }
```
frankkopp, line 246:
```suggestion
  static inline void printProcedure(const std::vector<ProcedureStep>& procedures) {
    for (const auto& p : procedures) {
      std::cout << p.id << " = " << p.description << std::endl;
    }
  }
```

MarcDirven:
All done I think!

Gurgel100 (approved):
LGTM.

frankkopp (approved):
Thanks for making the changes!

FlightSimmer7700:
Quality Assurance Tester

Discord: FlightSimmer7700#5438
Object of testing: #7574 
Tier of Testing: 1
Date: 04/11/22

Testing Process: Completed a full flight and used efb throughout the flight (have not used aircraft in quite some time so may have missed things since Im not so familiar)
Negatives: None
Testing Results: Passed

Conclusions: LGTM great work

frankkopp:
@MarcDirven can you please rebase/update this branch before it can get merged

MarcDirven:
Hi! Yes sorry I haven't had checked my messages, will do somewhere today/this week.

PR #7590, opened 2022-11-08
chore: strip debug info from systems release build

Saschl:
Fixes #[issue_no]

## Summary of Changes

This needs https://github.com/flybywiresim/a32nx/pull/7588 to be merged as the current dev-env's rust version is too old for that feature.

Added strip = true (https://doc.rust-lang.org/cargo/reference/profiles.html#strip) option to remove debug symbols.

This makes:
 a) MSFS happy as it does not detect it as DEBUG build anymore (see screenshot).
 b) Reduce the size wasm file to 709kb
 c) Reduces wasm compile time in MSFS from ~150 seconds to ~53 seconds (i7-10700k)

Note: I had to reduce wasm-opt to level O1, everything else crashed the sim but I think the benefits are higher now.

## Screenshots (if necessary)

Before:

systems_before

After:

wasm_after

## Additional context

## Testing instructions

Perform a flight and look for any abnormal behavior in systems (hydraulics, electrics, bleed...)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,4 @@ members = [
 
 [profile.release]
 lto = true
+strip = true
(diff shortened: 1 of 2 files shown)


Saschl:
Was included in dev-env update

PR #7585, opened 2022-11-06
fix: pass an id to un/focus_input_field

tracernz:
Fixes #[issue_no]

## Summary of Changes

These methods are supposed to be provided with an id argument, per `fs-base-ui\html_ui\Templates\uiInput\uiInput.js`. Add the ID to avoid any issues.

## Additional context

Discord username (if different from GitHub):

## Testing instructions

Test keyboard input on the EFB (calculators etc.), and MCDU keyboard input.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #7595, opened 2022-11-11
fix: init order for scratchpad

tracernz:
Fixes #[issue_no]

## Summary of Changes

Rookie mistake in #7585... calling this.scratchpad.guid before this.scratchpad is constructed. I also added the rest of the params for the FOCUS_INPUT_FIELD calls just in case.

## Additional context


Saschl (approved):
Good catch. Should've seen it during review 😢

PR #7597, opened 2022-11-12
fix(fms): various small bugs or erroneous warnings

tracernz:
Fixes #[issue_no]

## Summary of Changes

Tiny bugfixes for issues picked up by Sentry. 
- There's bad ILS approach navdata out there so we better handle it gracefully.
- Since VFR map is using the new nearest facility API we get it's search results, not a bug.
- Try to track which facilities are failing to load.
- Handle missing f-pln (not sure of caller).
- Don't log an error for the user aborting at the new waypoint page after NOT IN DATABASE.

## Additional context


src/fmgc/src/flightplanning/LegsProcedure.ts
```diff
@@ -232,7 +232,7 @@ export class LegsProcedure {
                   mappedLeg.speedConstraint = currentLeg.speedRestriction;
                   mappedLeg.turnDirection = currentLeg.turnDirection;
 
-                  const recNavaid: RawVor | RawNdb | undefined = this._facilities.get(currentLeg.originIcao);
+                  const recNavaid: RawVor | RawNdb | undefined = this._facilities.get(currentLeg.originIcao) as RawVor | RawNdb | undefined;
```
tracernz, line 235: Skipped this one as it's not an error for this to be missing.

Saschl (approved):
LGTM

PR #7611, opened 2022-11-23
fix(simbridge): use old type for fetch

tracernz:
Fixes #[issue_no]

## Summary of Changes

Some people are still using old typings which don't have `| URL`. This seems to make both happy I think anyway.

## Additional context


Saschl (approved):
Ayup

PR #7625, opened 2022-11-29
feat(hyd): added priority valves

crocket63:
Fixes #[issue_no]

## Summary of Changes

This PR adds priority valves in the hydraulic circuits. Non essential consumers will lose pressure when system pressure is too low.

## Additional context

## Testing instructions

Try retracting gear in flight using only yellow epump + ptu (so all edp off).
Pressures shall remain above 1700ish, and gear retraction might fail.

Try using flaps/slats in only blue circuit available scenario.
Pressures should not drop below 1700psi to keep flight controls alive.


.github/CHANGELOG.md
```diff
@@ -130,6 +130,7 @@
 1. [EFB] Added deboarding button to flyPad Payload - @frankkopp (Frank Kopp)
 1. [EFB] Improved simbridge-client connection handling - @frankkopp (Frank Kopp)
 1. [HYD] Tilting gear mechanism - @Crocket63 (crocket)
+1. [HYD] Added priority valves - @Crocket63 (crocket)
```
BlueberryKing, line 133: I think this should go under 0.10.0

src/systems/systems/src/hydraulic/mod.rs
```diff
@@ -1891,6 +1911,65 @@ impl CheckValve {
...
+        self.upstream_pressure = upstream_pressure;
+
+        self.update_open_state(context);
+
+        self.update_downstream_pressure();
+    }
+
+    fn update_open_state(&mut self, context: &UpdateContext) {
+        let opening_ratio = if self.upstream_pressure > self.fully_opened_threshold {
+            Ratio::new::<ratio>(1.)
```
BlueberryKing, line 1948: Do we need to cover this case separately? I feel like this is covered by the `.min` below.

src/systems/systems/src/hydraulic/mod.rs
```diff
@@ -1891,6 +1911,65 @@ impl CheckValve {
...
+        let current_open_state = self.open_ratio.output();
+
+        self.downstream_pressure = self.upstream_pressure * current_open_state * current_open_state;
+    }
+
+    fn downstream_pressure(&self) -> Pressure {
+        self.downstream_pressure
+    }
+}
+impl SimulationElement for PriorityValve {}
```
BlueberryKing, line 1971: I don't think this is necessary.

```suggestion
```

src/systems/systems/src/hydraulic/mod.rs
```diff
@@ -1668,6 +1676,10 @@ impl Section {
         if let Some(valve) = &self.leak_measurement_valve {
             self.pressure_switch
                 .update(context, valve.downstream_pressure());
+
+            if let Some(priority_valve) = &mut self.priority_valve {
+                priority_valve.update(context, valve.downstream_pressure())
+            }
```
BlueberryKing, line 1682: Are we sure this should only be called if a leak measurement valve exists? Does this assume that there can only be a priority valve if there's a leak measurement valve?

Also, I think we can move the pressure switch update call outside of the `if let Some(valve) = &self.leak_measurement_valve` by simply updating it with `self.pressure_downstream_leak_valve()`. The if statement is already there. This might be outside of the scope of this PR though.

src/systems/systems/src/hydraulic/nose_steering.rs
```diff
@@ -155,7 +155,10 @@ impl SteeringActuator {
         pushback_tug: &impl Pushback,
     ) {
         if !pushback_tug.is_nose_wheel_steering_pin_inserted() {
-            self.update_max_speed(context, section_pressure.pressure());
+            self.update_max_speed(
+                context,
+                section_pressure.pressure_downstream_priority_valve(),
+            );
```
BlueberryKing, line 161: I think you should pass on the `section_pressure: &impl SectionPressure` to the `SteeringActuator` call `pressure_downstream_priority_valve` on the actuator to unwrap instead of passing in the pressure directly.

BlueberryKing (changes requested):
Very nice!

src/systems/systems/src/hydraulic/mod.rs
crocket63: it is necessary but only for tests => will move this it in test module :)

src/systems/systems/src/hydraulic/mod.rs
crocket63: So last time I tried changing stuff around this I broke something for Luke. And in the meantime your point seems totally valid and furthermore current implementation is just wrong.....  Priority valve is not downstream of  leak measurement valves, so flaps for example should still move if  leak measurement valve is closed.

BlueberryKing (approved):

marcman86:
Quality Assurance Report

Discord: marcman86#4907
Object of testing: #7625
Tier of Testing: 1
Date: 3/02/2023

Testing Process:
- Spawned in air
- Tested gear with only yellow epump and PTU > gear down and up works, pressures about 2100 to 2300, L/G LGCIU 2 FAULT shows up sometimes
- Tested flaps/slats with only blue circuit > only slats work (image 1), flight controls work, pressure stays mostly at 3000

Negatives: Not sure about the flaps not working with only blue circuit.

Testing Results: Passed
Conclusions: Seems to work

Hope this helps :-)

Media:


image 1

PR #7627, opened 2022-12-01
Assisted checklists

Chryssie:
Fixes #[issue_no]

## Additional context


frankkopp:
Thanks for creating this.

Make sure your PR request starts with one of these:
 - feat: A new feature
 - fix: A bug fix
 - docs: Documentation only changes
 - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - refactor: A code change that neither fixes a bug nor adds a feature
 - perf: A code change that improves performance
 - test: Adding missing tests or correcting existing tests
 - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
 - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
 - chore: Other changes that don't modify src or test files
 - revert: Reverts a previous commit
 
 Also add your Discord username so we can also discuss these things on Discord :)
 Ask in #dev-support if you have any questions or need support.

Chryssie:
I had to adjust the scope a bit for two reasons:
1. I could quite determine some of the expected behaviours in the checklists that were in the game previously. And started referring back to the [FlyByWire Normal Procedures document](https://docs.flybywiresim.com/pilots-corner/SOP/). However, this wasn't easy to map around so I rewrote the two in-game check list pages entirely from that document.
   1. I have rebased this to prevent merge conflicts with another change that changed the landing gear checklist slightly.
   2. I had to make a fair number of structural changes to make the assisted checklist be coherent, since every action needs to map to a discrete checkpoint item.
2. I was originally planning on doing everything, however, that turned out to be a very large amount of work. So I decided to break it up into smaller chunks. In future I'd probably want to do one page-per-pr but I had done two pages before realising how big this was getting.

Chryssie:
List of changes since when I made this ready for review:
1. Added changelog entries.
2. Disabled the check for the Audio Switching knob as the knob stopped updating the variable.
3. Corrected the name of "Power Transfer Unit" step (was erroneously called "Yellow Electric Pump")
4. Changed all of the `<Action>` elements in the new library to use `Once="True"`. It fixes a few glitches (especially on actions that will toggle the state when run rather than setting it off, such as pushing an ECAM page button).

Chryssie:
Two more changes:
1. Fix the strobe light step. It should now set it to `AUTO` rather than `ON`.
2. Reduce some of the overhead panel steps. The GPWS panel had an expectation of `All On` of the `LDG FLAP3` button shows a white light when it is on rather than off. So rather than splitting it I reduced some of the steps to say something along the lines of "GPWS Panel White Lights" "OFF" which removes that grammatical error and reduces the number of steps.

Chryssie:
Last two force-pushes are for the following:
1. I managed to fix the programmatic control of the brake pedals so the checkpoint items that relied on them can autocomplete correctly now rather than just force validating.
2. I have rebased this onto the latest `master` branch to resolve the conflicts.

Chryssie:
The rebase missed a change in some folder structures for the checklist fixes (since it wasn't a rename for those files) which I have fixed.

However, I have taken the time to separate out the dangerous fix from the rest of the package. Whilst it will still be globally enabled this allows it to be installed and removed independently to the main package(s) allowing an end user to (with whatever config is required for the installer) to remove just the hotfix rather than the whole aircraft in the event of a conflict.

Chryssie:
Most recent changes are:
1. Updated the ECAM page checkpoint items to use the new mechanism for calling up pages.
2. Fix the checkpoint item for the bluepump overdrive not auto-completing if it is on.
3. Add instrument ids and actions for
   1. Speed Mach Window : Dashed
   2. HDG V/S - TRK FPA : HDG V/S
4. Remove redundant checkpoint items to call ECAM pages.

Chryssie:
I disabled the linter for the Asobo code for `/fbw-common/src/base/msfs-checklists-fix` as it's slightly modified Asobo code and making a bunch of changes will make it harder to keep up to date.

frankkopp:
I have started reviewing this but put it on hold as I'd like to clarify a few things first.

First - this is good work and appreciated - thank you in advance!

We just want to make sure this fits into how we think our project should be structured and work.

There are still concerns in the team because of overwriting core sim files which would impact every other aircraft in the sim. 
Also the way you structured your code is something I think we should look at in more detail.

### Therefore a question - how would your changes work if the sim-fix is not present - what would be the exact impact?

Is it only the visual issue you show in your description?

Reason behind the question is that some of us think this should in fact be a separate package and not necessarily be installed with the aircraft. E.g. could be a separate add-on which the user could install with the Installer. In this case it could also easily be uninstalled if it causes problems. 

The above also has impact on the build structure. I'm not a big fan of how you did it currently - fbw-common was not intended to have an `out` folder - I was expecting it might have a pre-built `lib` folder (or smth similar) to be included in the aircraft builds.

Also building an aircraft should be doable without this `igniter -r 'a32nx|checklists-fix'`.

And even if we leave your structure in fbw-common as it is in your PR I would like to generalize the build (igniter) to not do a "checklists-fix" but a `fbw-common` step to allow for future additions when indeed fbw-common needs additional build steps before any aircraft is build. 

Once we have an anser to the question above we will discuss what we think the best way forward would be. 

Discord: Cdr_Maverick#6475

tracernz:
I've been checking in on this from time to time as it looks like a great piece of work that's super helpful for those new to the A320! 

In addition to the above I'd also add it's a good idea to post details of the issue with the sim's checklist code on https://devsupport.flightsimulator.com. Splitting the checklist fix into it's own package alongside a32nx and a380x is a good idea I think. With luck it will receive triage and necessary fix on the Dev Support platform so we can drop the fix with a later update to the sim.

Chryssie:
@frankkopp 

> Therefore a question - how would your changes work if the sim-fix is not present - what would be the exact impact?

Without the checklist fix the only impact would be the rendering issue of the checklists as shown in the images above. They would still *kinda* work but quite awkwardly due to the phantom items.

> Reason behind the question is that some of us think this should in fact be a separate package (which it seems to be) and not necessarily be installed with the aircraft. E.g. could be a separate add-on which the user could install with the Installer. In this case it could also easily be uninstalled if it causes problems.

I actually agree with that. I added the extra part to the build process so that I could fully test have it in a separate package to see if it would work and it does. I have also, just now, tested moving the actual checklists themselves out and that also words. So it would be possible for me to change it to have no checklists (or reduced functionality checklists) and have the complete ones in a separate package that can be installed and uninstalled independently by the user.

> The above also has impact on the build structure. I'm not a big fan of how you did it currently - fbw-common was not intended to have an out folder - I was expecting it might have a pre-built lib folder (or smth similar) to be included in the aircraft builds.
> 
> Also building an aircraft should be doable without this igniter -r 'a32nx|checklists-fix'.
> 
> And even if we leave your structure in fbw-common as it is in your PR I would like to generalize the build (igniter) to not do a "checklists-fix" but a fbw-common step to allow for future additions when indeed fbw-common needs additional build steps before any aircraft is build.

That makes sense, I was trying to replicate some of what I saw for the package `flybywire-aircraft-a320-neo-lock-highlight` without tying it to a specific aircraft for testing and I didn't want to hand-write a manifest. So I'm open to any changes for the structure there. Although I might need some advice on how to structure that for the igniter build system as I'm not familiar with it.

@tracernz thanks, I have previously submitted a ticket for the issue including specific locations of the bugs. So hopefully sometime in the future they'll be able to fix it.

frankkopp:
I'll put together a few thoughts tomorrow as it is very late for me here.

frankkopp:
> Without the checklist fix the only impact would be the rendering issue of the checklists as shown in the images above. They would still _kinda_ work but quite awkwardly due to the phantom items.

This is a good argument towards having the fix in a separate add-on package users can install/uninstall independently. 

> I actually agree with that. I added the extra part to the build process so that I could fully test have it in a separate package to see if it would work and it does. I have also, just now, tested moving the actual checklists themselves out and that also words. So it would be possible for me to change it to have no checklists (or reduced functionality checklists) and have the complete ones in a separate package that can be installed and uninstalled independently by the user.

The checklist themselves can of course be part of the aircraft - that is where they belong. Just the msfs fix should be separate (and hopefully not required in the future if and when Asobo fixes this). 

> That makes sense, I was trying to replicate some of what I saw for the package `flybywire-aircraft-a320-neo-lock-highlight` without tying it to a specific aircraft for testing and I didn't want to hand-write a manifest. So I'm open to any changes for the structure there. Although I might need some advice on how to structure that for the igniter build system as I'm not familiar with it.

In your case I think it should be more user friendly than `flybywire-aircraft-a320-neo-lock-highlight`. Your use case is the other way round. Whereas `flybywire-aircraft-a320-neo-lock-highlight`is for user who want to deactivate something which is built into the aircraft, the checklist fix is something they should install separately for your checklist to work well and uninstall, if there is a problem/conflict with another add-on. 

I would like to recommend the following:

- Move the fix to the top-level folder (parallel to fbw-a32nx, fbw-a380x) - e.g. `fbw-msfs-checklist-fix`
- Have it build independently - no dependencies to the aircraft builds (add a `build_msfs_checklist_fix.sh`)
- Add a github workflow for this to build and upload this to the CDN (happy to support here or do that if you allow)
- Add this to the installer configuration as a separate package (I can do that then)

Makes sense?

Ping me on Discord for questions and collaboration on this: Cdr_Maverick#6475

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
```diff
@@ -0,0 +1,2610 @@
...
+                </TestValue>
+                <Action Copilot="False" Condition="TestValueFalse" Once="True" SpecialAction="WaitUntilTestValueValidated"/>
+            </Test>
+        </Checkpoint>
+        <Checkpoint Id="A32NX.TAKEOFF_BRIEFING:PERFORM">
+            <CheckpointDesc SubjectTT="TT:A32NX.CHECKLIST.TAKEOFF_BRIEFING" ExpectationTT="TT:GAME.CHECKLIST_PERFORM" />
+            <Clue name="TT:A32NX.CLUE.TAKEOFF_BRIEFING:PERFORM" />
+        </Checkpoint>
+    </Checklist.CheckpointLibrary>
+</SimBase.Document>
```
frankkopp, line 2610: Make sure your IDE adds a newline at the end of each file
```suggestion
</SimBase.Document>

```

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
```diff
@@ -0,0 +1,2610 @@
+<?xml version="1.0" encoding="utf-8"?>
```
frankkopp, line 1: What is the reason you made a Library_new.xml?
And would a more specific name make sense? (e.g. Libraray_A32NX.xml or Library_flybywire.xml)

frankkopp (approved):
I did not check each XML but overall it look good to me.
QA Tests will show if everything works well - will test in the next days as well. 
Two minor remarks added.
Good work!

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
Chryssie: The file currently exists within the a32nx package so adding a flybywire or a32nx suffix doesn't seem appropriate here. Additionally, making it generic (for the a380) may not be feasible, however, that depends on how much the variable names differ and how much the procedures differ. It's something that would probably have to be evaluated when it's closer to completion.

As for why it's a separate file, it's mostly because I ended up making large amounts of structural changes as I was going through it and it made it easier for me to modify what I had implemented and what was still, effectively, a stub. I was also considering splitting it into more files based on the location of the instrument being referred to, however, that doesn't work for all items so I left it in one file.

However, given how large the changes are the diff may look weird if it's added back into the same file (even as a reset). If you want I can go merge them back together and see how it looks.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
Chryssie: I've added the new line at the end of the file, however, my IDE didn't add it automatically as it is not required via the `.editorconfig`.

frankkopp: interesting - I have this setting set in my IDE (Clion) but it is not using the .editorconfig (I think I would have to enable this) - if you know how you could add it to the .editorconfig as it is always a good practice to have this.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
frankkopp: I just think we should avoid *_new* in filenames - "new" is always relative and quickly is "not new" any more 😀 
I'll leave it to you but maybe you can think of something more neutral/lasting.

brewers10:
Quality Assurance Trainer

Discord          : brewers_10#5706
Object of testing: #7627 
Tier of Testing  : 1
Date             : 18/02/2023

Testing Process: 
Loaded into KFBW.
Followed process as designated:
1. Once with assisted checklists off - **PASS**
2. Once using the "Evaluate" button - **PASS**
3. Once using the "Auto Complete Item" button - **PASS**
4. Once using the "Autocomplete Page" button - **PASS** (this was extremely satisfying to do xD)

Negatives:       
TOOK SOOOO LONG!!!!!!!!! (nothing wrong with aircraft tho)

Testing Results: 
Passed

Conclusions: 
Fantastic job. Everything worked as intended. LGTM

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Library_new.xml
tracernz: > Just for xml files or everything?

All text files please, else they're not valid text files :D.

Chryssie:
Given the discussions in Discord, I've put this back into draft until all of the checklists are done.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/Checklist/Airbus_A320neo_Checklist.xml
```diff
@@ -6,151 +6,445 @@
...
+                        <Block SubjectTT="TT:A32NX.CHECKLIST.CENTER_FUEL_LT_200KG">
+                            <BlockCondition>
+                                <Operator OpType="LESS">
+                                    <!-- Centre fuel tank -->
+                                    <Val SimVarName="FUELSYSTEM TANK WEIGHT:1" Units="kilograms" />
+                                    <Val Value="200" />
+                                </Operator>
+                            </BlockCondition>
+                            <Checkpoint ReferenceId="A32NX.FUEL_MODE:MAN" />
+                            <Checkpoint ReferenceId="A32NX.CENTER_FUEL_TANK_PUMPS:OFF" />
```
tracernz, line 279: This is the procedure for the old fuel system prior to 2014 (A320ceo).

.github/CHANGELOG.md
```diff
@@ -85,6 +85,7 @@
 1. [EFB/ATSU] Use MSFS METAR data rather than FSX cloud data from FBW API - @tracernz (Mike)
 1. [APU] Added xfeed APU fuel capabilities - @Taz5150 (TazX [Z+1]#0405)
 1. [ATSU] Fix ATIS Auto Update LSK6L to ATIS page instead of ATC MENU - @BravoMike99 (bruno_pt99)
+1. [CHECKLISTS] Rewrite Preliminary Cockpit Prep & Cockpit Prep pages with assisted checklists support @Chryssie (Chryssie)
```
tracernz, line 88: Please rebase/merge and move to 0.13.0.

PR #7631, opened 2022-12-02
fix: fix rust tests stack overflow

crocket63:
Fixes #[issue_no]

## Summary of Changes

Add some test profile optimisations so the stack memory is not  requested at each test bed affectation

## Additional context

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,9 @@ members = [
 [profile.release]
 lto = true
 strip = true
+
+[profile.test]
+opt-level=1
+lto = false
+debug-assertions=true
+debug=true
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -13,3 +13,9 @@ members = [
 [profile.release]
 lto = true
 strip = true
+
+[profile.test]
+opt-level=1
```
beheh, line 18:
```suggestion
```suggestion
# We enable some basic optimization to prevent stack overflows during testing (flybywiresim/a32nx#7631)
# However, if this interferes with your ability to debug locally you can remove it here
opt-level=1
```
opt-level=1
```

beheh (commented):
One suggestion to add a comment, and otherwise can merge this. No in-sim QA necessary as this only affects CI (and executing the Rust test suite locally).

beheh (approved):

PR #6647, opened 2022-02-04
feat: Implement Slats Flaps Control Computer

omryg2:
## Summary of Changes

This PR will aim to fully implement the SFCCs functionalities.
Feature list will be updated to be more comprehensive as I slowly refresh my memory on the system

- [ ] SFCC features
   - [x] Alpha lock mode (need to update ECAM indications to accommodate)
   - [ ] ...
- [ ] Implement dual SFCCs for the flaps/slats complex
- [ ] ...

## Additional context

Discord username (if different from GitHub): omryg#9274

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #7632, opened 2022-12-03
Slat flap model

Eagle941:
Fixes #6647

## Summary of Changes

Complete implementation of SFCC

## Additional context

### DIFF
--- a/src/systems/a320_systems/src/hydraulic/flaps_computer.rs
+++ b/src/systems/a320_systems/src/hydraulic/flaps_computer.rs
@@ -59,6 +59,7 @@ struct SlatFlapControlComputer {
     flaps_conf_index_id: VariableIdentifier,
     slats_fppu_angle_id: VariableIdentifier,
     flaps_fppu_angle_id: VariableIdentifier,
+    slat_flap_component_status_word_id: VariableIdentifier,
     slat_flap_system_status_word_id: VariableIdentifier,
     slat_flap_actual_position_word_id: VariableIdentifier,
     slat_actual_position_word_id: VariableIdentifier,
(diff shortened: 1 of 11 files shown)


fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -477,13 +477,42 @@ impl AdirsDiscreteOutputs for AirDataInertialReferenceSystem {
...
 impl AirDataSource for AirDataInertialReferenceSystem {
     fn computed_airspeed(&self, adiru_number: usize) -> Arinc429Word<Velocity> {
         self.adirus[adiru_number - 1].computed_airspeed()
     }
 
     fn alpha(&self, adiru_number: usize) -> Arinc429Word<Angle> {
         self.adirus[adiru_number - 1].alpha()
+
+impl AdirsMeasurementOutputs for AirDataInertialReferenceSystem {
+    fn is_fully_aligned(&self, adiru_number: usize) -> bool {
```
tracernz, line 490: This signal doesn't exist in this form on the real thing, so probably not the right way to go. Instead the discrete words are sent on the bus, by the ADR and the IR.

Benjozork:
@Eagle941 is this still being worked on?

Eagle941:
> @Eagle941 is this still being worked on?

@Benjozork 
I don't have any time at the moment to work on this. The main thing to finish is to rewrite the flap/slat actuator by integrating the physical model that was already there instead of the "pressure vs flow" interpolation that I introduced in this branch. Happy to guide people through if needed.
Plus some general cleanup.

PR #7633, opened 2022-12-03
feat: polar heading/track part 1

tracernz:
Fixes #[issue_no]

## Summary of Changes

This pull request is the first step to full polar nav capability for the A32NX (and by extension A380X). The ADIRU part is implemented, along with the PFD and ND parts. FCU, FMGC/MCDU, FWC/ECAM, and other systems are not touched in this PR.

Known issues:
- MCDU does not display true heading data, nor stop show "magnetic" data
- AP/FCU in heading mode is unaware of true headings/tracks
- Many other things in the systems that aren't touched yet

## Screenshots (if necessary)

(Graphical tweaks since this)
image

Flashing at slat extension on PFD, different TRUE flag with approach message on ND:
https://user-images.githubusercontent.com/9995998/206412158-93802570-650e-4e18-9b37-2d31d24c93e0.mp4

## Additional context

## Testing instructions

Fly into, out of and around polar areas (> 82 north or south and in the magnetic north region). The Lvar `L:A32NX_PUSH_TRUE_REF` can be used to simulate the TRUE REF pushbutton that isn't actually present in the Asobo A320 flight deck (this will allow you to enter TRUE REF mode outside of polar regions).


tracernz:
Ugh, forgot about rust tests.

src/systems/systems/src/navigation/adirs.rs
```diff
@@ -1365,6 +1440,99 @@ impl InertialReference {
...
+        // TODO excess motion during align
+
+        // TODO ADR data not received or parity error
+
+        if self.extreme_latitude {
+            maint_word |= IrMaintFlags::EXTREME_LATITUDE;
+        }
+
+        if self.is_aligning() {
+            if self
```
davidwalschots, line 1493: For some strange reason multiline comment no longer work. 

**suggestion**: You can replace this code with the following. Note that Rust 1.66 introduces half open ranges without lower bound. We're currently using 1.65. But later we can remove the lower bound on most of these matches and it would still work.

```rust
            maint_word |= match self
                .remaining_align_duration()
                .map(|duration| duration.as_secs())
            {
                Some(0..=60) => IrMaintFlags::ALIGN_1_MINUTES,
                Some(61..=120) => IrMaintFlags::ALIGN_2_MINUTES,
                Some(121..=180) => IrMaintFlags::ALIGN_3_MINUTES,
                Some(181..=240) => IrMaintFlags::ALIGN_4_MINUTES,
                Some(241..=300) => IrMaintFlags::ALIGN_5_MINUTES,
                Some(301..=360) => IrMaintFlags::ALIGN_6_MINUTES,
                Some(361..) => IrMaintFlags::ALIGN_7_10_MINUTES,
                None => IrMaintFlags::default(),
            };
```

src/systems/systems/src/navigation/adirs.rs
```diff
@@ -2137,6 +2318,7 @@ mod tests {
 
         fn assert_ir_heading_data_available(&mut self, available: bool, adiru_number: usize) {
             assert_eq!(self.heading(adiru_number).is_normal_operation(), available);
+            // assert_eq!(self.true_heading(adiru_number).is_normal_operation(), available);
```
davidwalschots, line 2321: **nitpick**: Intended to commit commented out code?

davidwalschots (commented):
I reviewed the Rust portion of code. All good, one suggestion and one nitpick thingie. 😀

src/systems/systems/src/navigation/adirs.rs
davidwalschots: You could in fact also remove the `self.is_aligning()` check, and add a `Some(0)` condition with IrMaintFlags::default():

```rust
            maint_word |= match self
                .remaining_align_duration()
                .map(|duration| duration.as_secs())
            {
                Some(1..=60) => IrMaintFlags::ALIGN_1_MINUTES,
                Some(61..=120) => IrMaintFlags::ALIGN_2_MINUTES,
                Some(121..=180) => IrMaintFlags::ALIGN_3_MINUTES,
                Some(181..=240) => IrMaintFlags::ALIGN_4_MINUTES,
                Some(241..=300) => IrMaintFlags::ALIGN_5_MINUTES,
                Some(301..=360) => IrMaintFlags::ALIGN_6_MINUTES,
                Some(361..) => IrMaintFlags::ALIGN_7_10_MINUTES,
                Some(0) | None => IrMaintFlags::default(),
            };
```

docs/a320-simvars.md
```diff
@@ -1431,6 +1431,7 @@ In the variables below, {number} should be replaced with one item in the set: {
     - 0 is CAPT, 1 is NORM, 2 is F/O
 
 - A32NX_ADIRS_ADIRU_{number}_STATE
+    - TODO delete
```
Benjozork, line 1434: I would here indicate what to use instead, and use wording like "deprecated"

docs/a320-simvars.md
```diff
@@ -1440,6 +1441,7 @@ In the variables below, {number} should be replaced with one item in the set: {
       Aligned | 2
 
 - A32NX_ADIRS_REMAINING_IR_ALIGNMENT_TIME
+    - TODO delete
```
Benjozork, line 1444: I would here indicate what to use instead, and use wording like "deprecated"

src/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -423,10 +424,12 @@ export class LnavDriver implements GuidanceComponent {
         // FIXME should be NCD if no FM position
 
         SimVar.SetSimVarValue('L:A32NX_EFIS_L_TO_WPT_BEARING', 'Degrees', efisBearing);
+        SimVar.SetSimVarValue('L:A32NX_EFIS_L_TO_WPT_TRUE_BEARING', 'Degrees', efisTrueBearing);
```
Benjozork, line 427: Is there a reason we have to transmit both? Since it's only used for ND display.

tracernz: We either need to fo that, or transmit the magnetic variation to use.

docs/a320-simvars.md
tracernz: Ah, good catch. I had intended to delete these simvars as part of this PR when I wrote those comments, so they were for me only. I will either do that (which will require changes in other systems), or most likely deprecate as you suggest...

Benjozork (approved):

PR #7635, opened 2022-12-05
feat(hyd): faster core hydraulics solver

crocket63:
Fixes #[issue_no]

## Summary of Changes

As wasm performances proves to be really good, this PR sets all the hydraulic calculations to 100hz, which should help in some fast flow transients use cases, such as gear or multiple flight control inputs.

## Additional context

## Testing instructions

1) Check no noticeable performance degradation.

2) Please check as many hydraulic cases as possible and search for any regression, including brakes/steering/cargo doors/hydraulic leaks...

3) On gear movements pressure down spikes should not go as low as before but should still trigger PTU.


src/systems/systems/src/hydraulic/linear_actuator.rs
```diff
@@ -2539,7 +2539,7 @@ mod tests {
     fn linear_actuator_not_moving_on_locked_rigid_body() {
         let mut test_bed = SimulationTestBed::new(|context| {
             let tested_object = cargo_door_assembly(context, true);
-            TestAircraft::new(context, Duration::from_millis(33), tested_object)
+            TestAircraft::new(context, Duration::from_millis(10), tested_object)
```
Gurgel100, line 2542: I don't know if it would be helpful to extract these values to an "global" constant so it is changed later one only has to change it at one point.

Gurgel100, line 2542: I don't know if it would be helpful to extract these values to an "global" constant so it is changed later one only has to change it at one point.

Gurgel100 (changes requested):

src/systems/systems/src/hydraulic/linear_actuator.rs
crocket63: Simpler better :D
Used to have different time steps for different things, hope we can have the same for all from now on

PR #7648, opened 2022-12-11
feat: upload to cloudflare cdn on exp

Benjozork:
## Summary of Changes

This PR:

* adds a script to upload to the CloudFlare R2 CDN;
* adds a step in the experimental pipeline to upload it there as well, keeping bunnycdn for now;
* changes the fragmenter module chunk size from 500MiB to 98MiB (R2 supports 100MB uploads at most).

### DIFF
--- a/scripts/cf-cdn.sh
+++ b/scripts/cf-cdn.sh
@@ -14,7 +14,7 @@ upload () {
 
     # Try to upload the file up to MAX_RETRY times before failing
     counter=0
-    until curl --fail -X PUT -H "X-FBW-Access-Key: $CLOUDFLARE_BUCKET_PASSWORD" --data-binary -T "$1" "$DEST"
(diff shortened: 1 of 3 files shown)


scripts/cf-cdn.sh
```diff
@@ -0,0 +1,39 @@
...
+MAX_RETRY=5
+
+upload () {
+    DEST="$CDN_URL/$CDN_DIR/$(basename -- "$1")"
+    echo "Syncing file: $1"
+    echo "Destination: $DEST"
+
+    # Try to upload the file up to MAX_RETRY times before failing
+    counter=0
+    until curl --fail -X PUT -H "X-FBW-Access-Key: $CLOUDFLARE_BUCKET_PASSWORD" --data-binary -T "$1" "$DEST"
```
Saschl, line 17: Afaik you should either use --data-binary or -T, but not both

Benjozork: Oops that wasn't intentional. Will fix

Benjozork: Done.,

Saschl (approved):
Noice

PR #7655, opened 2022-12-18
feat: Add aircraft version check and user message when using old version

frankkopp:
## Summary of Changes
This PR introduces a check of the version of the currently running aircraft against the latest github versions.

This PR adds a comprehensive version tag to the a32nx_build_info.json which is then used in the EFB code to check against the github version. A MSFS notification will be shown and the flyPad will also show the old version state  in the status bar.

Notification policy:

- when the version is older than the latest Stable release
- when version is equal or newer and branch is dev or exp then compare commit hash
  - if different and build date of aircraft is older than build date of latest dev/exp commit by 3 days

## Screenshots (if necessary)
image
image
image

https://user-images.githubusercontent.com/16833201/209550906-66b233c7-27be-4bbd-971f-ca5b1ceff3fa.mp4

## Testing instructions
To test this the a32nx_build-info.json file needs to be edited to "pretend" to be on an outdated version.

All editions; 
- To check for outdated main version change the version info to an older version (e.g. 0.9.0)

Dev and Exp
- Change the version tag to something like `v0.10.0-dev.124567` or `v0.10.0-exp.124567`
- As long as the commit sha is different to the last commit in dev or exp it will check for build time. 
  - If older than 3 days the message should appear - change the built date to test this
- If commit sha is identical no message should appear. (change the commit sha in the version string)


package.json
```diff
@@ -101,22 +101,23 @@
...
+    "@localazy/cli": "^1.6.0",
     "@reduxjs/toolkit": "^1.6.2",
     "@rollup/plugin-json": "^4.1.0",
     "@sentry/browser": "^6.17.7",
     "@sentry/integrations": "^6.17.7",
     "@sentry/tracing": "^6.17.7",
     "@tabler/icons": "^1.41.2",
     "@types/react-canvas-draw": "^1.1.1",
-    "@localazy/cli": "^1.6.0",
+    "@types/semver": "^7.3.13",
```
Benjozork, line 112: This should be in `devDependencies`, since it's not a runtime package.

frankkopp: fixed

src/instruments/src/EFB/Utils/AircraftVersionChecker.ts
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        // If the user's version is equal or newer than the latest release then check if
+        // the edition is Development or Experimental and if the commit is older than
+        // {maxAge} days after the latest release to show notification
+
+        const maxAge = 3;
+        const timestampAircraft: Date = new Date(this.buildInfo.built);
+
+        if ((branchName === 'Development')
```
Benjozork, line 180: I would store those branch names in an enum `KnownBranchNames` to avoid having magic values in the code

Benjozork (approved):

src/instruments/src/EFB/Efb.tsx
```diff
@@ -119,6 +118,9 @@ const Efb = () => {
 
     useEffect(() => {
         document.documentElement.classList.add(`theme-${theme}`, 'animationsEnabled');
+        if (!AircraftVersionChecker.isVersionChecked) {
```
Benjozork, line 121: Now that I think of it, we wanna delay this until the user is definitely loaded in the cockpit - some people have crashes when the popup shows before.

I think the sentry popup has some logic to do this.

frankkopp: Ok - will look into this

frankkopp: Added wait for A32NX_IS_READY which basically is the same.
Also moved the check to the index.tsx and our of the EFB.tsx - think it fits better there

src/instruments/src/EFB/Utils/AircraftVersionChecker.ts
frankkopp: Good idea - done

mico975:

Discord          : mico#3145
Object of testing: #7655
Tier of Testing  : 1
Date             : 01/01/2023

Testing Process: 
To check for outdated main version change the version info to an older version (e.g. 0.9.0) - **OK**

Change the version tag to something like v0.10.0-dev.124567 or v0.10.0-exp.124567
As long as the commit sha is different to the last commit in dev or exp it will check for build time.
If older than 3 days the message should appear - change the built date to test this - **OK**
If commit sha is identical no message should appear. (change the commit sha in the version string) - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

Media:       
image

tracernz:
I think the dev/exp limit should be increased to at least a week, if not a month. It will be tedious and annoying otherwise.

frankkopp:
> I think the dev/exp limit should be increased to at least a week, if not a month. It will be tedious and annoying otherwise.

From a support point of view I would disagree. We often get the most support requests after a bug/fix/new feature from users not having done an update - so it could even make sense to drop the age threshold altogether :) - which indeed might be annoying.
Although it is one click :)

1 week would be the max for my taste - but I would prefer the 1/2 week (3 days) tbh.
Haven't gotten any other feedback on this yet though.

PR #7670, opened 2022-12-26
feat(atsu): ATSU as standalone system with event-bus outside of the MCDU

svengcz:
Fixes report about missing reset of auto-ATIS update on discord

## Summary of Changes

This PR changes the software architecture of the ATSU to an event-based system that runs as a single system outside the MCDU.

- [x] Split up ATSU into common and system
- [x] Use EventBus for the communication with the DCDU/Mailbox
- [x] Implement MCDU client to handle ATSU events outside the MCDU-page logic
- [x] Create System-Host as instrument to load Typescript-systems as single components
- [x] Remove ATSU from MCDU and add to system-host
- [x] Change flight state observer based on events
- [x] Testing, testing testing :D
  - [x] AOC
  - [x] ATC
  - [x] DCDU

## Additional context

Discord username (if different from GitHub): Sven [de en]

## Testing instructions

Test all features of the current ATSU version.
- All AOC features (ATIS, METAR, TAF, Freetext for Hoppie & FBW)
- All ATC features
  - Connect, disconnect
  - Handover
  - Message from ATC (w & w/o semantic responses (i.e. REPORT SQUAWK))
  - Message to ATC (DCL, OCL, Request, etc.)
  - Position reports (manual & automatic)
  - D-ATIS


src/atsu/common/src/databus/Fms.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export interface FmsRouteData {
+    lastWaypoint: Waypoint;
+    activeWaypoint: Waypoint;
+    nextWaypoint: Waypoint;
+    destination: Waypoint;
+}
+
+export interface AtsuFmsMessages {
```
Benjozork, line 27: I think we should make the messages sent from ATSU -> FMS be a separate interface than the messages sent from FMS -> ATSU.

src/atsu/common/src/index.ts
```diff
@@ -0,0 +1,67 @@
+import { AtsuStatusCodes } from './AtsuStatusCodes';
+import { MailboxStatusMessage } from './databus';
+import {
```
Benjozork, line 3: You can replace this with `export * from './messages'` instead

src/atsu/common/src/messages/WeatherMessage.ts
```diff
@@ -61,4 +61,11 @@ export class WeatherMessage extends AtsuMessage {
 
         return message;
     }
+
+    public deserialize(jsonData: Record<string, unknown>): void {
```
Benjozork, line 65: I would generally make those a static method instead, for example:

```ts
public static deserialize(jsonData: Record<string, unknown>): WeatherMessage
```

src/atsu/common/src/types/Clock.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    public hour: number;
+
+    public minute: number;
+
+    public second: number;
+
+    public secondsOfDay: number;
+
+    constructor() {
```
Benjozork, line 16: You can make a shorthand for this by declaring properties in the constructor:

```ts
constructor(
    public year: number,
    public month: number,
    public dayOfMonth: number,
    public hour: number,
    public minute: number,
    public second: number,
    public secondsOfDay: number,
) {
}
```

TypeScript will automatically generate the assignments in the constructor body at compile-time. The properties in the class body can then be omitted.

src/atsu/common/src/types/Waypoint.ts
```diff
@@ -0,0 +1,11 @@
+export class Waypoint {
```
Benjozork, line 1: This looks like it could just be an interface.

src/atsu/system/src/index.ts
```diff
@@ -0,0 +1,13 @@
+import { Atsu } from './ATSU';
```
Benjozork, line 1: We can replace those with `export * from './X'`

src/instruments/src/DCDU/index.tsx
```diff
@@ -232,140 +237,132 @@ const DCDU: React.FC = () => {
...
                 }
+                setMessages(new Map<number, DcduMessageBlock>(messagesRef.current));
             }
-            setMessages(new Map<number, DcduMessageBlock>(messagesRef.current));
-        }
-    });
+        });
+
+        setSubscriber(newSubscriber);
+    }, []);
```
Benjozork, line 365: We need a way here to unsubscribe from all those things in the cleanup callback, but unfortunately our version of `msfssdk` does not have the `Subscription` type....

Need to think of a way to handle this, or add a TODO. This component will never unmount for now, but it is still unoptimal to leak subscriptions

Benjozork (changes requested):

.github/CHANGELOG.md
```diff
@@ -19,8 +19,10 @@
 1. [EFB] Fix and improve pushback system and add API documentation - @frankkopp (Frank Kopp)
 1. [RMP] RMPs navigation backup - Julian Sebline (Julian Sebline#8476 on Discord)
 1. [SEC] Fix GND SPLR logic, add missing GND SPLR partial extension condition - @lukecologne (luke)
-1. [FMGC] Improved importing flight plans from MSFS World Map - @frankkopp (Frank Kopp)  
-1. [EFB] Added boarding time indication to Payload page - @ChristianLutzCL (Christian Lutz) @frankkopp (Frank Kopp)
+1. [FMGC] Improved importing flight plans from MSFS World Map - @frankkopp (Frank Kopp)
+1. [EFB] Added boarding time indication to Payload page - @ChristianLutzCL (Christian Lutz) @frankkopp(Frank Kopp)
```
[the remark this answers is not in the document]
svengcz: No, reverted the commit

src/atsu/common/src/index.ts
svengcz: fixed

src/atsu/system/src/index.ts
svengcz: fixed

src/atsu/common/src/types/Waypoint.ts
svengcz: fixed

src/instruments/src/DCDU/index.tsx
svengcz: handled the leak in the cleanup-call of the effect

src/atsu/common/src/messages/WeatherMessage.ts
svengcz: fixed

src/atsu/common/src/databus/Fms.ts
svengcz: fixed

Benjozork (approved):

frankkopp:
I'd really love to see some documentation esp. around the systems-host - this is really a great addition and it should be clearly documented when to use it, how to use it and how additional modules should be structured. 

Either a README in the systems-host folder or directly in the index.tsx would work fine.

This way others could learn faster how this should be used in a way you intended it to be used. Maybe a minimal example would make sense - Atsu is a big system and it might be easier to understand a more "hello world" like example on systems-host usage.

PR #7696, opened 2023-01-07
fix(eslint): fix indent rule in src folder for TS

ExampleWasTaken:
## Summary of Changes

This PR adapts the eslint config in the src folder for TypeScript as that is the primary language used in there.

The reason for this PR is an error thrown by eslint when linting TS files. Instead of the default eslint indent rule the replacement rule from [@typescript-eslint/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) should be used.

The error as well as the fix introduced in this PR is described in eslint/eslint#13956.

Direct link to the fix: <https://github.com/eslint/eslint/issues/13956#issuecomment-751232188>

## Additional context

Discord username (if different from GitHub): ExampleWasTaken#0886

## Testing instructions

As this change only affects the dev environment there isn't really anything to test for QA.


ExampleWasTaken:
This is pretty weird as all linting errors except those related to the introduced `@typescript-eslint/indent` rule exist on master too. I wonder how other PRs into master got merged/didn't fail the lint.
As for the indent errors I think they should be fixed (won't do it without a go from a maintainer though).

tracernz:
This seems to be changing the actual indenting rules looking at the lint log. Some of the files I checked are indented correctly but this rule is causing them to fail.

> EDIT: I cannot reproduce the failing rust test locally (how are they failing I didn't touch that?)

A small number of the tests are not deterministic enough and fail randomly. I've re-run the job.

ExampleWasTaken:
As explained in <https://github.com/typescript-eslint/typescript-eslint/issues/1824> the `@typescript-eslint/indent` rule is not consistent. 
Based on that note I don't think there is an easy way to fix this issue at this time.

tracernz:
That’s unfortunate. Thanks for trying. 👍

PR #7708, opened 2023-01-13
feat(efb): Custom SimBridge IP Address 

auroraisluna:
Fixes #[issue_no]

## Summary of Changes

This PR adds the ability on the EFB to connect to a different IP address for Simbridge, allowing users to have Simbridge running on a different machine. 

## Screenshots (if necessary)

image

## Additional context

Discord username (if different from GitHub): Alepouna🌙#9824

## Testing instructions

During all testing, please ensure SimBridge is running.
- [ ] Try to enable SimBridge without any configuration change 
    - [ ] Test remote MCDU, terrain (no difference, ensure connection simply works)
- [ ] Change the SimBridge address from 127.0.0.1 to your computer network IP (192.168.1.X or different, depending on your network - find with /ipconfig on a windows terminal)
- [ ] Have SimBridge running on a different computer, change IP address to the new machine and test SimBridge connection (Terrain/MCDU etc)


Lucky38i (approved):
LGTM

src/simbridge-client/src/components/McduServerClient.ts
```diff
@@ -8,9 +8,10 @@ import { ClientState } from './ClientState';
  * Class to communicate with the SimBridge MCDU server
  */
 export class McduServerClient {
+    public static ip = ():string => NXDataStore.get('CONFIG_SIMBRIDGE_IP', 'localhost');
     public static port = ():string => NXDataStore.get('CONFIG_SIMBRIDGE_PORT', '8380');
 
-    public static url = ():string => `ws://127.0.0.1:${this.port()}/interfaces/v1/mcdu`.replace(/\s+/g, '');
+    public static url = ():string => `ws://${this.port()}:${this.port()}/interfaces/v1/mcdu`.replace(/\s+/g, '');
```
tracernz, line 14: I don't think that address is going to work...

src/instruments/src/EFB/Settings/Pages/SimOptionsPage.tsx
```diff
@@ -90,6 +91,16 @@ export const SimOptionsPage = () => {
                         </div>
                     </SettingItem>
 
+                    <SettingItem name={t('Settings.SimOptions.SimBridgeIp')}>
+                        <SimpleInput
+                            className="text-center w-30"
+                            value={simbridgeIp}
+                            onChange={(event) => {
+                                setSimbridgeIp(event.replace(/\s/g, ''));
```
tracernz, line 99: It's really just an address rather than an IP address, right?

src/simbridge-client/src/components/McduServerClient.ts
auroraisluna: Oops, thanks

src/instruments/src/EFB/Settings/Pages/SimOptionsPage.tsx
auroraisluna: Yep, you can have something like `simbridge.local` or if you have a proper dns setup they could have something like `computername.local` etc. Should support that :)

tracernz: What I mean is it probably shouldn't be labelled "IP Address" in the UI etc.

auroraisluna:
For whatever reason VSC messed up the commits so I lost a few typo fixes + the merge thingy. Sorry! Should be good to test now from what I can see.

auroraisluna:
Converting to draft: 
- [ ] Use simbridge.local
- [ ] Perform a Simbridge connection test when clicking "Auto" for SB to check it can connect at that point of entering the details

Lucky38i (approved):

Nufflee:
This will not work if `simbridge.local` is not available for whatever reason. I think the best way to go about this is to, if `simbridge.local` is in the text field, ping `simbridge.local/health` and if it is available, great, use simbridge.local. Otherwise, try to use `localhost`. This is a good default because it covers these 3 cases:
1. Running SimBridge on the local computer _with_ mDNS
2. Running SimBridge on the local computer _without_ mDNS
3. Running SimBridge on a remote computer with mDNS

An edge case I can think of is where the user may want to connect to SimBridge on a remote computer but they have also started it on the local computer (with mDNS), in which case we would connect to local SimBridge instead of remote, but I would consider that user error. 
A way to prevent this would be to indicate whether we are connected to local or remote SimBridge on the EFB, but that would require a disproportionate amount of effort (specifically comparing the current machine's IP with SimBrige's IP).

auroraisluna:
That was my initial concern with using `simbridge.local` instead of just `localhost` when it was brought up me to use s.local instead. On the edge case you happen to not have `simbridge.local` available, SimBridge it self should alert the user of that issue (in all cases) and the user can take action as required. [If the issue is from the SB host machine] 

I had an initial idea of implementing a common function that will check if `simbridge.local` is available and then try `localhost` either on: 
- Every function that is trying to parse the URL (currently 2 - which means 2 requests to simbridge.local for health check) 
- When you save the address / when the EFB loads (1 request for healthcheck, but needs a variable set somewhere to remember this change)
    **Potentially issue here**: 
    A user is running SB on another PC, types simbridge.local on EFB -> plane fails to connect (for any reason) and corrects to localhost. This will confuse the user when debugging, thinking its the EFB not saving. (If we do this on the background without updating the actual EFB setting, this is mitigated) 

Anyways, @Nufflee as we said on disc, feel free to propose a solution as you seem fit. 

Also, I am putting this here mostly so I don't forget: 
- [ ] Update the docs troubleshooting connection problems when using `simbridge.local)

PS. The PR still needs to be tested for if using SimBridge on another PC works, not having simbridge.local working is a separate issue, and so I recommend adding DoNotMerge instead. QA needs to be done for the connectivity and functionality part anyways.

auroraisluna:
Okay, idea change. Instead of using the currently proposed idea [ONLY FOR THE INTERFACE], which actually can potentially break systems, I am instead going to implement it like this: 
image
image

The connection is unaffected, that part of this PR can be tested.

auroraisluna:
Ready to test :)

2hwk:
Quality Assurance Report

Discord: 2Cas#1022
Object of testing: https://github.com/flybywiresim/a32nx/pull/7708
Tier of Testing : 1
Date : 12/02/2023

Testing Process:
Peripherals Tested - Flight Stick Trim Switch, Throttle Trim Wheel, Mouse (click and drag), Mouse (mousewheel), Keyboard
Loaded into KFBW.
- [x] Try to enable SimBridge without any configuration change 
    - [x] Test remote MCDU, terrain (no difference, ensure connection simply works)
- [x] Change the SimBridge address from 127.0.0.1 to your computer network IP (192.168.1.X or different, depending on your network - find with /ipconfig on a windows terminal)
- [x] Have SimBridge running on a different computer, change IP address to the new machine and test SimBridge connection (Terrain/MCDU etc)

Negatives:
- If a successful simbridge connection is already established with a localhost, there wasn't an obvious way to force the connection to the new remote host even after the IP Address was entered, in the end I had to restart the flight to successfully connect to the new remote host. Not a flow that would normally be encountered, but maybe in a future PR this could be improved.

Testing Results:
Passed

Conclusions:
LGTM

auroraisluna:
> If a successful simbridge connection is already established with a localhost, there wasn't an obvious way to force the connection to the new remote host even after the IP Address was entered, in the end I had to restart the flight to successfully connect to the new remote host. Not a flow that would normally be encountered, but maybe in a future PR this could be improved.

Nice catch. I thought it would cause this behavior as the address is hardstored in the functions once set, I think. I am thinking in a future PR to implement a third shared function that will return the current set address at any point its requested and maybe even terminate/close old connections as well so the performance isn't affected. Need to investigate this more. 

Thanks for testing!

PR #7714, opened 2023-01-15
build: improve c/c++ build for debugging

tracernz:
Fixes #[issue_no]

## Summary of Changes

Sets all compiler flags needed for sucessful debugging of C++ WASM in MSFS with VS Code. You also need to enable fast WASM compilation in the sim. Optisation set at -O2 after discussion.

Next step would be to setup a proper build system, perhaps in #7711 .

## Additional context

## Testing instructions

Check that the aircraft generally operates correctly. No functional change expected.


.vscode/launch.json
```diff
@@ -8,7 +8,6 @@
             "name": "Attach",
             "type": "cppvsdbg",
             "request": "attach",
-            "processName": "FlightSimulator.exe",
```
frankkopp, line 11:
```suggestion
            "processId": "${command:pickProcess}",
```
You can use this instead.

src/fadec/build.sh
```diff
@@ -6,15 +6,19 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 OUTPUT="${DIR}/../../flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/panel/fadec.wasm"
 
 if [ "$1" == "--debug" ]; then
+  WASMLD_ARGS=""
   CLANG_ARGS="-g"
 else
-  WASMLD_ARGS="--strip-debug"
+  WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
+  CLANG_ARGS="-flto -O2 -DNDEBUG"
 fi
```
frankkopp, line 14:
```suggestion
if [ "$1" == "--debug" ]; then
CLANG_ARGS="-g -DDEBUG"
WASMLD_ARGS="-O0"
else
CLANG_ARGS="-flto -O2"
WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
fi
```
I think it would be more useful to have DEBUG when in debug and nothing when not - wdyt?

src/fbw_a320/build.sh
```diff
@@ -7,20 +7,25 @@ COMMON_DIR="${DIR}/../fbw_common"
 OUTPUT="${DIR}/../../flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/panel/fbw.wasm"
 
 if [ "$1" == "--debug" ]; then
+  WASMLD_ARGS=""
   CLANG_ARGS="-g"
 else
-  WASMLD_ARGS="--strip-debug"
+  WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
+  CLANG_ARGS="-flto -O2 -DNDEBUG"
 fi
```
frankkopp, line 15:
```suggestion
if [ "$1" == "--debug" ]; then
CLANG_ARGS="-g -DDEBUG"
WASMLD_ARGS="-O0"
else
CLANG_ARGS="-flto -O2"
WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
fi
```
I think it would be more useful to have DEBUG when in debug and nothing when not - wdyt?

src/fbw_a380/build.sh
```diff
@@ -7,20 +7,25 @@ COMMON_DIR="${DIR}/../fbw_common"
 OUTPUT="${DIR}/../../flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/panel/fbw.wasm"
 
 if [ "$1" == "--debug" ]; then
+  WASMLD_ARGS=""
   CLANG_ARGS="-g"
 else
-  WASMLD_ARGS="--strip-debug"
+  WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
+  CLANG_ARGS="-flto -O2 -DNDEBUG"
 fi
```
frankkopp, line 15:
```suggestion
if [ "$1" == "--debug" ]; then
CLANG_ARGS="-g -DDEBUG"
WASMLD_ARGS="-O0"
else
CLANG_ARGS="-flto -O2"
WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
fi
```
I think it would be more useful to have DEBUG when in debug and nothing when not - wdyt?

src/flypad-backend/build.sh
```diff
@@ -6,15 +6,19 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 
 if [ "$1" == "--debug" ]; then
-  CLANG_ARGS="-g -DDEBUG"
+  WASMLD_ARGS=""
+  CLANG_ARGS="-g"
 else
-  WASMLD_ARGS="--strip-debug"
+  WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
+  CLANG_ARGS="-flto -O2 -DNDEBUG"
 fi
```
frankkopp, line 14:
```suggestion
if [ "$1" == "--debug" ]; then
CLANG_ARGS="-g -DDEBUG"
WASMLD_ARGS="-O0"
else
CLANG_ARGS="-flto -O2"
WASMLD_ARGS="-O2 --lto-O2 --strip-debug"
fi
```
I think it would be more useful to have DEBUG when in debug and nothing when not - wdyt?

frankkopp (changes requested):
Added two things - processID picker and -DDEBUG instead of -DNDEBUG

I use DEBUG in my code usually and nothing when not in debug. Rarely seen it the other way round. But we could also keep both if you have code using NDEBUG.

frankkopp:
> Next step would be to setup a proper build system, perhaps in #7711 .

Please go ahead and work directly on the branch/PR - happy to leave the build to you :)
Also happy to test it any time (I use Clion which uses CMake as a project base anyway).

src/fadec/build.sh
tracernz: NDEBUG comes from the MSFS SDK template… I assume it’s used in their headers.

frankkopp: Ok - I have added DEBUG on mine as well - so I have both now. Maybe you can do the same or it will be in my PR then.

tracernz: I'd rather keep this PR applicable to what's in master now, then do stuff for your PR in that branch.

.vscode/launch.json
tracernz: That is actually the default, but we can make it explicit I guess.

tracernz: https://code.visualstudio.com/docs/cpp/launch-json-reference#_additional-properties

Gurgel100:
> Added two things - processID picker and -DDEBUG instead of -DNDEBUG
> 
> I use DEBUG in my code usually and nothing when not in debug. Rarely seen it the other way round. But we could also keep both if you have code using NDEBUG.

Just as a note: The `NDEBUG` macro is also used in the C/C++ standard library to deactivate asserts. So I think it's good to have it defined in release builds.

frankkopp (approved):

PR #7715, opened 2023-01-15
refactor(w&b):  rust refactor of payload management

2hwk:
Fixes #[issue_no]

## Summary of Changes

Applied ontop of GSX Integration to ensure compatibility with new backend refactoring.
See: https://github.com/flybywiresim/a32nx/pull/7442 for more information and https://github.com/flybywiresim/a32nx/pull/7695

This refactor: 
- Optimisation and bug fixes for the EFB Payload page.
   - Better support for mixed seating classes and staggered rows 
- Moves the bitwise flags that control passenger seating from localStorage (Which is slow and can only be accessed via coherentJS) into LocalVars instead, as well as to deprecate the old ``TOTAL_ROWS_{}_{}`` Local:Vars which are redundant with this bitpacked pax representation.
- Moves the boarding logic from javascript into rust (WASM), allowing for more complex bit manipulation, which is a good optimization and slightly noticeable performance bump (when boarding and navigating the EFB page). We can also utilise ``trait`` to pipe this data into other modules. 

This has several benefits:

* Performance is much faster than using LocalStorage, where before frametimes could spike by 20ms or more with just 4 pax stations when trying to store several pax flag vars, now the impact is barely noticable to the end user (~1-2ms, which can be further improved). 

* Enables state saving/loading of seating by MSFS's default save file system

* Allows manipulation of passenger seating in WASM (rust/C++), and allows other systems to be able to read current passenger location on a per passenger level of granularity (vs per section/payload station as before)

* External programs can now set a seating layout:
   - Either by first setting the relevant ``PAX_FLAG_{}_DESIRED`` Local:Var with a bitpacked 64-bit long number, i.e. 7, or ``1111111`` would set the first 7 seats in a passenger section (station) as seated, then start the boarding by setting the ``BOARDING_STARTED_BY_USR`` Local:Var as ``true``. This will start the boarding process as though it was triggered by the EFB.
   - Or setting three variables ``PAX_FLAG_DESIRED_{}`` and``PAX_FLAG_{}`` directly alongside writing the relevant aircraft variable ``PAYLOAD STATION WEIGHT:{}``
   - Note that in this first iteration this is still fairly limited, and boarding time when using the boarding system is limited by the user's boarding time setting. *As such this is not the beginning of official support for this feature, this is a stepping stone only*

## Screenshots (if necessary)

No visual changes

## Additional context

## Testing instructions

Payload page should work as before

### New Payload Backend

~~1. Go to Ground > Services page and check that all elements are correctly rendered~~
2. Go to Payload page, 
3. Adjust seating by clicking on the seating layout. This should add new seats and increase the table value for planned passengers.
4. Add baggage by clicking on the 4 sliders, this should affect the desired cargo value in the table.
5. Change per passenger and per bag weight from 84 and 20 to higher/lower values (or leave at default)
6. Input desired passenger and desired cargo in the table
7. Check that ZFW calculation is accurate (Empty Weight is 42500KG, add passenger, passenger bags and cargo weight to this value and this should match desired ZFW)
8. Board passengers with desired time setting (Real/Fast/Instant). Should be able to change speed while passengers are boarding and have it transition between boarding times seamlessly.
9. Deboard passengers (either by inputting in the table, or clicking on the seating layout, or using the deboard button).
10. Go to the main EFB dashboard by clicking the FBW Logo in the top left, import a simbrief OFP
11. Go back to the payload page and click the cloud icon next to the table to load the OFP. Check that all relevant table values are correctly imported.
12. Load passengers/cargo again
~~13. Fill out INIT-A and INIT-B in MCDU (Init fuel pred). Check that MLDW indicator now shows on the performance envelope chart.~~
14. Proceed with normal flight operations. Check that when in air, MTOW envelope is replaced by the FLIGHT envelope.
15. Check that visuals in all themes (dark/light/blue) are correct and readable, with both bright daytime and at night (in sim time)
16. Press COND page and check that air conditioning is still functioning as expected when the air conditioning knobs are adjusted. 
~~17. Create a new save file with the new seating loadout, then load it. The seating layout should show when you go into the EFB's payload page.~~ <- Delayed/Not Officially supported for now (watch this space)
18. Check that all the above steps work for LBS as well as KG

### Check GSX Integration still works

There's a few things to be aware of when using GSX Sync. **PLEASE READ** each of these carefully, documentation is planned for release.

NOTE: There are some bugs with GSX core (not related to this PR) that occur when using jet bridges occasionally, so for the purposes of testing, if this is encountered please take note.

* There's a new page in the EFB settings labelled `3rd Party Options`. Inside there are two options to give you the freedom on what to sync with GSX. This activates the synchronisation. Without it GSX won't sync as expected. It's important to note various items and fields get disabled in the payload and fuel page with these options enabled.
* De-fueling Is NOT supported, GSX only checks if the fuel has changed to a number higher than the current value. If you call the refueling truck and attempt to defuel, the aircraft will defuel but GSX will nag you to change fuel and not actually do anything, constantly nagging you.
* Fueling is managed by A32NX, GSX simply acts as a visual indicator here which allows the user to still use Instant/Fast/Real options in the EFB. However, Fuel cannot be started until the fueling truck has connected the hose as shown by a grey start button. The fueling truck will remain and count for the duration of the fueling period set by A32NX, you will of course get a bill as usual.
* You must set your amount of passengers before GSX boarding, whether through simbrief sync in the EFB or manually in the payload page. Same goes for cargo. Once GSX boarding has been requested, these fields _CANNOT_ be edited until deboarding (meaning an entire flight)
* Boarding is managed by GSX (in a sense), this means you cannot adjust the boarding times in the EFB, you can only adjust boarding time, through GSX's own settings. If one physical passengers boards the aircraft through GSX, then A32NX's Boarding backend is made aware and acts accordingly. Passenger seats are still loaded in the same way they would without GSX Sync. 
* The same is regarded with Cargo/Baggage, however due to limitations with the simvars available from GSX, all stations are loaded at the same time. So if GSX says baggage is 33% loaded, all stations will be loaded to 33% of their given weight target. This removed a bit of the randomness to loading but the limitations is out of my hands.

An aircraft profile is provided, you don't need to do anything GSX picks up on this
* With this profile, the aft passenger door is disabled and the the aft cargo door as-well. This is because these doors can't be opened from the EFB. The aft service door is not disabled, but must be manually opened through the EFB. I'll try and notify the GSX team on the simvars to open this door. So you won't have to in the future.

- [ ] Ensure simbrief loading still performs as before in EFB
- [ ] Fueling
  - [ ] Test GSX properly plays the re-fueling animation when you begin refueling with the EFB
  - [ ] Test the refueling animation correctly stops when the EFB stops refueling the aircraft
  - [ ] Test the refueling animation is instant when you're using Instant refueling time
  - [ ] Ensure you can still refuel in-flight with GSX Fuel Sync enabled
- [ ] Payload
  - [ ] Set in the EFB how much passengers you want as-well as cargo, to notify GSX of these details. 
  - [ ] Ensure passenger current in EFB, matches GSX tooltip
  - [ ] Confirm as GSX baggage tooltip percentage increase, current cargo weight increase in EFB
  - [ ] Ensure baggage weight decreases when deboarding
  - [ ] Ensure passenger amount decreases when deboarding
  - [ ] Ensure passenger/cargo amount in EFB cannot be modified after boarding is requested and only appears again after deboarding is completed 


2hwk:
This will now probably be delayed until after ``boarding.rs`` rust implementation is completed

2hwk:
Note: Check for conflicts with https://github.com/flybywiresim/a32nx/pull/7804

Lucky38i (approved):
This is an approval for the general logic and GSX flow but I can't confirm the integrity of the bit logic as I'm not aquainted enough with it.

fbw-a32nx/src/wasm/systems/a320_systems/src/payload/mod.rs
```diff
@@ -0,0 +1,663 @@
...
+            "PAYLOAD_STATION_7_REQ",
+        ),
+        CargoInfo::new(
+            Mass::new::<kilogram>(1497.),
+            "CARGO_AFT_BULK_LOOSE",
+            "PAYLOAD_STATION_8_REQ",
+        )
+    ]);
+}
+
```
mjuhe, line 84: I don't really think you need this. You can `impl Display` for the `A320Pax` and `A320Cargo` enums to later construct the VariableIdentifiers.

Suggestion: 

impl Display (you'll need another one for `A320Cargo`)

```suggestion
impl Display for A320Pax {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            A320Pax::A => write!(f, "PAX_A"),
            A320Pax::B => write!(f, "PAX_B"),
            A320Pax::C => write!(f, "PAX_C"),
            A320Pax::D => write!(f, "PAX_D"),
        }
    }
}

```

You will have to deal with the max_pax somewhere else. Either in the enum itself by storing the value, or in `Pax`. If you do this, you might be able to remove `PaxInfo` and `CargoInfo` all together.

fbw-a32nx/src/wasm/systems/a320_systems/src/payload/mod.rs
```diff
@@ -0,0 +1,663 @@
...
+        let mut pax = Vec::new();
+
+        for ps in A320Pax::iterator() {
+            pax.push(Pax::new(
+                context.get_identifier(A320_PAX[ps].pax_id.to_owned()),
+                context.get_identifier(format!("{}_DESIRED", A320_PAX[ps].pax_id).to_owned()),
+                context.get_identifier(A320_PAX[ps].payload_id.to_owned()),
+                Rc::clone(&per_pax_weight),
+            ));
+        }
```
mjuhe, line 203: Linked to the comment above, you can directly use your enums here to construct the identifiers.

```suggestion
        for (id, ps) in A320Pax::iterator().enumerate() {
            pax.push(Pax::new(
                context.get_identifier(format!("{}", ps).to_owned()),
                context.get_identifier(format!("{}_DESIRED", ps).to_owned()),
                context.get_identifier(format!("PAYLOAD_STATION_{}_REQ", (id + 1)).to_owned()),
                Rc::clone(&per_pax_weight),
            ));
        }
```

fbw-a32nx/src/wasm/systems/a320_systems/src/payload/mod.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    fn per_pax_weight(&self) -> Mass {
+        self.per_pax_weight.get()
+    }
+}
+impl SimulationElement for A320Payload {
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+        for ps in 0..self.pax.len() {
+            self.pax[ps].accept(visitor);
```
mjuhe, line 631: You can use the `accept_iterable!` macro instead.

```suggestion
        accept_iterable!(self.pax, visitor);
        accept_iterable!(self.cargo, visitor);
```

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+        per_pax_weight: Rc<Cell<Mass>>,
+    ) -> Self {
+        Pax {
+            pax_id,
+            pax_target_id,
+            per_pax_weight,
+            payload_id,
+            pax_target: 0,
+            pax: 0,
+            payload: Mass::new::<pound>(0.),
```
mjuhe, line 93:
```suggestion
            payload: Mass::default(),
```

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+        let pax_diff = self.pax_target_num() - self.pax_num();
+
+        let n: u64 = if pax_diff > 0 {
+            !self.pax & self.pax_target
+        } else {
+            self.pax & !self.pax_target
+        };
+        let count = n.count_ones() as i8;
+        if count > 0 {
+            let mut skip: i8 = rand::thread_rng().gen_range(0..count);
```
mjuhe, line 152: You can use `random_from_range()` from systems::shared.

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+        cargo_target_id: VariableIdentifier,
+        payload_id: VariableIdentifier,
+    ) -> Self {
+        Cargo {
+            cargo_id,
+            cargo_target_id,
+            payload_id,
+            cargo: Mass::new::<kilogram>(0.),
+            cargo_target: Mass::new::<kilogram>(0.),
+            payload: Mass::new::<pound>(0.),
```
mjuhe, line 207:
```suggestion
            cargo: Mass::default(),
            cargo_target: Mass::default(),
            payload: Mass::default(),
```

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    pub fn load_cargo_percent(&mut self, percent: f64) {
+        self.cargo = (percent / 100.) * self.cargo_target;
+        self.load_payload();
+    }
+
+    pub fn reset_cargo_and_target(&mut self) {
+        self.cargo = Mass::new::<kilogram>(0.);
+        self.cargo_target = Mass::new::<kilogram>(0.);
```
mjuhe, line 258:
```suggestion
        self.cargo = Mass::default();
        self.cargo_target = Mass::default();
```

mjuhe (commented):
Left some comments on your rust code. I didn't review the Air Conditioning files as they will have to be modified when #7804 gets merged anyway.

fbw-a32nx/src/wasm/systems/a320_systems/src/payload/mod.rs
mjuhe: Note for `cargo` you'll need to have `id + 5` instead of 1.

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
Gurgel100: Also this doesn't work in the wasm environment. `random_from_range` works.

fbw-a32nx/src/wasm/systems/a320_systems/src/payload/mod.rs
2hwk: I'll try and think of a better solution for this overall, perhaps in this PR or in the next refactor

2hwk:
Note: Need to resolve an issue with unit conversion already noted by testers

2hwk:
QA Report

- [x] Go to Payload page, 
- [x] Adjust seating by clicking on the seating layout. This should add new seats and increase the table value for planned passengers.
- [x] Add baggage by clicking on the 4 sliders, this should affect the desired cargo value in the table.
- [x] Change per passenger and per bag weight from 84 and 20 to higher/lower values (or leave at default)
- [x] Input desired passenger and desired cargo in the table
- [x] Check that ZFW calculation is accurate (Empty Weight is 42500KG, add passenger, passenger bags and cargo weight to this value and this should match desired ZFW)
- [x] Board passengers with desired time setting (Real/Fast/Instant). Should be able to change speed while passengers are boarding and have it transition between boarding times seamlessly.
- [x]  Deboard passengers (either by inputting in the table, or clicking on the seating layout, or using the deboard button).
- [x]  Go to the main EFB dashboard by clicking the FBW Logo in the top left, import a simbrief OFP
- [x] Go back to the payload page and click the cloud icon next to the table to load the OFP. Check that all relevant table values are correctly imported.
- [x]  Load passengers/cargo again
- [x] Proceed with normal flight operations. Check that when in air, MTOW envelope is replaced by the FLIGHT envelope.
- [x] Check that visuals in all themes (dark/light/blue) are correct and readable, with both bright daytime and at night (in sim time)
- [x] Press COND page and check that air conditioning is still functioning as expected when the air conditioning knobs are adjusted. 
- [x] Create a new save file with the new seating loadout, then load it. The seating layout should show when you go into the EFB's payload page.
- [x] Check that all the above steps work for LBS as well as KG. For simbrief import, KG exports should import correctly into the flypad in both KG/LBs, LBS export should import correctly to the flypad in both KG/LBS

Negatives:
 - Bug with repeatedly mashing clicks on seating map is still not fixed, but this is a somewhat difficult case to crack and not addressed by this PR.
 
 Result:
  - Pass

PR #7725, opened 2023-01-20
fix: Fuel system center tank transfer

donstim:
Much of the current fuel system architecture was carried over from the default Asobo A320NEO. Asobo incorrectly put the A320CEO fuel system architecture into the A320NEO in terms of fuel burned directly from the center fuel tank. For the A320NEO, fuel is transferred from the center tank to the inner tanks from which it flows to the engines. This transfer takes place under certain conditions (fuel in the center tank and inner tank fuel level less than about 500kg lower than the respective inner tank fuel capacity.).
## Summary of Changes

Center tank fuel will now transfer to the inner tanks rather than being fed directly to the engines. This transfer will take place when the inner tank fuel level is 500 kg less than the inner fuel tank capacity. The transfer will stop when either there is no center tank fuel, or the respective inner tank is filled.
## References

Screenshot (2656)
(Note:  We do not have additional center tanks (ACTs).)
Screenshot (2657)

## Additional context

### DIFF
--- a/fbw-a380x/docs/a380-simvars.md
+++ b/fbw-a380x/docs/a380-simvars.md
@@ -1,0 +1,60 @@
+# A380 Local SimVars
+
+## Contents
+
+- [A380 Local SimVars](#a380-local-simvars)
+  - [Indicating/Recording ATA 31](#indicating-recording-ata-31)
+
+## Indicating/Recording ATA 31
+
+- A32NX_CDS_CAN_BUS_1_1_AVAIL
+  - Bool
+  - Indicates if the first CAN bus of the CDS on the captain's side is available
+
+- A32NX_CDS_CAN_BUS_1_2_AVAIL
+  - Bool
+  - Indicates if the first CAN bus of the CDS on the captain's side is available
+
+- A32NX_CDS_CAN_BUS_2_1_AVAIL
+  - Bool
+  - Indicates if the first CAN bus of the CDS on the first officer's side is available
+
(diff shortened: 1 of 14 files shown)


flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/flight_model.cfg
```diff
@@ -86,53 +86,49 @@ point.8 = 2, -75.20,   0,  32.3,   100,  0,       0,   0,      0,     0,     0,
 fuel_type = 2 ; 1 = OCTANE 100, 2 = JET_A, 3 = OCTANE 80, 4 = AUTO GAS, 5 = JET B
 
 [FUEL_SYSTEM]
+Version = Latest
```
tracernz, line 89: Best to fix this to the latest version number so it won't get broken by a future update to the sim.
[From Sylvain](https://devsupport.flightsimulator.com/questions/11585/fuel-system-bug-junctions-cannot-have-all-inputs-c.html?childToView=13404#answer-13404):
> Also switch your version from Latest to 4 for SU11. Latest is a very risky setting as it will change between updates.

donstim: Nit so sure about that. They talk out of both sides of their mouth. They say that, but notice that the default airplanes use "Version = Latest."

PR #7790, opened 2023-02-11
fix: ESLint for base JS files

Lucky38i:
## Summary of Changes

Currently VSCode reads from eslintignore whereas IntelliJ reads from eslintrc.js. This presents issues for users working inside of the base JS files of A32NX where they get no linting at all when using VSCode. 

This PR aims to ensure regardless of IDE there correct linting rules are adhered to.

There is also a custom eslintrc for base JS files that aims to maintain a useful level of rules for these files without require a rewrite of most classes.

## Additional context

Discord username (if different from GitHub): Lucky38i
## Testing instructions

Doesn't require Testing


.eslintrc
```diff
@@ -7,7 +7,32 @@
...
+    "fbw-a32nx/src/systems/fmgc/src/flightplanning/WaypointBuilder.ts",
+    "fbw-a32nx/src/systems/fmgc/src/flightplanning/WorldMagneticModel.ts",
+    "fbw-a32nx/src/systems/fmgc/src/utils/LzUtf8.js",
+    "target/*",
+    "CMakeFiles/*",
+    "cmake*",
+    "fbw-a32nx/out/*",
+    "fbw-a380x/out/*"
   ],
   // DO NOT CHANGE THIS FILE (unless you are an admin) (if admin, make sure you change the same file in the latest asobo branch as well)
```
frankkopp, line 37: This can be removed or changed I guess :)

frankkopp (approved):
If this still works for the VSCode users it lgtm

.eslintrc
Lucky38i: I don't think we want linting on our generated files

frankkopp: I meant the comment `// DO NOT....`

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+    rules: {
+        'strict': 'off',
+        'no-unused-vars': 'warn',
+        '@typescript-eslint/no-unused-vars': 'warn',
+        'no-underscore-dangle': 'warn',
```
tracernz, line 8: I suggest removing this. A ton of the Asobo files use leading underscores heavily. This rule also makes getters and setters a bit awkward in general.

tracernz (commented):
How do these rules look on the existing code? We might need to tweak them a bit to suit the existing conventions.

It's always worked fine in vscode for me without these changes, at least for brace style etc. Not sure what's doing it 😆 .

tracernz (dismissed):
Looking at the fixup commit, the rules are wrong. Our convention is double quotes for strings in JS files.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../../../.eslintrc",
```
Benjozork, line 2: Why not just put the rules in the root `.eslintrc` here? They are only meant for `base/` anyway.

tracernz: The A380 should not use any legacy JS.

PR #7832, opened 2023-02-27
chore: Added the --no-cache parameter to the build scripts for clean build

frankkopp:
## Summary of Changes
With the copying of base files from src/base to out/ files removed from src/base remain in out/ until manually removed in out/ as well our the complete out/ folder is removed. 

This PR changes the build scripts to also use the --no-cache parameter normally used by Igniter to to remove the out/ folder completely therefore forcing a fresh clean build.

## Testing instructions
QA can't test this
Devs can test is the build script respects the new paramter and that it otherwise works as before. 

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -14,7 +14,7 @@ fi
 # Loop through the arguments
 for arg in "$@"; do
   if [ "$arg" == "--no-cache" ]; then
-    echo "Cleaning build directories"
+    echo "Removing out directories..."
     rm -rf /external/fbw-a32nx/out
     rm -rf /external/fbw-a380x/out
     rm -rf /external/fbw-ingamepanels-checklist-fix/out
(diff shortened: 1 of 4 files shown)


2hwk (approved):

tracernz:
Please change this to use a separate flag rather than igniter's --no-cache, as it now makes --no-cache fail while the sim is running (due to not being able to erase the sound pcks).

frankkopp:
> Please change this to use a separate flag rather than igniter's --no-cache, as it now makes --no-cache fail while the sim is running (due to not being able to erase the sound pcks). Additionally, does it play well with `--no-cache -r xxx`?

Ok, can do this with an extra -clean parameter. That's what I had first bad thought it would be redundant. But you're right of course. Unfortunately already merged 😀   
Will do another PR.  

The second should work with no issues as Args are passed on as before.

PR #7880, opened 2023-03-18
fix(fms): approach names

tracernz:
Fixes #[issue_no]

## Summary of Changes

Fix approach names so they always appear correctly.

## Additional context

## Testing instructions

Check a few airports, especially those with circling approaches, and make sure there are no approaches in the ARRIVAL page without a name.


BlueberryKing (commented):
Looks good! I think a small improvement could be to separate the formatting from the parsing, i.e you have one function parsing `string | RawApproach` to `ApproachNameComponents` and then have `shortApproachName` and `longApproachName` simply format `ApproachNameComponents` as a `string`. However, the functions are small enough that this is not critical.

tracernz:
> Looks good! I think a small improvement could be to separate the formatting from the parsing, i.e you have one function parsing `string | RawApproach` to `ApproachNameComponents` and then have `shortApproachName` and `longApproachName` simply format `ApproachNameComponents` as a `string`. However, the functions are small enough that this is not critical.

Both options are available to the caller with the overloads. It's just more convenient to be able to pass whatever in and get the string out.

BlueberryKing:
> > Looks good! I think a small improvement could be to separate the formatting from the parsing, i.e you have one function parsing `string | RawApproach` to `ApproachNameComponents` and then have `shortApproachName` and `longApproachName` simply format `ApproachNameComponents` as a `string`. However, the functions are small enough that this is not critical.
> 
> Both options are available to the caller with the overloads. It's just more convenient to be able to pass whatever in and get the string out.

Ah, I see it now, you're totally right!

BlueberryKing (approved):

PR #7804, opened 2023-02-17
feat(press): merge pressurization and air conditioning system

mjuhe:
## Summary of Changes

This PR merges the pressurization and air conditioning systems into one (ATA21). It creates a single cabin simulation, so pressure changes affect temperature and temperature changes affect pressure. This PR is a big refactor that intends to improve the press system and make it scalable for future implementations (aka for the A380).

Most of the changes will be completely unnoticable to the end user, but here is a list of the main ones:
- [x] Pressurization system and pressurization overhead moved to the aircraft specific portion of the code
- [x] RCPU system moved to aircraft specific portion of the code
- [x] Adjusted the volume of the pressurized cabin from references
- [x] One single air simulation for the cabin: so changing cabin temperature affects cabin pressure, and changing pressure affects temperature
- [x] Moved all aircraft specific constants out of `systems`, passed as generic argument
- [x] Passenger quantity and number of open doors are now read in the aircraft specific portion of code
- [x] Opening doors affects cabin pressure
- [x] Improved outflow valve flow out calculation (separate functions for supersonic and subsonic flow)
- [x] Added baro correction value to ADIRS
- [x] Connected ADIRS as input to CPCs (Arinc429)
- [x] Changed LVAR used for FMS landing elev to `FM1_LANDING_ELEVATION`
- [x] Fixed landing and departure cabin altitude calculation
        - When in auto, it uses FMS landing elev -> baro correction -> standard
        - When in manual, it uses standard only
- [x] Added outflow valve motor and improved overall OFV code design
- [x] Connected outflow valve to electrical system
- [x] The cabin vertical speed transmitted to the ECAM is limitted to +/- 6400fpm as in the real aircraft
- [x] Many many other refactor improvements 😄 

## References

The usual: FCOM, AMM, and others.

## Additional context

This is a big refactor, but there are still things to do!
Future work:
- Connect CPCs to electrical system.
- Failures.
- Use ARINC for outputs.
- Use a single LVAR for landing elevation everywhere (now two are being used)
- Connect the baro correction in the ADIRS to the PDF

Discord username (if different from GitHub): Commander Ga#1548

## Testing instructions

As mentioned, not a lot of noticeable changes in this PR so the objective is to ensure the pressurization system still works normally. You can use the testing instructions in #5159 and #5504 as guideline, but you don't have to follow every single point there.

[...]

PR #7883, opened 2023-03-18
fix(press): cabin pressure when starting in the air

mjuhe:
## Summary of Changes

This PR fixes an issue introduced in #7804 where spawning in the air was triggering a CAB PRESS EXCESS CAB ALT warning. Three fixes were needed:
- Initialization of cabin temperature: ambient when starting on ground, 24degC when starting in the air.
- Outflow valve starting state: was open, now closed. Note that when starting on the ground the OFV will take a couple of seconds to fully open now as soon as batteries are started.
- ADIRS data capture initialization in the CPCs.

## Screenshots (if necessary)

image

## Additional context

## Testing instructions

Spawn in the air and ensure the cabin pressure is initialized at a _reasonable_ altitude (see below for guidance):

**Aircraft Alt -> Cabin Alt (ft)**

39000 -> 8000
30000 -> 6000
20000 -> 4000
5000 -> 1000

Spawn C&D and on the runway and ensure the OFV opens as soon as you load the sim or turn on the batteries (of course without power you won't be able to see the ECAM page - if the OFV is open when the ECAM turns on, then it worked - or you can have a look at the simvar).

### DIFF
--- a/fbw-a32nx/src/wasm/systems/a320_systems/src/air_conditioning.rs
+++ b/fbw-a32nx/src/wasm/systems/a320_systems/src/air_conditioning.rs
@@ -1540,6 +1540,30 @@ mod tests {
         }
 
         #[test]
+        fn pressure_initialization_works() {
+            let test_bed = test_bed()
+                .ambient_pressure_of(InternationalStandardAtmosphere::pressure_at_altitude(
+                    Length::new::<foot>(39000.),
+                ))
+                .ambient_temperature_of(ThermodynamicTemperature::new::<degree_celsius>(-50.))
+                .iterate(2);
+
+            assert!(
+                (test_bed.cabin_altitude() - Length::new::<foot>(7800.)).abs()
+                    < Length::new::<foot>(50.)
+            );
+            assert!(
+                (test_bed.cabin_pressure()
+                    - InternationalStandardAtmosphere::pressure_at_altitude(Length::new::<foot>(
+                        7800.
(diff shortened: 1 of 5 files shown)


Gurgel100:
I think for the initialisation of the cabin pressure and temperature you should just use the values from `UpdateContext`. I don't think you can get simvars already with the `InitContext` right?

2hwk (approved):

tracernz:
Probably want to use basic sim values like `ON GROUND` to determine the initial state. Other subsystems might not yet be in a good state to use them. That is okay as this is a sim-specific thing rather than real system behaviour.

2hwk:
Quality Assurance Report

Discord : 2Cas#1022
Tier of Testing  : 1
Date: 19/03/23

Testing ProcessSpawn in the air and ensure the cabin pressure is initialized at a reasonable altitude (see below for guidance):

Aircraft Alt -> Cabin Alt (ft)

- [x] 39000 -> 8000
- [x] 30000 -> 6000
- [x] 20000 -> 4000
- [x] 5000 -> 1000

- [x] Spawn C&D and on the runway and ensure the OFV opens as soon as you load the sim or turn on the batteries (of course without power you won't be able to see the ECAM page - if the OFV is open when the ECAM turns on, then it worked - or you can have a look at the simvar).

Negatives:
- N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7893, opened 2023-03-21
fix(press): departing elevation calculation in cpc

mjuhe:
## Summary of Changes

This PR fixes an issue introduced in #7804 where the CAB PR EXCESS CAB ALT was being triggered when departing from a high altitude airport (over 8000ft). The warning should be inhibited when the cabin altitude is within 1000ft from departing or landing airfields. This was not happening as the departing airfield data was not being computed correctly.

This PR moves the calculation of the departing altitude to the CPCs (instead of taking it from the FMS) - a more accurate representation of the real system.

## Screenshots (if necessary)

image

## References

FCOM, AMM.

## Additional context

Thanks to Gorka#9245 for raising the issue on Discord.

## Testing instructions

Take off from an airport with an elevation greater than 8000 ft. Validate that after take off:
1. The cabin altitude does not climb higher (stays at the take off altitude).
2. No CAB PR EXCESS CAB ALT warning goes off in the cabin.


beheh (approved):

fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
```diff
@@ -222,13 +223,18 @@ impl<C: PressurizationConstants> CabinPressureController<C> {
 
             new_exterior_altitude =
                 self.calculate_altitude(self.exterior_pressure.output(), self.reference_pressure);
-            self.exterior_vertical_speed.update(
-                context.delta(),
-                self.calculate_exterior_vertical_speed(context, new_exterior_altitude),
-            );
+            // When the reference pressure changes, we skip the update to the external
+            // V/S to avoid a jump
+            if self.previous_reference_pressure == self.reference_pressure {
```
beheh, line 228: A tiny note: It feels a bit icky to compare f64s with equality, even though this probably works. What maybe a little nicer is if you use [`f64::EPSILON`](https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.EPSILON) to check it.

Here's some possible inspiration for how our APU solves it:
https://github.com/flybywiresim/a32nx/blob/8b40cb9a9bca1ed962bdd0000f0e65ccf7e40540/fbw-common/src/wasm/systems/systems/src/apu/air_intake_flap.rs#L58-L59

mjuhe: Thanks for the review and the comment @beheh. I've changed it as suggested in two areas where I had direct comparisons of `f64`.

mico975:

Discord          : mico#3145
Object of testing: #7893
Tier of Testing  : 1
Date             : 29/03/2023

Testing Process: 
Take off from an airport with an elevation greater than 8000 ft. (Departed SKBO - 8200ft) Validate that after take off:

The cabin altitude does not climb higher (stays at the take off altitude). - **NOT OK**, climbed to 8450ft abruptly around passing FL150 (cabin V/S momentarily jumped to +2000ft/min), then stayed at 8450ft the whole time
No CAB PR EXCESS CAB ALT warning goes off in the cabin. - **OK**

Negatives:       
N/A

Testing Results: 
Not passed

Conclusions: 
Almost

mjuhe:
Thanks for the test @mico975! What you saw with the cabin altitude changing abruptly was normal. This is because the calculation of the cabin altitude is done using the local altimeter when the aircraft is within 5000ft from departing or landing airfields, and using standard sea level pressure when above.

When you crossed 14200ft (or thereabouts, depending on the local QNH that day), the cabin altitude re-calculated using 1013.25 HPa. 

What was not intended was the vertical speed jumping to +2000ft/min! I think this was due to the way turbulence affected the calculation above. I've added a fix that should solve this problem.

Interesting note: if the airport you are taking off from is around 8000ft in elevation, you might see that you end up with the cabin altitude being below 8000ft when it recalculates. This will trigger the cabin to pressurize again up to around 8050ft, then stop climbing. So you might see a positive cabin V/S when crossing 14200ft ish, depending on your QNH, but never over 1000ft/min.

brewers10:

Discord: brewers_10#5706
Object of testing: #7893 
Tier of Testing  : 1
Date: 06/04/2023

Testing Process: 
Followed steps above departing from SKBO

Negatives:       
N/A

Testing Results: 
Cabin altitude remained at 8100ft on departure with no master cautions. As mentioned in response to Miko's test re. cabin alt dipping around 14,000ft. The max fpm used was 450fpm to get the cabin alt back up to 8100ft, so no issues from me. Good job!

Conclusions: 
LGTM

Media:       
Just a little picture
Microsoft Flight Simulator Screenshot 2023 04 06 - 17 08 59 02

PR #7902, opened 2023-03-25
feat: barometric (b) key handling

### LINKED ISSUE #6786: [ISIS] pressing B changes ALT while in STD mode
### Aircraft Version

Experimental

### Build info

```json
{
  "built": "2022-02-22T10:29:25+00:00",
  "ref": "experimental",
  "sha": "c736b13b89efac7b7e819d1d4e010a350f777a0d",
  "actor": "aguther",
  "event_name": "manual",
  "pretty_release_name": "experimental:c736b13b"
}
```

### Describe the bug

I like to use the B key, to quickly set altimeters. Today I thought about, wether its possible to "preset" the QNH while on STD, so it would match the second I turn back to QNH. So i pressed B in cruise and I noticed the ALT on the ISIS changed. I changed the ISIS to QNH (which was still at previous setting) and back to STD and the ALT was correct again. Hitting B again changes the ALT again.

I guess the STD mode also has its own altimeter setting which you can change, because QNH mode had a different setting.

### Expected behavior

Nothing should change, when on STD.

### Steps to reproduce

1. Set QNH to something other than current QNH
2. Change to STD
3. Press B key -> ALT changes
4. Change to QNH (shows old QNH from 1.)

### References (optional)

IRL experience

_No response_

_No response_

tracernz:
Fixes #6786

## Summary of Changes

Handle the 'B' key ourselves, and set QNH/STD depending on the transition altitude and transition level entered in the MCDU, defaulting to 18000/FL180. This matches behaviour of default avionics like the CJ4, and improves the usefulness of the key for non-USA simmers. Also fixes the ISIS so that pressing the B key while in STD mode doesn't change the underlying baro correction value to a QNH.

## Additional context

## Testing instructions

Set up a short flight plan and ensure transition altitude is set on PERF TO, and transition level on PERF APPR. Check that pressing the B key on the takeoff runway sets the local QNH, press it during climb prior to trans alt and make sure it continues to set local QNH... press it after trans alt and ensure the altimeter switches to STD. Press at various points to make sure it stays STD. During descent press before and after trans level and ensure it changes to local QNH below trans level and keep trying to ensure the same behaviour at various points.


fbw-a32nx/src/systems/extras-host/index.tsx
```diff
@@ -40,18 +43,6 @@ class ExtrasHost extends BaseInstrument {
...
-        super();
-
-        this.bus = new EventBus();
-        this.hEventPublisher = new HEventPublisher(this.bus);
-
-        this.versionCheck = new VersionCheck(this.bus);
-        this.keyInterceptor = new KeyInterceptor(this.bus);
-
-        console.log('A32NX_EXTRASHOST: Created');
-    }
```
frankkopp, line 53: Is there a specific reason you prefer to avoid the construcor?

I'd at least want to keep the log output so finding issues is easier.

fbw-a32nx/src/systems/extras-host/modules/baro_handler/BaroHandler.ts
```diff
@@ -0,0 +1,123 @@
...
+                    this.altitude.setFromSimVar(`L:A32NX_ADIRS_ADR_${n}_BARO_CORRECTED_ALTITUDE_1`);
+                }
+                if (!climb && !this.altitude.isNormalOperation()) {
+                    // use pressure alt as we're descending through the transition level
+                    this.altitude.setFromSimVar(`L:A32NX_ADIRS_ADR_${n}_ALTITUDE`);
+                }
+            }
+
+            if (!this.altitude.isNormalOperation()) {
+                // we can't do anything as there are no healthy ADRs
```
frankkopp, line 46: Might warrant a console output (log, warning) - not fully clear to me if this is a kind of error condition or not.

frankkopp (changes requested):
You could have hooked this into the KeyInterceptor and done all the logic in a separate class. Why did you create another class intercepting keys?
I think it would be cleaner to keep these in one class. Of course then the Ctrl+E stuff could be moved out as well to be consistent.

fbw-a32nx/src/systems/extras-host/modules/baro_handler/BaroHandler.ts
tracernz: Just means all the ADRs are off or failed. Not a code error.

tracernz:
> You could have hooked this into the KeyInterceptor and done all the logic in a separate class. Why did you create another class intercepting keys? I think it would be cleaner to keep these in one class. Of course then the Ctrl+E stuff could be moved out as well to be consistent.

I don't think that's a good idea at all. I'd rather have each separate function encapsulated in it's own class rather than mixing a whole bunch of different things into a messy file full of unrelated stuff. The KeyInterceptManager (KeyEventManager with framework 0.4) is designed for exactly this.

tracernz:
> You could have hooked this into the KeyInterceptor and done all the logic in a separate class. Why did you create another class intercepting keys?
I think it would be cleaner to keep these in one class. Of course then the Ctrl+E stuff could be moved out as well to be consistent.

This is the style I am used to using with the framework in my day job. As you're the maintainer of this code, I'll defer to you, and put the intercept in KeyInterceptor.

frankkopp (approved):

2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #7902
Tier of Testing  : 1
Date             : 29/03/2023

Testing Process: 
Set up a short flight plan and ensure transition altitude is set on PERF TO, and transition level on PERF APPR. Check that pressing the B key on the takeoff runway sets the local QNH, press it during climb prior to trans alt and make sure it continues to set local QNH... press it after trans alt and ensure the altimeter switches to STD. Press at various points to make sure it stays STD. During descent press before and after trans level and ensure it changes to local QNH below trans level and keep trying to ensure the same behaviour at various points. - **All OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

.github/CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 1. [MCDU] Implemented TUNE BBB FFF.FF, SPECIF NDB UNAVAIL messages - @tracernz (Mike)
 1. [F/CTL] Add ELAC 2 Emergency Powersupply - @lukecologne (lukecologne#1156)
 1. [HYD] New aerodynamic model for RAT turbine / Added A320 RAT anti stall valve system - @Crocket63 (crocket)
+1. [MISC] Handle the BAROMETRIC (B) key event and set QNH/STD depending on FMS data - @tracernz (Mike)
```
tracernz, line 82: Changelog 0.12

PR #7904, opened 2023-03-25
feat(fms): handle non-msfs approach transition coding

tracernz:
Fixes #[issue_no]

## Summary of Changes

Navigraph does not separate approach transitions out as expected by MSFS, and as a consequence many transitions are not available when using their navdata. We now find IAFs embedded within other transitions (as allowed by ARINC424, not MSFS), and append them to the list of available transitions. These transitions cannot be synced to the MSFS flight plan system when selected as they do not exist there, but that is better than not having them available at all. Let's hope Navigraph will rethink this decision in future.

Also fixed a couple of small formatting issues on the page while I was there (it now shows "NO VIA" instead of "NO VIAS", and the selected VIA is highlighted in green.

## Screenshots (if necessary)

Before with Navigraph data, KIAD R19CY with DIMKE transition missing:
image

After:
image

Default MSFS data (CRVER transition is missing entirely from the Navigraph/Jeppesen data...):
image
image

## Additional context

## Testing instructions

- With Navigraph data installed, check an airport that was missing some approach transitions before, and see if they are now present.
- Check that approach transition page scrolling works
- Check that approach transitions are strung into the flightplan when selected


fbw-a32nx/src/systems/fmgc/src/flightplanning/RawDataMapper.ts
```diff
@@ -60,11 +61,36 @@ export class RawDataMapper {
...
+
+                    // Fix up navigraph approach transitions which hide IAPs inside other transitions rather
+                    // than splitting them out as they should be in MSFS data. Unfortunately this means
+                    // these transitions cannot be synced to the sim's flight plan system for ATC etc. as
+                    // they're not visible without this hack.
+                    // Note: it is safe to append to the array inside the forEach by the ECMA spec, and the appended
+                    // elements will not be visited.
+                    for (let i = 1; i < trans.legs.length; i++) {
+                        const leg = trans.legs[i];
+                        if ((leg.fixTypeFlags & FixTypeFlags.IAF) > 0 && (leg.type === LegType.TF || leg.type === LegType.IF)) {
```
2hwk, line 79: ``(leg.type === LegType.TF || leg.type === LegType.IF)``

Perhaps could be made into its own codeblock with a ``switch case`` instead which would be slightly more readable?
Although this also works and is understandable as is.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_AvailableArrivalsPage.js
```diff
@@ -305,15 +305,23 @@ class CDUAvailableArrivalsPage {
                 selectedStarCell = selectedArrival.name;
                 selectedStarCellColor = mcdu.flightPlanManager.getCurrentFlightPlanIndex() === 1 ? "yellow" : "green";
             }
+
+            // we need to sort the transitions alphabetically for display, but keep track of their original index for the flightplan
+            const availableTransitions = [];
+            if (selectedApproach) {
+                availableTransitions.push(...selectedApproach.transitions.map((transition, index) => [transition, index]));
+                availableTransitions.sort(([a], [b]) => a.name.localeCompare(b.name));
+            }
```
2hwk, line 314: This is quite a big object assignment (potentially), maybe need to test the performance impact in QA

2hwk (changes requested):

2hwk:
Also needs testing instructions

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_AvailableArrivalsPage.js
tracernz: It's creating an array with maybe 5-15 elements containing an array with 2 elements each, one being a reference to an object that already exists and the other a number. That should be pretty cheap. Ideally we wouldn't need to allocate at all, but in practice there isn't a good away around it, and this page loads nominally a couple of times per flight.

fbw-a32nx/src/systems/fmgc/src/flightplanning/RawDataMapper.ts
tracernz: I'd agree for more options, but 2 is more readable like this.

2hwk (approved):

PR #7921, opened 2023-03-29
fix: only update ewd memo lines on change

Saschl:
Fixes #[issue_no]

## Summary of Changes

Adds a check for the pfwc memos to make sure we only send updates when the messages actually change. This might improve performance and prevent  the memos from disappearing for a short moment (as we clear the messages on every update as well).

## Additional context

## Testing instructions

 - Make sure memos on right and left side of the EWD still appear and become updated correctly (check T.O. Memos and LDG memos as an example for the left messages. Parking Brake and Landing lights for the right messages).
 - Check failures appear and can be cleared.


Saschl:
> You could just put them into subjects...

What exactly? The messages are already `ArraySubjects`, but as we replace the whole array each time we also get notified each time. I think I know what you mean but I didn't figure out how to implement that yet.

tracernz:
```ts
private static readonly ewdMessageSimvars = Array.from({ length: EWD_MESSAGE_LINES }, (_, i) => `L:A32NX_EWD_LINE_${i}`);
private readonly ewdMessageLines = Array.from({ length: EWD_MESSAGE_LINES }, (_, i) => Subject.create(0));

// ...

// onAfterRender
this.ewdMessageLines.forEach((line, i) => line.sub((v) => SimVar.SetSimVarValue(ClassName.ewdMessageSimVars, 'number', v);

// ...

// calculating messages
this.ewdMessageLines[2].set(blah.code);
```
Avoid allocating all the time, and only write the simvar when needed (the sub takes care of that for you). I'd do the same on the other side with the strings as well if not already done so you only update the DoM when needed.

fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
```diff
@@ -1594,7 +1590,7 @@ export class PseudoFWC {
         this.failuresRight.push(...failureKeysRight);
 
         if (tempFailureArrayLeft.length > 0) {
-            this.memoMessageLeft.set(orderedFailureArrayLeft);
+            this.ewdMessageLinesLeft.forEach((l, i) => l.set(orderedFailureArrayLeft[i]));
```
tracernz, line 1593: Doesn't it need to set them to zero also if the tempFailureArrayLeft has nothing in it?

tracernz (commented):
I like it.

fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
Saschl: I think that's being dealt with in https://github.com/flybywiresim/a32nx/pull/7921/files/7c80f4c7011d3ac61914ff47d97fc2fd3a798c7c#diff-2439d534e1bce3f0a530cd3d6eefdf08aae38febfb1f9300a2ae75afed4661d5R1633 and following. At least I think the logic of this  isn't touched by my changes.

2hwk (changes requested):
Needs changelog

Saschl:
> Needs changelog

It’s just a fix m/improvement of a previous item (pfwc rewrite) so I don’t think a changelog is necessary here 👍

tracernz (approved):

PR #7927, opened 2023-04-01
fix(mcdu): no overfly when forced turn, return to f-pln on tmpy insert

tracernz:
Fixes #[issue_no]

## Summary of Changes

Couple of small tweaks that never got in before:
- Inserting the TMPY from the APPR VIAs page returns to F-PLN rather than ARRIVALS.
- The overfly symbol is not shown on the F-PLN page when a forced turn is shown.
  - Originally put both because we didn't know which has priority, in the hopes that somebody would notice and bring us the correct information. 😀 

## Additional context

## Testing instructions

Setup a flightplan with FROM/TO, select an approach, and a VIA. Insert the TMPY from the VIA page and ensure you are returned to the F-PLN page. Select a SID with forced turns and overflys and make sure you only get on or the other symbol next to the waypoint, not both.


2hwk (approved):

mico975:

Discord          : mico#3145
Object of testing: #7927
Tier of Testing  : 1
Date             : 03/04/2023

Testing Process: 
Setup a flightplan with FROM/TO, select an approach, and a VIA. Insert the TMPY from the VIA page and ensure you are returned to the F-PLN page. - **OK**
Select a SID with forced turns and overflys and make sure you only get on or the other symbol next to the waypoint, not both. - **OK**

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

PR #7939, opened 2023-04-02
fix: update mass units conversion (KG/LBS)

donstim:
Fixes #[issue_no]

## Summary of Changes
Updates units conversion in https://github.com/flybywiresim/a32nx/blob/master/fbw-a32nx/src/systems/instruments/src/Common/NXUnits.ts to be consistent with that used in https://github.com/flybywiresim/a32nx/blob/master/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/A32NX_Utils/NXUnits.js and https://github.com/flybywiresim/a32nx/blob/master/fbw-a32nx/src/systems/instruments/src/EFB/Ground/Pages/FuelPage.tsx.
## Additional context


donstim:
I don't know why we have units conversions coded separately in different places and don't know if this particular one is called anywhere anymore, but in case it is it should be consistent with what is used elsewhere.

tracernz (approved):

2hwk:
Needs Testing instructions. Note that this might have some deep (unintentional) impact so testing potentially will be quite extensive.

donstim:
> Needs Testing instructions. Note that this might have some deep (unintentional) impact so testing potentially will be quite extensive. Unfortunately I think the .ts library might be more used than the .js one, but I would need to assess this by looking at the code.

If that is true, this PR is not worth it. It changes the weight in lbs by 1 pound at the MTOW of 79T. Essentially round off error. This was intended just to make all weight cinversiobs consistent. Closed.

PR #7957, opened 2023-04-08
fix: off-by-one error in truncateWaypoints

icalder:
## Summary of Changes

- Fixes changing flight plan destination using NEW DEST from LAT REV page
- Screenshot below shows stack trace for error encountered in `ManagedFlightPlan.removeWaypoint` without the fix

## Screenshots (if necessary)
image

## Additional context
I've been struggling to make diverts work using the LAT REV page.  Several YT videos show this functionality on older releases of the a32nx and on other sim 320s.  Tried develop and experimental versions.  Without the fix, because of the JS errror the MCDU stays stuck on the LAT REV page and does not return to the flight plan page.

## Testing instructions
Create a flight plan with some waypoints.  Pick a waypoint and go to LAT REV.  Enter NEW DEST.  MCDU should return to the flight plan page with a temp flight plan ready for review and the new destination inserted at the end.


2hwk (approved):

fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanManager.ts
```diff
@@ -1032,7 +1032,7 @@ export class FlightPlanManager {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     public truncateWaypoints(index: number, thenSetActive = false, callback = () => { }): void {
         const fp = this._flightPlans[this._currentFlightPlanIndex];
-        for (let i = fp.length; i >= index; i--) {
+        for (let i = fp.length; i > index; i--) {
```
tracernz, line 1035: Shouldn't it start at `fp.length - 1` (i.e. the last index in the plan), and the original condition is then correct?

icalder: That would also work (but is one extra op ;-)).  `index` doesn't change so the loop is calling `removeWaypoint(index)` n times.  The first time in the loop the waypoint[index] is removed, that matches the doc. Remaining waypoints "shift left" so each time after that the succeeding waypoint is removed.  Just the calculation of n was wrong.

Letreivchen:

Discord          : Letreiv#9404
Object of testing: #7957 
Tier of Testing  : 1
Date             : 11/04/2023

Testing Process: 
Create a flight plan with some waypoints. Pick a waypoint and go to LAT REV. Enter NEW DEST. MCDU should 

 return to the flight plan page with a temp flight plan ready for review -- OK

 the new destination inserted at the end. --OK


Negatives:       
The new destination automatically has an approach selected, but only when the previous flight plan has an approach selected

Testing Results: 
Passed 

Conclusions: 
Testing instructions passed, maybe needs an accuracy check if the point under negatives needs to be corrected

Media:       
none

2hwk:
Congratulations on your first merged contribution to the A32NX repo, if you are on discord you can receive the contributor role on our discord, contact a moderator for help with this.

PR #7966, opened 2023-04-14
feat(efb): Added rounding up to nearest 100 for simbrief fuel auto fill

### LINKED ISSUE #7923: Round up fuel quantity
### Aircraft Version

Experimental

### Description

Currently the total fuel is being imported with an exact number as in simbrief. On the refuel panel of a real thing there are only the tenths place values (e.g. 7.4). So it would be more realistic if the imported values rounded up to the next hundred (e.g. 7349 -> 7400).

### References (optional)

_No response_

_No response_

K8#4795

MikioDK:
Fixes #7923

## Summary of Changes

When using the "Fill block fuel from Simbrief" button on EFB, the value is now rounded up to nearest 100 kg for the internal value. While the OFP in Dispatch still shows the simbrief value

## Screenshots (if necessary)

image
image

## Additional context

## Testing instructions

Ensure that the EFB is still functional and refueling works as expected for both KG and LBS setting.


2hwk (approved):

2hwk:
QA Test Report:
- [x] Ensure that the EFB is still functional and refueling works as expected for both KG and LBS setting

FlightSimulator_G6hToIDulS
FlightSimulator_WMblNogvOW
FlightSimulator_7w5g8Xazu4


Negatives:
- LBS can be off in block fuel display on upper ECAM by 20, however in reality you do not receive fuel to such a high degree of granularity to begin with, and this is the result of a rounding error between LBS and KG that is also dependent on the sim's unit conversion value for LBS to KG. 

This would also be an issue in dev branches outside of this PR, and is not caused by this change.

Result:
- Passed

Conclusions:
- LGTM

PR #7980, opened 2023-04-22
feat: custom wingflex simulation

crocket63:
Fixes #[issue_no]

## Summary of Changes

MSFS wing flex only provides one value depending on global lift for a single down to up animation.

In order to get maximum realism of a 500 tons plane, this PR will try to simulate the wing flex physically by splitting the wing in multiple flexible sections. 

This way we can have a wing with different behavior depending on fuel load, and flight control positions, as well as visually different oscillation modes (tip oscillating while other part of the wing is up or down).

- [x] Physical simulation of N elements of each wing
- [x] Each wing section has correct fuel mass
- [x] Plane global lift estimation in flight
- [x] Plane global lift estimation on ground
- [ ] Lift spread on the wing depending on flight controls positions
- [ ] Animations using multiple bones per wing
- [ ] Model tuning

## Additional context


fbw-a380x/src/wasm/systems/a380_systems/src/lib.rs
```diff
@@ -258,18 +256,18 @@ impl Aircraft for A380 {
             [self.lgcius.lgciu1(), self.lgcius.lgciu2()],
         );
 
-        self.engines_flex_physics.update(context);
-        self.elevators_flex_physics.update(
+        self.cds.update();
+
+        self.egpwc.update(&self.adirs, self.lgcius.lgciu1());
+
+        self.structural_flex.update(
```
2hwk, line 263: Admittedly I'm not that familiar with the code at such a low level, but shouldn't ``self.structural_flex.update()`` belong in ``update_before_power_distribution()`` and not ``update_after_power_distribution()`` ?

fbw-a380x/src/wasm/systems/a380_systems/src/structural_flex.rs
```diff
@@ -0,0 +1,77 @@
...
+            wing_flex: WingFlexA380::new(context),
+
+            surface_vibrations: SurfaceVibrationGenerator::default(),
+        }
+    }
+
+    pub fn update(
+        &mut self,
+        context: &UpdateContext,
+        outter_inner_elevator_aero_torques: [(Torque, Torque); 2],
```
2hwk, line 40:
```suggestion
        outer_inner_elevator_aero_torques: [(Torque, Torque); 2],
```

Typo, fix all associated usages of this param too

2hwk (changes requested):

fbw-a380x/src/wasm/systems/a380_systems/src/lib.rs
crocket63: No real rule about this when it comes to something that doesn't have electrical, but as it has dependencies on hydraulics which is updated after elec, better to put it there so it's updated at frame N with latest hydraulic values

fbw-a380x/src/wasm/systems/a380_systems/src/structural_flex.rs
crocket63: God! wrong everywhere :D

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+        let mut all_heights_meters = [0.; NODE_NUMBER];
+
+        for (idx, height) in all_heights_meters
+            .iter_mut()
+            .enumerate()
+            .take(NODE_NUMBER)
+            .skip(1)
+        {
+            *height = self.nodes[idx].position().get::<meter>();
+        }
```
Gurgel100, line 1130: This should be the same:
```suggestion
        for (height, node) in all_heights_meters[1..].iter_mut().zip(&self.nodes) {
            *height = node.position().get::<meter>();
        }
```

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+        all_heights_meters
+    }
+
+    fn acceleration_at_node_idx(&self, node_idx: usize) -> Acceleration {
+        assert!(node_idx < NODE_NUMBER);
```
Gurgel100, line 1135: Just as note:
Actually not really needed because indexing an array in Rust already does bound checks (only in debug mode though). So maybe it's a good idea to still have it.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+    fn fuel_masses(&self, fuel_tanks_masses: [Mass; FUEL_TANK_NUMBER]) -> [Mass; NODE_NUMBER] {
+        let mut masses = [Mass::default(); NODE_NUMBER];
+        for (idx, fuel) in fuel_tanks_masses.iter().enumerate() {
+            masses[self.fuel_tank_mapping[idx]] += *fuel;
+        }
+        masses
+    }
+}
+
+enum A380fuelTanks {
```
Gurgel100, line 503: Shouldn't the A380 specific enum and structs reside in the a380 systems crate?

Gurgel100 (commented):
Fantastic work. Only a couple of comments.

Gurgel100:
Probably needs a little bit of changing due to merging of #8143

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+    right_body_compression_id: VariableIdentifier,
+
+    center_compression: Ratio,
+    left_wing_compression: Ratio,
+    right_wing_compression: Ratio,
+    left_body_compression: Ratio,
+    right_body_compression: Ratio,
+}
+impl LandingGearWeightOnWheelsEstimator {
+    const GEAR_CENTER_COMPRESSION: &'static str = "CONTACT POINT COMPRESSION";
```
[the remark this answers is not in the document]
crocket63: None is :0 for simconnect

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+            + self.spoilers_right_position[6]
+            + self.spoilers_right_position[7])
+            / 6.;
+
+        let left_ailerons_mid = ((self.ailerons_left_position[0] - 0.5) * 2.
+            + (self.ailerons_left_position[1] - 0.5) * 2.)
+            / 2.;
+        let right_ailerons_mid = ((self.ailerons_right_position[0] - 0.5) * 2.
+            + (self.ailerons_right_position[1] - 0.5) * 2.)
+            / 2.;
```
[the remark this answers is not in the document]
crocket63: Problem with those is in the end I don't understand what it's doing while with "too much" multiplications compiler would probably end up with the same machine code :p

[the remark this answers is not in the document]
Gurgel100: Actually the compiler will not optimize the unecessary multiplications away because of floating point semantics. What I suggest instead is to write your formula as a comment before the statement.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
crocket63: Doubt it might hurt in release mode?

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+    LeftInner,
+    LeftFeed2,
+    LeftMid,
+    LeftFeed1,
+    LeftOuter,
+    RightInner,
+    RightFeed3,
+    RightMid,
+    RightFeed4,
+    RightOuter,
```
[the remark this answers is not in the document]
crocket63: I feel it's easy to get lost in those when doing out board to inboard iterators and so on.... Can't hurt to have a bit too much informations in the naming as you would more easily notice a left and a right in a left only formula whereas just mixing 2 and 3 might go under the radar :p

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+                    Self::FLEX_COEFFICIENTS,
+                    Self::DAMPING_COEFFICIENTS,
+                ),
+                FlexPhysicsNG::new(
+                    context,
+                    empty_mass,
+                    Self::FLEX_COEFFICIENTS,
+                    Self::DAMPING_COEFFICIENTS,
+                ),
+            ],
```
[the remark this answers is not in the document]
crocket63: Is the [1,2] just a way to make a dummy array of size 2 there?  I have issues to understand the .map stuff :D

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+
+        self.flex_physics[1].update(
+            context,
+            self.wing_lift_dynamic
+                .per_node_lift_right_wing_newton()
+                .as_slice(),
+            self.fuel_mapper
+                .fuel_masses(self.wing_mass.right_tanks_masses()),
+            surface_vibration_acceleration + self.left_right_wing_root_position[1].acceleration(),
+        );
```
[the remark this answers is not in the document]
crocket63: I must admit I'm unable to understand the code :D. I like it the old way there (we should stay at max 2 wings isn't it? :D )

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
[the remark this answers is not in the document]
Gurgel100: Yes. Another way (probably even better) would be `[0; 2]`. The map in this case just ignores the value it's passed.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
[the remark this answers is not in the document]
Gurgel100: I accept that it's a "little" bit less readable ;). Can stay like it is now.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
Gurgel100: Not really.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -568,25 +545,17 @@ impl WingMassA380 {
...
-                * MassDensity::new::<kilogram_per_cubic_meter>(Self::FUEL_MASS_DENSITY_KG_M3);
-        }
+        Self::tanks_masses(&self.right_tank_volumes)
+    }
 
-        masses
+    fn tanks_masses<const N: usize>(tank_volumes: &[Volume; N]) -> [Mass; N] {
+        tank_volumes.map(|volume| {
+            volume * MassDensity::new::<kilogram_per_cubic_meter>(Self::FUEL_MASS_DENSITY_KG_M3)
+        })
```
Gurgel100, line 558: Not quite the right place but still: You should probably get the fuel mass using the fuel system struct introduced with #8143 instead of reading the same variables again.

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+                .get_identifier("HYD_AIL_LEFT_MIDDLE_DEFLECTION".to_owned()),
+            aileron_left_3_position_id: context
+                .get_identifier("HYD_AIL_LEFT_OUTWARD_DEFLECTION".to_owned()),
+
+            aileron_right_1_position_id: context
+                .get_identifier("HYD_AIL_RIGHT_INWARD_DEFLECTION".to_owned()),
+            aileron_right_2_position_id: context
+                .get_identifier("HYD_AIL_RIGHT_MIDDLE_DEFLECTION".to_owned()),
+            aileron_right_3_position_id: context
+                .get_identifier("HYD_AIL_RIGHT_OUTWARD_DEFLECTION".to_owned()),
```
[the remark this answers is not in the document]
crocket63: Phew! Finally done

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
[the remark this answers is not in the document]
crocket63: Went for the new fuel payload system

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
```diff
@@ -0,0 +1,2089 @@
...
+        self.left_tank_volumes[1] = reader.read(&self.left_tank_2_id);
+        self.left_tank_volumes[2] = reader.read(&self.left_tank_3_id);
+        self.left_tank_volumes[3] = reader.read(&self.left_tank_4_id);
+        self.left_tank_volumes[4] = reader.read(&self.left_tank_5_id);
+
+        self.right_tank_volumes[0] = reader.read(&self.right_tank_1_id);
+        self.right_tank_volumes[1] = reader.read(&self.right_tank_2_id);
+        self.right_tank_volumes[2] = reader.read(&self.right_tank_3_id);
+        self.right_tank_volumes[3] = reader.read(&self.right_tank_4_id);
+        self.right_tank_volumes[4] = reader.read(&self.right_tank_5_id);
```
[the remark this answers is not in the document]
crocket63: Sorted through fuel payload system

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
crocket63: Done quickly by moving everything, but there's obviously one peace of abstraction still missing: I'd want to have a generic wing flex mechanism in common place, whereas now there's core logic in the A380 part.
Will do for now, will have time to think about this later!

fbw-common/src/wasm/systems/systems/src/structural_flex/wing_flex.rs
crocket63: Done

fbw-a380x/src/wasm/systems/a380_systems/src/structural_flex.rs
```diff
@@ -75,3 +87,1166 @@ impl SimulationElement for A380StructuralFlex {
...
+            lateral_offset: Ratio::default(),
+
+            left_wing_lift: Force::default(),
+            right_wing_lift: Force::default(),
+
+            standard_lift_spread: Vector5::from(Self::NOMINAL_WING_LIFT_SPREAD_RATIOS),
+            lift_left_table_newton: Vector5::default(),
+            lift_right_table_newton: Vector5::default(),
+        }
+    }
```
Gurgel100, line 126: Move to `Default` trait implementation.

Gurgel100 (commented):
One small change then all is good.

fbw-a380x/src/wasm/systems/a380_systems/src/structural_flex.rs
crocket63: Too much time without any line of Rust :D

Gurgel100 (approved):
Great work as always.

PR #7983, opened 2023-04-23
[DOCS] Some tiny tweaks to README.md

Simmer1234569:
Added a little bit more detail in the 'Is feature X from the A320neo going to be implemented?" and fixed a grammar mistake.

## Summary of Changes
Added a bit more detail in the "Is X feature from the A320 NEO coming to the FBW?' section and fixed a grammar mistake in "Is the A32NX payware?".

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Read [Contributing.md](.github/Contributing.md) and join our Discord to get star
 
 ### Is the A32NX payware?
 
-No, it is a completely free aircraft, open-source.
+No, it is a completely free aircraft,and open-source.
 
 ### How do we report bugs?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,7 +77,7 @@ While many in the team dislike the term "study-level" as its use is mostly appli
 
 ### Is feature X from the A320neo going to be implemented?
 
-Chances are, yes! While we do not guarantee every single detail of the aircraft will eventually be represented, our goal is to produce an extremely accurate simulation based on technical data and real-world testing. This means you can be almost certain every feature of the aircraft will eventually be simulated to the best of the simulator's ability as long as the technical data backs it up.
+Chances are, yes! While we do not guarantee every single detail of the aircraft will eventually be represented, our goal is to produce an extremely accurate simulation based on technical data and real-world testing. This means you can be almost certain every feature of the aircraft will eventually be simulated to the best of the simulator's ability as long as the technical data backs it up. This project is constantly getting updated, don't believe me then just look at our pull requests, so if you just suggest your idea either on the issues section of this repository or on our discord, we will try our best to include it.
```
frankkopp, line 80: I think the is not a good addition. First everything that needs to be said is already covered in the paragraph before. Also the "don't believe me" is really unnecessary. 
And last we do not include any feature just because someone writes an Issue. Why not is explained in the paragraph existing already.

frankkopp (changes requested):

README.md
Simmer1234569: I was saying that if someone suggests a feature, we will try our best to add it to add a sense of community and so they know how to help out.

frankkopp: But we won't try "our best"" if it will be outside of what it is said in the current paragraph. And if it is within the meaning of current paragraph we already do "our best" to "to produce an extremely accurate simulation based on technical data and real-world testing". 
So your addition just might give a wrong expression that "do our best" to add everything to the aircraft that is requested via issue. This is simply not the case. For me this new paragraph is completely redundant and also misleading.

Simmer1234569: Alright, I guess I won't merge this then.

PR #7996, opened 2023-04-30
fix(fadec): add correct debug options and match flags to other gauges

tracernz:
Fixes #[issue_no]

## Summary of Changes

Changes necessary to debug the FADEC properly in VSCode. Build the FADEC with `--debug` flag, launch the sim with `-AllowDebug` flag. If you want to catch the first frame, attach VSCode at the world map screen (hit start and choose FlightSimulator.exe as the process to attach), set your breakpoints and launch a flight. You can also attach at any later time and set/unset breakpoints as you wish. Note, "Fast WASM" option is also required in sim to avoid stripping debug info.

This also fixes the source code paths after the repo re-org, and aligns the FADEC flags with the other gauges (-O2, and NDEBUG defined).

## Additional context

## Testing instructions

No QA required.


Taz5150 (approved):
Looks good from my side.

PR #8025, opened 2023-05-15
feat(hyd): edp clutch disconnection

crocket63:
Fixes #[issue_no]

## Summary of Changes

A380 EDP disconnection mechanism

## Additional context


.github/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 1. [FMS] Implement vertical navigation functions - @BlueberryKing (BlueberryKing)
 1. [EFB] Added pause at T/D function - @2hwk (2Cas#1022)
+1. [HYD] Engine pumps disconnection mechanism on A380- @Crocket63 (crocket)
```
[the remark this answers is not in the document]
crocket63: Oh yeah just mechanically added it and even later it will never benefit to 320.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/mod.rs
```diff
@@ -3227,6 +3261,9 @@ impl A380EngineDrivenPumpController {
         self.are_pumps_disconnected = self.are_pumps_disconnected
             || overhead_panel.engines_edp_disconnected(self.pump_id.into_engine_num());
```
Gurgel100, line 3262: Maybe you should use `self.disconnection_mechanism.is_connected()` instead of `overhead_panel.engines_edp_disconnected(self.pump_id.into_engine_num())` here too (and move it after update)?

crocket63: Had a look and I think it would be different. In fact the mechanism.is_connected is the actual mechanical state of the mechanism, whereas self.are_pumps_disconnected is the logical state from cockpit/button point of view.
So if we have a failure in mechanism, you should then have self.are_pumps_disconnected = true while mechanism.is_connected staying false. I believe there's no actual feedback from the clutch system to the overhead, so my current understanding is that it can become out of sync

Gurgel100 (approved):

PR #8066, opened 2023-06-08
feat(bleed): crossbleed starts and more

### LINKED ISSUE #820: [BUG] Engine Bleed air not available for engine start
**Mod Version**
Tested using 0.2.0 version the 23th of september 2020.

**Describe the bug**
For engine start on ground, it is not possible to perform the start with engine air bleed. It is working only with APU bleed.

**To Reproduce**
1. Start the game with aircraft on runway with engine running
2. Shutdown Engine1 with Master Lever Off
3. Try to restart it : Rotary to IGN and MasterLever 1 to ON, nothing is happening. We can see on ECAM SD page there is no air pressure.
(If we start the APU, and switch on the APU bleed, ten we can start.) 

**Expected behavior**
As engine 2 is running, they shall be bleed air available for the start.

**Actual behavior**
We can see on ECAM SD page the starter air valve opening, but still "0" air pressure.
The N2 is not increasing.

BlueberryKing:
Fixes #820

## Summary of Changes

Detailed list of changes:
- Allow engine start from crossbleed air
- Implement pressure sensors for regulated, transfer, and precooler inlet pressure
- Implement precooler outlet temperature sensor
- Implement fully open/fully closed sensors for the crossbleed valve.
- Update SD Bleed page indications
  - Implement crossbleed valve indications for open, closed and in-transfer status
  - Precooler inlet pressure green/amber logic adjusted
  - Use 5s confirmation before showing closed pressure regulating valve indication
- Add bleed air overpressure/underpressure/overheat detection in the BMC (no ECAM warnings yet)
- Improve fan air valve regulation logic
- Rework engine compressor physics
- Implement overpressure valve
- Rework high pressure valve regulation
- Increase precooler cooling efficiency
- Tune APU bleed pressure
- Rework pack fault light logic

## Additional context

Discord username (if different from GitHub): BlueberryKing

## Testing instructions

- Load up a flight
- Make sure starting an engine from APU bleed works
- Make sure performing a crossbleed start works
  - Turn off APU Bleed off
  - Turn the engine bleed pushbutton off on the receiving engine
  - Open the crossbleed valve
  - Set the started engine to about 30% N1
  - Set the engine mode selector to IGN
  - Set the engine master switch to ON
- Fly in various conditions, make sure indications on the BLEED SD page, stay green


Benjozork (approved):
For JS part

fbw-common/src/wasm/systems/systems/src/pneumatic/valve.rs
```diff
@@ -73,6 +74,122 @@ impl Default for PurelyPneumaticValve {
...
+    }
+}
+impl<const N: usize> PneumaticValve for SolenoidValve<N> {
+    fn is_open(&self) -> bool {
+        self.open_amount().get::<percent>() > 0.
+    }
+}
+impl<const N: usize> SimulationElement for SolenoidValve<N> {
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+        self.solenoid.accept(visitor);
```
Gurgel100, line 189:
```suggestion
        self.solenoid.accept(visitor);
        visitor.visit(self);
```

fbw-common/src/wasm/systems/systems/src/pneumatic/valve.rs
```diff
@@ -156,6 +273,85 @@ impl SimulationElement for ElectroPneumaticValve {
...
+impl<const N: usize> OverpressureValve<N> {
+    pub fn new(
+        characteristics: PneumaticValveCharacteristics<N>,
+        protection_threshold: Pressure,
+    ) -> Self {
+        Self {
+            connector: PneumaticContainerConnector::new(),
+            characteristics,
+            protection_threshold,
+            open_amount: Ratio::new::<ratio>(0.),
```
Gurgel100, line 292:
```suggestion
            open_amount: Ratio::default(),
```

fbw-common/src/wasm/systems/systems/src/pneumatic/valve.rs
```diff
@@ -156,6 +273,85 @@ impl SimulationElement for ElectroPneumaticValve {
...
+        } else {
+            1.
+        };
+
+        let current_open_amount = self.open_amount.get::<ratio>();
+
+        self.open_amount = Ratio::new::<ratio>(if target_open_amount > current_open_amount {
+            target_open_amount.min(
+                current_open_amount
+                    + context.delta_as_secs_f64() * self.characteristics.valve_speed,
```
Gurgel100, line 323: Extract calculation to separate variable to reduce code duplication.

fbw-common/src/wasm/systems/systems/src/pneumatic/valve.rs
```diff
@@ -156,6 +273,85 @@ impl SimulationElement for ElectroPneumaticValve {
...
+    }
+}
+
+trait FullyOpen {
+    fn is_fully_open(&self) -> bool;
+}
+
+trait FullyClosed {
+    fn is_fully_closed(&self) -> bool;
+}
```
Gurgel100, line 353: This is never used?

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -858,8 +858,44 @@ impl<const ENGINES: usize> PackFlowController<ENGINES> {
...
+    fn update_pressure_condition(
+        &mut self,
+        context: &UpdateContext,
+        pneumatic: &impl PackFlowValveState,
+    ) {
+        self.inlet_pressure_below_min.update(
+            context,
+            pneumatic
+                .pack_flow_valve_inlet_pressure(self.id + 1)
+                .map_or(false, |p| {
```
Gurgel100, line 875: Am I correct that this means that when the inlet pressure sensor fails/is unpowered the inlet pressure is assumed to be above min?

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -858,8 +858,44 @@ impl<const ENGINES: usize> PackFlowController<ENGINES> {
...
+    }
+
+    fn update_fcv_monitoring(
+        &mut self,
+        context: &UpdateContext,
+        pneumatic: &impl PackFlowValveState,
+    ) {
+        self.fcv_failed_open_monitor.update(
+            context,
+            !self.should_open_fcv && pneumatic.pack_flow_valve_is_open(self.id + 1),
```
Gurgel100, line 888: Extract logic to separate variable to avoid code duplication.

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -708,6 +713,13 @@ impl<const ENGINES: usize> PackFlowController<ENGINES> {
             operation_mode: ACSCActiveComputer::None,
 
             fcv_timer_open: Duration::from_secs(0),
+            fcv_failed_open_monitor: DelayedTrueLogicGate::new(Duration::from_secs(
+                Self::FCV_FAILED_OPEN_TIME_LIMIT_SECOND,
+            )),
+            fcv_failed_closed_monitor: DelayedTrueLogicGate::new(Duration::from_secs(
+                Self::FCV_FAILED_CLOSED_TIME_LIMIT_SECOND,
+            )),
+            inlet_pressure_below_min: DelayedTrueLogicGate::new(Duration::from_secs(5)),
```
Gurgel100, line 722: Extract literal value into constant

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -694,6 +695,10 @@ impl<const ENGINES: usize> PackFlowController<ENGINES> {
     const FLOW_CONSTANT_C: f64 = 0.5675; // kg/s
     const FLOW_CONSTANT_XCAB: f64 = 0.00001828; // kg(feet*s)
 
+    const PACK_INLET_PRESSURE_MIN_PSIG: f64 = 8.;
+    const FCV_FAILED_OPEN_TIME_LIMIT_SECOND: u64 = 30;
+    const FCV_FAILED_CLOSED_TIME_LIMIT_SECOND: u64 = 17;
```
Gurgel100, line 700: You can directly do constants of type `Duration`.

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1053,12 +1354,12 @@ impl SimulationElement for EngineBleedAirSystem {
         writer.write(
             &self.transfer_pressure_transducer_pressure_id,
             self.transfer_pressure_transducer_pressure()
-                .map_or(-1., |p| p.get::<psi>()),
+                .map_or(-1., |p| p.get::<psi>().max(0.).min(512.)),
```
Gurgel100, line 1357:
```suggestion
                .map_or(-1., |p| p.get::<psi>().clamp(0., 512.)),
```

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1053,12 +1354,12 @@ impl SimulationElement for EngineBleedAirSystem {
...
         );
         writer.write(
             &self.regulated_pressure_transducer_pressure_id,
             self.regulated_pressure_transducer_pressure()
-                .map_or(-1., |p| p.get::<psi>()),
+                .map_or(-1., |p| p.get::<psi>().max(0.).min(512.)),
```
Gurgel100, line 1362:
```suggestion
                .map_or(-1., |p| p.get::<psi>().clamp(0., 512.)),
```

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1247,21 +1605,25 @@ pub struct PackComplex {
...
             pack_container: PneumaticPipe::new(
-                Volume::new::<cubic_meter>(2.),
+                Volume::new::<cubic_meter>(1.),
                 Pressure::new::<psi>(14.7),
                 ThermodynamicTemperature::new::<degree_celsius>(15.),
             ),
             exhaust: PneumaticExhaust::new(0.3, 0.3, Pressure::new::<psi>(0.)),
             pack_flow_valve: DefaultValve::new_closed(),
+            // TODO: Check electrical supply of this
+            // I'm pretty sure PACK 1 is DC 1 and PACK 2 is DC 2, but happy to be corrected
```
Gurgel100, line 1625: I think this comment is better placed where the `PackComplex` is instantiated (i.e. in `A320Pneumatic`).

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1359,41 +1739,45 @@ impl CrossBleedValve {
...
         container_two: &mut impl PneumaticContainer,
     ) {
-        if !self.is_powered_for_manual_control && !self.is_powered_for_automatic_control {
-            self.set_open_amount_from_pressure_difference(
-                container_one.pressure() - container_two.pressure(),
-            )
-        }
+        self.open_amount = if self.target_open_amount > self.open_amount {
+            self.target_open_amount
+                .min(self.open_amount + context.delta_as_secs_f64() * self.valve_speed)
```
Gurgel100, line 1744: Extract calculation into separate variable to avoid code duplication (the `time * valve_speed` part).

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
```diff
@@ -1405,8 +1789,62 @@ impl SimulationElement for CrossBleedValve {
...
+            is_hold_active: false,
+
+            vertical_speed_greater_140: DelayedTrueLogicGate::new(Duration::from_secs(10)),
+            vertical_speed_less_80: DelayedTrueLogicGate::new(Duration::from_secs(10)),
+            vertical_speed_within_140: DelayedTrueLogicGate::new(Duration::from_secs(10)),
+        }
+    }
+
+    pub fn update(&mut self, context: &UpdateContext) {
+        // TODO: Use correct signals here
```
Gurgel100, line 1813: You want to implement this in a future PR?

Gurgel100 (changes requested):
Good work. This is my review of the Rust part

fbw-common/src/wasm/systems/systems/src/pneumatic/mod.rs
```diff
@@ -849,6 +849,101 @@ impl ControllerSignal<Pressure> for DifferentialPressureTransducer {
...
+            powered_by,
+        }
+    }
+
+    fn update(&mut self, controller: &impl ControllerSignal<SolenoidSignal>) {
+        if !self.is_powered {
+            self.is_energized = false;
+        } else if let Some(signal) = controller.signal() {
+            self.is_energized = signal.should_energize;
+        }
```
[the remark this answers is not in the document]
BlueberryKing: I don't think this works because the assignment doesn't check for an exhaustive set of conditions.

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
BlueberryKing: Yes, this is correct and intended 👍.

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
BlueberryKing: Good catch! The real answer is that the code should not be duplicated at all. Silly copy-paste error 😄.

fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic.rs
BlueberryKing: Yeah, need some kind of EIU implementation for it.

fbw-common/src/wasm/systems/systems_wasm/src/engines.rs
```diff
@@ -0,0 +1,46 @@
...
+                let is_sim_bleed_air_active = to_bool(values[1]);
+
+                // These values are very arbitrary. Whether crossbleed starts work or not is binary for now,
+                // until we have a custom engine model
+                if (starter_pressure_psig > 10. && !is_sim_bleed_air_active)
+                    || (starter_pressure_psig < 5. && is_sim_bleed_air_active)
+                {
+                    toggle_sim_engine_bleed_air(engine_number);
+                }
+            }),
```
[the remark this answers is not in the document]
BlueberryKing: Yeah, I'm not too happy with this piece of code, but couldn't find a better way. `variable_to_event` only lets me read one variable, but I need other variables to determine whether the event should be emitted. Would love to hear suggestions on how to improve this code.

fbw-common/src/wasm/systems/systems/src/pneumatic/mod.rs
```diff
@@ -849,6 +849,101 @@ impl ControllerSignal<Pressure> for DifferentialPressureTransducer {
...
+    fn is_energized(&self) -> bool {
+        self.is_energized
+    }
+}
+impl SimulationElement for Solenoid {
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+        visitor.visit(self);
+    }
+
+    fn receive_power(&mut self, buses: &impl ElectricalBuses) {
```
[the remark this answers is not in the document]
BlueberryKing: Good idea. I think I will tackle this for the entire bleed system in a future PR.

fbw-common/src/wasm/systems/systems/src/pneumatic/mod.rs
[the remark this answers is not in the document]
Gurgel100: Ok. Is it correct then that the solenoid stays energized when it is powered and there is no controller signal?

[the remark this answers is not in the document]
BlueberryKing: Hmm, probably not 😅 .

fbw-common/src/wasm/systems/systems_wasm/src/engines.rs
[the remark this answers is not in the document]
Gurgel100: I actually don't know either. But at least change the `AMBIENT PRESSURE` variable read to an unindexed one (use index 0).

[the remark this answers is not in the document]
Gurgel100: I really like the solution you came up for this.

Gurgel100 (approved):

pinatacolada:

Discord			: pinatacolada
Object of testing	: #8066 
Tier of Testing  	: 2
Date            		: 30/08/2023

Testing Process: 

- [x]     Load up a flight
- [x]     Make sure starting an engine from APU bleed works
- [x]     Make sure performing a crossbleed start works
- [x]         Turn off APU Bleed off
- [x]         Turn the engine bleed pushbutton off on the receiving engine
- [x]         Open the crossbleed valve
- [x]         Set the started engine to about 30% N1
- [x]         Set the engine mode selector to IGN
- [x]         Set the engine master switch to ON
- [x]     Fly in various conditions, make sure indications on the BLEED SD page, stay green

Performed several starts in different combinations, including starting 2 first on APU, starting 1 first on APU, starting 1 then 2 both on APU, starting 2 on APU then 1 on crossbleed, and then shutting down 2 then starting 2 on crossbleed.

Also made a full flight.

Negatives:

Pack flow control valve logic is flawed. Engine bleed pressures at high altitude (> 320ish) higher than expected.

Testing Results: 
Passed, with remarks

Conclusions: 
Normal engine starts and crossbleed starts work fine, however the pack flow control valve logic is flawed.

Tested at LPPT elev. 374ft, temp +19ºC.
       
Engine 2 was the supplying engine. When closing bleed 1, bleed 2 gets up to 48 PSI and bleed 1 42 PSI, at idle 19.8% N1, which albeit I don't remember the last time I did a real crossbleed start on a Neo, seem high at idle. FCOM mentions adjusting N1 to 30 PSI before starting, and given that there should be an idle N1 increase commanded by the FADEC that is not modeled yet, 48 PSI at "normal" idle might be too much. But, again, I have to reiterate I don't remember the last real crossbleed on a neo, so take it with a pinch of salt.

A very nice positive is that without increasing the supplying engine N1, the receiving engine pressure falls to around 20 PSI, and while the supplying N1is not increased to provide at least around the 25 PSI that FCOM requests as minimum, which is very good!

May be beyond the scope of the PR, but:

> After 30 s, if the flight crew does not set the ENG MASTER lever to ON (Refer to DSC-70-80-40 Second Step), pack valves automatically open again.

Pack valves remain closed forever if no engine master is not set to on, which is inaccurate.

and

> • The engine start valve closes
> • The pack valves reopen if another engine is not started within 60 s.

The available engine's pack opens around 6 seconds after the AVAIL indication. The opposite pack remains closed forever. Both are inaccurate. Assuming the crossbleed valve is open, both packs should reopen after a 60 second period from N2 > 63%. If the crossbleed is closed, the avail side pack remains on, and only the starting side pack goes through the off then on cycle.

For the flight test, everything was green, however higher up in in cruise, the PSI is higher than expected. For example, during climb at around FL370, with N1 at 88.3%, the fbw was giving 42PSI and175ºC, while I was expecting around 36 PSI and 200ºC. Some other measured points, now at cruise in FL390:

N1 | FBW PSI | Expected PSI
80 | 40          | 28
81 | 40          | 28
82 | 42          | 30
85 | 42          | 32

The HP valves were also closing at lower N1 values than expected at high cruise levels, closing at around 65 % N1, instead of the expected 75ish N1. Bleed pressures looked as expected once the HP valves were open, though.

As a bonus, I also tried to apply the AIR CONDITIONING DURING AUTOMATIC ENGINE START procedure, which calls for closing the crossbleed after engine 2 is available:

Engine 1 was started successfully without any issue. However, because the pack flow valves are opening sooner than they should, pack 2 valve opened at around the same 6 second gap from AVAIL, instead of when closing the crossbleed.

Also, after engine 1 start, when setting the crossbleed to auto (which will open it because it is feeding from the APU bleed), the pack 2 valve briefly closed for a second before reopening, which is not accurate.

pinatacolada:
tested again after changes, made a full flight + an apu start followed by a crossbleed start:

bleed pressure and temperatures are all as expected through out all stages of flight! HP valves are still opening and closing around 65 % N1 which is the same as before, instead of the expected 75.

The only negative is that the pack compressor outlet temperature is always matching the bleed temperature, lagging by some +/- 5ºc ish, usually hotter than the bleed. On the real thing the pack compressor outlet is around 10-20ºc cooler than the bleed.

The negative outcome of it, that is only reproducible at FL390, is that an engine deceleration to a point that opens the HP followed by an acceleration that leads to bleed temperatures north of 225ºC (which for the bleed itself is within expected), will also lead to the same compressor outlet temperatures, which were displayed as amber in some cases, and that's obviously not the case on the real thing in a normal operation.

On a side note, the compressor outlet temperature should be amber when equal to or above 260 °C, and it seems to be happening at 240ºC on the fbw.

engine starts were all normal with bleed pressures within the expected values!

PR #7960, opened 2023-04-10
feat(cond): a380 flow control and monitoring

mjuhe:
## Summary of Changes

This PR introduces the first building blocks of the air conditioning system for the a380. Although my goal is to reuse as much of the a320 code as possible, in this case not much has been reused because the system architecture and the logic between the two planes is different.

I've structured it like this:
- CPIOM B is in `a380-systems`. It checks if the corresponding CPIOMs are active (introduced in #7731) and contains all the applications needed for the control of the air conditioning system. This PR only has the AGS Application (Air Generation System Application) for now.
- FDAC (Full Digital AGU Controller) is in `systems`. This is a local controller, and although it's not used by the A320, maybe other planes in the future might use it (?). It controls the flow control valves using the demand provided by the AGS.

## Additional context

I've decided to duplicate the tests from `acs_controller` into the top level a380 air conditioning file instead of splitting and duplicating them in the FDAC/CPIOM B files. I think it makes more sense to have all tests in the top level to ensure everything interacts as it should.

Discord username (if different from GitHub): Commander Ga#1548

## Testing instructions

This should not affect the a320 systems at all. A quick test that the air conditioning system still works should be more than enough.

For internal a380 testing: you'll need the air-cond branch for the overhead buttons to work. Let me know if you want detailed test instructions 😃.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8086, opened 2023-06-24
feat(cond): a380 temperature control and ventilation

mjuhe:
## Summary of Changes

This PR finishes the first iteration of the Air Conditioning system for the A380. It adds:
- CPIOM B Applications:
     - Temperature Control System Application
     - Ventilation Control System Application
- Local controllers:
     - Trim Air Drive Device (TADD)
     - Ventilation Control Module (VCM FWD and AFT)
- Other elements and features:
     - Bulk Cargo Heater
     - High Pressure cabin fans
     - Isolation Valves for FWD and BULK Cargo compartments

The temperature simulation and heat exchange will still need to be adjusted later on for accuracy when we can load passengers and do full flights, but this should provide a foundation with all pieces in place.

## Screenshots (if necessary)

I created this to help whoever wants to work on the COND SD page 😉 

image

## Additional context

As per the A320 air cond system, this doesn't incorporate failures, although most elements can be "failed" by removing power.
It also doesn't include avionics ventilation.

## Testing instructions

As per #7960 this shouldn't have any effect on the A320. A quick courtesy test that everything works would be much appreciated but shouldn't be required.

If you want to test this in the A380, you should use the air-cond-system branch.


2hwk (approved):

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/cpiom_b.rs
```diff
@@ -311,9 +359,259 @@ impl PackFlow for AirGenerationSystemApplication {
...
+        let bulk_heater_on_allowed = (cargo_door_open.aft_cargo_door_locked()
+            || !lgciu.iter().all(|a| a.left_and_right_gear_compressed(true)))
+            && vcm_shared.bulk_duct_heater_on_allowed();
+        let temperature_difference = cabin_temperature.cabin_temperature()[ZoneType::Cargo(2).id()]
+            .get::<degree_celsius>()
+            - acs_overhead
+                .selected_cargo_temperature(ZoneType::Cargo(2))
+                .get::<degree_celsius>();
+        (temperature_difference < -1.
+            || (self.should_switch_on_bulk_heater && temperature_difference < 1.))
```
Gurgel100, line 571: Are you sure this condition is correct? Because only `temperature_difference` determines the result of it.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/cpiom_b.rs
```diff
@@ -311,9 +359,259 @@ impl PackFlow for AirGenerationSystemApplication {
...
+    }
+}
+
+impl DuctTemperature for TemperatureControlSystemApplication {
+    fn duct_demand_temperature(&self) -> Vec<ThermodynamicTemperature> {
+        let mut duct_demand_temperature = Vec::new();
+        for controller in self.zone_controllers.iter() {
+            duct_demand_temperature.extend(&controller.duct_demand_temperature())
+        }
+        duct_demand_temperature
```
Gurgel100, line 450:
```suggestion
        self.zone_controllers
            .iter()
            .flat_map(|controller| controller.duct_demand_temperature())
            .collect()
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
```diff
@@ -345,10 +341,62 @@ impl A380AirConditioningSystem {
                     ],
                 ),
             ],
-            cabin_fans: [CabinFan::new(ElectricalBusType::AlternatingCurrent(1)); 2],
+            tadd: TrimAirDriveDevice::new(
+                context,
+                vec![
```
Gurgel100, line 346:
```suggestion
                [
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
```diff
@@ -345,10 +341,62 @@ impl A380AirConditioningSystem {
...
+                    ElectricalBusType::AlternatingCurrent(2), // 117XP
+                    ElectricalBusType::AlternatingCurrent(4), // 206XP
+                ],
+            ),
+            vcm: [
+                VentilationControlModule::new(
+                    context,
+                    VcmId::Fwd,
+                    vec![
```
Gurgel100, line 355:
```suggestion
                    [
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
```diff
@@ -345,10 +341,62 @@ impl A380AirConditioningSystem {
...
+                        ElectricalBusType::DirectCurrent(1),       // 411PP
+                        ElectricalBusType::DirectCurrentEssential, // 109PP
+                    ],
+                ),
+                VentilationControlModule::new(
+                    context,
+                    VcmId::Aft,
+                    vec![
```
Gurgel100, line 363:
```suggestion
                    [
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
```diff
@@ -345,10 +341,62 @@ impl A380AirConditioningSystem {
...
+                ), // Left Hand - 100XP2
+                CabinFan::new(
+                    VolumeRate::new::<liter_per_second>(Self::CAB_FAN_DESIGN_FLOW_RATE_L_S),
+                    ElectricalBusType::AlternatingCurrent(3),
+                ), // Right Hand - 200XP3
+                CabinFan::new(
+                    VolumeRate::new::<liter_per_second>(Self::CAB_FAN_DESIGN_FLOW_RATE_L_S),
+                    ElectricalBusType::AlternatingCurrent(4),
+                ), // Right Hand - 200XP4
+            ],
```
Gurgel100, line 387: A small suggestion to reduce code duplication:
```suggestion
            cabin_fans: [
                ElectricalBusType::AlternatingCurrent(1), // Left Hand - 100XP1
                ElectricalBusType::AlternatingCurrent(2), // Left Hand - 100XP2
                ElectricalBusType::AlternatingCurrent(3), // Right Hand - 200XP3
                ElectricalBusType::AlternatingCurrent(4), // Right Hand - 200XP4
            ]
            .map(|powered_by| {
                CabinFan::new(
                    VolumeRate::new::<liter_per_second>(Self::CAB_FAN_DESIGN_FLOW_RATE_L_S),
                    powered_by,
                )
            }),
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
```diff
@@ -398,17 +428,27 @@ impl A380AirConditioningSystem {
...
 
-        let pack_flow: [MassRate; 2] = [
-            self.acsc.individual_pack_flow(Pack(1)),
-            self.acsc.individual_pack_flow(Pack(2)),
-        ];
-        let duct_demand_temperature = self.acsc.duct_demand_temperature();
         for (id, pack) in self.packs.iter_mut().enumerate() {
-            pack.update(pack_flow[id], &duct_demand_temperature)
+            pack.update(pack_flow[id], &cpiom_b.duct_demand_temperature())
         }
```
Gurgel100, line 452:
```suggestion
        for (pack, pack_flow) in self
            .packs
            .iter_mut()
            .zip(self.fdac.iter().map(|fdac| fdac.pack_flow()))
        {
            pack.update(pack_flow, &cpiom_b.duct_demand_temperature())
        }
```
Unfortunately this suggestion is a bit weirdly formatted because it also includes the deleted lines.

Gurgel100 (changes requested):
Good job.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/cpiom_b.rs
mjuhe: The idea is that the heater switches on when temperature difference is less than -1deg, and switches off when it's above +1deg (this condition is essentially keeping the heater on if it was already on and the temperature difference is between -1 and +1)

fbw-common/src/wasm/systems/systems/src/air_conditioning/trim_air_drive_device.rs
```diff
@@ -0,0 +1,204 @@
...
+        }) || (engines
+            .iter()
+            .any(|e| e.corrected_n1() >= Ratio::new::<percent>(15.))
+            && pneumatic_overhead
+                .engine_bleed_pushbuttons_are_auto()
+                .iter()
+                .any(|pb| pb == &true)
+            && pneumatic.engine_crossbleed_is_on()))
+            || pneumatic.apu_bleed_is_on())
+            && hot_air_is_enabled
```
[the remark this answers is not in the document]
mjuhe: Yes... There's been a couple of discussions about this already on Discord, so I've decided to move the three local controllers to the a380 crate now. If we ever need to move them back to `systems` again it's an easy change.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
mjuhe: Thanks for the suggestion, I sort of applied it but in a slightly different way as the `CabinFan` constructor has changed since this comment.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/mod.rs
mjuhe: I like it! Thanks.

mjuhe:
Thanks for the review @Gurgel100 ! It's taken a while but all comments should've been addressed now. Really good suggestions as always! 😄 🚀

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -242,17 +245,55 @@ impl<const ZONES: usize, const ENGINES: usize> AirConditioningSystemController<Z
...
+            .collect();
+
+        // Because each ACSC calculates the demand of its respective zone(s), we fill the vector for the trim air system
+        let mut filler_vector = [ThermodynamicTemperature::new::<degree_celsius>(24.); ZONES];
+        if matches!(self.id, AcscId::Acsc1(_)) {
+            filler_vector[..1].copy_from_slice(&demand_temperature);
+        } else {
+            filler_vector[1..].copy_from_slice(&demand_temperature);
+        };
+        filler_vector.to_vec()
```
Gurgel100, line 296:
```suggestion
        let demand_temperature = self
            .zone_controller
            .iter()
            .map(|zone| zone.duct_demand_temperature()[0]);

        // Because each ACSC calculates the demand of its respective zone(s), we fill the vector for the trim air system
        if matches!(self.id, AcscId::Acsc1(_)) {
            demand_temperature.take(ZONES).collect()
        } else {
            demand_temperature.skip(1).take(ZONES).collect()
        }
```

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -632,20 +664,18 @@ impl ZoneController {
...
-                    10.
-                },
-            )
-        } else {
-            self.calculate_duct_temp_demand(context, pressurization, zone_measured_temperature)
-        };
+        self.duct_demand_temperature =
+            if self.galley_fan_failure.is_active() && matches!(acsc_id, AcscId::Acsc2(_)) {
+                // Cabin zone temperature sensors are ventilated by air extracted by this fan, cabin temperature regulation is lost
+                // Cabin inlet duct is constant at 15C, cockpit air is unnafected
```
Gurgel100, line 670:
```suggestion
                // Cabin inlet duct is constant at 15C, cockpit air is unaffected
```

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -1280,21 +1303,54 @@ impl PneumaticValveSignal for TrimAirValveSignal {
...
+        }
+    }
+}
+
+impl Default for TrimAirPressureRegulatingValveController {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
Gurgel100, line 1338: Remove this implementation but instead derive it.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/local_controllers/trim_air_drive_device.rs
```diff
@@ -100,22 +95,20 @@ impl<const ZONES: usize, const ENGINES: usize> TrimAirDriveDevice<ZONES, ENGINES
...
+                Some(TaddFault::BothChannelsFault)
+            } else {
+                self.switch_active_channel();
+                Some(TaddFault::OneChannelFault)
             }
+        } else if self.stand_by_channel.has_fault() {
+            Some(TaddFault::OneChannelFault)
+        } else {
+            None
         };
```
Gurgel100, line 112: Another possibility (don't need to accept):
```suggestion
        self.fault = match (
            self.active_channel.has_fault(),
            self.stand_by_channel.has_fault(),
        ) {
            (true, true) => Some(TaddFault::BothChannelsFault),
            (false, false) => None,
            (ac, _) => {
                if ac {
                    self.switch_active_channel();
                }
                Some(TaddFault::OneChannelFault)
            }
        };
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/local_controllers/trim_air_drive_device.rs
```diff
@@ -128,41 +121,18 @@ impl<const ZONES: usize, const ENGINES: usize> TrimAirDriveDevice<ZONES, ENGINES
         acs_overhead: &impl AirConditioningOverheadShared,
         should_close_taprv: bool,
         hot_air_id: usize,
+        pneumatic: &impl PackFlowValveState,
     ) -> bool {
         acs_overhead.hot_air_pushbutton_is_on(hot_air_id)
             && !self.active_channel.has_fault()
-            && !matches!(self.fault, Some(TaddFault::BothChannelsFault))
+            && ((pneumatic.pack_flow_valve_is_open(1)) || (pneumatic.pack_flow_valve_is_open(2)))
```
Gurgel100, line 128: Unnecessary parenthesizes:
```suggestion
            && (pneumatic.pack_flow_valve_is_open(1) || pneumatic.pack_flow_valve_is_open(2))
```

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/local_controllers/trim_air_drive_device.rs
```diff
@@ -128,41 +121,18 @@ impl<const ZONES: usize, const ENGINES: usize> TrimAirDriveDevice<ZONES, ENGINES
...
             && !should_close_taprv
+        // && !self.duct_overheat_monitor()
+        // && !any_tav_has_fault
```
Gurgel100, line 131: Is this for later?

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/cpiom_b.rs
Gurgel100: I didn't see that once it's -1 and the other time +1. My bad.

fbw-common/src/wasm/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -263,49 +281,72 @@ impl From<Channel> for usize {
...
         }
     }
 
-    fn has_fault(&self) -> bool {
+    pub fn update_fault(&mut self) {
+        let failure_is_active = if self.failure.is_some() {
+            self.failure.as_ref().unwrap().is_active()
+        } else {
+            false
+        };
```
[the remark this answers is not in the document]
mjuhe: Thanks for the suggestions! I like the first one. In fact I wanted to use `.is_some_and()` but I don't think it works with the version of Rust we are using.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/local_controllers/trim_air_drive_device.rs
mjuhe: I love this suggestion, thanks! I've applied it to all the local controllers that use the same `fault_determination` function. This could probably be further abstracted with a macro but I'll leave it for a future refactor 😄.

fbw-a380x/src/wasm/systems/a380_systems/src/air_conditioning/local_controllers/trim_air_drive_device.rs
mjuhe: Yes, this is for when failures are implemented in the a380.

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
mjuhe: This suggestion doesn't work. The problem is that the resulting vector here needs to be the length of `ZONES` but `zone_controller` is a different length in each ACSC. For the a320:
- There are 3 ZONES
- ACSC1 has a `zone_controller` of length 1 (cockpit)
- ACSC2 has a `zone_controller` of length 2 (cabin zones)

What we want is to "fill" the remaining positions of the resulting vector with dummy data, so for ACSC1 you end up with [real, dummy dummy] and for ACSC2 [dummy, real, real]. Hope this makes sense, but I'm open to other suggestions as I don't particularly like this piece of code either!

fbw-common/src/wasm/systems/systems/src/air_conditioning/mod.rs
```diff
@@ -656,7 +698,8 @@ impl<const ZONES: usize, const ENGINES: usize> TrimAirSystem<ZONES, ENGINES> {
         &mut self,
         context: &UpdateContext,
         mixer_air: &MixerUnit<ZONES>,
-        tav_controller: &impl TrimAirControllers,
+        taprv_controller: &[&impl ControllerSignal<TrimAirValveSignal>],
```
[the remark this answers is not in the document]
mjuhe: Thanks for pointing this out, in fact the logic was not correct. The a320 has 2 controllers for a single valve, and the a380 has 2 controllers for 2 valves. I've added a condition check and implemented your suggestion, not the cleanest ever but it does the job for now.

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
Gurgel100: Does it have to return ZONES entries? Because actually the function return type doesn't specify that. (I know it was a constant size array before, but now the caller of this function cannot rely on that)

mjuhe: Yeah I see your point... I could introduce const generics in the trait itself and then return an array `[ThermodynamicTemperature; ZONES]` instead of a vector. But this makes the whole code a lot more verbose because I need to pass the generics around more elements and/or hard code the number of zones, which I don't really like either. I could add a comment explaining why this function needs to return a vector with the length of ZONES here. Or do you think it's worth it having the generics in the return type of this trait?

mjuhe: Comment added 😃. Thanks again @Gurgel100

fbw-common/src/wasm/systems/systems/src/air_conditioning/acs_controller.rs
```diff
@@ -280,6 +280,14 @@ impl<const ZONES: usize, const ENGINES: usize> AirConditioningSystemController<Z
...
     for AirConditioningSystemController<ZONES, ENGINES>
 {
+    /// This function needs to return a Vector of size ZONES.
+    ///
+    /// Because the `ZoneController` is different for each ACSC, we create a vector with "dummy" data for the zones
+    /// not being calculated by the relevant ACSC.
+    ///
+    /// ACSC1 calculates the duct demand temperature of the cockpit.
+    ///
+    /// ACSC2 caluclates the duct demand temperature of the fwd and aft cabin zones.
```
Gurgel100, line 290:
```suggestion
    /// ACSC2 calculates the duct demand temperature of the fwd and aft cabin zones.
```

Gurgel100 (approved):
Great work.

PR #7834, opened 2023-02-28
fix: increase outer to inner tank transfer rate

donstim:
Fixes #7833 

## Summary of Changes
Increases the fuel transfer rate from the outer to inner tanks from 500 gallons/hour to 1200 gallons/hour.

## Additional context
Performed several takeoffs in different conditions and observed a maximum fuel flow of about 3500 kg/hr/engine. This roughly corresponds to the fuel burn to 1500 ft in the FCOM climb performance data.

The fuel transfer from outer to inner tanks is by gravity feed. It is currently set to 500 gallons/hour/tank, which is what was included in the Asobo A320NEO that we started with. To accomodate a takeoff fuel flow of 3500 kg/hour/engine, this PR changes the outer to inner tank transfer rate to 1200 gallons/hour (3647 kg/hr). This change will prevent low fuel warnings from appearing on takeoff or go-around (where fuel flow is highest) when there is fuel in the outer tanks to transfer to the inner tanks.

The trigger for both the low fuel warning and the transfer of fuel from outer tanks to inner tanks is the uncovering of the low fuel sensor. This will occur at 750 kg of fuel in each inner tank (although this value can be affected by airplane attitude and acceleration IRL). The low fuel warning does not appear until 30 seconds after the sensor is uncovered, which provides time for the fuel transfer to occur and cover up the sensor again.

Discord username (if different from GitHub): donbikes#4084

## Testing instructions
Conduct takeoffs and/or go-arounds starting with 750 kg of fuel in each inner tank and full outer tanks. Confirm that the inner tank fuel level is always increasing, preventing a low fuel warning from occurring.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8091, opened 2023-06-28
fix: Outer to inner tank valve closure

### LINKED ISSUE #8090: Outer tank fuel transfer valves close during flight
### Aircraft Version

Development

### Build info

```json
{
    "built": "2023-06-26T14:24:53+00:00",
    "ref": "refs/heads/master",
    "sha": "e540d80e7f33b0d133412069ea683a7472e20dc6",
    "actor": "Saschl",
    "event_name": "manual",
    "pretty_release_name": "master:e540d80e",
    "version": "v0.11.0-dev.e540d80"
}
```

### Describe the bug

On flights of sufficient length, the outer to inner fuel tank transfer valves may close after the fuel is transferred.  This should not occur. These valves should not close until the airplane is refueled.

### Expected behavior

Outer to inner fuel tank transfer valves to remain open until the airplane is refueled.

### Steps to reproduce

1. Load fuel such that the outer to inner fuel tank transfer valves will open around the time descent is initiated.
2. Plan a route of at least 22 minutes.
3. Fly an idle thrust descent with fuel transferring until all fuel is transferred from outer to inner wing fuel tanks (descent of about 6 minutes).
4. Observe closing of outer to inner fuel tank fuel transfer valves

### References (optional)

_No response_

The outer to inner tanks fuel transfer valves open whenever one inner fuel tank has less than 750 kg in it. The valves should not close until the airplane is refueled. Because there is no coding in the A32NX to close the valves when the airplane is fueled, it is currently set up to close the valves when a certain level of fuel is reached in each inner tank. The triggering fuel level (1422 kg) was selected such that if all fuel was transferred essentially immediately, it would not cause the valves to close.

However, since the engines are coded such that one engine always burns, on average, 20 kg/hr more than the other, and both transfer valves open when the lowest inner tank fuel level reaches 750 kg, there can be enough of a fuel tank asymmetry for the inner tank with the higher fuel level to reach the 1422 kg triggering level to close the valves due to the transferred fuel.

To have a difference in fuel flow between engines. one engine is randomly picked at the beginning of the flight to have a random 0-36 kg/hr higher fuel flow. This averages to about 20 kg/hr. The trigger level for opening the transfer valves is the first inner tank to reach less than 750 kg of fuel.  There is about 693 kg of fuel in each outer tank, and it takes about 5:40 minutes:seconds to transfer.

If a transfer occurs, it is likely to happen during descent (and is most critical for this issue), where on idle thrust, fuel flow is about 10 kg/hr (or 5 kg/hr/engine). The trigger level for the valve to close is current 1422 kg (468 gallons). From this information, you can determine about how long a flight needs to be for the tank imbalance to result in triggering the valve for the inner tank with the most fuel to close.

For example, if there was no imbalance, the transfer valves would open when both tanks are at 750 kg. 693 kg of fuel is transferred in, while 5 kg/min times 5.67 minutes is burned. This results in 1415 kg of fuel in each tank, which is very very close to the trigger level for closing the transfer valve. The time it would take for an imbalance to exceed the trigger level to close the valve can be determined from [(20 kg/hr) * (X hours)] + 750 kg + 693 kg - [(5 kg/min) * (5.67)] = 1422. X = 7.4/20 = 0.37 hours. Any flight longer than about 22.2 minutes that has the fuel transfer valve open during descent will also have the fuel transfer valve in one tank close.

donbikes

donstim:
Fixes #8090 

## Summary of Changes
This PR makes 2 changes to the fuel system in the flight model.cfg file:
1. It changes the triggering condition for closing a given outer to inner fuel tank transfer valve from reaching a fuel level in the respective inner tank of over 468 gallons (1422 kg) to reaching a fuel level in the respective outer tank of over 227.9 (692.5 kg)
2. It changes the fuel transfer rate from each outer tank transfer line from 1200 gallons/hour to 600 gallons/hour
## Additional context

The outer to inner fuel tank transfer valves will only open when needed to replenish the inner tanks. This will cause the fuel level to drop in the outer tanks. Because the outer tanks are filled first when refueling, filling the outer tanks is a more robust indication of refueling than increasing the inner tank fuel amount above a certain level. This not only fixes issue #8090, but also makes it much less likely that a user on a turnaround flight will not add enough fuel to cause the valves to close.

The fuel transfer rate change from 1200 gallons/hour to 600 gallons/hour in each transfer line is to correct an oversight when these rates were implemented by PR #7834. The overall 1200 gallons/hour from each outer tank to each inner tank is correct, but there are 2 lines between each tank. Therefore, the rate from each line only needs to be 600 gallons/hour to achieve an overall rate of 1200  gallons/hour between each outer tank to each inner tank.

Discord username (if different from GitHub):donbikes

## Testing instructions
1. Confirm that the outer to inner tank transfer valves close (after being opened) whenever enough fuel is added through EFB fueling that fuel is added to the inner tanks. This implies that the outer tanks are filled by the EFB fueling process before fuel is added to the inner tanks. If this does not occur, the situations where this happens should be noted. The FUEL page in the lower ECAM can be used to determine the position of the fuel transfer valves.
2. Since the fuel transfer rates have been changed, the testing for PR #7834 should be performed again - Conduct takeoffs and/or go-arounds starting with 750 kg of fuel in each inner tank and full outer tanks. Confirm that the inner tank fuel level is always increasing, preventing a low fuel warning from occurring.


2hwk (approved):

tracernz (approved):

BlueberryKing:
Dev Team Tester Report

Discord          : BlueberryKing
Object of testing: #8091
Tier of Testing  : 1
Date             : 09/07/23

Testing Process: 
- Load up at KFBW.
- Load 3000 kg of fuel in total. Observe full outer tanks and inner tanks at about 800 kg each.
- Perform a few circuits, and observe transfer valves opening after inner levels drop below 750 kg. 
- Observe fuel levels consistently rising in the inner fuel tanks until the outer tanks are empty.
- Land, refuel to 3000 kg, observe transfer valves closing (see two images below).

Negatives:       
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

Media:       
After a few circuits having let the fuel transfer from the outers to the inners.
image
After refueling to 3000 kg.
image

PR #8120, opened 2023-07-17
refactor(l/g): refactor file structure of the L/G systems

lukecologne:
## Summary of Changes

This PR intends to refactor the layout of the files of the L/G systems, to be able to accomodate different L/G
hydraulic and avionic layouts for other aircraft.
This PR also refactors the LandingGear struct which previously provided the shock absorber compression values,
and integrates it into the `GearSystemSensors` trait. It also changes the shock absorber compression detection method
from using `GEAR ANIMATION POSITION` which is shared for both compression and gear position, to `CONTACT POINT COMPRESSION` which is only for compression. Also, `GEAR ANIMATION POSITION` is not capable of giving values for indices
higher than 2 (it is just stuck at 0.5, which is an issue when we have more than three gears). `CONTACT POINT COMPRESSION` does not have this problem.

## Additional context


fbw-a380x/src/wasm/systems/a380_systems/src/landing_gear/mod.rs
```diff
@@ -0,0 +1,1683 @@
...
+            self.status = LgciuStatus::FailedNotPowered;
+        } else if self.status == LgciuStatus::Ok {
+            if self.transition_duration > Self::MAX_TRANSITION_DURATION {
+                self.status = LgciuStatus::FailedNoChangeOver;
+            }
+
+            if self.internal_error_failure.is_active() {
+                self.status = LgciuStatus::FailedAutoChangeOver;
+            }
+        }
```
2hwk, line 737:
```suggestion
       self.status = if !self.is_powered {
            LgciuStatus::FailedNotPowered
        } else {
            if self.internal_error_failure.is_active() {
                LgciuStatus::FailedAutoChangeOver
            } else if self.transition_duration > Self::MAX_TRANSITION_DURATION {
                LgciuStatus::FailedNoChangeOver
            } else {
                LgciuStatus::Ok
            }
        };
```

2hwk: Should be equivalent (use simple assignment)

2hwk (changes requested):

fbw-a380x/src/wasm/systems/a380_systems/src/landing_gear/mod.rs
tracernz: Shouldn't clippy do this? Also, you have an else with if -> else if -> else inside; that can still be simplified more :P.

lukecologne:
Closing, as there are a large amount of conflicts, and the restructure should probably be done in smaller PR steps

PR #8145, opened 2023-07-30
feat: apu fuel consumption

tracernz:
Fixes #[issue_no]

## Summary of Changes

This patch allows the APU to consume fuel (from the left inner tank).

Known issues not fixed by this PR:
- The APU has no fuel flow when the left inner tank fuel pumps are off (https://github.com/flybywiresim/aircraft/issues/8146).

## Additional context

## Testing instructions

Check that the left inner tank fuel level reduces when the APU is running with the left inner tank pumps on.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -31,6 +31,7 @@
 1. [MCDU] Add basic annunciator support - @tracernz (Mike)
 1. [MCDU] Add basic MCDU MENU functionality - @tracernz (Mike)
 1. [MCDU] Split subsystem scratchpads - @tracernz (Mike)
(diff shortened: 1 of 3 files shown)


Taz5150 (approved):
Checked - APU consumes fuel from left tank at a rate of 33 gallons/ hour as expected.

Saschl:
Tested as well. fuel consumption works as expected. (Already existing) Issue with fuel pumps necessary might be solved in another PR.

donstim:
Just noticed that APU fuel consumption here is always drawn from the left inner tank. If the crossfeed valve is open, it should draw from both tanks.

PR #8155, opened 2023-08-04
feat(nd): support ZOOM and 640NM for A380

svengcz:
## Summary of Changes

Makes the NDv2 configurable to run on A32NX and A380X.
An aircraft type will be introduced to differentiate between different range values

## Additional context

## Testing instructions

Test all ND features during a test flight


fbw-a32nx/src/systems/fmgc/src/efis/EfisCommon.ts
```diff
@@ -25,6 +25,9 @@ export function withinEditArea(lla: Coordinates, range: EfisNdRangeValue, mode:
 export function calculateEditArea(range: EfisNdRangeValue, mode: EfisNdMode): [number, number, number] {
     switch (mode) {
     case EfisNdMode.ARC:
+        if (range <= 0.5) {
```
Benjozork, line 28: The FMS doesn't need to know about OANS ranges as it doesn't transmit symbols while OANS is active as far as I know. Or at least they aren't displayed.

fbw-a32nx/src/systems/shared/src/NavigationDisplay.ts
```diff
@@ -2,11 +2,20 @@
...
+import { AircraftType } from '@shared/aircraft';
 
 export type EfisSide = 'L' | 'R'
 
-export type EfisNdRangeValue = 10 | 20 | 40 | 80 | 160 | 320;
-export const rangeSettings: EfisNdRangeValue[] = [10, 20, 40, 80, 160, 320];
+export type EfisNdRangeValue = 0.5 | 10 | 20 | 40 | 80 | 160 | 320 | 640;
+export const rangeSettingValues: EfisNdRangeValue[] = [0.5, 10, 20, 40, 80, 160, 320, 640];
+
+export const getRangeSettingsValue = (aircraft: AircraftType, index: number): EfisNdRangeValue => {
```
Benjozork, line 12: I think I would prefer that we just change the A380 modelbehaviors to use the same indices for the range settings that do overlap, and negative values for ZOOM ranges. Passing around an aircraft type isn't an idea I like a lot

Benjozork (changes requested):

fbw-a32nx/src/systems/shared/src/NavigationDisplay.ts
Benjozork: Besides we're doing OANS on A320 anyway, and headwind are too, so we need support on both.

fbw-a32nx/src/systems/fmgc/src/efis/EfisCommon.ts
tracernz: Also I think rather than editing the edit areas (ha!) each aircraft should have their own. The current definitions are exact for the A320 family FMS and with this PR they would no longer be.

fbw-a32nx/src/systems/shared/src/NavigationDisplay.ts
svengcz: I don't like to map the A380 modelbehaviors of the A380 and have special values for zoom, etc.
This is not really clean for the underlying systems (EGPWC).

I like tracernz's solution much more to use aircraft specific files during compile time.
Would require to move the ND to fbw-common and special implementations for fbw-a320 and fbw-a380.
Is the cleaner solution in the end.

PR #7835, opened 2023-02-28
perf: optimize pfd and simconnect requests

Saschl:
Fixes #[issue_no]

## Summary of Changes

 - Removes usages of SIMCONNECT_PERIOD_ONCE as it's not recommended for regularly requesting data (https://docs.flightsimulator.com/html/Programming_Tools/SimConnect/API_Reference/Structures_And_Enumerations/SIMCONNECT_PERIOD.htm?rhhlterm=SIMCONNECT_PERIOD_ONCE&rhsearch=SIMCONNECT_PERIOD_ONCE)
 - PFD drawing optimizations
   - Reduce amount of unnecessary draws
   - Use translate3d for translation to make use of GPU acceleration

## Additional context

Discord username (if different from GitHub):

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8169, opened 2023-08-14
fix: use VISUAL_FRAME instead of SIM_FRAME for simconnect requests

lukecologne:
## Summary of Changes

This PR replaces usages of `SIMCONNECT_PERIOD_SIM_FRAME`, introduced by #7835 with `SIMCONNECT_PERIOD_VISUAL_FRAME`.
`SIMCONNECT_PERIOD_SIM_FRAME` seems to create issues, where simconnect only sends every 1s instead of every sim frame, which of course causes all sorts of issues. VISUAL_FRAME does not appear to have this issue, so we will use that instead.

## Additional context


Saschl:
Thanks 👍  Can be merged IMO.

PR #8208, opened 2023-09-12
fix(efb): amend dashboard weather widget to better reflect APPR phase weather input order

oim1:
## Summary of Changes
This PR changes the order weather is displayed on the flyPad's weather widget. This reordering makes it easier to read and input weather into the MCDU for the approach perf page. 

## Screenshots (if necessary)

Before:
image

After:
image

MCDU APPR PERF page:
image

## References
N/A

## Additional context

Open to further discussion if this change is not necessary. 

Discord username (if different from GitHub): oim#0

## Testing instructions
1. Once loaded into aircraft, turn on flyPad. 
2. Import weather from Simbrief and check that temperature is to the left of wind speed/direction. 


FoxtrotSierra6829:
In my opinion, where both temperature and dew point are given, they should strictly be next to each other.
In general, I‘d rather be in favor of the order being derived from standard weather reporting formats used in aviation, such as the METAR and the weather information in the ATIS rather than to adapt it to the FMS of a specific aircraft. When listening to the ATIS in real life, the correct order ensures that all relevant information is being copied and is easily identifiable.
The order matching these formats would then be wind, temperature, dew point, QNH

oim1:
> In my opinion, where both temperature and dew point are given, they should strictly be next to each other.
> 
> In general, I‘d rather be in favor of the order being derived from standard weather reporting formats used in aviation, such as the METAR and the weather information in the ATIS rather than to adapt it to the FMS of a specific aircraft. When listening to the ATIS in real life, the correct order ensures that all relevant information is being copied and is easily identifiable.
> 
> The order matching these formats would then be wind, temperature, dew point, QNH

This is a fair point. I did not realise there was a standard format. 

I could change it to match that format if that is preferred.

.github/CHANGELOG.md
```diff
@@ -8,46 +8,47 @@
...
+33. [FWC] Added "twenty five hundred" and "two thousand" callouts - @tracernz (Mike)
+34. [FMS] Load procedures that reference a runway correctly - @tracernz (Mike)
+35. [MCDU/ND] Corrected altitude formatting with respect to transition altitude and transition level - @tracernz (Mike)
+36. [PFD] Adjusted size of pitch trim messages - @tracernz (Mike)
+37. [BLEED] Allow engine start from crossbleed air - @BlueberryKing (BlueberryKing)
+38. [BLEED] Improve bleed page visuals on the SD - @BlueberryKing (BlueberryKing)
+39. [BLEED] Use more accurate pneumatic valves and sensors - @BlueberryKing (BlueberryKing)
+40. [EFB] flyPad OS 3.2 - Cosmetic changes for throttle configuration settings and support for multiple airframes - @2hwk (2cas)
+41. [SD] Connect SD BLEED page APU bleed pressure indication to correct ADRs - @BlueberryKing (BlueberryKing)
+42. [EFB] Change the order of the weather dashboard elements to reflect the order of the MCDU APPR PERF phase - @oim (oim1)
```
2hwk, line 51: fix diff

tracernz:
The only concern I have (which is already valid, but further reinforced now) is that people will enter the true wind direction from this METAR into the PERF APPR page rather than the magnetic wind from an ATIS.

FoxtrotSierra6829:
> The only concern I have (which is already valid, but further reinforced now) is that people will enter the true wind direction from this METAR into the PERF APPR page rather than the magnetic wind from an ATIS.

Maybe we can write the wind in the dashboard as e.g. `320°(T)/03kt` ?

tracernz:
I think we should do that. The other idea I had is to allow toggling between true and magnetic if we have the airport lat/lon available (so we can get the mag var).

Lauvaan:
This is probably a bit out of scope of this pr, but a possibility to rearrange the order the weather is displayed. If it would be possible, maybe just dragging them to different positions could be a nice idea. This feature could also be used to rearrange the other widgets(like rearranging the whole widgets) on the main screen. 

But as I said, may be a bit out of scope for this pr, for now I'd say best to either use what we have now, or probably better use the same order they're in metars/atis's.

.github/CHANGELOG.md
tracernz, line 51: Please keep all these 1.... markdown handles the numbering automatically. Also, move your own one to v0.12.0 (after a rebase/merge of master).

tracernz (changes requested):

Benjozork:
@oim1 Is this still being worked on?

Benjozork:
Closing due to inactivity.

PR #8214, opened 2023-09-13
feat(eis): move trans alt/lvl to arinc bus

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Utilize the FMGC DMC bus to transmit transition altitude and transition level. This should not have any visible effect. However, it gets rid of the undocumented `L:AIRLINER_TRANS_ALT` and `L:AIRLINER_APPR_TRANS_ALT` variables.

## Additional context

## Testing instructions

- Check that the QNH/STD indication on the PFD still blinks once the aircraft crosses the transition altitude/level in the climb/descent.
- Check that there is no blinking when the corresponding fields are empty in the MCDU.


Sioux812:
Just to further clarify this topic further, Thales logic for the APPR Perf page will auto populate whatever the typical Transition FL is for that country/region plus 500ft. This is to prevent the PFD altimeter settings flashing if pilots were instructed to level off at the Transition Level during descent (pilots should be on the provided altimeter anytime at and below the transition level). So, for example we would see on the APPR performance page "TRANS FL" and below FL45 (or FL65, FL185, etc.).
APPR PERF Page
IMG_0393

BlueberryKing:
> Just to further clarify this topic further, Thales logic for the APPR Perf page will auto populate whatever the typical Transition FL is for that country/region plus 500ft. This is to prevent the PFD altimeter settings flashing if pilots were instructed to level off at the Transition Level during descent (pilots should be on the provided altimeter anytime at and below the transition level). So, for example we would see on the APPR performance page "TRANS FL" and below FL45 (or FL65, FL185, etc.). APPR PERF Page IMG_0393

Thanks for the info! Yes, on Honeywell, the transition altitude and transition level should be imported from the navigation database as well. Unfortunately, we don't have access to this through MSFS. So in the A32NX, it's imported from the flight plan if one is uplinked from SimBrief.

Sioux812:
Is it possible to make the FBW version auto add the 500ft to the transition level if it successfully uplinks one then?

fbw-a32nx/src/systems/instruments/src/PFD/AltitudeIndicator.tsx
```diff
@@ -620,12 +620,12 @@ class AltimeterIndicator extends DisplayComponent<AltimeterIndicatorProps> {
...
     private handleBlink() {
         if (this.mode.get() === 'STD') {
-            if (this.flightPhase > 3 && this.transAltAppr > this.props.altitude.get() && this.transAltAppr !== 0) {
+            if (this.flightPhase > 3 && this.transLvlAr.isNormalOperation() && 100 * this.transLvlAr.value > this.props.altitude.get()) {
                 this.stdGroup.instance.classList.add('BlinkInfinite');
             } else {
                 this.stdGroup.instance.classList.remove('BlinkInfinite');
             }
-        } else if (this.flightPhase <= 3 && this.transAlt < this.props.altitude.get() && this.transAlt !== 0) {
+        } else if (this.flightPhase <= 3 && this.transAltAr.isNormalOperation() && this.transAltAr.value < this.props.altitude.get()) {
```
2hwk, line 628: ``  100 * this.transLvlAr.value > this.props.altitude.get() `` vs ``this.transAltAr.value < this.props.altitude.get()``

Is there a reason one value is multiplied by 100 and in the other case it is not? It seems like a mistake

2hwk: also clearer variable naming might help i.e. is this altitude in ft or flight level

2hwk (changes requested):

fbw-a32nx/src/systems/instruments/src/PFD/AltitudeIndicator.tsx
tracernz: > ` 100 * this.transLvlAr.value > this.props.altitude.get()` vs `this.transAltAr.value < this.props.altitude.get()`
> 
> Is there a reason one value is multiplied by 100 and in the other case it is not? It seems like a mistake

No, this is correct. One is a flight level while the other is an altitude.

tracernz: > also clearer variable naming might help i.e. is this altitude in ft or flight level

One is transAlt(itude) and the other is transLevel... what could be clearer?

fbw-a32nx/src/systems/fmgc/src/components/EfisLabels.ts
```diff
@@ -17,14 +17,25 @@ export class EfisLabels implements FmgcComponent {
...
 
         if (transitionLevel !== this.lastTransitionLevel) {
-            SimVar.SetSimVarValue('L:AIRLINER_APPR_TRANS_ALT', 'Number', (transitionLevel ?? 0) * 100);
+            Arinc429Word.toSimVarValue(
+                'L:A32NX_FM1_TRANS_LVL', transitionLevel ?? 0, transitionLevel !== undefined ? Arinc429SignStatusMatrix.NormalOperation : Arinc429SignStatusMatrix.NoComputedData,
+            );
+            Arinc429Word.toSimVarValue(
+                'L:A32NX_FM2_TRANS_LVL', transitionLevel ?? 0, transitionLevel !== undefined ? Arinc429SignStatusMatrix.NormalOperation : Arinc429SignStatusMatrix.NoComputedData,
+            );
+
```
tracernz, line 38: These should all go FW when the respective FM is not powered. The old FMC has some stuff that takes care of handling the bus outputs automatically, including proper quantisation and range limiting per the word encoding, but nothing exists yet in the TS code.

fbw-a32nx/src/systems/instruments/src/PFD/AltitudeIndicator.tsx
```diff
@@ -589,15 +589,15 @@ class AltimeterIndicator extends DisplayComponent<AltimeterIndicatorProps> {
             this.handleBlink();
         });
 
-        sub.on('transAlt').whenChanged().handle((ta) => {
-            this.transAlt = ta;
+        sub.on('fmTransAltRaw').whenChanged().handle((ta) => {
+            this.transAltAr = new Arinc429Word(ta);
```
tracernz, line 593: Would be better to use an Arinc429Register so you don't need to allocate a new object every update (applies to all of these words below too).

tracernz (changes requested):

tracernz (approved):

fbw-a32nx/src/systems/instruments/src/PFD/AttitudeIndicatorHorizon.tsx
```diff
@@ -357,8 +357,10 @@ class RadioAltAndDH extends DisplayComponent<{ bus: ArincEventBus, filteredRadio
                 const raHasData = !this.radioAltitude.isNoComputedData();
                 const raValue = this.filteredRadioAltitude;
                 const verticalOffset = calculateVerticalOffsetFromRoll(this.roll.value);
-                const chosenTransalt = this.fmgcFlightPhase <= 3 ? this.transAlt : this.transAltAppr;
-                const belowTransitionAltitude = chosenTransalt !== 0 && (!this.altitude.isNoComputedData() && !this.altitude.isNoComputedData()) && this.altitude.value < chosenTransalt;
+                const useTransAltVsLvl = this.fmgcFlightPhase <= 3;
+                const chosenTransalt = useTransAltVsLvl ? this.transAltAr : this.transLvlAr;
+                const belowTransitionAltitude = chosenTransalt.isNormalOperation() && (!this.altitude.isNoComputedData() && !this.altitude.isNoComputedData())
```
Saschl, line 362: `(!this.altitude.isNoComputedData() && !this.altitude.isNoComputedData())` isn't this comparison redundant?

BlueberryKing: Of course, good catch!

Saschl:
QA Report

Tester: saschl
Date: 20.10.2023

Report:

 1. Check that the QNH/STD indication on the PFD still blinks once the aircraft crosses the transition altitude/level in the climb/descent. 
 PASS

2. Check that there is no blinking when the corresponding fields are empty in the MCDU.
 PASS

Summary:
Passed

PR #8178, opened 2023-08-25
feat(cond): air conditioning failures

mjuhe:
## Summary of Changes

This PR introduces failures for the Air Conditioning System. In order to do this properly, the system has been modified slightly to more accurately model an A320 equiped with an ACSC. Other additions/changes are:
- ARINC discrete communication
- Valve simulation and travel time (TAPRV and TAV) and pack realistic temperature response time (until packs are modelled)

Failable items:
- ACSC 1 & 2 (each lane individually)
- Trim Air Pressure Regulating Valve
- Trim Air Valves
- Cabin Fan 1 & 2
- Lavatory and Galley Fan

Other failures:
- Trim air system high pressure
- Duct overheat

All associated ECAM warnings have been added as well.

## Screenshots (if necessary)

EFBFailures
image

## Additional context

This PR will have conflicts with #8086, this one should be merged first as it has the fix for the temperature jumping when the valves are open and closed.

Discord username (if different from GitHub):

## Testing instructions

What to look for in each of the failures:
-	ACSC -> One lane fault: No effect as the second lane takes over. You will see a “COND CTL 1(2)-A(B) FAULT” ECAM warning (no chime) if you have engines off, otherwise no warning.
-	ACSC -> Two lane fault: The related pack is lost, and the hot air pressure-regulating valve and trim air valves close. You will get a “AIR PACK 1(2) FAULT” ECAM warning (single chime). Temperature regulation is lost.
-	Trim Air Pressure Regulating Valve: The “hot air” valve is stuck in position. If it fails open this has no effect. If it fails closed temperature optimisation is lost. You will only get a “HOT AIR FAULT” when the valve disagrees with the commanded position. The best way to test this is to fail the valve and then press the overhead button (if it was open, this will command closure). You will get the warning after a certain period of time. Observe the FAULT light illuminate on the overhead.
-	Trim Air System High Pressure: Hot air pressure increases but no observable effect. You will get a “COND TRIM AIR SYS FAULT” and “TRIM AIR HI PR” ECAM message (no chime).
-	[ZONE] Trim Air Valve: Optimised temperature regulation of the corresponding zone is lost. You will get a “COND TRIM AIR SYS FAULT” and “[ZONE] TRIM VALVE” ECAM message (no chime).
[...]

PR #8234, opened 2023-09-28
fix(cond): Fix SD COND page duct ovht indicator

Eearslya:
## Summary of Changes

PR #8178 added a few failures for air conditioning, including a few changes to the status page. One of these was to indicate duct temperatures as amber if they exceed the overheat threshold. This is meant to happen at 80 degrees Celsius, but because of the way that the COND page converts all temperatures to Fahrenheit, it was triggering at 80 Fahrenheit instead.

This fixes the indicator to properly trigger above 80C (176F).

## Additional context

Discord username (if different from GitHub): Eearslya

### DIFF
--- a/fbw-a32nx/src/systems/instruments/src/SD/Pages/Cond/Cond.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/SD/Pages/Cond/Cond.tsx
@@ -130,7 +130,7 @@ type CondUnitProps = {
 
 const CondUnit = ({ title, trimAirValve, cabinTemp, trimTemp, x, y, offset, hotAir } : CondUnitProps) => {
     const rotateTemp = offset + (trimAirValve * 86 / 100);
-    const overheat = trimTemp > 80;
+    const overheat = trimTemp > 176; // Overheat indicator at 80C, temperature is given in F
 
     return (
         <SvgGroup x={x} y={y}>


fbw-a32nx/src/systems/instruments/src/SD/Pages/Cond/Cond.tsx
```diff
@@ -130,7 +130,7 @@ type CondUnitProps = {
 
 const CondUnit = ({ title, trimAirValve, cabinTemp, trimTemp, x, y, offset, hotAir } : CondUnitProps) => {
     const rotateTemp = offset + (trimAirValve * 86 / 100);
-    const overheat = trimTemp > 80;
+    const overheat = trimTemp > 176; // Overheat indicator at 80C, temperature is given in F
```
tracernz, line 133: This doesn't make sense. The temp is only F here if you select alternate units. The problem is really that the unit conversions are done too early. Should be at the point they're used in the UI only.

tracernz (changes requested):

fbw-a32nx/src/systems/instruments/src/SD/Pages/Cond/Cond.tsx
Eearslya: You...are absolutely correct, and I'm not sure how I managed to mix that logic up in my head. Somehow I had it flipped and thought it would convert to F if the units were set to C. I'll work on a more complete change which converts the units later.

Benjozork:
@Eearslya Are you still planning to work on this?

Eearslya:
@Benjozork Unfortunately at this point, not likely to be any time soon. Probably better if someone else wanted to take over this one.

PR #8238, opened 2023-10-01
build: add typechecking to instruments and modules

Benjozork:
## Summary of Changes

* [x] Adds `tsc` type-checking to both instruments and standalone modules, via a custom esbuild pluigin
* [x] Get type-checking working in iginiter
* [x] Enables type-checking in PR builds
* [x] Fix issues with type-checking
* [x] Fix existing type errors found

## Philosophy and design compromises 

Type-checking is implemented via a custom esbuild plugin which invokes `tsc` before actually building the instrument/module. The plugin is integrated in both the Mach bundler (for instruments) and our custom build script from our `build-utils` module for standalone modules.

This has a few advantages:

- No need for discrete type-checking tasks or scripts that need to be run separately, requiring one more command to remember;
- Requires less modifications to our build system to conditionally turn off;
- Allows developers to locally enable type-checking for local development work if they wish, without requiring two programs to be running at once, and natively supporting watch mode.

Bringing `tsc` into the mix again has some drawbacks, notably:

- It requires every instrument and standalone module to have its own `tsconfig.json` file. Note that this file can simply extend another `tsconfig.json` file.
- It breaks some type resolutions that previously worked when using editors' built-in TS support, requiring some changes to where some types are located;
- It removes the possibility of building a singular bundle mixing both React and FSComponent, due to `tsc` not taking into account `tsconfig.json` files in subdirectories of the built project.

## Testing instructions

- Ensure the aircraft systems and avionics still behave correctly. Of particular note to verify are the EFB and FMS LNAV/VNAV.


Benjozork:
Blocked by https://github.com/synapticsim/mach/issues/7

fbw-common/src/systems/shared/src/ApproachUtils.ts
```diff
@@ -117,7 +118,7 @@ export class ApproachUtils {
         (approach: RawApproach): string;
     } = ApproachUtils.formatLongApproachName;
 
-    public static approachTypeString(type: ApproachType): string {
+    public static approachTypeString(type: ExtendedApproachType): string {
```
tracernz, line 121: We shouldn't want ExtendedApproachType here. That is WT specific (has their extensions that MSFS doesn't use). Instead the plain old MSFS ApproachType is good.

tracernz: This is the one enum that actually exists in the MSFS JS code, so we can just use the existing declaration (I've pushed a fix).

tracernz (approved):

Saschl:
QA Report

Discord: saschl
Object of testing: https://github.com/flybywiresim/aircraft/pull/8238
Tier of Testing : 2
Date : 08/10/2023

Testing Process:

Full flight from EDDS to LOWW (full local rebuild).

1. EFB functions as usual
2. LNAV, MCDU and FMS functions as expected
3. Everything else looks good as well.
4. Local build works as well with typechecking enabled :)

Testing Results:
Passed

PR #8245, opened 2023-10-08
build: remove nuking of node_modules/ in setup.sh

Benjozork:
## Summary of Changes

- Removes `rm -rf node_modules/` in `setup.sh`, which is overkill and was mostly used for when we built our own version of `msfs-sdk`. Speeds up process for switching branches locally. `npm ci` already does a clean install.
- The option is conserved via a `--clean` flag which can be passed to `setup.sh`, which is done in all CI workfflows

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -3,5 +3,4 @@
 set -ex
 
 cd /external
-rm -rf node_modules
 npm ci
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -3,4 +3,12 @@
 set -ex
 
 cd /external
+
+for arg in "$@"; do
+  if [ "$arg" = "--clean" ]; then
+    echo "Removing node_modules..."
+    rm -rf node_modules/
+  fi
+done
+
 npm ci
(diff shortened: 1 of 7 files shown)


scripts/setup.sh
```diff
@@ -3,5 +3,12 @@
 set -ex
 
 cd /external
-rm -rf node_modules
+
+for arg in "$@"; do
+  if [ "$arg" = "--clean" ]; then
+    echo "Removing node_modules..."
+    rm -rf node_modules/
```
Saschl, line 10: I‘d suggest the following:

Instead of using rm rf we can use `npm ci` which does a clean install for the `—clean` case.

Otherwise let’s use regular `npm i`

tracernz: You don't need to trust npm is doing the right thing and cleaning up properly if you `rm -rf`.

Saschl: Fair enough, then let’s just remove the npm ci command, otherwise we won’t gain much 👍

Saschl (approved):

PR #8260, opened 2023-10-22
feat(APU): Add APU start motor failure

uX-Pizza:
Fixes #[issue_no]

## Summary of Changes

The goal of this PR is to add a failure to the APU start motor which, when active, prevents the APU from beeing started.

## References

https://www.smartcockpit.com/docs/A320-Auxiliary_Power_Unit.pdf
This states the FAULT light illuminates for "no accelleration", so I dont know exactly if the FAULT light should come on or not in this case, an accuracy check would be highly appreciated.

## Additional context

This is my first contribution to this project, so dont judge me too hard.

Discord username (if different from GitHub): pizza187

## Testing instructions

1. Spawn in
2. Energize the aircraft
3. Observe the APU starting without the failure selected / Observe the APU not starting with the failure selected
4. Observe the N value drop back down when the failure is selected below 55% during the start cycle
5. When the failure is deselcted the FAULT light should disappear when the APU master switch is deselected and power to the electronic control box is cut
6. Test this in combination with the APU fire pushbutton


tracernz (commented):
Please add a changelog entry.

tracernz:
Hows does the failure actually manifest here? I think it should fail to reach the speed threshold after the first 30 seconds then the slow start fault should be triggered, causing the ECAM "<ins>APU</ins> AUTO (EMER) SHUT DOWN"?

uX-Pizza:
> Hows does the failure actually manifest here? I think it should fail to reach the speed threshold after the first 30 seconds then the slow start fault should be triggered, causing the ECAM "APU AUTO (EMER) SHUT DOWN"?

I dont know exaclty. I thought the FAULT light would come on immediately as the starter motor wouldnt use any power and it is not really a slow start but just no start at all. However I didnt think about the ECAM memo, it most likely should be triggered because the APU (in the current implementation) does an auto shutdown when it had time to spool up a bit.

If you have any sources for information, please reach out to me as I think they are pretty hard to find.

Benjozork:
What is the status on this?

PR #8267, opened 2023-10-24
fix(fwc): Fixes the Master Warning on startup

Maximilian-Reuter:
Fixes #[issue_no]

## Summary of Changes

On startup in C/D the aircraft plays a master warning until the startup EWD messages are empty. This was caused by the pseudo FWC checking whether the aircraft was on ground only after checking for dual engine failure which meant that the FWC didn't update the on ground status until after the check for dual engine failure and since the on ground status was initialized with false, the dual eng failure triggered on initial load.

This PR changes the order of those checks so that the setting of the on ground status happens first.

## Additional context

Discord username (if different from GitHub): \_chaoz_

## Testing instructions

just a quick fix that shouldn't affect anything beyond the fwc initialization.

1. start C/D
2. power up the aircraft
3. no master warning should occur

### DIFF
--- a/fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
+++ b/fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
@@ -920,6 +920,9 @@ export class PseudoFWC {
 
         this.computedAirSpeed.set(Arinc429Word.fromSimVarValue('L:A32NX_ADIRS_ADR_1_COMPUTED_AIRSPEED'));
 
+        // needs to happen before dual eng failure check
+        this.aircraftOnGround.set(SimVar.GetSimVarValue('SIM ON GROUND', 'Bool'));
+
         /* ENGINE AND THROTTLE */
 
         this.engine1State.set(SimVar.GetSimVarValue('L:A32NX_ENGINE_STATE:1', 'Enum'));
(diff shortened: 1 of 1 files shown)


fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
```diff
@@ -1594,9 +1597,11 @@ export class PseudoFWC {
                     }
 
                     if (value.failure === 3) {
+                        console.log(`Playing Master warning Source:${key}`);
```
Saschl, line 1600: Pls remove the logs and we're good to go :)

Saschl:
Also might be worth a changelog entry I think. Would go to 0.11 too imo

tracernz:
> Also might be worth a changelog entry I think. Would go to 0.11 too imo

I’d suggest keeping all the master changelogs under 0.12 and once the 0.11 release is done tidy the cherry-picked ones up all at once.

Maximilian-Reuter:
> ~Also might be worth a changelog entry I think. Would go to 0.11 too imo~ EDIT: Tracers suggestion is better

is this even an issue on stable right now ?

fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
Maximilian-Reuter: duh, knew I forgot smth

tracernz:
> > ~Also might be worth a changelog entry I think. Would go to 0.11 too imo~ EDIT: Tracers suggestion is better
> 
> is this even an issue on stable right now ?

0.11 stable is master as of 2 days ago.

Saschl (approved):

Saschl:
QA Report

Discord: saschl
Object of testing:
Tier of Testing : 1
Date : 26/10/2023

Testing Process:

Loaded the plane and powered up the plane. No master warning audible.
Slewed into the air with engines off. Master warning comes up as expected.

Testing Results:
Passed

PR #8272, opened 2023-10-27
feat(efcs): add precontrol to roll angle controller

lukecologne:
## Summary of Changes

This PR adds a precontrol to the roll angle controller of the lateral law. Due to this change, the dynamic response `roll rate command -> roll rate` is improved, without affecting the stability of the control law. Due to this improvement in dynamic response, it is possible to reduce the characteristic frequency of the roll rate controller, which will:
* improve the stability margin of the controller, and
* decrease the "strength" of the roll angle hold behaviour of the control law, increasing the effect of wind.

Since the stability margin is improved, it is possible to also increase the maximum simulation rate without affecting the stability of the lateral law. Thus:
* the default maximum simrate was increased to 8
* the default minimum effective fps was reduced to 5.

On my setup, I was able to maintain a stable 8x sim rate, while having around 40fps (at 1x sim rate), though this could vary so should be subject to testing.

## Additional context

## Testing instructions

* Fly the aircraft (in various situations: slow, fast, cruise, approach etc.) and check that the roll law is stable and has good dynamic response.
* Test the stability of the roll law at higher than normal sim rates, and also test the stability of the AP (during cruise at high sim rate for example).

If you want to test the stability of the law at sim rates higher than currently allowed, you can change the settings and enable/disable the performance monitoring in the `ModelConfiguration.ini` file under the `[AUTOPILOT]` section, with the keys `MINIMUM_SIMULATION_RATE=rate`, `MAXIMUM_SIMULATION_RATE=rate`, `LIMIT_SIMULATION_RATE_BY_PERFORMANCE=true/false`, `SIMULATION_RATE_REDUCTION_ENABLED=true/false`.

Further, this PR should also be tested by pilots to verify the changed handling characteristics of the law.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -16,6 +16,8 @@
 1. [MCDU] Dash alternate time predictions if alternate fuel is manually inserted @BravoMike99 (bruno_pt99)
 1. [BLEED] Add Air starter unit to enable ground starts - @Maximilian-Reuter (\_Chaoz_)
 1. [MCDU] FUEL PRED layout improvemnts @BravoMike99 (bruno_pt99)
+1. [EFCS] Add precontrol to roll angle controller to improve dynamic response - @lukecologne (luke)
+1. [EFCS] Increase maximum sim rate limit to 8 due to improvements in roll law stability - @lukecologne (luke)
 
 ## 0.11.0
 
--- a/fbw-a32nx/docs/Configuration/ModelConfiguration.ini
+++ b/fbw-a32nx/docs/Configuration/ModelConfiguration.ini
@@ -20,7 +20,7 @@
 
 ; !! WARNING CHANGE AT YOUR OWN RISK !!
 ; maximum allowed simulation rate
-;maximum_simulation_rate = 4
+;maximum_simulation_rate = 8
 
 ; !! WARNING CHANGE AT YOUR OWN RISK !!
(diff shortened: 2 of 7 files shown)


tracernz (approved):

tracernz:
Please add a changelog entry, especially mentioning the sim rate availability. I think that will make many people happy!

lukecologne:
Done. Added two separate entries since it would have been to long for only one

Valastiri:
Would likely need a docs update + bot team update to accommodate this PR and support efforts on our Discord

lukecologne:
Ok, so this was test-flown by @IbrahimK42 for a few flights, and @pinatacolada has also tested it. The model review by @IbrahimK42 will follow shortly, after which I'd say we can merge and see if there are any issues in master. About the docs, @Valastiri, what do you think will need changing? Only the max simrate/min fps, correct?

IbrahimK42 (approved):
Review of PR #8272

Reviewer: IbrahimK42

Review date: 12/11/2023

Roll maneuvers of varying intensity were flown at low speed (IAS=140 kn) and high speed (IAS=320 kn) at low altitude (1000 ft), medium altitude (14000 ft) and high altitude (35000 ft) with and without wind. There is a small, negligible deviation in the steady-state roll rate (2-3%).

No other findings were detected. 

Review passed!

Well done :)

PR #8288, opened 2023-11-07
fix: fms v2 geometry guards

flogross89:
## Summary of Changes
When using FlightPlanService for the a380x MFD, some things had to be changed in order for it to be working:

- Introduce some guards around usages of geometries, if they aren't available (yet)
- Connect FlightPlanServices' event bus to existing one
- Alter instantiation of FlightPlanService in A32NX_NavSystem (really inexperienced on that part, don't know if that works) 

**Couldn't test it in the A32NX on my machine unfortunately.**

## Additional context
@Benjozork 

floridude


fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -316,7 +316,7 @@ export class EfisSymbols {
             // eslint-disable-next-line no-lone-blocks
 
             // ALTN
-            if (this.flightPlanService.hasActive) {
+            if (this.flightPlanService.hasActive && this.guidanceController.activeGeometry) {
```
BlueberryKing, line 319: Maybe using `this.guidanceController.hasGeometryForFlightPlan(...)` is nicer here in the the other if statements below.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -352,7 +352,9 @@ export class EfisSymbols {
             }
 
             // TMPY
-            if (this.flightPlanService.hasTemporary) {
+            if (this.flightPlanService.hasTemporary && this.guidanceController.temporaryGeometry) {
+                const b = this.guidanceController.getGeometryForFlightPlan(FlightPlanIndex.Temporary);
+                console.log(b);
```
BlueberryKing, line 357:
```suggestion
```

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/NavSystems/A320_Neo/A32NX_NavSystem.js
```diff
@@ -62,7 +62,7 @@ class NavSystem extends BaseInstrument {
         if (this.nodeName.includes('CDU')) {
             this.currFlightPhaseManager = Fmgc.getFlightPhaseManager();
 
-            this.currFlightPlanService = new Fmgc.FlightPlanService();
+            this.currFlightPlanService = new Fmgc.FlightPlanService(new EventBus());
```
BlueberryKing, line 65: Am I correct that this change is only relevant for the A380? I don't see how passing in a new event bus instance in the constructor is different from just instantiating it in the constructor of the flight plan service.

BlueberryKing (changes requested):

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/NavSystems/A320_Neo/A32NX_NavSystem.js
flogross89: Yep, this is only relevant for the A380.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
flogross89: You're right, didn't know this method existed. Changed it.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
flogross89: Sorry, don't know how that slipped through. Thanks for noticing.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/NavSystems/A320_Neo/A32NX_NavSystem.js
tracernz: There should only really be one event bus for each entire instrument unless you're trying to do something very specific.

fbw-a32nx/src/systems/fmgc/src/guidance/geometry/GeometryFactory.ts
```diff
@@ -144,13 +144,15 @@ export namespace GeometryFactory {
 
                 const prevLeg = geometry.legs.get(i - 1);
 
-                const oldInboundTransition = geometry.transitions.get(i - 1);
-                const newInboundTransition = TransitionPicker.forLegs(prevLeg, newLeg);
+                if (prevLeg && newLeg) {
```
BlueberryKing, line 147: Do you what situations would lead to `prevLeg` or `newLeg` being null here? Just so we don't cover up an underlying issue with a nullcheck.

BlueberryKing (approved):
LGTM. See the comment about `prevLeg` or `newLeg` being null.

fbw-a32nx/src/systems/fmgc/src/guidance/geometry/GeometryFactory.ts
flogross89: geometry.legs seems to be missing indices in the middle of legs. In my case, when I create a new city pair from EDDM to EHAM, a map is created with legs at key 0 and key 2, but 1 is missing. So as soon as either prevLeg or newLeg is referring to key 1, the error is thrown

flogross89: image
debug output when i = 0

PR #8296, opened 2023-11-10
fix(hoppie): Remove ATSU/AOC source checks 

auroraisluna:
Fixes #[issue_no]

## Summary of Changes

This PR removes the EFB METAR and ATIS source check that prevent Hoppie from connecting. There seems to be no reason for this lock other than enforcing the VATSIM or IVAO weather source if using Hoppie. However, there is no specific check for if the Hoppie key is a VATSIM or IVAO key, which means either way there is no proper check for this. In addition, the Hoppie system can be used outside of VATSIM or IVAO networks, which means if the airplane is used with other networks and the ATIS or METAR source is different, Hoppie will be completely disabled. 

There is no reason for such lock to exist in my opinion, as it limits for no reason, the usage of the system. I quote a thought of a CS member that I agree with and could have not worded better than them: 
> Why we force it on the user to pick one?
> Sure, ensuring they have the right weather is a logical case - but is that really for the FBW aircraft to 'enforce' ?
> Do we check their AIRAC is updated to the latest so to be sure they're capable of flying updated airways if you select VATSIM/IVAO sources? Reject the flight plan if not? (more extreme example, but same principle)

This PR is especially relevant now that the FBW API is down and people use MSFS sources for weather, but will not be able to use Hoppie. 

## Additional context

Discord username (if different from GitHub): alepoun

## Testing instructions

Test Hoppie functions work. (CPDLC, Telex etc) with and without VATSIM or IVAO set in EFB ATSU/AOC settings. 


auroraisluna:
If someone can re-run the PR title action as the title has been updated

Maximilian-Reuter (approved):

beheh:
I don't think there's any reason to QA this, as this is a piece of code removal only. I'll just poll the team to see if anybody objects.

auroraisluna:
> I don't think there's any reason to QA this, as this is a piece of code removal only. I'll just poll the team to see if anybody objects.
> 
> You could add a CHAGELOG entry still.

Roger, what changelog entry is this considered to fit under? Just Hoppie or Datalink?

auroraisluna:
> I don't think there's any reason to QA this, as this is a piece of code removal only. I'll just poll the team to see if anybody objects.
> 
> You could add a CHAGELOG entry still.

Done

PR #8299, opened 2023-11-12
feat(efcs): update yaw damper with calculated gain

lukecologne:
## Summary of Changes

This PR adjusts the yaw damper (normal law only) gains. Previously, the gains have been determined by trial-and-error, through a lookup table as a function of TAS.
Now, the gains are calculated as a function of CAS, and from aerodynamic coefficients (similarly as the roll law). The target parameter is the damping ratio, which is now chosen to be 1/sqrt(2) independent of CAS. Previously, the damping ration was overdamped (>1), with the exact damping ratio varying with the CAS.
This reduction in damping ratio results in an effective reduction of the gain across the whole speed range, which results also
in a reduced strength in turn coordination, which has been noted as too high.
Further, the yaw damper gains on the ground only, have been reduced. Below 40kts, the gain is 0 (yaw damper disabled), which prevents the rudder from being deflected when turning while taxiing. Above 40kts, the gain is lower than before, which should reduce oscillations that sometimes occur during taxi/rollout at high speed while steering.

## Screenshots (if necessary)

Demonstration of the new vs old damping in the air. Plotted parameter is yaw rate:
Before (notice how the yaw rate tends to increase sometimes when releasing the rudder): https://streamable.com/0e82ys
After (now, the yaw rate smoothly reduces to 0): https://streamable.com/20pkhv

Demonstration of the reduced oscillations and deflection on ground:
Before: https://streamable.com/fmel3f
After: https://streamable.com/2o9lyf

## Additional context


tracernz (approved):
Please add changelog.

lukecologne:
I just realized that the rudder elasticity table is still there. It might be worth to actually integrate this into this PR, as the yaw damper needs the rudder response to be linear, otherwise, the damping will b

lukecologne:
Oops, accidentally closed it, sorry

lukecologne:
Actually, you can see that the damping is too low at high speeds in the "after" video, the yaw rate overshoots first and then settles to 0 again.

donstim:
> I just realized that the rudder elasticity table is still there. It might be worth to actually integrate the removal of the table into this PR, as the new calculated yaw damper gain needs the rudder response to be linear, otherwise, the damping will be too low at high speeds.

The rudder elasticity table has always been there, that is, it was never removed. It was only the elevator elasticity table that was removed. I think there may have been discussion about removing the rudder elasticity table, but it was never done. That table originally was used to approximate the reduction in rudder deflection with speed. That would be better handled by the yaw control system, but I don't know what's in there and whether/how it might be affected by the rudder elasticity table.  I would be for removing that table if it's not needed or interfering. Just want make everything works/flies properly without it.

lukecologne:
For manual flight, there is only the yaw damper which this PR modifies. With this PR, the gain is based on flight mechanical formulas, and assumes that there is no elasticity tables. It will still works with the table, but the effectiveness is reduced. In automatic flight, we also have a sideslip controller, which is only used for decrab on autoland, but it's always active (and commands 0° beta when not decrabbing). For decrabbing (at landing speeds), the elasticity table is anways just 1, so it will not have an effect.

lukecologne:
For reference, the gain yaw rate->rudder position is now: `k_r = -(2*damping_factor*sqrt(N_beta) + N_r) / N_zeta`

lukecologne:
> Please add changelog.

Done

donstim:
Will the rudder travel limitations with increasing speed continue to be adequately simulated after removal of the rudder elasticity table?

lukecologne:
The rudder travel limiter just uses the limits given in the FCOM, so it should be correct, if that's what you mean

donstim:
> The rudder travel limiter just uses the limits given in the FCOM, so it should be correct, if that's what you mean

Yes, that's exactly what I mean. The rudder elasticity table was only used to try to approximate the rudder travel limiter. This was before there was an actual limiter in the code. If there is a rudder travel limiter in the code, there is no need for the rudder elasticity table. Thanks.

lukecologne:
Yep, we have a simulated rudder travel limiter, which the FAC commands. The values are simply taken straight out of the FCOM.

tracernz (approved):

donstim:
Is there anything holding this up?

lukecologne:
Not really, I was going to wait for some QA testing but then kind of forgot about it a little... I think we can just merge in dev tbh

PR #8283, opened 2023-11-04
fix(hoppie): Updated Hoppie Poll rate

auroraisluna:
Fixes #[issue_no]

## Summary of Changes

Changes the polling rate to Hoppie Network from a low amount to the recommended 45-75 seconds (randomized) to comply with rate limits. 

## Additional context

Rate limits requests can be found here: https://www.hoppie.nl/acars/system/tech.html
& here: https://www.hoppie.nl/acars/system/cpdlc.html
Quoting: 
> Many clients have set their poll rate at 5 seconds and this is simply never needed, please SET IT TO A MINUTE.

> It is heavily recommended to poll once between every 45 and 75 seconds, randomly timed so that the average server load is constant. If you expect a message, such as right after you sent a request, you may increase the polling rate to once per 20 seconds for a short time.

Discord username (if different from GitHub): alepouna

## Testing instructions

1. Spawn in a runway/gate 
2. Make sure you have Hoppie integration enabled (https://docs.flybywiresim.com/fbw-a32nx/feature-guides/hoppie/) and available (align ADIRS, weather to VATSIM, etc. as docs detail)
3. Create a flight number in INIT A 
4. Test you can receive Hoppie requests by sending a TELEX via MDCU AOC MENU to your own callsign. 
5. You should receive this message within 45-75 seconds after sending it. 
6. Keep sending requests ever so often. You should NOT receive a COMPANY MESSAGE by SYSTEM warning you about the rate limit, if everything is correct. You should test this for at least 5 minutes. 

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8303, opened 2023-11-15
feat(nxapi): Decrease TELEX API poll rate to random number 45-75. 

alepouna:
Fixes #[issue_no]

## Summary of Changes

Changes the polling rate to FBW API from static 15 seconds to a random number between 45 & 75 (randomized) to deload the API.

## Additional context
Discord username (if different from GitHub): alepouna

## Testing instructions

1. Spawn in a runway/gate
2. Make sure you have Telex integration enabled https://docs.flybywiresim.com/fbw-a32nx/feature-guides/freetext/
3. Create a flight number in INIT A
4. Test you can receive FBW requests by sending a TELEX via MDCU AOC MENU to your own callsign.
5. You should receive this message within 45-75 seconds after sending it.

Recommend trying this with another plane as well not just self-tests.


fbw-common/src/systems/datalink/router/src/webinterfaces/NXApiConnector.ts
auroraisluna, line 1: On line 190
>             if (NXApiConnector.updateCounter++ % 4 === 0) {
This condition from my understanding is so that the code "CreateAircraftStatus" runs only every 4th time, with the old code that is every minute. Now that the API is polled every 45-70 seconds, should this be removed completely or be reduced to 2?

pdellaert: Hmm, this call seems to be used to update the status of the aircraft in the TELEX, which is used for the live flight map and status information of the aircraft (assuming position etc, ...). At least that's my quick thought, haven't dug deep. 

If that's the case, right now, the aircraft details get updated every 1 minute. With your change, plus changing this to 2, we change that to random between 1 minute and 2.5 minutes. 

I don't have an issue with that, but might need some people to sign off on that.

auroraisluna: Could also run the aircraft status update on a different loop as well, if 1 minute should always be achieved.

beheh:
Duplicate of #8283

auroraisluna:
> Duplicate of #8283

It is not, my other PR #8283 changes the Hoppie rate, this one changes the FBW API rate. Not a duplicate, just same fashion applied.

beheh:
I see, sorry about that.

auroraisluna:
> I see, sorry about that.

no worries, can we remove the duplicate tag and add QA 1? unless QA not needed

fbw-common/src/systems/datalink/router/src/webinterfaces/NXApiConnector.ts
Maximilian-Reuter: In my opinion, if we just remove the line then on average nothing changes compared to the status quo.
I feel like once a minute is fine. Less code the better IMHO

auroraisluna: Could also do this on another PR as well, to also update the name of the functions to be slightly more descriptive to their purpose. Thoughts?

Maximilian-Reuter: I think doing it here might make more sense as those changes are related

auroraisluna:
Removed the CreateAircraftStatus condition, RTT

pilotseyea350:

Discord          : (PilotEyesA350)
Object of testing: (#PR 8303)
Tier of Testing  : (1)
Date             : (07/01/2024)

Testing Process: 
Spawn at gate 308 at EDDM
Flight Number DEV123 created in INIT A
Sent  "HI ILY" via MDCU AOC MENU to DEV123
Message Received after 17 seconds

Negatives:
N/A

Testing Results: 
Passed

PR #8336, opened 2023-12-22
fix(lights): Fixed Lighting Auto Config on spawn

Maximilian-Reuter:
Fixes #[issue_no]

## Summary of Changes

This PR fixes the auto lighting config, that tries to set the brightness of the cockpit screens and the dome, flood and integral lights based on the ambient brightness in the sim.

This now leads to dome lights, flood light and integral lights to be on when spawning at night or dusk and dawn.

For people who don't like the auto lighting config, they can use the EFB light presets to load their own preset automatically.

## Additional context

## Testing instructions

1. Spawn either C/D
2a. IF not DAY: Capt dome light should be on dim
2b. IF DAY: nothing should be on
3. Turn on BAT and GPU
4a. IF not DAY: floods should be at 20% and integral lights should be on and screens at either 15% or 35% depending on how dark it is
4b. IF DAY: Only screens should come on at 85% brightness


tracernz:
We have code already that sets the brightness all based on the ambient light level but it seems to not be working. Better to fix that instead.

Maximilian-Reuter:
> We have code already that sets the brightness all based on the ambient light level but it seems to not be working. Better to fix that instead. It probably needs to wait for the GameStateProvider to say we're ingame, as the .flt files are probably overriding it if it's too quick.

but what would the knobs then do ?

tracernz:
> 
> but what would the knobs then do ?
> 
> EDIT: also doesnt that code only work on screens ? Not the integ lights or the dome light ?

It's supposed to set everything once on spawn, so the knobs are still good. It does do integ lights, not but not dome.

becas22:
> I have had multiple people complain about having to always turn up the brightness on every screen and also remark that it feels unrealistic as screens don't need to be turned down anymore since they aren't CRTs.

FYI, the parking flow still orders to dim the DUs

fbw-a32nx/src/systems/systems-host/index.ts
```diff
@@ -30,6 +32,12 @@ class SystemsHost extends BaseInstrument {
         this.hEventPublisher = new HEventPublisher(this.bus);
         this.powerSupply = new PowerSupplyBusses(this.bus);
         this.atsu = new AtsuSystem(this.bus);
+        KeyEventManager.getManager(this.bus).then((manager) => {
+            this.keyInterceptManager = manager;
+            if (this.gameState === 3) {
```
tracernz, line 37:
```suggestion
            if (this.gameState === GameState.ingame) {
```

tracernz (commented):
You should probably really await both GameStateProvider and KeyEventManager before calling initLighting. There should then be no need to call it in the update loop.

tracernz (approved):

Maximilian-Reuter:
Got multiple reports that it works so I will merge

PR #8351, opened 2023-12-30
chore: remove github release artefacts

tracernz:
Fixes #[issue_no]

## Summary of Changes

Our build artefacts no longer fit within GitHub's 2 GiB file size limit, causing the CI steps to fail. Remove these steps so we no longer try as we know it will fail.

## Screenshots (if necessary)

image

## Additional context

### DIFF
--- a/.github/workflows/experimental.yml
+++ b/.github/workflows/experimental.yml
@@ -50,35 +50,3 @@ jobs:
           CDN_BUCKET_DESTINATION: addons/a32nx/experimental
         run: |
           ./scripts/cf-cdn.sh $CDN_BUCKET_DESTINATION ./fbw-a32nx/out/build-modules
-      - name: Delete old GitHub Pre-Release assets
-        uses: mknejp/delete-release-assets@v1
-        with:
-          token: ${{ github.token }}
-          tag: ${{ env.EXPERIMENTAL_PRE_RELEASE_TAG }}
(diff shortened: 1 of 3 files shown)


Saschl (approved):

FoxtrotSierra6829:
I'd prefer to keep the files uploaded and just split it at 2GB. The standard zip tool should be capable of this. I didn't get to do that yet however

Maximilian-Reuter:
> I'd prefer to keep the files uploaded and just split it at 2GB. The standard zip tool should be capable of this. I didn't get to do that yet however

What is the thinking there? We have a perfectly good installer. Plus it is not the nicest thing to use GitHub for this. If we really do need a separate download we can just host that on the cdn.

FoxtrotSierra6829:
> > I'd prefer to keep the files uploaded and just split it at 2GB. The standard zip tool should be capable of this. I didn't get to do that yet however
> 
> What is the thinking there? We have a perfectly good installer. Plus it is not the nicest thing to use GitHub for this. If we really do need a separate download we can just host that on the cdn.

We should have a direct download, some people don't like to use the installer for whatever reasons. I think for FBW products it makes sense to use it because of the modular updating but for other stuff I dislike to have a separate installer for everything. Then I think some regions have issues with accessing cloudflare services.

PR #8357, opened 2023-12-30
feat: Plugged-Window mesh added

Henrik2k17:
## Summary of Changes
Adds the ability to plug windows (livery specific).

## Screenshots
FlightSimulator_9G3Kv3QxOw copy

## Testing instructions
-Only the FBW house livery should have the first two rows plugged.
-Make sure 3rd party liveries are displayed correct (no plugs).
-Make sure the custom plugging works.


tracernz:
It would be better to add this in the blender project for the fuselage I think. Do you have a blend file?

Benjozork:
We should probably document this on our docs as well

TJCAviation:
I would reccomend making the Plug COMP the exact same as the Fuse COMP

With the Plug COMP being full red, it can cause some issues with the lighting/shadows as the base comp is slightly different around it...so the plugs tend to be a bit brighter then the fuse with in a shadow.

See my attatched image for what I mean
FlightSimulator 2024-02-16 02-05-21

tracernz:
We need to get the blend file and png textures for this so we can integrate into our new model build.

tracernz:
Can this texture be a 2x2, transparent one by default? I don't think it's worth a bunch of VRAM for the tiny number of liveries that will use it.

pilotseyea350:
This could be more relevant for the A380X, there isn't any A380 operator (current or former) that doesn't have some plugged windows

PR #8372, opened 2024-01-03
docs: Specification of A380X Public Flight Deck API

frankkopp:
## Summary of Changes
Specification and documentation for the future public Flight Deck API using mainly B-Events.

This is a working document to define and document the public API mainly for Flightsim Hardware and Cockpit Builder usage. 

## Testing instructions
N/A

### DIFF
--- a/fbw-a380x/docs/a380x-public-api-spec.md
+++ b/fbw-a380x/docs/a380x-public-api-spec.md
@@ -1,0 +1,1025 @@
+# A380X Public API Specification
+
+## Cockpit Controls API
+
+| Section                    | Panel                      | Name                  | Side | Var or Event | Remark |
+|----------------------------|----------------------------|-----------------------|------|--------------|--------|
+| Overhead Panel Maintenance | OXYGEN                     |                       |      |              |        |
+|                            | GND HYD                    |                       |      |              |        |
+|                            | FUEL                       |                       |      |              |        |
+|                            | GND HF DATALINK            |                       |      |              |        |
+|                            | CKPT DOOR LOCK SYS         |                       |      |              |        |
+|                            | BAT Check                  |                       |      |              |        |
+|                            | ENG                        |                       |      |              |        |
+|                            | MAINT                      |                       |      |              |        |
+|                            | ELEC                       |                       |      |              |        |
+|                            | NSS                        |                       |      |              |        |
+|                            | AIR                        |                       |      |              |        |
+|                            | VENT                       |                       |      |              |        |
+|                            |                            |                       |      |              |        |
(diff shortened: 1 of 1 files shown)


tracernz:
The table format will be much more difficult to maintain than the section format in our current simvar docs. It is much too wide to fit in a typical browser window.

frankkopp:
> The table format will be much more difficult to maintain than the section format in our current simvar docs. It is much too wide to fit in a typical browser window.

I wouldn't even call the current simvar docs a "format". It has none - is is an unstructured continuous unordered text mess. 

The table format allows for structured access and reformatting into any format if required. Also easily reused for docs where we can the apply any css we want. 

Separate structure from formatting - this allows us to reuse the data much more easily. E.g. put it into a database for a dynamic search or so. 

Also priority is not to make it easy to maintain (although it actually is very easy with a good IDE like Jetbrains that supports md tables) but reusibilty and easy to read for end users.

tracernz:
I'd make a couple of points on this:
- Sections of a doc that existed before a format was decided on are hardly a good counter-example
- A key feature of markdown is that it's made to be hand-written and read; wide tables defeat that. and
- The reasonable line length limit is somewhere around 120 chars; editor windows are often not wider than that
- The docs in the repo are for maintainers, not for end-users, and they need to be easy to work with for maintainers
- Jetbrains is not the recommended IDE for this project, and I don't think a proprietary software ever should be

frankkopp:
> I'd make a couple of points on this:
> 
> - Sections of a doc that existed before a format was decided on are hardly a good counter-example
> 
> - A key feature of markdown is that it's made to be hand-written and read; wide tables defeat that. and
> 
> - The reasonable line length limit is somewhere around 120 chars; editor windows are often not wider than that
> 
> - The docs in the repo are for maintainers, not for end-users, and they need to be easy to work with for maintainers
> 
> - Jetbrains is not the recommended IDE for this project, and I don't think a proprietary software ever should be recommended or optimised for

And it is still a data table and not formated text/table. If it wouldn't bring even more problems (to edit the data) I would have done this in Excel/Google Sheets or even CSV. It is about the structure and not the formatting. Formatting comes last. Using MD was just a middle ground as it is available within GitHub and most editors (btw. I did not say we have to use JetBrains although they do have free products which people could use - I said a good editor could handle ms tables - probably there is a plugin for VSC as well)

And I strongly disagree that I did this for maintainers only but mainly for the end users (cockpit builders, etc.)! The title allone clearly states "Public Flight Deck API" and not "FBW Internal Systems API". I hope you can see that it makes little sense that you try to convince me what my own intention was when creating this ;).

PR #8381, opened 2024-01-04
feat(fm/fbw): enable CFD simulation and adapted gyro stability

aguther:
## Summary of Changes
This PR enabled the CFD simulation of the modern flight model. It additionally tries to improve the impact of turbulence for a more realistic feeling in flight.

## Screenshots (if necessary)
n/a

## References
n/a

## Additional context
n/a

## Testing instructions
- fly the airplane in different scenarios and conditions
- check if Pitch angles and AoA values with different weights and conditions are still valid


aguther:
What I personally observed is a significant impact on the pitch when enabling the horizontal tail reinjection, but it seems to settle at the same values in the end. But that is still to be fully checked.

PR #8385, opened 2024-01-06
feat(flt_model): Adjust some 380 flight model geometry/contact point/max speed

crocket63:
Fixes #[issue_no]

## Summary of Changes

-Adjustment of geometry positions so the sim forces applies on correct positions according to 3D model
  -Wing forces are set down / wing thickness reduced to match 3D model
  -Horizontal surface further back
  -Vertical surface apply forces higher to the top of the rudder plane
-Moved contact points of main gears so they are at the middle of bogeys so tilt mechanism works correctly
-Nose steering limit set to 70° max as per 3D animation
-Max speed definition set to 390 to have wing overspeed oscillations at correctish speed.

## Screenshots (if necessary)

Before:
image
image
image

After
image
image
image

## Additional context

## Testing instructions

Check plane pitch still matches what we had before at different configurations / speeds
Check that main gears contact points a bit more forward do not cause abnormal pitch up of the plane on ground depending on CG


fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -606,7 +606,7 @@ cruise_speed = 487.5 ; Knots True (KTAS)
 cruise_mach = 0.85
 crossover_speed = 320 ; Knots Indicated (KIAS)
 max_mach = 0.89 ; MMO
```
lukecologne, line 608: The max design mach is 0.96, and max design speed is 375kts. So with 390 we have 15kts of buffer even above the Vd, which is probably reasonable, as IIRC, the sim immediately starts flutter above these speeds. We should also adjust max_mach to be above Md with a similar buffer, I think Vd+0.01 is reasonable, so I'd say max_mach=0.97

lukecologne (dismissed):

2hwk (approved):

Issue #8387, opened 2024-01-07
Wing Anti-Ice is Inop


donstim:
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-01-05T14:10:10+00:00",
    "ref": "refs/heads/master",
    "sha": "5843e8e6d5bb4ea0850d23378a01f313ccaa7569",
    "actor": "tracernz",
    "event_name": "manual",
    "pretty_release_name": "master:5843e8e6",
    "version": "v0.12.0-dev.5843e8e"
}
```

### Describe the bug

The wing anti-ice switch no longer works properly. It does not reduce the ice accreting on the wings in icing conditions.  See the following video:
https://1drv.ms/v/s!Ajsy7Ey4_8uFgZBj4J1YqtRD50yCJA?e=q8d4O7

At the beginning of the video, ice is accreting on the wing as can be seen by the increasing "Structural Ice PCT" simvar. After wing anti-ice is turned on, the "Structural Ice PCT" simvar continues to increase at the same rate. Although the "Structural Deice Switch"
simvar has changed on "1" (on), the activation of wing A/I has no effect on the rate of ice accretion.

This issue appears to be a bleed-related bug. Continuing with the video, when the APU bleed is opened with the wing A/I on, the "Structural Ice PCT" simvar starts decreasing. When the APU bleed is closed, it goes back to increasing. (Wing A/I must be activated, the APU must be on, and APU bleed open for the wing A/I to be operating properly.)

### Expected behavior

Accreted ice on the wings should be steadily reduced until it is eliminated when wing A/I is selected. 

### Steps to reproduce

1. Start at any reasonable altitude (I used 5000 feet)
2. Use the "Snow" weather preset in MSFS
3. Set a cloud layer that includes the altitude(s) you will be using
4. Set the cloud coverage to 100%
5. Use simvar watcher as shown in my video to observe the change in the "Structural Ice PCT" simvar to evaluate the effect of different selections on ice accretion.

### Additional info (optional)

I have an old A32NX PR build from 13 July 2023 where this worked properly.

### Discord Username (optional)

donbikes

MicahBCode:
This issue is fixed by #8388 isn't it?

donstim:
> This issue is fixed by #8388 isn't it?

No, that is an temporary disabling of icing effects until this bug is fixed.

tracernz:
The issue is that the crossbleed start PR disables the bleed air source when not starting engines.

[...]

PR #8388, opened 2024-01-07
fix: Interim anti ice fix

donstim:
Temporary interim "fix" for #8387 until the wing A/I system bug is fixed

## Summary of Changes
1. Sets the icing scalar in the flight model cfg file flight tuning section to 0 to prevent any effects of icing on airplane aerodynamics or weight. When the wing A/I system is fixed, this scalar should be set back to 1.
2. Increases the rate at which ice is cleared from the windscreen such that all ice will be cleared in 30 seconds. This change should remain in place after the wing A/I system is fixed.

## Screenshots (if necessary)
Video of windscreen ice clearing in action. (I hit the chrono timer button a bit late, so the windscreen clears before the chrono gets to 30 seconds.)
(https://1drv.ms/v/s!Ajsy7Ey4_8uFgZBkhTeKoCyp_F8KCg?e=jD1OGL)

## References
MSFS allows one rate of windshield clearing although the actual rate of clearing will depend on the conditions. Also, it is unlikely that only windshield heat will be used for deicing a fully iced windscreen. The rate chosen for this PR is based on the approximate time one of our pilots said it takes to defrost a frosted over windscreen. At the default MSFS rate of clearing, it takes nearly 8 minutes for the windscreen to clear.

This rate is quicker than the very slow default rate MSFS provides. From a sim standpoint, this should be helpful to sim pilots who have to wait until the windscreen heat clears a fully iced windscreen when starting cold and dark on the stand.

## Additional context

## Testing instructions
This really should not need testing. However, if testing is to be done, icing conditions can be set up by using the MSFS snow weather preset with 100% cloud coverage at the altitudes to be used for testing. The dev mode speed debug window can be used to verify that drag and weight do not increase with increasing ice. (Use the structural ice PCT simvar in simvar watcher to observe increasing ice.)

For the windscreen effects, the ice visibility slider under options in the dev mode dropdown can be used to ice up the windscreen as was done in the video above


frankkopp (approved):

Saschl:
Change is quite small and should not affect anything but icing. Merging without QA is fine from my side.

PR #8400, opened 2024-01-10
fix: remove upload of fragments to github

Saschl:
Installer uses cf and we upload the full package to github as an alternative for the installer.

## Summary of Changes

Removing the upload of fragments to github as I could not find/remember any usage currently

## Additional context


Benjozork (approved):

Saschl:
Merging, in case we still want GitHub as fallback, we can restore the changes and make the necessary adaptions.

PR #8407, opened 2024-01-12
feat(GSX): Automatically start Refuel Process & allow Instant Refuel with GSX Fuel Sync enabled

Fragtality:
Fixes #8406 

## Summary of Changes

- FBW Refuel Process is started automatically when GSX Fuel-Hose is connected
- Instant-Refuel is still possible with GSX Fuel Sync enabled
- EFB Fuel Page Reflects State of the GSX Refuel Service (and indicates more clearer that the User-Interaction is locked)

## Screenshots (if necessary)

GSX Fuel Sync enabled, Real (or Fast) Refuel is selected:
image

GSX Refuel was called (Truck on its way to Fuel-Hose Connected)
image

Refueling is active
image

Instant Refuel allowed while GSX Fuel Sync is enabled in the Settings
instant-refuel-with-sync-enabled

## Additional context

The Motivation is to remove manual Steps in the EFB when used in Conjunction with GSX. A User requesting Refuel in the GSX Menu already formulated/stated the Intention to Refuel the Plane. So an additional Confirmation is not really necessary.
On Top of that these Kind of Automation is a Preliminary to solve Issues like #7772.
With that Users using external GSX Service Automations (like included in my PilotsDeck/StreamDeck Profile for the FBW) would need less manual Steps in the EFB to enjoy these Automations.

I'm aware that this Change would also start the Refuel-Process even when the User did not make a single Interaction on the EFB (even not powered on). In that Case the FBW Default of 9524kg desired Fuel would be applied. But please consider:
- The User is still able to correct such a Situation with Instant-Refuel (because that is now possible with GSX Fuel Sync enabled)
- The User is still able to set the correct desired Fuel and then call GSX Refuel again, if so desired
- Looking into Auto-Importing the SimBrief Data to Fuel & Payload would be my next Target
- It is not exactly a "best Practice" Flight-Preparation to call Refuel straight away and then don't care about the required Fuel 😅 

Discord username (if different from GitHub): Fragtality

## Testing instructions

1. Load into the Plane
2. Ensure GSX Fuel Sync is enabled in the Settings>3rd Party
3. Do your normal Preparation Flow, but ensure you set/import your desired Fuel on the Ground>Fuel Page in the EFB.
4. Call Refuel in GSX Menu => No further manual Interaction should be needed for the Refuel Process to start. Also: Start Button should not be visible with GSX Fuel Sync on and Fast/Real selected.
[...]

PR #8330, opened 2023-12-16
refactor(efb/fuel): refueling system changes for future airframe systems

2hwk:
Fixes #[issue_no]

## Summary of Changes

- Optimisation
- More boring changes
- Draft v1 of new refueling system for a certain mysterious object

## Screenshots (if necessary)

rip

## Additional context

Discord username (if different from GitHub):

## Testing instructions

- Test that A32NX refueling flows work as per normal with no changes (only slight refactor that is cross-cutting in rust systems)
- IMPORTANT ! : Test that fuel is being consumed correctly by the engines and in flight! 

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8415, opened 2024-01-15
feat(EFB/SimBrief): Automatically import SimBrief Fuel & Payload

### LINKED ISSUE #7772: Auto-import simbrief data for GSX integration
### Aircraft Version

Experimental

### Description

Now having the gsx integration could you consider adding an option to automatically load simbrief payload data? Similarly to the one we have for dashboard page currently? As for the refuellig ideally would be to import simbrief fuel and start refueling once the hose connected (there is a gsx lvar). The idea is to minimize steps interacting with efb and automate the process. So you spawn at gate, turn on the efb, request boarding and refuelling via gsx, and everything happens on its own.

### References (optional)

_No response_

_No response_

K8#4795

Fragtality:
Fixes #7772
(Together with PR #8407)

## Summary of Changes

- When SimBrief Data was imported (either manual or automatic), the Fuel & Payload Pages automatically import their Data too. Approach:
  - The Fuel and Payload Page have an useEffect to import the Data when visited (only when SimBrief Data is loaded)
  - The Import Function navigates through the Pages (Dashboard->Fuel->Payload->Dashboard)
  - Two additional Flags in the SimBrief Store prevent that Payload or Fuel are imported every time the respective Page is revisited
- Auto-Import of SimBrief Data moved from EFB.tsx to FlightWidget.tsx (to not maintain two Copies of the same Code)
- Added Setting to enable/disable that Behavior

## Screenshots (if necessary)

flightwidget-toast

settings-atsu

## Additional context

Although the originating Issue is in the GSX Context, this Feature improves Comfort for everyone (using SimBrief). Only one Click needed - if at all - to get all relevant Data imported 😃 

Discord username (if different from GitHub): Fragtality

## Testing instructions

Test 1

1. Prepare a Flightplan on SimBrief for the FBW
2. Load into the Plane and enable the new Setting in the EFB (Settings > ATSU/AOC). Ensure that your SimBrief Account is set.
3. Navigate to the Dashboard and hit "Import SimBrief Data" => A Notification should appear (like shown in the Screenshots)
4. Navigate to the Fuel & Payload Page => Each Page should display the planned Fuel / Payload from your SimBrief FP

Test 2

1. If you don't use the (already existing) Auto-Import Setting: Please enable it
2. Poweroff the EFB
3. Power it on again => You should see the same Effects as before (Notification, Fuel & Payload Page have the planned Values)

For additional Verification you can monitor the L-Vars `A32NX_FUEL_DESIRED` and `A32NX_AIRFRAME_ZFW_DESIRED` => these should immediately Change when the SimBrief Data was imported (without leaving the Dashboard).
(If you have a Tool at Hand to monitor L-Vars quick & easy)

### DIFF
--- a/fbw-a32nx/src/localization/flypad/en.json
+++ b/fbw-a32nx/src/localization/flypad/en.json
@@ -55,6 +55,7 @@
         "RearrangeWidgets": "Rearrange Widgets"
       },
       "Title": "Important Information",
+      "ToastPayloadImported": "Fuel & Payload imported!",
       "Weather": {
         "AirPressure": "Air Pressure",
         "DewPoint": "Dew Point",
@@ -508,6 +509,7 @@
     "AtsuAoc": {
       "AtisAtcSource": "ATIS/ATC Source",
       "AutomaticallyImportSimBriefData": "Automatically Import SimBrief Data",
+      "ImportSimBriefPayload": "Set Fuel & Payload when SimBrief Data was imported",
       "Enable": "Enable",
       "EnableTelex": "Enable TELEX",
       "ErrorReporting": "Error Reporting",
--- a/fbw-a32nx/src/localization/flypad/en.json
+++ b/fbw-a32nx/src/localization/flypad/en.json
@@ -509,7 +509,7 @@
     "AtsuAoc": {
       "AtisAtcSource": "ATIS/ATC Source",
       "AutomaticallyImportSimBriefData": "Automatically Import SimBrief Data",
-      "ImportSimBriefWeights": "Automatically Import SimBrief Fuel & Payload when SimBrief Data was imported",
+      "ImportSimBriefWeights": "Automatically set SimBrief Fuel & Payload on Import",
       "Enable": "Enable",
       "EnableTelex": "Enable TELEX",
       "ErrorReporting": "Error Reporting",
--- a/fbw-a32nx/src/localization/flypad/en.json
+++ b/fbw-a32nx/src/localization/flypad/en.json
@@ -55,7 +55,6 @@
         "RearrangeWidgets": "Rearrange Widgets"
       },
       "Title": "Important Information",
-      "ToastPayloadImported": "Fuel & Payload imported!",
       "Weather": {
         "AirPressure": "Air Pressure",
         "DewPoint": "Dew Point",
@@ -82,6 +81,7 @@
       "CostIndex": "Cost Index",
       "CruiseAlt": "Cruise Alt.",
       "ImportSimBriefData": "Import SimBrief Data",
+      "ToastFuelPayloadImported": "Fuel & Payload imported!",
       "Route": "Route",
       "SimBriefDataNotYetLoaded": "SimBrief data not yet loaded.",
       "Title": "Your Flight",
@@ -509,7 +509,7 @@
     "AtsuAoc": {
       "AtisAtcSource": "ATIS/ATC Source",
       "AutomaticallyImportSimBriefData": "Automatically Import SimBrief Data",
-      "ImportSimBriefPayload": "Set Fuel & Payload when SimBrief Data was imported",
+      "ImportSimBriefWeights": "Automatically Import SimBrief Fuel & Payload when SimBrief Data was imported",
       "Enable": "Enable",
       "EnableTelex": "Enable TELEX",
       "ErrorReporting": "Error Reporting",
--- a/fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
@@ -141,7 +141,7 @@ export const FlightWidget = () => {
                 })
                 .then(() => {
                     history.push('/dashboard');
-                    toast.success(t('Dashboard.ImportantInformation.ToastPayloadImported'));
+                    toast.success(t('Dashboard.YourFlight.ToastFuelPayloadImported'));
                 });
         }
 
--- a/fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
 /* eslint-disable max-len */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { useHistory } from 'react-router-dom';
 import { IconPlane } from '@tabler/icons';
 import { CloudArrowDown } from 'react-bootstrap-icons';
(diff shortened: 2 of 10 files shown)


fbw-a32nx/src/localization/flypad/en.json
```diff
@@ -81,6 +81,7 @@
       "CostIndex": "Cost Index",
       "CruiseAlt": "Cruise Alt.",
       "ImportSimBriefData": "Import SimBrief Data",
+      "ToastFuelPayloadImported": "Fuel & Payload imported!",
```
2hwk, line 84:
```suggestion
      "ToastFuelPayloadImported": "Fuel & Payload imported!",
```
Have these strings been added to localezy? Currently this is a manual process, if you were not aware for future reference.

For Mav / Localezy Maintainers: If these keys do not already exist, please added to localezy for this to show up in a production build.

fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
```diff
@@ -141,72 +170,72 @@ export const FlightWidget = () => {
                     </h1>
                 )}
             </div>
-            <div className="overflow-hidden relative p-6 w-full rounded-lg border-2 h-content-section-reduced border-theme-accent">
-                <div className="flex flex-col justify-between h-full">
+            <div className="h-content-section-reduced border-theme-accent relative w-full overflow-hidden rounded-lg border-2 p-6">
+                <div className="flex h-full flex-col justify-between">
```
2hwk, line 174: May be better not to commit linting changes like this, maybe remove from the diff for this PR and at some point there should be a PR to clean this up.

fbw-a32nx/src/systems/instruments/src/EFB/Store/features/simBrief.ts
```diff
@@ -119,6 +119,8 @@ export const initialState: {data: SimbriefData} = {
         loadsheet: '',
         costInd: '',
     },
+    payloadImported: false,
+    fuelImported: false,
```
2hwk, line 123: In the longer term, we may have to consider a bit more whether managing the state like this is a good approach, or if it should be a simvar or some other form of persistence as if I'm not wrong, currently this would maintain the same behaviour that when the EFB is powered off, that all simbrief import is lost. 

Maybe not in this PR though.

fbw-a32nx/src/systems/instruments/src/EFB/Ground/Pages/Payload/A320_251N/A320Payload.tsx
```diff
@@ -138,6 +144,15 @@ export const A320Payload: React.FC<A320Props> = ({
...
 
+    const dispatch = useAppDispatch();
+
+    useEffect(() => {
+        if (simbriefDataLoaded === true && simbriefWeightsImport === 'ENABLED' && payloadImported === false) {
+            setSimBriefValues();
+            dispatch(setPayloadImported(true));
+        }
+    }, []);
+
```
2hwk, line 155: It is probably possible to have the simbrief components be common between both the A320 and A380, unless there is some blocking thing that I have forgotten about since it has been a while, in which case disregard. 

Is it possible to refactor this to reside in the common Payload.tsx element?

2hwk (commented):
See comments, not a hard requirement but I still need to scruntinise this more before approval.

fbw-a32nx/src/systems/instruments/src/EFB/Dashboard/Widgets/FlightWidget.tsx
Fragtality: Well, I can't stop that from happening unless I change the provided vscode Settings. Wasn't an Issue on PR 8407.
(Not that I would like going over Diffs on Areas that I have not changed^^)

fbw-a32nx/src/systems/instruments/src/EFB/Store/features/simBrief.ts
Fragtality: Yes, these two new States would the same Behavior! (One Reason I've added them there. The other being: I just don't know how/where else^^)
So it fits in the current State the User already knows. Power Off => SimBrief Data lost. Power On => It is imported again. I just "hooked on" to that. (The Fuel & Payload is also then reimported)

To be clear: I'm totally clueless. So I would say: definitely not in this PR respectively not by me 😅

fbw-a32nx/src/systems/instruments/src/EFB/Ground/Pages/Payload/A320_251N/A320Payload.tsx
Fragtality: Uhm ... I honestly don't know 😅 
I've gone by the Patterns/Stuff seen in the existing Code. And in none of the "Common/Base Page including AC specific Pages" Structures I've seen, an useEffect was used on the Common Page.

But I surely can test it later! ("What can possibly go wrong?" ^^)

Fragtality:
> See comments, not a hard requirement but I might still want to scruntinise this more before approval, if I have the time, if not I don't see anything obviously wrong

All good! Thanks 👍

fbw-a32nx/src/systems/instruments/src/EFB/Ground/Pages/Payload/A320_251N/A320Payload.tsx
Fragtality: Looked again at it: doing it on the Common Page has the same Problem which lead me to this Approach in the first Place! The Code/Logic to import SimBrief's Payload and Fuel are handled by Functions in the specific Pages (setSimBriefValues() and handleFuelAutoFill() ) which I need to call. Which call other Functions on that (specific) Page. And I just don't know how to call a Function of Page from somewhere else (and did not see any Pattern in that Direction).

Fragtality:
@2hwk 
How do we proceed on this Topic?

2hwk:
I've realised it might have a lot of implications on my fuel refactor PR https://github.com/flybywiresim/aircraft/pull/8330 but at this point I don't know when I will get back to it, so this should be merged first

Fragtality:
> I've realised it might have a lot of implications on my fuel refactor PR #8330 but at this point I don't know when I will get back to it, so this should be merged first

Thank you very much! 😃 

Just did a quick Look on PR 8330, but it would be basically the same Pattern as with the Payload Page (~copy & pasting the Code). Sure that it is not great, but without a Way to centralize Functions that are common (and these be callable by any Pages) I don't see a better Way currently.
But: my Knowledge is very limited, I just don't know how something like that could be implemented 😅

Maximilian-Reuter:
Maybe I am a bit late in the process, but I am wondering if we need the extra option in the EFB.

To me "Automatically import Simbrief Data" means all of the data from Simbrief which does include fuel and payload. 
It would also mean existing users wouldnt have to change anything and still experience the better experience.

Thoughts ?

Fragtality:
> I don't really see the usecase for only importing flightplan but not payload and fuel. If someone wants to change their payload or fuel planning from simbriefs values they can still do that. But for the large majority of users this simply removes two steps.
Thoughts ?

I just added that Option for the "default Behavior" not being changed. Keeping it "minimal" not "disturbing" much with my Changes.
Up to FBW to decide - I don't need or want that Option (Import should have always imported Fuel & Payload right away), so we can get rid of that additional Option!

Fragtality:
Untied Fuel & Payload import from autoSimbriefImport again - I'm dumb 🤣

fbw-a32nx/src/localization/flypad/en.json
frankkopp: > No was not added - I don't know how :/ Sorry!
> 
> On that Matter: would a changed Text also need to be added manually? Cause my former PR 8407 included something like that (Key: GSXFuelSyncEnabled).

Text changes when the key remains the same are best done in Localazy.
Key changes are best done with simply a new key. There is new feature to change even existing keys to not loose translations - ping me if required.

PR #8418, opened 2024-01-16
feat(A380X): Added systems-host and extras-host to A380X

frankkopp:
## Summary of Changes
Added the systems-host like in the A32NX to the A380X.
Added the extras-host to A380X.
Moved the IS_READY LVAR to the extras-host index.ts

Any modules are commented out as they would need to be migrated and adapted separately. 

## Testing instructions
n/a


fbw-a380x/src/systems/systems-host/build.js
```diff
@@ -0,0 +1,15 @@
...
+
+'use strict';
+
+const esbuild = require('esbuild');
+const path = require('path');
+const { createModuleBuild } = require('#build-utils');
+
+const outFile = 'fbw-a380x/out/flybywire-aircraft-a380-842/html_ui/Pages/VCockpit/Instruments/A380X/SystemsHost/index.js';
+
+// process.env.FBW_TYPECHECK = "1";
```
Benjozork, line 13: **style:** remove commented code

Benjozork (changes requested):

fbw-a380x/src/systems/systems-host/systems/powersupply.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface PowerSupplyBusTypes {
+    acBus1: boolean,
+    acBus2: boolean,
+    acBusEss: boolean,
+    dcBus1: boolean,
+    dcBus2: boolean,
+    dcBusEss: boolean,
+}
+
+export class PowerSupplyBusses {
```
tracernz, line 45: I don't think this class is needed. You can map the values in the simvar publisher simvar definitions. What this class ends up doing is forces all the Lvars to be read every frame even when nobody is actually subbed to the topics (the SimVarPublisher will normally not start reading anything that hasn't actually been subscribed). So, I'd suggest moving the map function (powered !== 0) to the ` PowerSupplySimvarPublisher` definitions, and plugging it into the InstrumentBackplane for the systems host instrument so it automatically gets init, startPublish, and onUpdate called.

frankkopp: That's fine - I just brought it over from the A32NX and commented it out. 
I'll look into it but I did not want to touch the systems stuff itself in this PR - just provide the systems-host so that everybody can use it.

tracernz: Oh, right, fair enough I guess.

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/panel/panel.cfg
```diff
@@ -151,5 +151,11 @@ htmlgauge00=WasmInstrument/WasmInstrument.html?wasm_module=systems.wasm&wasm_gau
...
-# Uncomment when A380 ExtrasHost exists in monorepo
-#htmlgauge04=A380X/ExtrasHost/extras-host.html,0,0,1,1
+
+[VCockpit21]
+size_mm = 0,0
+pixel_size = 0,0
+texture = NO_TEXTURE
+background_color = 0,0,0
+
+htmlgauge00 = A380X/SystemsHost/systems-host.html,0,0,1,1
```
tracernz, line 161: Add to the existing headless "systems" VCockpit above.

Benjozork (approved):

PR #8454, opened 2024-01-29
fix(flight model / bleed air / ice prot): Restore wing anti-ice functionality and simplify bleed air

### LINKED ISSUE #8387: Wing Anti-Ice is Inop

ngreatorex:
Fixes #8387

## Summary of Changes

1. Reverts the temporary fix from #8388
1. Removes the `systems_wasm` `engines` aspect and sets bleed air source to auto.
1. Introduces 2 new LVars for engine anti ice push button states.
1. Modifies the `WING ANTI ICE` push button to control their local variables rather than the MSFS aircraft variable directly.
1. Modifies the `ENGINE ANTI ICE` push buttons to control the new local variables rather than the MSFS aircraft variable directly.
1. Modifies the rate that the fuselage deices.
1. Adds an `anti_ice` aspect that monitors the push button LVars and toggles the aircraft variables appropriately.

## Screenshots (if necessary)

N/A

## References

N/A

## Additional context

I have tested this with GSX's de-icing. GSX will toggle on all aircraft anti-ice/de-ice systems in MSFS itself, but will not affect the state of the systems simulation or the overhead push buttons. This means that the next time you turn on the anti ice systems using the overhead push buttons, it won't actually do anything to the MSFS aircraft variables. However, they will then be back in sync, and turning the push buttons off will turn off the MSFS deice.

Discord username (if different from GitHub): fatboyfat_uk

## Testing instructions

This PR makes changes to the bleed air systems as well as re-enabling the anti-ice. The tests below will check that cross-bleed starts still work while testing the anti ice:

1. Spawn somewhere with icing conditions. I've found the easiest way to guarantee this on the ground is to set the custom weather to Snow, move the lowest cloud layer to encompass ground level, then set scatter and density to maximum.
1. Observe the `STRUCTURAL ICE PCT` SimVar somehow - best way is SimvarWatcher or similar. Alternatively, Flow's weather widget shows icing percentage, but it will only drop slowly.
1. Ensure that the variable is slowly increasing i.e. you definitely have icing conditions.
1. Start Engine 1 using APU or ASU.
1. Remove APU or ASU.
1. Perform a cross-bleed start of Engine 2 (following the docs at https://docs.flybywiresim.com/pilots-corner/advanced-guides/engines/crossbleed-start/)
1. Turn on `WING ANTI ICE` with the push button on the overhead panel.
1. While the aircraft is on the ground, this will only work for 30 seconds while the aircraft runs a test. During this time, check that the `STRUCTURAL ICE PCT` SimVar decreases. Once the test is completed, it should start to increase again.
1. When in the air, turn `WING ANTI ICE` back on and check that the `STRUCTURAL ICE PCT` SimVar (eventually) reduces to 0.


tracernz:
I don’t think GSX should be able to turn a button on the overhead on or off. We only simulate actual aircraft systems.

alepouna:
> I don’t think GSX should be able to turn a button on the overhead on or off. We only simulate actual aircraft systems.

GSX is turning our EXT POWER switch on when calling the GPU and turning it off when dismissing it. No idea if this is limited to FBW only.

ngreatorex:
> I don’t think GSX should be able to turn a button on the overhead on or off. We only simulate actual aircraft systems. 
> 
> 
> 
> If GSX is using a hack for their de-icing system that’s too bad. 

The problem is that if we don't do that the status of the overhead light doesn't match what MSFS is doing. 

Before this change, GSX was already toggling the button on the overhead as the button was directly tied to the SimVar. The code still to be added here is only to keep that functionality as it was after the button is switched to controlling the LVar instead.

tracernz:
The thing is GSX is not actually trying to use the aircraft wing anti-ice, but instead simulate the effect of the de-icing fluid that's physically on the wings, isn't it? In that case it doesn't really make sense for the aircraft wing anti-ice system to turn on. It also should not be manipulating the EXT PWR switch.

Maximilian-Reuter:
In my opinion the engines.rs part should be able to be removed. If the msfs doesnt switch to the engines as bleed source automagically then we need to investigate why since that should work as far as I am concerned.

BlueberryKing:
> In my opinion the engines.rs part should be able to be removed. If the msfs doesnt switch to the engines as bleed source automagically then we need to investigate why since that should work as far as I am concerned.

No, that's probably needed. It only switches to the engines as bleed source if they are toggled on to be a bleed source, which is what the aspect does.

Maximilian-Reuter:
> > In my opinion the engines.rs part should be able to be removed. If the msfs doesnt switch to the engines as bleed source automagically then we need to investigate why since that should work as far as I am concerned.
> 
> No, that's probably needed. It only switches to the engines as bleed source if they are toggled on to be a bleed source, which is what the aspect does.

but we talked about this, to my knowledge there is nothing in our systems that cares about what msfs thinks about where bleed air comes from. I just looked at the APU aspect code and thought the same thing about the APU bleed source toggle. 

The fadec controlls whether engines start based on our pneumatics system so it couldnt care less what msfs thinks it is. To me removing all the bleed toggle code and setting the source selection to auto in the flt files should be all that is neccessary.

BlueberryKing:
> > > In my opinion the engines.rs part should be able to be removed. If the msfs doesnt switch to the engines as bleed source automagically then we need to investigate why since that should work as far as I am concerned.
> > 
> > 
> > No, that's probably needed. It only switches to the engines as bleed source if they are toggled on to be a bleed source, which is what the aspect does.
> 
> but we talked about this, to my knowledge there is nothing in our systems that cares about what msfs thinks about where bleed air comes from. I just looked at the APU aspect code and thought the same thing about the APU bleed source toggle.
> 
> The fadec controls whether engines start based on our pneumatics system so it couldnt care less what msfs thinks it is. To me removing all the bleed toggle code and setting the source selection to auto in the flt files should be all that is neccessary.

Yes, you can also enable them as bleed source from the flt files of course, but you need to enable them somewhere.

Maximilian-Reuter:
> > > > In my opinion the engines.rs part should be able to be removed. If the msfs doesnt switch to the engines as bleed source automagically then we need to investigate why since that should work as far as I am concerned.
> > > 
> > > 
> > > No, that's probably needed. It only switches to the engines as bleed source if they are toggled on to be a bleed source, which is what the aspect does.
> > 
> > 
> > but we talked about this, to my knowledge there is nothing in our systems that cares about what msfs thinks about where bleed air comes from. I just looked at the APU aspect code and thought the same thing about the APU bleed source toggle.
> > The fadec controls whether engines start based on our pneumatics system so it couldnt care less what msfs thinks it is. To me removing all the bleed toggle code and setting the source selection to auto in the flt files should be all that is neccessary.
> 
> Yes, you can also enable them as bleed source from the flt files of course, but you need to enable them somewhere.

I am a big fan of removing code if we can acomplish it otherwise. 
image
To my knowledge we can remove all the entries in this section and just have the BleedAirSource set to 0 for AUTO and it should just have bleed air available as long as engines or apu is running.

Whether the actual systems work is smth we define ourselves in our individual systems code. This just makes sure that the effect is there for msfs when we want it to be.

ngreatorex:
> I am a big fan of removing code if we can acomplish it otherwise. 
> To my knowledge we can remove all the entries in this section and just have the BleedAirSource set to 0 for AUTO and it should just have bleed air available as long as engines or apu is running.
> 
> Whether the actual systems work is smth we define ourselves in our individual systems code. This just makes sure that the effect is there for msfs when we want it to be.

I've taken out `engines.rs` locally and set `BleedAirSource` to 0 in the *.flt files. The only issue is that now the WING ANTI ICE works without bleed air LOL. It seems the PNEU_WING_ANTI_ICE_SYSTEM_ON does not check there's pressure, even though the comment in https://github.com/flybywiresim/aircraft/blob/b4c2044e8bf3e13ec90fd5711d520f82e65526e6/fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic/wing_anti_ice.rs#L247 says "we need to check" the pressure. Guess I need to change how I detect the anti ice is actually on...

Also, I was wondering if there's an existing way to set an LVar from within an `on_change` closure in an aspect? The problem I'm having is that if I use something like:
```rust
let variable = Variable::named("BUTTON_OVHD_ANTI_ICE_WING_POSITION");
let value: VariableValue = (&variable).into();
value.write(...);
```
then the variable doesn't get the prefix added (because it doesn't get run through the `MsfsVariableRegistry`. 

I could obviously add this functionality (perhaps by adding something like an `on_change_write_to_variable` function alongside `on_change`) but wanted to check if I was missing something first.

Maximilian-Reuter:
> > I am a big fan of removing code if we can acomplish it otherwise.
> > To my knowledge we can remove all the entries in this section and just have the BleedAirSource set to 0 for AUTO and it should just have bleed air available as long as engines or apu is running.
> > Whether the actual systems work is smth we define ourselves in our individual systems code. This just makes sure that the effect is there for msfs when we want it to be.
> 
> I've taken out `engines.rs` locally and set `BleedAirSource` to 0 in the *.flt files. The only issue is that now the WING ANTI ICE works without bleed air LOL. It seems the PNEU_WING_ANTI_ICE_SYSTEM_ON does not check there's pressure, even though the comment in
> 
> https://github.com/flybywiresim/aircraft/blob/b4c2044e8bf3e13ec90fd5711d520f82e65526e6/fbw-a32nx/src/wasm/systems/a320_systems/src/pneumatic/wing_anti_ice.rs#L247
> 
> says "we need to check" the pressure. Guess I need to change how I detect the anti ice is actually on...
> Also, I was wondering if there's an existing way to set an LVar from within an `on_change` closure in an aspect? The problem I'm having is that if I use something like:
> 
> ```rust
> let variable = Variable::named("BUTTON_OVHD_ANTI_ICE_WING_POSITION");
> let value: VariableValue = (&variable).into();
> value.write(...);
> ```
> 
> then the variable doesn't get the prefix added (because it doesn't get run through the `MsfsVariableRegistry`.
> 
> I could obviously add this functionality (perhaps by adding something like an `on_change_write_to_variable` function alongside `on_change`) but wanted to check if I was missing something first.

hmm, I feel like you should get there if you also reference the PNEU_WING_ANTI_ICE_HAS_FAULT LVAR since that should come up if the system is supposed to be on but doesnt work for example for low pressure.

As to your second question, what exactly are you trying to do with that variable write ? Like what and why are you trying to write there ?

ngreatorex:
> hmm, I feel like you should get there if you also reference the PNEU_WING_ANTI_ICE_HAS_FAULT LVAR since that should come up if the system is supposed to be on but doesnt work for example for low pressure.
> 

I'll try that tomorrow, thanks for the suggestion. 

> 
> 
> As to your second question, what exactly are you trying to do with that variable write ? Like what and why are you trying to write there ?

Well initially I was doing it for the GSX thing, so was wanting to write to the anti ice push button LVar. But if the consensus is that GSX shouldn't illuminate the overhead switch it's now just curiosity and trying to understand Rust a bit better!

(On that note GSX deicing also illuminates the engine anti ice push buttons so will probably take a look at decoupling that too)

Maximilian-Reuter:
> Well initially I was doing it for the GSX thing, so was wanting to write to the anti ice push button LVar. But if the consensus is that GSX shouldn't illuminate the overhead switch it's now just curiosity and trying to understand Rust a bit better!
> 
> (On that note GSX deicing also illuminates the engine anti ice push buttons so will probably take a look at decoupling that too)

To me the only workable solution to support GSX deicing is to monitor the state of the STRUCTURAL DEICE SWITCH Simvar and basically allow it to be on on the ground. Because if we arent setting it on the ground past the 30s selftest then it will in all likelyhood only be triggered by GSX. Since deicing is usually only relevant for ground operations I think after takeoff we can then set the deicing according to our systems code. 

Least that is how I can imagine it working out.

tracernz:
Deicing fluid doesn't stop being effective immediately after the wheels leave the ground. When does GSX turn off the de-icing? I think we can just intercept the events from them with KeyEventManager's passthrough disabled mechanism and then control the sim according to both them and our systems again by bypassing key intercepts with KeyEventManager.

Maximilian-Reuter:
> Deicing fluid doesn't stop being effective immediately after the wheels leave the ground. When does GSX turn off the de-icing? I think we can just intercept the events from them with KeyEventManager's passthrough disabled mechanism and then control the sim according to both them and our systems again by bypassing key intercepts with KeyEventManager.

I am not sure GSX does anything past turning them on, hence why we need a time where we take over

ngreatorex:
I've just pushed my latest changes. To summarise:

1. The overhead anti-ice push buttons now control LVars instead of SimVars (both wing and engine).
2. The new `anti_ice` aspect handles any changes to the LVars, updating the SimVars appropriately.
3. If something external (e.g. GSX) modifies the SimVars, the systems simulation does not react and the overhead push buttons do not light.
4. The next time you do toggle the push buttons, the LVars will be back in sync with the SimVars so you can continue as normal.

What's left? Tests, tests and tests 🙂

PS. I added the lines to add the aspect to the A380 but obviously without the model / behaviours I cannot test. Let me know if I should take those lines out.

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
```diff
@@ -0,0 +1,77 @@
...
+                vec![
+                    Variable::named(&format!(
+                        "BUTTON_OVHD_ANTI_ICE_ENG_{}_POSITION",
+                        engine_number
+                    )),
+                    Variable::aircraft("ENG ANTI ICE", "bool", engine_number),
+                ],
+                Box::new(move |prev_values, new_values| {
+                    let was_eng_anti_ice_push_button_on = to_bool(prev_values[0]);
+                    let is_eng_anti_ice_push_button_on_now = to_bool(new_values[0]);
```
Maximilian-Reuter, line 26:
```suggestion
                    let is_eng_anti_ice_push_button_on = to_bool(new_values[0]);
```
now isnt needed since it is alread `is`

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
```diff
@@ -0,0 +1,77 @@
...
+                    let is_eng_anti_ice_on_now = to_bool(new_values[1]);
```
Maximilian-Reuter, line 27:
```suggestion
                    let is_eng_anti_ice_on = to_bool(new_values[1]);
```
same as above

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
```diff
@@ -0,0 +1,77 @@
...
+        builder.on_change(
+            ExecuteOn::PostTick,
+            vec![
+                Variable::named("PNEU_WING_ANTI_ICE_SYSTEM_ON"),
+                Variable::named("PNEU_WING_ANTI_ICE_HAS_FAULT"),
+                Variable::aircraft("STRUCTURAL DEICE SWITCH", "Bool", 0),
+            ],
+            Box::new(move |prev_values, new_values| {
+                let was_wing_anti_ice_on = to_bool(prev_values[0]) && !to_bool(prev_values[1]);
+                let is_wing_anti_ice_on_now = to_bool(new_values[0]) && !to_bool(new_values[1]);
```
Maximilian-Reuter, line 63:
```suggestion
                let is_wing_anti_ice_on = to_bool(new_values[0]) && !to_bool(new_values[1]);
```
same as above

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
```diff
@@ -0,0 +1,77 @@
...
+                let is_now_deicing = to_bool(new_values[2]);
```
Maximilian-Reuter, line 64:
```suggestion
                let is_deicing = to_bool(new_values[2]);
```
same as above

Maximilian-Reuter (changes requested):
See suggestions on less verbose varaible naming. Take care to change all occurences of those variables.

Additionally there are some (pre existing) bugs associated with GSX deicing. Currently our FADEC and our FWS checks the MSFS ENG anti ice simvars to determine if ENG anti ice is in use. Since GSX sets those, this leads to the ENG A. ICE memo being shown erroneously.
Thrustlimits also get calculated incorrectly though since you would usually have it on anyways in icing conditions I dont think that part is really an issue.

Specifically https://github.com/flybywiresim/aircraft/blob/6f3af7b4b7b5013d03dc5e14b638c466d4e4ac59/fbw-a32nx/src/wasm/fadec_a320/src/SimVars.h#L104
and
https://github.com/flybywiresim/aircraft/blob/6f3af7b4b7b5013d03dc5e14b638c466d4e4ac59/fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts#L992-L993

In order to better solve this ideally one would implement a rudimentary eng anti ice system in rust that can then represent the actual state of the system. Since right now we only have the MSFS simvar or our button LVARs both can be misleading.

Though I am not sure this needs to be solved in this PR and could be handled by a later GSX specific deicing PR that could also include some rudimentary holdover simulation.

ngreatorex:
Thanks for the review!

> See suggestions on less verbose varaible naming. Take care to change all occurences of those variables.

Done (although not tested in the sim yet - will do that over the weekend).

> Additionally there are some (pre existing) bugs associated with GSX deicing. Currently our FADEC and our FWS checks the MSFS ENG anti ice simvars to determine if ENG anti ice is in use. Since GSX sets those, this leads to the ENG A. ICE memo being shown erroneously. Thrustlimits also get calculated incorrectly though since you would usually have it on anyways in icing conditions I dont think that part is really an issue.
> 
> Specifically
> 
> https://github.com/flybywiresim/aircraft/blob/6f3af7b4b7b5013d03dc5e14b638c466d4e4ac59/fbw-a32nx/src/wasm/fadec_a320/src/SimVars.h#L104
> 
> 
> and
> https://github.com/flybywiresim/aircraft/blob/6f3af7b4b7b5013d03dc5e14b638c466d4e4ac59/fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts#L992-L993
> 
> In order to better solve this ideally one would implement a rudimentary eng anti ice system in rust that can then represent the actual state of the system. Since right now we only have the MSFS simvar or our button LVARs both can be misleading.
> 
> Though I am not sure this needs to be solved in this PR and could be handled by a later GSX specific deicing PR that could also include some rudimentary holdover simulation.

I would suggest the rust engine anti ice sim is better as a later PR. However, I can certainly update the FWS and FADEC to reference the push button LVar rather than the SimVar. That way they will work off the overhead push button at least rather than the SimVar. I will again look at this over the weekend.

Maximilian-Reuter:
> I would suggest the rust engine anti ice sim is better as a later PR. However, I can certainly update the FWS and FADEC to reference the push button LVar rather than the SimVar. That way they will work off the overhead push button at least rather than the SimVar. I will again look at this over the weekend.

Yeah that doesnt really make sense, cause then you will have the same issue with erroneous ECAM memos. The only real solution is adding a seperate LVAR that is set when the ENG anti ice actually is running correctly.

But that can be a later PR since the main scope here is the anti ice system working again.

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
ngreatorex: Done

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
ngreatorex: Done

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
ngreatorex: Done

fbw-common/src/wasm/systems/systems_wasm/src/anti_ice.rs
ngreatorex: Done

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/systems.cfg
```diff
@@ -517,6 +517,7 @@ stick_shaker = 1
 
 [DEICE_SYSTEM]
 structural_deice_type = 1 ; 0 = None, 1 = Heated Leading Edge, 2 = Bleed Air Boots, 3 = Eng Pump Boots
+structural_deice_rate = 0.01666 ; Pct reduction of ice per second -- Clears airframe of ice in 60 seconds
```
ngreatorex, line 520: Does anyone have an opinion on this? The MSFS default is 0.00128 %/sec, which seems incredibly slow. Of course I don't think this has a real world equivalent - as any deice would be done by external vehicles rather than the anti-ice systems.

ngreatorex:
I've made the suggested changes and marked this as ready for review. I've not resolved the comments - not sure the etiquette here as to whether the reviewer should do it or the author 😃 

I've also added a comment about the deice rate - I'd appreciate feedback on whether my suggested speed up is a good idea.

Maximilian-Reuter (approved):
LGTM
Fine with the rate change and can always be tuned later. Good work!

fbw-a32nx/docs/a320-simvars.md
```diff
@@ -3049,6 +3049,14 @@ In the variables below, {number} should be replaced with one item in the set: {
...
     - Bool
 
+- A32NX_BUTTON_OVHD_ANTI_ICE_ENG_1_POSITION:
+    - Indicated whether the overhead ENG ANTI ICE button is illuminated for engine 1
+    - Bool
+
+- A32NX_BUTTON_OVHD_ANTI_ICE_ENG_2_POSITION:
+    - Indicated whether the overhead ENG ANTI ICE button is illuminated for engine 2
+    - Bool
+
```
Benjozork, line 3059:
```suggestion
- A32NX_BUTTON_OVHD_ANTI_ICE_ENG_1_POSITION:
    - Indicates whether the overhead ENG ANTI ICE button is illuminated for engine 1
    - Bool

- A32NX_BUTTON_OVHD_ANTI_ICE_ENG_2_POSITION:
    - Indicates whether the overhead ENG ANTI ICE button is illuminated for engine 2
    - Bool

```

ngreatorex: Good spot 😄

frenchairbuspilot:
QA Trainee Report

Discord          : jp320#3336
Object of testing: #8454 
Tier of Testing  : 1
Date             : 04/02/2024

**Testing Process:** 
_Spawn somewhere with icing conditions. I've found the easiest way to guarantee this on the ground is to set the custom weather to Snow, move the lowest cloud layer to encompass ground level, then set scatter and density to maximum._
**Done in BGSF as explained.**

_Observe the STRUCTURAL ICE PCT SimVar somehow - best way is SimvarWatcher, Ensure that the variable is slowly increasing i.e. you definitely have icing conditions._
**Check ice build up on simwar watcher**

_Start Engine 1 using APU or ASU._
_Remove APU or ASU._
_Perform a cross-bleed start of Engine 2_
_Turn on WING ANTI ICE with the push button on the overhead panel._
_While the aircraft is on the ground, this will only work for 30 seconds while the aircraft runs a test. During this time, check that the STRUCTURAL ICE PCT SimVar decreases. Once the test is completed, it should start to increase again._
**All checked and behaves as expected.**

_When in the air, turn WING ANTI ICE back on and check that the STRUCTURAL ICE PCT SimVar (eventually) reduces to 0._
**again behaves as expected, de icing rate might be kind of abrupt but i never really experienced real moderate to severe icing IRL so i can't really give more feedback about this. usually it takes about 10 seconds to see chips of ice shearing of the slats.**

Negatives:
appart from the anti icing rate, which is debatable, no remarks.

Testing Results: 
Passed

Conclusions: 
Good job =)

Media:
I have a MKV file which is too heavy so.. 10 days for the video here:
[https://we.tl/t-yjoSFhFQDo](url)

PR #8457, opened 2024-01-30
fix(TCAS/ND): TCAS targets disappearing

### LINKED ISSUE #8219: TCAS targets disappearing under different map ranges
### Aircraft Version

Development

### Build info

```json
{
    "built": "2023-09-16T06:20:39+00:00",
    "ref": "refs/heads/master",
    "sha": "4783ebcadace8462bc71af4f8c1bb4c1ebdb322c",
    "actor": "2hwk",
    "event_name": "manual",
    "pretty_release_name": "master:4783ebca",
    "version": "v0.11.0-dev.4783ebc"
}
```

### Describe the bug

While flying in busy airspace, I noticed that the ND was rather empty. When I reduced the range, some TCAS targets appeared that were not there. When I increased map range they disappeared again.

### Expected behavior

The top 8 targets should appear in the ND regardless of selected range. They should not disappear with higher ranges.

### Steps to reproduce

1. Fly in busy airspace
2. Increase/decrease the ND range and notice the TCAS targets

This was noted in VATSIM

Bernardo R

ArtemisPlayer:
Fixes #8219

## Summary of Changes
The mask for displaying TCAS targets on the ND was misplaced and that made some targets to disappear/not being displayed. This especially happened when changing ND range.

## Screenshots (if necessary)

Plot of ARC mask (using blue diamonds) on the ND:
image

Full mask on a python plot:
image

ROSE mask on a python plot:
image

Corrected mask:
image

I aligned the ARC mask on the ROSE one.

## Additional context

Discord username : ArtemisPlayer

## Testing instructions
Fly in a busy airspace, such that there is a plane at roughly the same altitude right ahead (about 10 NM), set the ND to 20 NM and ARC mode, and then increase the range: the target should no longer disappear.

### DIFF
--- a/fbw-common/src/systems/shared/src/parseMetar.tsx
+++ b/fbw-common/src/systems/shared/src/parseMetar.tsx
@@ -26,6 +26,7 @@ export function parseMetar(metarString: string): MetarParserType {
     const metarArray = metarString
         .trim()
         .replace(/^METAR\S*?\s/, '')
+        .replace(/=$/, '')
         // convert visibility range like `1 1/2 SM`
         .replace(/(\s)(\d)\s(\d)\/(\d)(SM)/, (all, a, b, c, d, e) => `${a + (Number(b) * Number(d) + Number(c))}/${d}${e}`)
         .split(' ');


ArtemisPlayer:
And a screenshot showing a target in a previously "dead" zone:
image

2hwk (approved):

2hwk:
@tracernz You might want to have a look as well

tracernz:
This should just need shifting to suit NDv2 rather than any changes to the mask itself. Is that what was done here, or have the points been shifted relative to each other?

ArtemisPlayer:
> This should just need shifting to suit NDv2 rather than any changes to the mask itself. Is that what was done here, or have the points been shifted relative to each other?

Yes that's what has been done. You can see it on the pictures (matplotlib plots)

tracernz:
> Yes that's what has been done. You can see it on the pictures (matplotlib plots)

Okay, that is not clear from your PR description. What I wanted to know is that you haven't modified the size/shape at all, but merely offset all the existing points by the same amount. We were already happy with the mask as it was in NDv1 from our refs.

ArtemisPlayer:
> Okay, that is not clear from your PR description. What I wanted to know is that you haven't modified the size/shape at all, but merely offset all the existing points by the same amount to suit whatever has been moved in NDv2. We were already happy with the mask as it was in NDv1 from our refs. Matplotlib plots are cool, but don't tell me anything in terms of accuracy against refs 😛.

Yes it is only an offset. I don't think that the UI has been so significantly changed, the position of the mask was just wrong, as you can see on my first screenshot (I displayed the mask's points using small blue diamonds, the bottom of the mask can be seen right il the middle of the screen. On the plot below, the Y axis has been flipped, but the shape can be recognised) It is not a slight offset, but quite a move : almost halt the length of the screen along the Y axis. That led to some issues as described in #8219.

PR #8477, opened 2024-02-09
feat(efcs): restore original flare law with optional new one

aguther:
## Summary of Changes
This PR restores by default the original flare law for manual flight.

The new one is still available and can be activated in `ModelConfiguration.ini`:
```
[model]
tailstrike_protection_enabled = true
```

This option was used because the tailstrike protection on rotation is disabled by default and there are no known uses anyway. Therefore, tailstrike protection is always off now.

## References
Various feedback on Discord indicates, that the new law might be correct implemented based on paper work, but doesn't feel like the real airplane. Therefore, the old one is used for the time being with the new one available so further tuning can be done in parallel.

## Testing instructions
- do various manual landings
- do a takeoff to ensure no impact on rotation mode happened


2hwk (approved):

aguther:
Should be no longer needed due to the bugfix PR (https://github.com/flybywiresim/aircraft/pull/8480)

PR #8479, opened 2024-02-10
fix(flt): Update 380 mass distribution and ground weights

crocket63:
Fixes #[issue_no]

## Summary of Changes

Correct mass of cargo and fuel tanks Z position so it matches model.
Adjust wing gear contact point to allow weight shift on nose gear

## Additional context


fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
Benjozork, line 1: There's a changelog at the top of file, we should probably update it

2hwk (approved):

PR #8501, opened 2024-02-16
fix(fbw): delay init of fbw wasm to align with systems

lukecologne:
## Summary of Changes

This PR implements a delay before starting to update the fbw wasm:
* Delay until `A32NX_IS_READY` is true
* Delay until `SIMULATION TIME` is greater than 2s.

This is done to be more in line with the sytems wasm, which implements the same delay.
Previously, the fbw wasm started updating earlier than the systems wasm (particularly in air starts, as `SIMULATION TIME` does not update there until the "Ready to Fly" button is pressed), which resulted in some data generated by the systems, particularly electrical supplies, not being properly initialized when the fbw wasm starts updating. Now, the fbw wasm starts in sync with the systems wasm. This fixes two issues:
1. The FACs now are immediately powered when the fbw wasm starts to update, which gets rid of the FAC fault when starting in the air
2. The spoilers 3 and 4 now start properly retracted when spawning on the runway.

## Additional context

### DIFF
--- a/fbw-a32nx/src/wasm/fbw_a320/src/FlyByWireInterface.cpp
+++ b/fbw-a32nx/src/wasm/fbw_a320/src/FlyByWireInterface.cpp
@@ -74,6 +74,9 @@ bool FlyByWireInterface::update(double sampleTime) {
   // get data & inputs
   result &= readDataAndLocalVariables(sampleTime);
 
+  // get sim data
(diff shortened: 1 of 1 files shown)


2hwk (approved):

aguther (approved):

aguther:
I tested the PR and it works as expected.

PR #8511, opened 2024-02-18
feat(press): pressurization failures

mjuhe:
## Summary of Changes

This PR introduces failures for the Pressurization System. All variables have been converted to ARINC type, and the deprecated `A32NX_PRESS_AUTO_LANDING_ELEVATION` LVar has been removed and replaced by the correct ARINC value coming from the CPCs.

Failable items:
   - Cabin Pressure Controller 1 and 2
   - Outflow Valve
   - Safety Valve

Other failures:
   - Rapid decompression

ECAM warnings added:
   - CAB PR SYS 1 FAULT
   - CAB PR SYS 2 FAULT
   - CAB PR SYS 1 + 2 FAULT
   - CAB BR LO DIFF PR
   - CAB PR OFV NOT OPEN
   - CAB PR SAFETY VALVE OPEN
   - CAB PR EXCES RESIDUAL PR

## Screenshots (if necessary)

Screenshot 2024-02-01 223158

## References

Amongst others, this video, which is cool: https://youtu.be/iORXooxhYN8?si=nJb2QE0xOitw50NR

## Additional context

A couple of small a380 pressurization fixes will be required. These will come in a separate PR.

Discord username (if different from GitHub): Commander Ga

## Testing instructions

What to look for in each of the failures:
   - CPC 1 or 2: The corresponding cabin pressure controller fails, the system transfers to the other CPC if it was active. You should get an amber warning on the ECAM "SYS 1(2) FAULT", but no aural sound. The press page shows the failed system in amber.
   - CPC 1 and 2: Both cabin pressure controllers fail. The automatic pressurization system is no longer working. The outflow valve freezes in position and is not responsive. You should get a "SYS 1+2 FAULT" warning on the ECAM. 
   - Outflow valve: the automatic motor of the outflow valve fails. The outflow valve is unresponsive in automatic operation (it will still move under manual mode). No associated warning. To get the "OFV NOT OPEN" warning, follow these steps:
          1. Fail the OFV as you are approaching to land, so the OFV is "frozen" in position
          2. Ensure the pressurization system is in auto mode. If it's in manual the Residual Pressure Controller will kick in after 55 seconds and open the outflow valve automatically.
          3. After landing, wait 70 seconds
   - Safety valve: the safety valve fails open. If the aircraft is pressurized, you might see a slow loss of pressure. An amber warning "SAFETY VALVE OPEN" should appear on the ECAM after a 60s delay, or instantly on the ground.
   - Rapid decompression: we simulate "structural damage" so the aircraft loses pressure at a fast rate. After a few seconds you should get the red "EXCESS CAB ALT" warning. If you perform an emergency descent, you should eventually also get the "LOW DIFF PR" amber warning when the pressure inside the aircraft equalises with the outside.


Benjozork:
cc @frankkopp lvar changes

fbw-a32nx/src/systems/instruments/src/SD/Pages/Crz/Crz.tsx
```diff
@@ -119,38 +119,35 @@ export const OilComponent = () => {
...
+    const cabinAlt = useArinc429Var('L:A32NX_PRESS_CABIN_ALTITUDE', 500);
+    const cabinVs = useArinc429Var('L:A32NX_PRESS_CABIN_VS', 500);
+    const deltaPsi = useArinc429Var('L:A32NX_PRESS_CABIN_DELTA_PRESSURE', 1000);
 
     const vsx = 440;
     const y = 385;
     const radius = 50;
 
-    const deltaPress = splitDecimals(deltaPsi);
+    const deltaPress = splitDecimals(deltaPsi.value);
```
Benjozork, line 142: **question:** it seems like in this page, there is no handling of the Arinc429Word actually holding a valid value. Are we missing some logic to check the SSM?

fbw-a32nx/src/systems/instruments/src/SD/Pages/Press/Press.tsx
```diff
@@ -13,20 +13,30 @@ import { SvgGroup } from '../../Common/SvgGroup';
...
     const [flightPhase] = useSimVar('L:A32NX_FWC_FLIGHT_PHASE', 'enum', 1000);
-    const [systemNumber] = useSimVar('L:A32NX_PRESS_ACTIVE_CPC_SYS', 'number', 1000);
     const [safetyValve] = useSimVar('L:A32NX_PRESS_SAFETY_VALVE_OPEN_PERCENTAGE', 'percentage', 500);
 
     const [cabinAltTextCss, setCabinAltTextCss] = useState('');
     const [cabinAltGaugeCss, setCabinAltGaugeCss] = useState('');
 
     useEffect(() => {
-        if (Math.round(cabinAlt / 50) * 50 >= 8800 && Math.round(cabinAlt / 50) * 50 < 9550) {
+        if (Math.round(cabinAlt.value / 50) * 50 >= 8800 && Math.round(cabinAlt.value / 50) * 50 < 9550) {
```
Benjozork, line 36: Ditto in this file.

Benjozork (changes requested):
Review of JS part

fbw-a32nx/src/systems/instruments/src/SD/Pages/Crz/Crz.tsx
mjuhe: This is a great point and something I thought about. The issue is that when both CPCs fail in the real aircraft, the data for the ECAM pages is taken from a backup pressure sensor which is not modelled at the moment. 

For now I've decided to take the values from the active CPC regardless of whether it is sending a Failure Warning signal (as forcing a failure doesn't affect the calculation of the values anyway).

fbw-a32nx/src/systems/instruments/src/SD/Pages/Press/Press.tsx
mjuhe: As above :) let me know your thoughts.

2hwk:
Approved for rs

fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
```diff
@@ -575,29 +644,39 @@ impl<C: PressurizationConstants> ControllerSignal<PressureValveSignal>
...
+            writer.write_arinc429(
+                &self.cabin_vs_id,
+                self.cabin_vertical_speed_out().get::<foot_per_minute>(),
+                ssm,
+            );
+            writer.write_arinc429(
+                &self.landing_elevation_id,
+                self.landing_elevation_out(),
+                ssm,
+            );
```
Gurgel100, line 678: You probably want to write these values even if `self.is_active` is false else the ssm will always be normal operation. Or do I miss something?

fbw-common/src/wasm/systems/systems/src/failures/mod.rs
```diff
@@ -13,7 +13,11 @@ pub enum FailureType {
     TrimAirOverheat(ZoneType),
     TrimAirFault(ZoneType),
     TrimAirHighPressure,
+    CpcFault(usize),
+    OutflowValveFault,
+    SafetyValveFault,
```
Gurgel100, line 18: Move these after `GalleyFans` to have them in the same order as everywhere else.

fbw-a32nx/src/systems/instruments/src/SD/Pages/Crz/Crz.tsx
Benjozork: So, if I understand correctly, right now those vars would never have an SSM that isn't NormalOperation?

If that is right, let's add a TODO comment explaining this in the relevant UI files.

fbw-a32nx/src/wasm/systems/a320_systems/src/air_conditioning.rs
```diff
@@ -694,13 +692,14 @@ impl<const ZONES: usize> SimulationElement for A320AirConditioningSystemOverhead
...
-    active_cpc_sys_id: VariableIdentifier,
-
     cpc: [CabinPressureController<A320PressurizationConstants>; 2],
+    cpc_interface: [PressurizationSystemInterfaceUnit; 2],
     outflow_valve: [OutflowValve; 1], // Array to prepare for more than 1 outflow valve in A380
     safety_valve: SafetyValve,
     residual_pressure_controller: ResidualPressureController,
     active_system: usize,
+
+    pressurization_overhead: A320PressurizationOverheadPanel,
```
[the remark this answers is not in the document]
mjuhe: Mainly to be able to update the overhead FAULT light when both CPCs fail. It's neater to have it all contained within the `A320PressurizationSystem` struct. This is also how it's set up for the air conditioning system.

fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
mjuhe: Since we have one ARINC variable for both of the CPCs, I wanted to avoid CPC 2 overwriting the values with `NoComputedData` when CPC1 was active. But you are right that this always returned `NormalOperation` because the conditional was incorrect. I've changed it now so if both CPCs fail the SSM is `FailureWarning`.

In reality both CPCs should be sending separate ARINC words, and these "cleaned" within the SDAC depending on which system is active. As we don't have the SDAC modelled yet this can be left for future work imo as we probably don't want to add all of the logic into the display js code.

fbw-a32nx/src/systems/instruments/src/SD/Pages/Crz/Crz.tsx
mjuhe: After conversations with @lukecologne on discord I've changed this and added the analog signals coming from the manual partition of CPC1. I've had to use a couple of workarounds since we don't have the SDAC logic implemented. The manual partition at the moment sends cabin altitude and differential pressure, which is not true to real life (these should be calculated within the SDAC).

fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
Gurgel100: I understand what you're doing here. Just generally it's a bad idea to write a simvar from 2 different systems.

lukecologne: > In reality both CPCs should be sending separate ARINC words, and these "cleaned" within the SDAC depending on which system is active

This is actually not correct, both SDACs retransmit the A429 words from both CPCs, under different SDIs

Gurgel100: So how do we proceed?

tracernz: The CPCs should write their own separate vars. It's not their job to implement logic for other systems, and IMO it's better to create a stub for the other system to do it if needed.

mjuhe: I can make the CPCs write separate vars that's no problem. The issue is all the logic to select which one to use will end up in the javascript display end, but I'll implement this 👍

tracernz: We have much better tools to deal with data from the MSFS SDK on the typescript side, so this is a positive IMO.

mjuhe: Fixed now, each CPC sends its own arinc vars and the instruments select based on which CPC is active. 👍

fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
```diff
@@ -1220,6 +1220,7 @@ export class PseudoFWC {
         this.cpc2DiscreteWord.setFromSimVar('L:A32NX_PRESS_CPC_2_DISCRETE_WORD');
 
         const activeCpc = this.cpc1DiscreteWord.bitValueOr(11, false) ? this.cpc1DiscreteWord : this.cpc2DiscreteWord;
+        const activeCpcNumber = activeCpc === this.cpc1DiscreteWord ? 1 : 2;
```
Gurgel100, line 1223: Why not do it this way?
```suggestion
        const activeCpcNumber = this.cpc1DiscreteWord.bitValueOr(11, false) ? 1 : 2;
        const activeCpc = activeCpcNumber  === 1 ? this.cpc1DiscreteWord : this.cpc2DiscreteWord;
```

fbw-a32nx/src/systems/instruments/src/SD/Pages/Crz/Crz.tsx
```diff
@@ -124,8 +124,10 @@ export const PressureComponent = () => {
 
     const cpcDiscreteWordToUse = cpc1DiscreteWord.getBitValueOr(11, false) ? cpc1DiscreteWord : cpc2DiscreteWord;
 
+    const activeCpcNumber = cpcDiscreteWordToUse === cpc1DiscreteWord ? 1 : 2;
```
Gurgel100, line 127: Same as above.

fbw-a32nx/src/systems/instruments/src/SD/Pages/Press/Press.tsx
```diff
@@ -307,7 +310,8 @@ const PressureComponent = () => {
     const cpcDiscreteWordToUse = cpc1DiscreteWord.getBitValueOr(11, false) ? cpc1DiscreteWord : cpc2DiscreteWord;
     const landingElevationIsMan = cpcDiscreteWordToUse.getBitValueOr(17, false);
 
-    const cpcLandingElevation = useArinc429Var('L:A32NX_PRESS_LANDING_ELEVATION', 500);
+    const activeCpcNumber = cpcDiscreteWordToUse === cpc1DiscreteWord ? 1 : 2;
```
Gurgel100, line 313: Same as above.

Gurgel100 (commented):
Some last remarks.

fbw-a32nx/src/systems/instruments/src/EWD/PseudoFWC.ts
mjuhe: This also works 😛 Changed now.

Gurgel100 (approved):

Benjozork (approved):

2hwk:
Fix merge conflicts, then reset ready to tag tag.

noonebling:

Discord          : noonebling
Object of testing: 8511
Tier of Testing  : 2
Date             : 20/04/2024

Testing Process: 
1. Start flight and climb up to 8000ft
2. Enabled CPC1 to fail. 
3. Checked for an amber warning on ECAM for "SYS 1 FAULT" with no aural sound
4. Repeated the same process with CPC2 and checked ECAM for the aforementioned warning
5. Enabled CPC1 and 2 to fail simultaneously
6. Checked for an amber warning on ECAM for "SYS 1+2 FAULT" with a single chime
7. Switched on MAN mode in the pressurization system and waited for the Residual Pressure Controller to kick in after 55 seconds
8. Reset the simulation and performed a circuit flight with CPC1+2 set to fail. 
9. Ensured the pressurization system is in auto mod, and landed
10. After landing, checked for "OFV NOT OPEN" amber warning on ECAM
11. Reset the simulation and climbed up to FL330 with CPC1+2 set to fail after reaching cruising altitude
12. The MASTER WARN lights flash with "EXCESS CAB ALT" in red, shown in ECAM.
13. When performing an emergency decent, checked for "LOW DIFF PR" amber warning, when the pressure inside aircraft equalised with the outside pressure
14. Checked for "SAFETY VALVE OPEN" amber warning if excessive positive or negative differential pressure occurs with a single chime


Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
Works as expected

Media:
image
image

PR #8519, opened 2024-02-23
feat: Chartfox support

### LINKED ISSUE #5768: Chartfox Support
### Aircraft Version

Experimental

### Description

I'd really like to see support for Chartfox because I can't afford navigraph and would really like to use charts in the aircraft.

### References (optional)

_No response_

_No response_

Scrub Man#4699

BenJuan26:
Fixes #5768 

## Additional context

### DIFF
--- a/fbw-a32nx/src/localization/flypad/en.json
+++ b/fbw-a32nx/src/localization/flypad/en.json
@@ -239,6 +239,21 @@
       "ScanTheQrCodeOrOpen": "Scan the QR Code or Open",
       "Title": "Navigraph"
     },
+    "ChartFox": {
+      "AirportDoesNotExist": "Airport Does Not Exist",
+      "GoToThirdPartyOptions": {
+        "Title": "Go to 3rd Party Options to link ChartFox account",
+        "Button": "3rd Party Options"
+      },
+      "AuthenticateWithChartFox": "Authenticate with ChartFox",
+      "InsufficientEnv": "Insufficient .env File",
+      "IntoYourBrowserAndEnterTheCodeBelow": "into your browser and enter the code below\n",
+      "LoadingMsg": "Loading",
+      "NoAirportSelected": "No Airport Selected",
+      "ResetChartFoxAuthentication": "Reset ChartFox Authentication",
+      "ScanTheQrCodeOrOpen": "Scan the QR Code or Open",
+      "Title": "ChartFox"
+    },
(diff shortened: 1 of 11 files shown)


ZigTag (commented):
I haven't done a full review but this is some awesome work, glad you got it to work as I've always had problems with the old API, also glad to see the new one is working properly!

Benjozork:
Could you get in touch with us on Discord in the #dev-support channel so we can coordinate regarding API keys?

BenJuan26:
This is ready for early review. If you want to test in-sim, you'll need the sample data, which I can provide in Discord.

In particular I need opinions: I've put in loading toasts for PDF charts, since they have to go through SimBridge. When loading is fast, they can be a little loud. 1) Should I leave them in? 2) Should I add them for image-based charts? They usually load pretty quickly.

FoxtrotSierra6829:
>- Figure out what's happening with phantom runway associations, e.g. CYYZ IKLEN THREE DEP IKLEN3 SID-10C (10C is not a runway)
>	- This is just bad data from them. It looks the same way on their website. I reported the inaccurate data for this particular case.

This is the chart index number, not bad data from them. In this case, `CYYZ-SID-10C` is the full index designator as issued by Nav Canada. For CYYZ, the SIDs are numbered 1A till 28D. The index number is used to quickly identify the correct chart for a given procedure. Jeppesen also numbers their charts, tho it has a slightly different system.
See below:
grafik

BenJuan26:
Thanks for the detailed reply. I don't have a Navigraph subscription, so despite seeing `chartIndex` in the code, I didn't know what that data looked like.. I could tell that in this case the `10C` was an identifier and not a runway, which is why it confused me that it was listed as a runway in the data:

```typescript
meta: [
    {
        type: 0,
        type_key: 'Runways',
        value: [
            '10C',
        ],
    },
],
```

That being said, this is just an API and is for machines to read, not humans, and it serves the purpose of grouping the charts. More importantly, it works the same way in my implementation as it does on their website, so no action is needed. I do think I'll review how the Navigraph implementation utilizes chart indices and see if I can do something similar here.

FoxtrotSierra6829:
Ah I see, yeah in that case their system must be misinterpreting the data indeed.

BenJuan26:
I'll be pulling back from this until we hear back from ChartFox. There are a number of problems that can't be solved until we hear from them.

PR #8523, opened 2024-02-25
fix: Pushback manual control improvement

frankkopp:
## Summary of Changes
The last pushback PR the way controller input resets buttons controlled unput so that rudder or elevator input would reset both speed and heading of pushback. This PR undoes this change as using the button for speed and rudders for direction was no longer possible. But as this is convenient for some users this change is reverted. 

## Testing instructions
Check that if using a button for speed (forward/backward) a rudder input does not reset the speed. 
Also using button for heading then elevator input should not reset the heading. 


FlightSimmer7700:
Quality Assurance Tester

Discord: flightsimmer7700
Object of testing: #8523 
Tier of Testing: 1
Date: 25/02/2024

Testing Process: 
Loaded in with both aircraft and tested that the pushback now works with preselecting a tug speed and then using controls to turn and when using controls for tug speed and selecting the turn using the slider
Negatives: N/A
Testing Results: Passed

Conclusions: Worked perfectly for both aircraft, great work and thanks for the quick fix!

2hwk (approved):

PR #8529, opened 2024-02-27
fix(actions): Fix QA Label action to use new FBW Action

vulpesfeggari:
## Summary of Changes

Changes the exiting QA action that was my using my action 'auroraisluna/qa-labels-for-prs' to use a FBW Action instead.

## Screenshots (if necessary)
Testing process:
msedge_ZOrTkvuzdt

## Additional context

Discord username (if different from GitHub): alepouna

## Testing instructions
No QA needed.


.github/workflows/qa-labels-for-prs.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       pull-requests: write # For adding labels, otherwise set to 'read'
     steps:
       - name: "PR QA Comments Helper"
-        uses: "auroraisluna/qa-labels-for-prs@v1.0.5" # Make sure to use the latest version available!
+        uses: "flybywiresim/actions-qa-labels@master" # Make sure to use the latest version available!
```
Gurgel100, line 12: Do we really want to use `@master`?

alepouna: Answering from my main - Considering this is a FBW maintained action now, there is 'pushing to master' protection, so master would only be updated when it has been 'approved'. The point of using `@master` is to not require a new PR every time there is an update on the action.

 I personally consider this just a faster way to ensure the action stays up2date in this (and other, in the future:tm:) repos without the friction of opening a PR every time. Of course, if this action was still under my account (like it is now) `@master` would not be an option, but since it is maintained by FBW, I don't see the reason not to have it automatically be kept up to date. 

Also, this action being 'broken' by a perhaps bad update (as it was for the past few weeks due to my mistaken account suspension) does not in any form mess with the workflow of the aircraft repository, the users, or devs, just makes checking for QA'd PRs slightly slower, so I would say this action not being functional due to a bad update, is almost not a problem but a minor time inconvenience.

Gurgel100: Good points.

Gurgel100 (approved):

PR #8543, opened 2024-03-01
fix(GSX): Fix GSX Integration not disabling the a/c provided cones 

### LINKED ISSUE #8241: [EFB] GSX Integration does not visually disable the chocks & cones buttons
### Aircraft Version

Development

### Build info

```json
{
    "built": "2023-09-30T05:19:00+00:00",
    "ref": "refs/heads/master",
    "sha": "ad7d65e41bf91472aeda0268a24a82be2a1e83db",
    "actor": "Benjozork",
    "event_name": "manual",
    "pretty_release_name": "master:ad7d65e4",
    "version": "v0.11.0-dev.ad7d65e"
}
```

### Describe the bug

When enabling the GSX integration, the Chocks & Cones buttons from the Sim options settings page, does not actually visually disable the buttons. You cannot click them, but I believe, similar to how users get confused why they cannot activate fuel or control the payload due to the GSX integration; some users may be confused why they can't turn on this setting, due to the integration. 

### Expected behavior

The buttons disable when the GSX integration is activated.

### Steps to reproduce

1. Activate GSX integration
2. Enter the Sim Options page
3. See Chocks and Cones buttons

### References (optional)

_No response_

_No response_

alepouna

alepouna:
Fixes #8241

## Summary of Changes

GSX has the ability to show its own cones if enabled by the aircraft configuration. Currently in the A320 GSX config this feature isn't enabled (unknown why). The A380 GSX Config has this enabled. 

This PR fixes the GSX Integration not disabling the aircraft(s) cones and also allowing the aircraft cones from being toggled in the EFB. If the integration is disabled, based on previous user setting, the cones will re-enable (or stay off). 

Note for localization: This is reusing the string from the payload page, so new strings are not required. 

A380 Note: Due to lack of the XML for the cones, there is no change to be done there. A new PR will be required once those are added. 

**Thoughts**: ~~Do we want to listen for the GSX cones Lvars to toggle this setting on/off instead of relying on the user enabling the GSX integration?~~ (EDIT: GSX does not currently have Lvars for the cones - automatic on/off not possible 😦 )
~~In addition, do we want to perhaps force the GSX cones state on via the EFB GSX integration (so payload will also turn on the aircraft configuration for cones)?~~ (EDIT: We can do that via the default configuration - addition not needed) 

## Screenshots (if necessary)

FlightSimulator_NNA4QL6pzb

https://github.com/flybywiresim/aircraft/assets/98479040/48831317-a257-4063-956f-99e8962d9c51

## Testing instructions

**Requires GSX**
1. Spawn on a gate 
2. Ensure GSX custom cones are disabled (via GSX Aircraft Settings)
3. Disable the GSX Payload integration from EFB Settings > 3rd Party
4. Ensure you can enable/disable the cones via EFB Settings 
5. Enable the GSX Payload integration 
6. Check you do not see a switch on cones setting
7. [Optionally] Confirm via Behaviors tool that `A32NX_GSX_PAYLOAD_SYNC_ENABLED` is 1.0. Cones  `A32NX_MODEL_CONES_ENABLED` ) should be 0.0. 
8. Enable GSX custom cones (via GSX Aircraft Settings)
9. Repeat steps 3 to 7, ensure the aircraft cones visually disappear when the GSX integration is enabled and only GSX cones are visible

**Does not require GSX**
- Check for any issues with the GSX Integration being disabled and check if the cones can be toggled on/off. 

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -73,7 +73,7 @@
 1. [FCU] Fix the QFE indicator not illuminating - @BenJuan26 (BenJuan26)
 1. [FCU] Disabled QFE mode - @tracernz (Mike)
 1. [FUEL] Lowered starting fuel on C/D spawn, will only load last saved fuel on C/D spawn, center tank refuel now happens simultaneous with wing refuel - @Maximilian-Reuter
-1. [GSX] Fix GSX Integration not disabling wheel chocks and cones
+1. [GSX] Fix GSX Integration not disabling the a/c provided cones
 
 ## 0.11.0
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -73,6 +73,7 @@
 1. [FCU] Fix the QFE indicator not illuminating - @BenJuan26 (BenJuan26)
 1. [FCU] Disabled QFE mode - @tracernz (Mike)
 1. [FUEL] Lowered starting fuel on C/D spawn, will only load last saved fuel on C/D spawn, center tank refuel now happens simultaneous with wing refuel - @Maximilian-Reuter
+1. [GSX] Fix GSX Integration not disabling wheel chocks and cones
 
 ## 0.11.0
 
--- a/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO.xml
+++ b/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO.xml
@@ -500,15 +500,15 @@
             <!-- GND CHOCKS -->
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_NOSE_GEAR</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_MAIN_GEAR_RIGHT</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_MAIN_GEAR_LEFT</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
             </UseTemplate>
 
             <!-- GND CONES -->
--- a/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO.xml
+++ b/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO.xml
@@ -500,37 +500,37 @@
             <!-- GND CHOCKS -->
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_NOSE_GEAR</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_MAIN_GEAR_RIGHT</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CHOCKS_MAIN_GEAR_LEFT</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED)</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_WHEELCHOCKS_ENABLED) (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
 
             <!-- GND CONES -->
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CONE_B</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) and</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) && (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CONE_R</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) and</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) && (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
             <UseTemplate Name="FBW_Gnd_Equipment">
                 <NODE_ID>CONE_L</NODE_ID>
-                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) and</GND_ENABLE_OVRD>
+                <GND_ENABLE_OVRD>(L:A32NX_MODEL_CONES_ENABLED) (L:A32NX_IS_STATIONARY, bool) && (L:A32NX_GSX_PAYLOAD_SYNC_ENABLED) ! &&</GND_ENABLE_OVRD>
             </UseTemplate>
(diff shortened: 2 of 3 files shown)


Fragtality:
> GSX has the ability to show its own wheel chocks and cones if enabled by the aircraft configuration. Currently in the A320 GSX config this feature isn't enabled (unknown why). The A380 GSX Config has this enabled.

That is incorrect. GSX only can show Cones - but no Chocks.
This was fully intentional with my PR (8405) - just because GSX has Cones, does not mean the Aircraft Developer has to use them. Especially when own Cones have been designed.

From a GSX Standpoint, there is absolut zero need to disable the Aircraft Cones & Chocks.

I generally don't know what the Issue should be - with or without GSX, the Chocks & Cones are placed and removed automatically. Do some Users really feel the Need that they have to set them manually and no one/nothing else?!

Sleinmaster:
I agree with fragtality. The cones and especially chocks from the A32NX should be preferred over the once from gsx.

alepouna:
> That is incorrect. GSX only can show Cones - but no Chocks.
Oh yeah you are right, edited description. Will look into removing the chocks toggle lock and keeping it only for the cones.

Considering we have the integration available, for GSX users who will enable the integration I do not see a reason to not ensure there isn't duplication of the cones. In my opinion the entire feature should auto-detect if GSX is enabled and disable the A32NX cones and enable the GSX ones, but from quick testing I don't see a reliable way to get this information and set them on GSX.

Fragtality:
Duplication of Cones?
That only happens if Users made a custom Aircraft Profile before my PR was merged. The Profile that comes with the current master has the GSX Cones disabled in the Profile.
And fixing that is only a Matter of GSX Menu -> Aircraft Customization -> Remove Checkbox

Fragtality:
And I still don't genuinely understood what the Problem is that would need a Solution.
Fired up the master, have Chocks & Cones (starting Cold & Dark). They are indicated in the EFB. When preparing the Plane fur Push-Back they are removed as soon as Push-Back is called or I switch on the Beacon.
Regardless of the GSX Sync Settings.

So what I am missing? :/

2hwk:
So essentially if I am following correctly, what this comes down to is do we want GSX Chocks and Cones or our own implementation when using GSX

alepouna:
Please also note my thoughts listed in the PR
> **Thought**: Do we want to listen for GSX Lvars to toggle this setting on/off instead of relying on user configuration? In addition, do we want to perhaps force the cones state on via the EFB GSX integration (so payload will also turn on the aircraft configuration for cones)?

2hwk:
> Please also note my thoughts listed in the PR
> 
> > **Thought**: Do we want to listen for GSX Lvars to toggle this setting on/off instead of relying on user configuration? In addition, do we want to perhaps force the cones state on via the EFB GSX integration (so payload will also turn on the aircraft configuration for cones)?

+1 on not relying on users to set a flag manually

alepouna:
Oh sidenote, on my 'thoughts' I've written
> (so payload will also turn on the aircraft configuration for cones)?

The PR already does this (but turns it OFF not on, typo there)

Fragtality:
> EDIT: GSX has cones, but not chocks, so we'll be removing the functional changes and keeping the EFB Settings changes in this PR

Well my former PR had the exact Motivation to remove the Connection to GSX Payload/Fuel Synch. I just never saw a logical Reason to hook Chocks/Cones to that. For Payload Synchronization, GSX does not care if there are Cones. Or Chocks - all it cares is that the Plane can't move.
<br/>

> +1 on not relying on users to set a flag manually

With that Approach (+1 from me too), it would make more Sense to me if the EFB Options for Chocks and Cones are really needed at all. I mean they are already placed & removed automatically (regardless of GSX). So from the Top of my Head I can't see any Use-Case where a User does not want them? What would be a negative Impact (considering they don't need manual Handling)?
<br/>

> Please also note my thoughts listed in the PR

Sure, but discussing an Approach for a Solution when the originating Problem to be solved is not clear isn't really productive 😅 
But yeah, I would like the Idea to configure something automatically so the User does not have to care about that (I mean that is somewhat the Motivation of all my PRs so far^^). Problem is: There are no GSX Lvars that would have anything to do with Chocks or Cones. You could maybe test for something like `FSDT_GSX_COUATL_STARTED != 0` if GSX is active - but beware, when you restart GSX this will be momentarily 0.
<br/><br/>


At the End of the Day, this is your "Baby" here ... I'm just a random Dude making very narrow Changes with only 50% knowing what I'm doing 🤣 
BUT, as a User, which lead to my Involvement, my View on Chocks & Cones:
- I want Chocks & Cones, especially the Ability to release the Parking Brake on the Gate (because Chocks are in Place and detected by GSX)
- I don't want to be excluded as GSX User to not have them
- Although I want them, I don't want to care about them. Meaning: they should be placed automatically on Startup and Arrival and should be removed automatically on Departure.

Maximilian-Reuter:
I do have to say I agree with fragtality here, we are providing a GSX profile that is customized for the plane. If somone really wants to use GSX cones they will have to change the profile and they will need to deactivate the option in the EFB. 

Only question I have is if we really need to support that use case rare as it probably is or simply always have chocks and cones enabled as they dont really have any downsides as they are automatic already.

alepouna:
> BUT, as a User, which lead to my Involvement, my View on Chocks & Cones:
> - I want Chocks & Cones, especially the Ability to release the Parking Brake on the Gate (because Chocks are in Place and detected by GSX)
> - I don't want to be excluded as GSX User to not have them
> - Although I want them, I don't want to care about them. Meaning: they should be placed automatically on Startup and Arrival and should be removed automatically on Departure.

For the record, this PR does not change the behavior of the above. The entire point of the PR is to prevent visual overlaps with the FBW cones, since GSX provides cones itself, as this was the assumed behavior with my bug reported in #8241. Non GSX users, will still be able to enable/disable the cones (and chocks) as they desire, this does not affect them. 

Now granted, the A320 configuration does not include the cones by default, I can see an argument being made that there will not be conflicts, if the default configuration is used. I can also see the argument of "but the user can just disable either of them then" and yes that is true, but I will also use the same reasoning for disabling the EFB fuel and boarding if GSX is enabled.

Considering GSX unfortunately does not have an LVAR *yet* for the cones, there is no way for the aircraft to know if the configuration is enabled so the only solution to prevent potential visual conflicts is to lock the feature (as again, we do for Boarding and Fueling - granted they have some LVARs). 
It is of my opinion, that since the user is enabling the GSX Integration, they should also have access to all the GSX related features by default. Other than personal preference of the user, I do not see any reason why they should be managed by the A32NX, IF the GSX integration is enabled. (and I will make the argument again that if the user wants to use GSX, they wish to use the integration, otherwise they would opt out by disabling the setting on the Third Party Options)

> we are providing a GSX profile that is customized for the plane.

For the above reason I also think this PR is good groundwork to enable the GSX cones by default without having visual conflicts on merge - If we provide already some GSX integration, I don't see why we should also not let GSX handle the cones. 

> Only question I have is if we really need to support that use case rare as it probably is or simply always have chocks and cones enabled as they don't really have any downsides as they are automatic already.

I don't really understand this question. What do you mean by having the cones always enabled? If a user has the GSX cones enabled, and we force the aircraft ones, there will just be a visual conflict. The automatic part of the cones (/chocks) is when the hide based on aircraft state. Unless you mean forcing the GSX cones off and forcing the NX ones on?

alepouna:
I have updated the PR to not include chocks anymore as those are not managed by GSX.

Sleinmaster:
In that case I'd say the gsx cones should be enabled in our aircraft profile in this pr as well

Maximilian-Reuter:
> > BUT, as a User, which lead to my Involvement, my View on Chocks & Cones:
> > 
> > * I want Chocks & Cones, especially the Ability to release the Parking Brake on the Gate (because Chocks are in Place and detected by GSX)
> > * I don't want to be excluded as GSX User to not have them
> > * Although I want them, I don't want to care about them. Meaning: they should be placed automatically on Startup and Arrival and should be removed automatically on Departure.
> 
> For the record, this PR does not change the behavior of the above. The entire point of the PR is to prevent visual overlaps with the FBW cones, since GSX provides cones itself, as this was the assumed behavior with my bug reported in #8241. Non GSX users, will still be able to enable/disable the cones (and chocks) as they desire, this does not affect them.
> 
> Now granted, the A320 configuration does not include the cones by default, I can see an argument being made that there will not be conflicts, if the default configuration is used. I can also see the argument of "but the user can just disable either of them then" and yes that is true, but I will also use the same reasoning for disabling the EFB fuel and boarding if GSX is enabled.
> 
> Considering GSX unfortunately does not have an LVAR _yet_ for the cones, there is no way for the aircraft to know if the configuration is enabled so the only solution to prevent potential visual conflicts is to lock the feature (as again, we do for Boarding and Fueling - granted they have some LVARs). It is of my opinion, that since the user is enabling the GSX Integration, they should also have access to all the GSX related features by default. Other than personal preference of the user, I do not see any reason why they should be managed by the A32NX, IF the GSX integration is enabled. (and I will make the argument again that if the user wants to use GSX, they wish to use the integration, otherwise they would opt out by disabling the setting on the Third Party Options)
> 
> > we are providing a GSX profile that is customized for the plane.
> 
> For the above reason I also think this PR is good groundwork to enable the GSX cones by default without having visual conflicts on merge - If we provide already some GSX integration, I don't see why we should also not let GSX handle the cones.
> 
> > Only question I have is if we really need to support that use case rare as it probably is or simply always have chocks and cones enabled as they don't really have any downsides as they are automatic already.
> 
> I don't really understand this question. What do you mean by having the cones always enabled? If a user has the GSX cones enabled, and we force the aircraft ones, there will just be a visual conflict. The automatic part of the cones (/chocks) is when the hide based on aircraft state. Unless you mean forcing the GSX cones off and forcing the NX ones on?

Do you prefer the GSX cones over ours ? Has a user complained about having to see our cones instead of GSXs ? 

It feels like you are trying to solve an "issue" where none exists. We provide the default and really intended configuration for GSX with our aircraft. IF a user wants to deviate from that they can do so by modifying the GSX config. But I then also expect them to do the additional changes that entails like toggling the cones off in the EFB. 

An additional point. Our EFB shows the state of our cones but won't be able to detect the GSX cones as such this is another functionality downgrade.

Just because someone uses GSX doesn't mean they want to use "all" of GSX.

Also as a point of reference, our plans are to decouple GSX from our services more and more and make it more of a thing that gets detected and not toggled on and off strictly.

So from my side I don't see a reason to change the current behaviour.

Fragtality:
> For the record, this PR does not change the behavior of the above. The entire point of the PR is to prevent visual overlaps with the FBW cones, since GSX provides cones itself, as this was the assumed behavior with my bug reported in https://github.com/flybywiresim/aircraft/issues/8241. Non GSX users, will still be able to enable/disable the cones (and chocks) as they desire, this does not affect them.
Now granted, the A320 configuration does not include the cones by default, I can see an argument being made that there will not be conflicts, if the default configuration is used. I can also see the argument of "but the user can just disable either of them then" and yes that is true, but I will also use the same reasoning for disabling the EFB fuel and boarding if GSX is enabled.
Considering GSX unfortunately does not have an LVAR yet for the cones, there is no way for the aircraft to know if the configuration is enabled so the only solution to prevent potential visual conflicts is to lock the feature (as again, we do for Boarding and Fueling - granted they have some LVARs).
It is of my opinion, that since the user is enabling the GSX Integration, they should also have access to all the GSX related features by default. Other than personal preference of the user, I do not see any reason why they should be managed by the A32NX, IF the GSX integration is enabled. (and I will make the argument again that if the user wants to use GSX, they wish to use the integration, otherwise they would opt out by disabling the setting on the Third Party Options)

Well, I never understood the Issue report in the first Place to be honest. The FBW Chocks and Cones are used and their current State can be seen in the EFB. Why is that a Bug?

Why do User need manual Control over Chocks & Cones Placement at all (when they are handled automatically for everyone with or without GSX)? IRL the Flight Crew does not need to do such a Micromanagement. What Aspect is simulated here?

Where is the visual Overlap? The included Aircraft Profile does not have Chocks enabled. If the Aircraft is used as indented there is no Overlap. User have all sorts of Ways to mess up their Aircraft Profile (Exit Positions, Brake Checks). There is no way I would see that FBW could protect Users from their self here. I don't see a Problem in Actions required by Users, if their Actions had caused a Problem in the first Place.

No one gets GSX because of the Cones. Other Aircraft also just don't never use GSX Cones (Fenix e.g.) and I never saw a Complaint about that. The only "Cone Discussion" I had on the Fenix was because of Fenix2GSX where user wanted their Placement to be more "realistic" - something I could not have accomplished with GSX Cones since there is 0 Control.
Users care about them being there, about their correct Placement and Removal. Not so much from where these Cones are provided.
The GSX Cones exist, so that even Planes that don't have them can have them through GSX. But the FBW has Cones.

I'm with you that some Behavior has to change when "opting in" for GSX. But I think that should only change what is necessary. Disabling the Instant Options imho is not, that's why another PR of mine already changed that for Refueling (allowing Instant Refueling with the Synch Option enabled in the EFB - 8407).

And binding Cones to Payload Synchronization still does not make any Sense to me. That's why I decoupled that. They are not related to each other.
On a Car, it is as if you would say: you can't use the Trunk while having a Cup in the Cup Holder.

<br/><br/>

> It feels like you are trying to solve an "issue" where none exists. We provide the default and really intended configuration for GSX with our aircraft. IF a user wants to deviate from that they can do so by modifying the GSX config. But I then also expect them to do the additional changes that entails like toggling the cones off in the EFB.

+1 on that.

<br/><br/>

My View/Recommendation would be to use the FBW Cones. You can detect them, you can toggle them dynamically - you have full Control over them.
It would be the same Result/Behavior for everyone - regardless of GSX is installed or used or what is set in the EFB. And that is the *current* State of master. Nothing to worry about with EFB Settings, Chocks & Cones are handled automatically with or without GSX. You can see their State in the EFB (which is not a Bug).
It just works. So much so, that the Need for keeping the Chocks & Cones Settings in the EFB are even debatable at this Point.

PR #8557, opened 2024-03-07
feat(fac): improve beta estimation

lukecologne:
## Summary of Changes

This PR immproves the beta estimation by employing a new method.
The method utilizes the fact that the lateral acceleration is approximately proportional to the sideslip $\beta$:
```math
\beta = \frac{b_y}{V Y_\beta}
```
where $b_y$ is the lateral acceleration as measured by the onboard accelerometer, and $Y_\beta$ is the proportionality factor for the sideforce generated by the sideslip.
The sideslip estimated by this method is then low-pass-filtered with a time constant of 4s.

Further, the PR fixes the calculations of the body acceleration calculations in the IRs, which is required for this to work. The vertical acceleration was correct and remains unchanged, and the other components are not yet used (except for the sideslip indicator on ground, which is unaffected by the change). This supersedes #7417.
Also, the PR cleans up some unused filters in the PFD, in the sideslip indicator component.

## Screenshots (if necessary)

Demonstration of the new sideslip estimation on the sideslip indicator: https://streamable.com/fakahu
You can see that, when using the rudder, the indication first goes in the opposite direction and then settles to the correct direction. This is caused by the acceleration value given by MSFS, which shows the exact same behaviour. I'm not sure if this is correct, but I can't think of a way around it (if it's not correct)

## Additional context

## Testing instructions

Check if the sideslip indicator works and behaves as expected


2hwk (approved):

noonebling:

Discord          : noonebling
Object of testing: #8557
Tier of Testing  : 1
Date             : 18/03/2024

Testing Process: 
Log into sim
Start aircraft as usual
Reached 10,000ft
Check if the sideslip indicator works and behaves as expected

Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
Works as expected

Media:
https://streamable.com/i2vivc

PR #8466, opened 2024-02-05
style: lint and styling overhaul

Benjozork:
## Summary of Changes

* Bump eslint-related package versions
* Consolidate eslint configs for modern TS code under one config
* Introduce Prettier as main formatting style, integrated with ESLint
* Fix linting of package base JS
* Fix lint issues now coming up after cleanup of configs/upgrade of packages

## Testing instructions

- Should not affect code, but make sure nothing is obviously broken
- Focus on that the guidance (LNAV/VNAV) does not have any new bugs 
  - Routes (or similar):
    - EDDM-LOWI LOC26R (circle to land) 
    - KCRQ24 PADRZ2 CHKNN AVE ROBIE5 BORED KSJC30L (ILS)
    - EDDF/25L F250 SULUS1F SULUS Z12 LASGA T202 ATGUP ATG25L EDDB/25L
    EBBR-EGKK26L 
- EFB does not have incorrect/broken formatting

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8581, opened 2024-03-23
feat: Improved expedited presets with quick ADIRS/APU/ENG start/stop

frankkopp:
## Summary of Changes
Improve the expedited quick Aircraft Preset loading with making ADIRS/APU/ENG starting and stopping instantly.

Major changes
- changes to ADIRS, APU, ENG, flaps/slats to allow a quick mode for very fast or instant startup and shutdown
- Improvements to the expediated mode to leverage the quick mode - e.g. new step types
- refactor the configuration to use a static XML file sitting in the aircraft package to allow future modding of presets

Additional changes:
- fixing and improving the imbalance calculation for the A32NX fadec
- Changed two wrongly named LVars (missing prefix):
  - AIRLINER_CRUISE_ALTITUDE ==> A32NX_AIRLINER_CRUISE_ALTITUDE
  - AIRLINER_TO_FLEX_TEMP ==> A32NX_AIRLINER_TO_FLEX_TEMP

## Screenshots / Video
https://github.com/flybywiresim/aircraft/assets/16833201/bcd0c3da-7f3d-4166-8057-e17cd2c1a9d5

## Testing instructions
A32NX:

Test all possible combinations of presets and presets sequences. 
Test expedited or normal speed - even mix or change while loading
Regression Test and check normal operation without using presets and ensure this has not regressed in any way.

Check:
- presets must load without hanging (hanging means the xml config is not correct)
- any intermittent error and alarms that comes up goes away after a short time without manual interaction
- no errors after a preset is loaded
- no unusual state of the aircraft after a preset is loaded - example flaps/slats not retracted/extended
- check if the state of the cockpit for each preset makes sense 
	- these are roughly done along the SOP but not exactly as this wouldn't make sense 
	- if you think a setting/step is incorrect in general or for a certain preset then report this - but this is not an error and does 
	  not fail QA
	- the steps are easily changed based on user and pilot feedback

Try to break it :D

Also check for regressions - there have been some changes to the ADIRS, APU, FADEC, flaps/slats code to make these instant or near-instant. So a full flight is recommended where these systems are also monitored/checked for any issues. 

A380X:
Same but with less priority


fbw-common/src/wasm/systems/systems/src/apu/electronic_control_box.rs
```diff
@@ -348,6 +357,34 @@ impl<C: ApuConstants> ElectronicControlBox<C> {
...
+    /// If the APU is not in quick mode, the method checks if the bleed air valve was open
+    /// in the last cooldown duration or if the APU is in a cooldown period.
+    /// If either of these conditions is true, the method returns true indicating that
+    /// a cooldown is required.
+    /// Otherwise, it returns false.
+    ///
+    /// # Returns
+    ///
+    /// * `bool` - True if a cooldown is required, false otherwise.
+    fn cooldown_required(&self) -> bool {
```
tracernz, line 374:
```suggestion
    fn is_cooldown_required(&self) -> bool {
```

fbw-common/src/wasm/systems/systems/src/apu/electronic_control_box.rs
```diff
@@ -64,7 +64,7 @@ pub(super) struct ElectronicControlBox<C: ApuConstants> {
     on_ground: bool,
     /** Absolute air pressure sensor in the air intake assembly. */
     inlet_pressure: Pressure,
-    apu_quick_mode: bool,
+    aircraft_preset_quick_mode: bool,
```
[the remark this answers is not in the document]
frankkopp: Will do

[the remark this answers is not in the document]
frankkopp: I will also prob refactor a little bit as the var could be introduced on mod.rs level and handed down via the update calls.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
```diff
@@ -1965,7 +1965,7 @@ class FMCMainDisplay extends BaseAirliners {
 
     set cruiseFlightLevel(fl) {
         this._activeCruiseFlightLevel = Math.round(fl);
-        SimVar.SetSimVarValue("L:AIRLINER_CRUISE_ALTITUDE", "number", this._activeCruiseFlightLevel * 100);
```
2hwk, line 1968: L:AIRLINER_CRUISE_ALTITUDE -> L:A32NX_AIRLINER_TO_FLEX_TEMP

We may want to keep this L:Var change in mind for supporting legacy external applications

fbw-a32nx/src/systems/fmgc/src/flightphase/Phase.ts
```diff
@@ -96,7 +96,7 @@ export class ApproachPhase extends Phase {
     landingConfirmation = new ConfirmationNode(30 * 1000);
 
     init() {
-        SimVar.SetSimVarValue('L:AIRLINER_TO_FLEX_TEMP', 'Number', 0);
+        SimVar.SetSimVarValue('L:A32NX_AIRLINER_TO_FLEX_TEMP', 'Number', 0);
```
2hwk, line 99: Same as above

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -1382,10 +1433,16 @@ impl InertialReference {
         &mut self,
         context: &UpdateContext,
         overhead: &AirDataInertialReferenceSystemOverheadPanel,
+        aircraft_preset_quick_mode: bool,
     ) {
+        let initialisation_duration: Duration = if aircraft_preset_quick_mode {
+            Self::ATTITUDE_INITIALISATION_QUICK_DURATION
+        } else {
+            Self::ATTITUDE_INITIALISATION_DURATION
+        };
```
2hwk, line 1442: See above about having ``aircraft_preset_quick_mode`` at this level

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -939,22 +973,33 @@ impl AirDataReference {
...
         &mut self,
         context: &UpdateContext,
+        aircraft_preset_quick_mode: bool,
         overhead: &AirDataInertialReferenceSystemOverheadPanel,
     ) {
+        let initialisation_duration: Duration = if aircraft_preset_quick_mode {
+            Self::INITIALISATION_QUICK_DURATION
+        } else {
+            Self::INITIALISATION_DURATION
+        };
```
2hwk, line 999: See above about having ``aircraft_preset_quick_mode`` at this level

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -534,11 +555,23 @@ impl AirDataInertialReferenceUnit {
...
+            simulator_data,
+        );
+        self.ir.update(
+            context,
+            &self.adr,
+            overhead,
+            align_time,
+            aircraft_preset_quick_mode,
+            simulator_data,
+        );
```
2hwk, line 574: Perhaps add a note here that this is to be refactored (abstracted) away from the `AirInertialReferenceUnit` struct, I don't there is a need for this struct to require awareness of ``aircraft_preset_quick_mode`` at this level. 

However solving this would require building a higher level wrapper and driver which would pass a mutable reference e.g. ``&mut adirs`` as a parameter in its update function, and perform manipulation like this, which is maybe a deep refactor and re-architecture that may be beyond the scope of this PR.

2hwk (changes requested):

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
frankkopp: I will add a note - I don't want to mess too much with the rust stuff.

fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/FMC/A32NX_FMCMainDisplay.js
frankkopp: I can add a note to the PR description - but as this LVar is not even documented in the fbw-a32nx/docs/a320-simvars.md file I think we are fine. 
I'll add these to the a320-simvars,md though.

2hwk (approved):

2hwk:
Could be impacted by the potential future merge freeze. https://github.com/flybywiresim/aircraft/pull/8466 may be required to be merged into this PR at some point. We're not final on this. Impact will be minor in this case.

2hwk:
On reflection this one is closer to a tier 2 than a tier 1 test, make sure also to check normal operation without using presets and ensure this has not regressed.

2hwk:
Quality Assurance Report

Discord : 2cas
Object of testing: https://github.com/flybywiresim/aircraft/pull/8581
Tier of Testing : 2
Date : 4.05.2024

Testing Process:

A32NX:

- [ ] Test all possible combinations of presets and presets sequences.

Intermittent HOT AIR issue, able to reproduce 100% by
 * Spawning in on the runway
 * Instantly loading the Cold and Dark preset
 * Batteries on + Power on APU
 
image
FlightSimulator_XfrBCw8Tce

This disappears if you disable the HOT AIR, then reeanble it (following the ECAM message), but does not disappear on its own.

- [x] Test expedited or normal speed - even mix or change while loading
- [x] Regression Test and check normal operation without using presets and ensure this has not regressed in any way.

Check:
 - [x] presets must load without hanging (hanging means the xml config is not correct)
 - [ ] any intermittent error and alarms that comes up goes away after a short time without manual interaction
 - [ ] no errors after a preset is loaded
 - [x] no unusual state of the aircraft after a preset is loaded - example flaps/slats not retracted/extended
 - [x] check if the state of the cockpit for each preset makes sense
        these are roughly done along the SOP but not exactly as this wouldn't make sense
        if you think a setting/step is incorrect in general or for a certain preset then report this - but this is not an error and does
        not fail QA
        the steps are easily changed based on user and pilot feedback

Try to break it :D

- [x] Also check for regressions - there have been some changes to the ADIRS, APU, FADEC, flaps/slats code to make these instant or near-instant. So a full flight is recommended where these systems are also monitored/checked for any issues.


Negatives:
- HOT AIR

Testing Results:
Not Passed

Conclusions:
Other than the above, works as expected

Pruznak:
I'm quite sure this has happened to me before even on the current dev build and I also had to disable and reenable hot air.

frankkopp:
> Quality Assurance Report
> 
> Discord : 2cas Object of testing: #8581 Tier of Testing : 2 Date : 4.05.2024
> 
> 
> This disappears if you disable the HOT AIR, then reeanble it (following the ECAM message), but does not disappear on its own.

> Negatives:
> 
> * HOT AIR

@2hwk - this should be fixed due to @mjuhe's help - if possible pls retest

2hwk:
Still observed this sometimes with the steps I listed after spawning on runway.

Also in general, The Ready for Taxi state does not seem to work (as per the video), it now waits for the APU to start up which seems like a regression.

frankkopp:
> Still observed this sometimes with the steps I listed after spawning on runway.
> 
> 
> 
> Also in general, The Ready for Taxi state does not seem to work (as per the video), it now waits for the APU to start up which seems like a regression.

I could not reproduce the hit air warning issue anymore. 

But I found another issue when starting on runway which I don't understand yet and I have little time currently. 
(Somehow when starting on runway and using preset 1 the xml loaded looses a part (powered_off) - raw xml has it, the parsed data structure not. Loading preset 1 again (xml is read every time) it there and works. Can't explain the issue yet. Might be the lib I'm using idk. When I find more time I will dig into this some more.

frankkopp:
I have refactored this to use another XML library as there was a strange issue with the previously used one which skipped one XML section for no apparent reasons when using preset 1 directly after loading into sim (either on gate or runway). 
The issue is fixed with using the other xml library (tinyxml2) - although this has a worse API than the previous pugixml it at least works without any issues. 

I could not reproduce the "hot air" issue any more a believe this to be fixed. 

@2hwk if you have time I would appreciate another test - thanks in advance.

alepouna:
QA Report

Discord: alepouna
Object of testing: #8581
Tier of Testing: 1
Date: 26/6/2024

Testing Process:
- Test all possible combinations of presets and presets sequences.
- Test expedited or normal speed - even mix or change while loading
- Regression Test and check normal operation without using presets and ensure this has not regressed in any way.
- [FAIL] presets must load without hanging (hanging means the xml config is not correct)
- [FAIL] any intermittent error and alarms that comes up goes away after a short time without manual interaction
- no errors after a preset is loaded
- no unusual state of the aircraft after a preset is loaded - example flaps/slats not retracted/extended

Testing Results:
Not Passed

Negatives / Concerns: 
1. When going from C&D to R-T-T, if you toggle the speed from slow to fast when the APU fire tests, the test pushbutton seems to remain pressed and thus the APU fire alarm does not go away. 
image
2. There seems to be a delay ('Waiting') for a few moments after the `Await FWC Init` step (about 1-2 seconds).
3. There is an about 4 second waiting time for the cockpit door to close, ironically locking the completion process.
4. In any procedure that ends with an engine start, there is always a 2 second delay `Await ENG 2 Avail`, not a ENG 1 one though.
5. There is a 1 sec delay waiting for `Autobrake Max`
6. There is a 1 sec delay for `Await Flaps 1+F` 
7. There is a 1 sec delay for `Await Slats 1+F`
8. There is a 1 sec delay for `Await Flaps 0` 
9. There is a 1 sec delay for `Await Slats 0`
10. There is a 2.5 sec delay for `ENG N1 <3%` 

Notes: 
Consider showing a popup notification when a state changes, since now they can happen extremely quickly 😉

frankkopp:
>     1. When going from C&D to R-T-T, if you toggle the speed from slow to fast when the APU fire tests, the test pushbutton seems to remain pressed and thus the APU fire alarm does not go away.

this might indeed be a small bug in the config (XML) - didn't account for a change in the middle of the process there - this is an easy fix
 
>     2. There seems to be a delay ('Waiting') for a few moments after the `Await FWC Init` step (about 1-2 seconds).
>     3. There is an about 4 second waiting time for the cockpit door to close, ironically locking the completion process.
>     4. In any procedure that ends with an engine start, there is always a 2 second delay `Await ENG 2 Avail`, not a ENG 1 one though.
>     5. There is a 1 sec delay waiting for `Autobrake Max`
>     6. There is a 1 sec delay for `Await Flaps 1+F`
>     7. There is a 1 sec delay for `Await Slats 1+F`
>     8. There is a 1 sec delay for `Await Flaps 0`
>     9. There is a 1 sec delay for `Await Slats 0`
>     10. There is a 2.5 sec delay for `ENG N1 <3%`
 
These are all intentional - partly because our systems can't really handle "instant" because of dependencies and algorithm (they have not been designed that way) and I had a lot of issues without some extra pauses. For Flaps and ENG I have accelerated these a lot and the check also needs a short delay. This could be shortened a little but should not be removed completely. 
Looking at other aircraft though I think the overall time to set a preset is really good - some other aircraft take much longer. 

> Notes: A. If the above negatives/concerns are expected or do not require any work, consider this PR as passed. B. Did not notice the behavior 2hwk previously reported C. Consider showing a popup notification when a state changes, since now they can happen extremely quickly 😉

I will fix 1, will look at the others again but most likely won't change much. I'll let you know then you can decide if you want to retest.

Aleloplt:

Discord: Alelo
Object of testing: #8581 
Tier of Testing: 1
Date:  26/07/24

Testing Process: 
Testing as per the instructions. Done multiple changes to panel state on normal and accelerated time. 
Did a flight between EGKK/EGCC and no secondary issues were found

Negatives:       
NA

Testing Results: 
Passed

Conclusions: 
Couple of things to note based on company SOPT I saw but that could change per airline
**Powered state:** 
All lights and switches seem to have been done as per the cockpit flow. If this is the case fuel pumps in the wing tanks should probably be switched on as well.
Evac switch to CAPT only 
Pack 2 off
**Ready for push**
XPNDR in auto on the before the push is how we do it. Not when you are ready to taxi.

mattking247:
Non-QA Team Report - Test was requested to be published by @alepouna

Discord          : Matt
Object of testing: #8581
Tier of Testing  : 1
Date             : (30/07/2024)

Testing Process: 
Aircraft located on the ramp in various stages of readiness. 

Negatives: Overall, test failed. 

NORMAL MODE: Test passed
Testing in normal mode powered the aircraft up at each step in a time you would expect. IRS alignment etc generally followed whatever was set up as the option in the EFB for instant v realistic. 

FAST MODE: Test failed
During fast loading, all times on startup were accelerated. However, this still relies on what is selected in the EFB as alignment preferences. I think if someone selects Fast mode for Powered to Ready for Taxi loading, chances are the want instant alignment of the IRS also. In fast loading for Cold and Dark to Powered, the APU fails to start. It starts correctly when loaded in normal speed, however in fast mode, the APU master button is pressed, but the start button is not pressed, meaning the whole process stalls until the APU is manually started. The next issue when selecting Fast start is that when going from Ready for Pushback to Ready for Taxi, the engine start time is still a normal engine start time. This is fully acceptable when someone selects a normal startup time. However, in fast startup, could this be sped up? 

Testing Results: Not Passed

Conclusions: 
Test failed for sequence and timeliness of some processes in Fast Mode. This included the APU failing to start each time in Fast Mode, the engine start sequence taking the usual amount of time to start and not being sped up and IRS alignment timings blindly following what is selected in the EFB. 

Recommendations…

1.    During Fast Starts, IRS alignment option in EFB automatically moves to instant.
2.    During Fast Starts, APU starter button needs to be engaged (not currently working) and the APU start sequence sped up. 
3.    During Fast Starts, Engine start sequence is sped up significantly.

frankkopp:
> FAST MODE: Test failed During fast loading, all times on startup were accelerated. **However, this still relies on what is selected in the EFB as alignment preferences.** I think if someone selects Fast mode for Powered to Ready for Taxi loading, chances are the want instant alignment of the IRS also. **In fast loading for Cold and Dark to Powered, the APU fails to start.** It starts correctly when loaded in normal speed, however in fast mode, the APU master button is pressed, but the start button is not pressed, meaning the whole process stalls until the APU is manually started. The next issue when selecting Fast start is that when going from Ready for Pushback to Ready for Taxi, the engine start time is still a normal engine start time. This is fully acceptable when someone selects a normal startup time. However, in fast startup, could this be sped up?
 
None of this could be reproduced - I re-tested this again - the fast loading is independent from the EFB setting as it actually completely ignores this setting and does the fast mode within the systems code itself (ADIRS/APU/ENG).
APU startup works every time for me. 
I would need need additional instruction to reproduce this as with the ones given in this report everything works as intended on my side. 

Please retest without anything else in Community and make sure to have the latest build of this PR.

tylerclt:

Discord Username: Tylerb780
Object of testing : #8581  
Aircraft: A32NX
Tier of Testing: 2
Date: 13/09/2024 

Testing Process: 
1. Loaded into SKBO with the A32NX
2. Tested all presets with normal loading 
3. Tested all presets with fast loading
4. Tested with switching between fast and normal loading to try and break it.
5. Checked to make sure nothing was broken
6. Set the aircraft up for a flight from SKBO-SKBG
7. Checked for any regressions/abnormalities with the the ADIRS, APU, FADEC, flaps/slats 


Testing Results: 
Passed  

Conclusions: 
Works as intended!

Media:

PR #8583, opened 2024-03-25
fix(reverser): Fix wrong MSFS plane acceleration seen from systems wasm

crocket63:
Fixes #[issue_no]

## Summary of Changes

Reverser thrust uses a hack so we have control over the amount of thrust relative to N1. However this was causing inconsistency between msfs plane acceleration and real simulation acceleration, causing auto brake wrong behaviour.

Now more reverser during auto brake will cause less brake force as expected.

## Additional context

## Testing instructions

Use auto brakes in various modes. Check in all modes that using reversers will cause less brake input from the auto brake low and med modes. Max should be unaffected (less braking distance but same brake force)


Maximilian-Reuter (approved):

noonebling:

Discord          : noonebling
Object of testing: 8583
Tier of Testing  : 1
Date             : 26/03/2024

Testing Process: 
1. Start up aircraft
2. Take off with 0 payload and 4000kg of fuel
3. Autoland with no reverses with low and medium brake
4. Autoland with reverses with low and medium brake
5. Repeat with High brake


Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
Works as expected

2hwk (approved):

donstim:
Stopping distance tests before this PR showed that stopping distance at medium autobrakes on a dry runway were significantly shorter when using reverse thrust than without using reverse thrust. These stopping distances should be approximately the same since the medium autobrake deceleration level can be achieved on a dry runway without using reverse thrust.

This PR fixes that issue, and the stopping distance using medium autobrakes are the same with or without using reverse thrust.

donstim:
I should have mentioned that the way in this is implemented, by essentially removing the acceleration MSFS sees from the "hacked" reverse thrust, it does lead to an issue for wet and slippery runways where the autobrake level cannot maintain its deceleration target without reverse thrust. For these cases, the effect of using reverse thrust is similar to the cases where the decel target can be achieved, such that reverse thrust does not reduce the stopping distance. The issue is that in these cases, it should reduce the stopping distance, at least to the point where the autobrake decel target level is achieved.

For the medium autobrake setting, this happens on runway condition code 5 (good) or worse, which is normally associated with a wet runway. For the low autobrake setting, it would occur on runway conditions code 4 (good to medium) or worse.

PR #8530, opened 2024-02-27
fix(A380): Fixed ADIRS alignment time flyPad setting

frankkopp:
## Summary of Changes
The flyPads ADIRS alignment time settings was broken as the persistent config was not synced to the Lvar any more (it was in the A320 code but not the A380 code). 
This PR puts this sync into the flyPad sync.ts where it works for the A380X as well. 

Discord username (if different from GitHub): cdr_maverick

## Testing instructions
A32NX: 
Check if flyPad settings alignment times are still respected. (Regression test)
A380X:
Check if flyPad settings alignment times are respected.

## How to download the PR for QA

Every new commit to this PR will cause a new A32NX artifact to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, click on the bottom **PR** tab
1. Click on the **A32NX** download link at the bottom of the page

PR #8602, opened 2024-04-01
fix: boarding rate and align time defaults

tracernz:
Fixes #[issue_no]

## Summary of Changes

Fixes the default values for boarding rate and align time. Fixes a console error that resulted from the invalid defaults. There is a changelog for boarding rate as it was broken in 0.11, but the align time was only broken in https://github.com/flybywiresim/aircraft/pull/8530 which hasn't been in a stable release.

## Additional context

## Testing instructions

None.

### DIFF
--- a/fbw-common/src/systems/instruments/src/EFB/Settings/sync.ts
+++ b/fbw-common/src/systems/instruments/src/EFB/Settings/sync.ts
@@ -50,12 +50,12 @@ const settingsToSync: Map<string, SimVar> = new Map([
 
 const settingEnumToSync: Map<string, SimVarEnum> = new Map([
     ['CONFIG_BOARDING_RATE',
-        ['L:A32NX_BOARDING_RATE', 'number', '0',
+        ['L:A32NX_BOARDING_RATE', 'number', 'REAL',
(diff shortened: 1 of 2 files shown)


Gurgel100 (approved):

fbw-common/src/systems/instruments/src/EFB/Settings/sync.ts
```diff
@@ -50,12 +50,12 @@ const settingsToSync: Map<string, SimVar> = new Map([
 
 const settingEnumToSync: Map<string, SimVarEnum> = new Map([
     ['CONFIG_BOARDING_RATE',
-        ['L:A32NX_BOARDING_RATE', 'number', '0',
+        ['L:A32NX_BOARDING_RATE', 'number', 'REAL',
```
2hwk, line 53: This does look pretty odd (`number`, `'REAL')  but maybe a subject for a refactor PR not a fix one.

2hwk (approved):

fbw-common/src/systems/instruments/src/EFB/Settings/sync.ts
tracernz: I think you designed it? :D

PR #8606, opened 2024-04-03
feat(nws): add better NWS hacks

lukecologne:
## Summary of Changes

This PR fixes two things related to the NWS:
* Enables the sim NWS to move when standing still (was previously only possible visually, but not functionally)
* Enhance the previous hack required to cancel out the rudder effect on the NWS, by limiting the NWS amplitude to the commanded value. This removes jittering that happens when the rudder is moved.

NOTE: There is still the issue that when giving opposite rudder and tiller orders, the sim's NWS command due to the rudder deflection can only be cancelled, but the NWS doesn't steer in the tiller direction as it should. AFAIK. there is nothing we can do about that.

## Additional context

## Testing instructions

Confirm that
* There is no more slight oscillation in the taxi path when using the rudder (very noticeable when using pedal disc, but also happens when using the rudder pedals to command the NWS)
* The NWS turns at a standstill (not only visually, which it did before, but also functionally)
* The NWS otherwise works correctly (except the note above)

This was changed both on the 380 and the 320.


2hwk (approved):

2hwk:
No changelog.md required, covered by line 49 from 0.11.0 -> 0.12.0.

aguther:
There is no doubt that this PR is a good improvement, and the situation is better after the merge than before.

For the sake of completeness, however, it should be said that it does not completely solve the problems. There is still a certain amount of jerkiness when changing the rudder deflection from left to right or vice versa. This is particularly noticeable when using VR.

noonebling:

Discord          : noonebling
Object of testing: 8606
Tier of Testing  : 1
Date             : 07/04/2024

Testing Process: 
1. Started flight with default joystick setup
2. Tested NWS at a standstill to check the functional
3. No more slight oscillation in the taxi path

Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
Works as expected

PR #8619, opened 2024-04-11
fix: Fixed FLX-CLB thrust limit transition

frankkopp:
## Summary of Changes
Fixes the FLX-CLB transition which was happening instantly but should be very gradual. 

## Testing instructions
Check if FLX-CLB transition after a FLX takeoff when reducing to the CLB detent is slow and gradual - first there is a several seconds phase where FLX and CLB limit are still identical then CLB limit raises slowly.


tracernz (approved):

frankkopp:
Covered by #8622

PR #8641, opened 2024-05-01
feat(efb): Simrate and Pause at T/D in EFB Quick Settings

2hwk:
Fixes #[issue_no]

## Summary of Changes

* Adds Pause At Top of Descent toggle to quick settings
* Adds Simulation Rate Controller to quick settings

## Screenshots (if necessary)

image
FlightSimulator_hHGan5IoVq

## Additional context

## Testing instructions

- Check flyPad Settings > Realism - Check that MCDU Input and Pause at TOD Settings options still work as per normal and that there is no regression
- Quick Settings -> New options: Pause at TOD and Simrate Controller
   - Ensure all localisation screens and hover text work
   - Ensure visually css is not broken (including with other localisation i.e. RU/PL
   - Check that the toggle for pause on TOD works and that this syncs with what is in the Settings > Realism page.
   - Check that the new simrate indicator in the status bar shows the simrate accurately
   - Check that you can increase and decrease the simrate (until limited by our built-in simrate reduction, which should be around 4x for most people)
      - Note that the automatic simrate reduction feature will still affect this, so if simrate is reduced because of low frame rate or performance issues, this is as designed and expected. 

Pause on TOD Tests:
   - Expected Behavior/Notes: 
      - Button Status/Color: When you first activate this in the Quick Controls the button's background will be *orange* (and say Standby) 
      - **NOTE: Critical MSFS Sim Bug possible if you ALT-TAB while panning the camera with the mouse look option (MMD by default). Do NOT Alt-TAB while panning the camera. This will make the pop-up unable to be dismissed and you will have to restart.** - As this is already a bug on master this is not considered a regression.
      - Armed State (Green): This will only update to Armed (green background), when all the pre-conditions for Pause on TOD are met.
         - A valid flight plan with a valid T/D must be first calculated (i.e. fill in INIT-B, PERF pages on MCDU).
         - Plane must be in the air - i.e. FMGC flight phase is on either CLIMB or CRUISE.
         - Distance to T/D *is equal to or less than* the configured T/D pause distance (default 10nm)
         - T/D point has not already been reached.
      - A/P guarding - If the A/P is disconnected ABOVE the inputted transition flight level (default 10000 FT if not specified) the sim will pause and for *10* seconds, pause on T/D will be on standby. **This 10s cooldown will be reflected in the Quick Setting button's status.**
        - Why A/P guarding: Necessary if A/P is disconnected to pause to prevent plane from going off course. 
        - 10s cooldown: This is to prevent repeated activation after unpause that interfere with the user's ability to fly and re-engage A/P as desired. 
        - Quirk: This A/P guarding logic i.e. above the transition flight level logic is hidden to the user and is not displayed in the button's status. 
          - In the future this may be displayed somehow, but as the two logic gates are independent of each other, this is just for tester's awareness that this exists.
        - After A/P guarding pause, for *10* seconds, Pause on T/D will not be armed. This will be reflected in the button's background colour. 
        - Pause due to A/P guarding can be repeated, unlike the pause that occurs when T/D is reached. 
      - Important: After T/D is reached and the sim paused. T/D Pause will activate and is then both T/D pause and A/P guarding will be *disabled* for the rest of the flight, the quick settings button returning to a Standby/orange state until the next turnaround, based on FMGC status (READY/DONE). This means engines must be shutdown and cockpit returned to a turnaround state on the ground after landing.
   - Do a short flight (with init-b/perf pages filled in). Ensure Pause at TOD feature works as per normal after arming it in quick settings menu (green top border visible)
   - Turnaround -> After completing the full proper turnaround procedure, and after departure with another complete flight plan, check that T/D pause is now fully re-enabled and that the Armed (green) state can be entered again, and that pause on T/D works on this turnaround flight.
   - Retry -> Change distance (Settings > Realism - Pause At TOD DIstance) and ensure that pause on T/D correctly pauses at this distance.


2hwk:
Still needs localisation keys, to be added before merging

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/descent/TodGuidance.ts
```diff
@@ -44,61 +47,68 @@ export class TodGuidance {
...
+    this.tdArmed =
+      this.cooldown <= 0 &&
+      !this.tdPaused &&
+      this.observer.get().flightPhase >= FmgcFlightPhase.Climb &&
+      this.observer.get().flightPhase <= FmgcFlightPhase.Cruise &&
+      Simplane.getAutoPilotAirspeedManaged();
+
+    if (SimVar.GetSimVarValue('L:A32NX_PAUSE_AT_TOD_ARMED', 'bool') !== this.tdArmed)
+      SimVar.SetSimVarValue('L:A32NX_PAUSE_AT_TOD_ARMED', 'bool', this.tdArmed);
+
```
Saschl, line 66: You could do this without reading the simvar, save the previous state of this.tdArmed before updating it and only set the simvar if it has changed. Minor optimization but should also be quick to do.

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/descent/TodGuidance.ts
```diff
@@ -44,61 +47,68 @@ export class TodGuidance {
...
+    if (this.tdArmed) {
+      // Check T/D pause first
       if (
-        !this.tdPaused &&
-        this.observer.get().flightPhase >= FmgcFlightPhase.Climb &&
-        this.observer.get().flightPhase <= FmgcFlightPhase.Cruise &&
-        Simplane.getAutoPilotAirspeedManaged()
+        (this.aircraftToDescentProfileRelation.distanceToTopOfDescent() ?? Number.POSITIVE_INFINITY) <
+        parseFloat(NXDataStore.get('PAUSE_AT_TOD_DISTANCE', '10'))
```
Saschl, line 71: Would be good to use `NXDataStore.getAndSubscribe(...)` (e.g. in the constructor) and set the values in a member variable and use that in the loop. I think NXDataStore.get is relatively slow and the value does not change often

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/descent/TodGuidance.ts
2hwk: Oh I must have forgotten to upstream the commit that changed this. I'm using LocalSimVar that essentially does this (in an abstracted function).

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/descent/TodGuidance.ts
2hwk: > Would be good to use `NXDataStore.getAndSubscribe(...)` (e.g. in the constructor) and set the values in a member variable and use that in the loop. I think NXDataStore.get is relatively slow and the value does not change often

Unfortunately I tried this once and it introduced a bug that somehow broke the entire functionality. I could try again at some point but as this part has essentially been in the codebase for months without causing significant observed performance degradation, I won't push this for now.

Probably would be better to write a new utility for this that could be abstract enough to be used everywhere, then replace all such NXDataStore operations in all js/ts.

2hwk: The change in this PR merely rearranges the logic slightly. Firstly to decouple the logic gating of the backoff timer from the NXDataStore check, such that this function ``updateTdPause`` is not called if ``PAUSE_AT_TOD`` is disabled. 

This enables this ``tdArmed`` state to be set (regardless of the status of the backoff timer)

package.json
```diff
@@ -7,23 +7,18 @@
   },
   "scripts": {
     "====== Generic===============": "==========================================",
-
-    "build:cpp-wasm-cmake":             "scripts/build-cmake.sh",
-    "build:cpp-wasm-cmake-clean":       "scripts/build-cmake.sh --clean",
-    "build:cpp-wasm-cmake-debug":       "scripts/build-cmake.sh --debug",
+    "build:cpp-wasm-cmake": "scripts/build-cmake.sh",
```
Saschl, line 10: We should decide whether we want to have the package.json formatter as I believe the line breaks were added intentionally (I noticed VSCode is happily formatting this file that's what I think happened here). Let's check on discord

Saschl (approved):
LGTM, let's just check package.json formatting before merging, but not relevant for testing.

2hwk:
> LGTM, let's just check package.json formatting before merging, but not relevant for testing.

Will fix

ExampleWasTaken:
I think the simrate box needs UI improvements. At the moment it doesn't align with anything. The clock is neither in the middle of the box nor below the gap between the simbridge and keyboard button.

I suggest centering the clock and displaying the next simrate in the buttons instead of the chevrons.

So if you're at normal simrate, the "previous" button (i.e. decrease) button is disabled. At 4x simrate the left button reads `2x` and the right button reads `8x`.

The current simrate could be displayed as part of the box title, meaning below the clock. E.g. `4x Simrate`

2hwk:
UI improvements will probably need to be in another PR at this stage, I don't anticipate having the bandwidth to adjust it further in the next 2 months. Potentially a good first issue for new contributors.

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/descent/TodGuidance.ts
```diff
@@ -44,61 +62,65 @@ export class TodGuidance {
...
+    this.tdArmed.setVar(
+      this.cooldown <= 0 &&
         !this.tdPaused &&
         this.observer.get().flightPhase >= FmgcFlightPhase.Climb &&
         this.observer.get().flightPhase <= FmgcFlightPhase.Cruise &&
-        Simplane.getAutoPilotAirspeedManaged()
+        Simplane.getAutoPilotAirspeedManaged(),
+    );
+
+    if (this.tdArmed) {
```
Saschl, line 80: Just a comment as I just tested this PR: I think it should be `if (this.tdArmed.getVar()) {` as I got spammed with Dialogs and the current condition is always true (as this.tdArmed is not undefined)

Saschl: I just pushed it as the change was minimal, feel free to revert if you don't like it 👍

alepouna:
> Ensure visually css is not broken (including with other localisation i.e. RU/PL

paniks in greek 
image

noonebling:

Discord          : noonebling
Object of testing: 8641
Tier of Testing  : 1
Date             : 06/07/2024

Testing Process: 
Booted up flight
Checked that MCDU input and Pause at TOD Settings options still work as per normal and there is no regression
Ensured all localisation screens and hover text works, ensured visually css is not broken
Checked that the toggle for pasue on TOD works that this syncs with what is in the Settings > Realism page
Checked that the new simrate indicator in the status bar shows the simrate accurately. 

Pause on TOD Tests:
Checked for the status button to be orange before the MCDU is filled in
Checked for the status to turn green when the MCDU is filled in
Checked for A/P Guarding 
Checked for Turnaround instructions

Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
Works as expected

2hwk:
T/D bugs out when:
- Keyboard/Mouse controls are used
- Toggle Free Look (Using middle mouse button by default) is used
- While Toggle Free Look is enabled, Keyboard Alt+TAB is used to background the sim
- The flight continues and T/D is reached.
- The pop-up will appear but be unable to be dismissed.

The only known workaround is to delete the entire pause menu, which will allow the flight to continue but you won't be able to exit to the menu and thus is a very unsafe practice.

alepouna:
Reminder: Make an issue once merged regarding CSS issue w/ long translations

2hwk:
Still need to update images + text in https://github.com/flybywiresim/docs/pull/981

2hwk:
Docs ready

PR #8648, opened 2024-05-06
fix(press): delta pressure calculation before adirs on

mjuhe:
## Summary of Changes

This PR fixes an issue with the pressurization system. When spawning at an airport high above sea level the delta pressure was not being calculated correctly and the safety valved opened (triggering the corresponding ECAM warning).

The pressurization system uses the outside static pressure from the ADIRS to calculate delta pressure. Before the system is powered, the delta P cannot be computed. Now before ADIRS power up, the ECAM system page shows "XX" for delta pressure, no warning comes on, and the safety valve does not open.

## Screenshots (if necessary)

image

## Additional context

Discord username (if different from GitHub): Commander Ga

## Testing instructions

Spawn at an airport high above see level (KDEN for example). Before powering up the ADIRS, ensure the delta pressure is not shown on the ECAM (it shows "XX" as above). Observe the safety valve is closed and no warrning comes on.

When setting the ADIRS to align, after a few seconds the delta pressure indication comes on, and is 0.


fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
```diff
@@ -255,17 +260,20 @@ impl<C: PressurizationConstants> CabinPressureController<C> {
         let (_, adirs_ambient_pressure) = self.adirs_values_calculation(adirs);
         let new_exterior_altitude: Length;
 
+        // If the ADIRUS are not aligned we take the ambient pressure from context
```
tracernz, line 263: It's nothing to do with IR alignment, as the data comes from the ADR.

fbw-common/src/wasm/systems/systems/src/air_conditioning/cabin_pressure_controller.rs
```diff
@@ -255,17 +260,20 @@ impl<C: PressurizationConstants> CabinPressureController<C> {
...
+            self.exterior_pressure
+                .reset(adirs_ambient_pressure.unwrap_or(context.ambient_pressure()));
             new_exterior_altitude =
                 self.calculate_altitude(self.exterior_pressure.output(), self.reference_pressure);
             self.is_initialised = true;
         } else {
             self.exterior_pressure.update(
                 context.delta(),
-                adirs_ambient_pressure.unwrap_or_else(|| Pressure::new::<hectopascal>(Self::P_0)),
+                adirs_ambient_pressure.unwrap_or(context.ambient_pressure()),
```
lukecologne, line 276: How does the CPC get the exterior pressure? Does it have an independent pressure sensor? If not, it kind of defeats the point of using the ADR data output if we just end up magically getting the pressure anyway

mjuhe: I added a comment to explain this in the code :) The CPC does not have an independent pressure sensor, we get the ambient pressure from context before ADIRS power up because:
1. The safety valve is a physical, mechanical valve that opens based on pressure differential, this pressure is not a "signal" but a physical pressure difference between the inside and the outside of the aircraft.
2. I have a low pass filter on the outside pressure to avoid delta P being too jerky when flying. Because of this, initialising external pressure to 1013 might result in you seeing delta P move when turning on the ADIRS, which you wouldn't in the real aircraft.

As the comment states, when the ADIRS are not powered the delta P signal has a No Ccomputed Data SSM, so any users of this value would not be able to get this info outside of the CPC.

lukecologne: Hm, but why is this then the physical ambient pressure one time, and the ADR reported pressure the other time? That doesn't really make sense. If it's a physical quantity, it should always be the physical quantity. If it's data reported by the ADR, it should always be that (or a default if not valid). Maybe it would make sense to extract the safety valve into it's own component that always uses the physical quantity, and the CPC can then correctly always just use ADR data.

mjuhe: Yeah I see what you mean and it's a good point, the safety valve should never use the ADR data. I'll change this. 👍

mjuhe: As you suggested, I've separated the safety valve "signal" away from the CPC (and I've also done the same with the negative relief valve signal of the A380).

The CPC now only "sees" the ambient pressure from the ADIRS.

Thanks again for the suggestion!

fbw-a32nx/src/systems/instruments/src/SD/Pages/Press/Press.tsx
```diff
@@ -54,6 +54,7 @@ export const PressPage: FC = () => {
   }, [cabinAlt]);
 
   const deltaPress = splitDecimals(MathUtils.clamp(deltaPsi, -9.9, 9.9));
+  const deltaPressNotAvail = arincDeltaPsi.isNoComputedData();
```
lukecologne, line 57: Normally the DMC should use the cabin pressure subtracted by the corrected static pressure as received directly from the ADR, as a backup if the delta pressure field is not available. I'm not entirely sure which failure case this is supposed to cover, as (outside of an isolated ARINC transceiver failure on the CPCs or something) the delta pressure field should be available when both the CPC and the ADR is functioning normally. Maybe something to add as a comment, but not critical

lukecologne (approved):
Except the note I left above, looks to be good

LunakisDev:

Discord: LunakisLeaks
Object of testing: #8648
Tier of Testing: 1
Date: 18/5/2024

Testing Process: 
1) Spawned at KDEN
2) Started the aircraft 
3) Checked if the delta pressure shows "XX"
4) Observed that the safety valve is closed and no warning comes on
5) ADIRS to align
6) Checked if the delta pressure indication comes on, and is 0

Negatives:
N/A

Testing Results: 
Passed

Conclusions: 
LGTM

fbw-a32nx/src/systems/instruments/src/SD/Pages/Press/Press.tsx
mjuhe: Yeah I see what you mean. So the "XX" would only appear when the data from the SDAC is not valid (which, in this case, would be when the CPC is not transmitting delta pressure AND the ADR is not transmitting static ambient pressure).

Since the SDAC is not modelled I think this logic is a good placeholder for now as the behaviour is correct, but I've added a comment as suggested. Thanks!

PR #8651, opened 2024-05-10
fix(a380x): Correct flap/slats positions

flogross89:
**A380X only, no QA needed.**

## Summary of Changes
This should be an improvement to the current SFCC implementation for the A380X.

In scope: 
- [x] Actuate to correct flap angles, modify FPPU angles
- [x] Actuate to correct slats angles, modify FPPU angles
- [x] Modify PFD lower part flaps/slats display to match new FPPU angles
- [x] Update Automatic extension system (AES)
- [x] Update Automatic retraction system (ARS)
- [x] Introduce new config 2*
- [x] Implement Flap load relief system (FLRS) + tests
- [x] Slat alpha/speed lock function + tests
- [x] Slats/flaps cruise baulk function + tests

Out of scope: 
- Aileron droop (should remain the same as the A32NX)
- 2 out of 8 slats are droop noses
- Electric actuation of slats via SLAT SYS 1 (via AC ESS, only if G HYD pressurized)
- Power-off brake
- Wing tip brakes
- Asymmetry position pick-off units (APPU)
- Flap disconnect sensor
- Slat/Flap fault for HYD faults + tests

## Additional context
No appropriate references for A380 FPPU, hence it's assumed the same FPPU as in the A320 is used. 
Resulting FPPU angles:
| CONF | Flap surface angle | Flap FPPU angle | Slat surface angle | Slat FPPU angle |
|------|--------------------|-----------------|--------------------|-----------------|
| 0    | 0                  | 0               | 0                  | 0               |
| 1    | 0                  | 0               | 20                 | 247.27          |
| 1+F  | 8                  | 108.28          | 20                 | 247.27          |
| 2    | 17                 | 154.65          | 20                 | 247.27          |
| 2*    | 17                 | 154.65          | 23                 | 284.65          |
| 3    | 26                 | 194.03          | 23                 | 284.65          |
| FULL | 32                 | 218.91          | 23                 | 284.65          |


fbw-a32nx/docs/a320-simvars.md
```diff
@@ -3280,27 +3280,27 @@ In the variables below, {number} should be replaced with one item in the set: {
...
+      | 20  | Flaps >= 14° (254° > FPPU > 140.7)       | Flaps >= 7° (254° > FPPU > 102.1°)       |
+      | 21  | Flaps >= 19° (254° > FPPU > 163.7°)      | Flaps >= 16° (254° > FPPU > 150.0°)      |
+      | 22  | Flaps >= 39° (254° > FPPU > 247.8°)      | Flaps >= 25° (254° > FPPU > 189.8°)      |
+      | 23  | Flaps Extended 40° (254° > FPPU > 250°)  | Flaps Extended 32° (254° > FPPU > 218°)  |
+      | 24  | Flap WTB engaged                         |                                          |
+      | 25  | Flap Fault                               |                                          |
+      | 26  | Spoiler Lift Demand                      |                                          |
+      | 27  | Spoiler Limit Demand                     |                                          |
+      | 28  | Slat System Jam                          |                                          |
+      | 29  | Flap System Jam                          |                                          |
```
tracernz, line 3303: This should go in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/docs/a380-simvars.md instead.

fbw-common/src/wasm/systems/systems/src/simulation/update_context.rs
```diff
@@ -306,6 +308,7 @@ impl UpdateContext {
     pub(crate) const LOCAL_VERTICAL_SPEED_KEY: &'static str = "VELOCITY BODY Y";
     pub(crate) const ALT_ABOVE_GROUND_KEY: &'static str = "PLANE ALT ABOVE GROUND";
     pub(crate) const LATITUDE_KEY: &'static str = "PLANE LATITUDE";
+    pub(crate) const ANGLE_OF_ATTACK_KEY: &'static str = "ANGLE OF ATTACK INDICATOR";
```
tracernz, line 311: We should not use the "INDICATOR" simvars at all, they are meant for cockpit gauges. This one is definitely not the actual AOA, but instead AOA as a ratio of stall AOA for one of these type of indicators:
garmin-aoa-small

Any aircraft systems that consume AOA almost definitely consume the ADR AOA words from their ARINC input bus anyway. No other system can read the AOA vanes directly.

tracernz (changes requested):

fbw-common/src/wasm/systems/systems/src/simulation/update_context.rs
flogross89: I'm now reading in the ADR AOA value

flogross89:
@tracernz Reworked some bits, your comments should now be addressed

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
```diff
@@ -210,7 +211,7 @@ impl SlatFlapControlComputer {
                 if context.is_in_flight()
                     && (context.indicated_airspeed().get::<knot>()
                         < Self::ALPHA_SPEED_LOCK_IN_AIRSPEED_THRESHOLD_KNOTS
-                        || context.angle_of_attack().get::<degree>()
+                        || adirs_output.angle_of_attack(1).value().get::<degree>()
```
tracernz, line 214: You should check the SSM before using the value. Should it just blindly always use ADR1, or should there be some selection logic to choose an ADR based on validity of the data/SSM?

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
```diff
@@ -695,6 +711,68 @@ mod tests {
...
+                    Angle::new::<degree>(0.0),
+                    SignStatus::NormalOperation,
+                ),
+            }
+        }
+
+        fn set_angle_of_attack(&mut self, v: Angle) {
+            self.angle_of_attack = Arinc429Word::new(v, SignStatus::NormalOperation);
+        }
+    }
```
tracernz, line 742: This feels very weird. Don't we already have the stuff to mock the ADR in the test bed? @Gurgel100 will know better than me.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
flogross89: Good point, forgot about the SSM, didn't know how these are handled in the rust libs. 
The selection logic for ADR data should be the same everywhere: 
Median value if alle three operational,
mean if only two operational
don't use it of only one operational

But I haven't seen this selection logic implemented somewhere, and I don't think it makes sense to handle this on every consumer's side (nor do I have the Rust skills to implement a selector). I'll maybe go with the approach seen in air cond, which is chose the first one with valid SSM

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
flogross89: I looked everywhere, could only find similar boilerplate code everywhere, but no re-usable test bed. Don't know how it's done in the Rust world

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
tracernz: > The selection logic for ADR data should be the same everywhere:

No, many systems have different logic for specific reasons.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
tracernz: Hmm, we have a comprehensive ADIRU test bed but it seems it's not public. I withdraw the comment here as out of scope, but I think there's an opportunity to avoid having many copies of the test bed setup code in future.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
flogross89: just pushed the agreed selection logic (pick first one with valid data)

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
```diff
@@ -159,24 +196,122 @@ impl SlatFlapControlComputer {
...
+        flaps_handle.position() == 0
+            && (self.flaps_conf == FlapsConf::Conf1 || self.flaps_conf == FlapsConf::Conf1F)
+    }
+
     fn surface_movement_required(demanded_angle: Angle, feedback_angle: Angle) -> bool {
         (demanded_angle - feedback_angle).get::<degree>().abs() > Self::EQUAL_ANGLE_DELTA_DEGREE
     }
 
+    fn angle_of_attack(&self, adirs: &impl AdirsMeasurementOutputs) -> Option<Angle> {
+        [1, 2, 3]
```
Gurgel100, line 297: Maybe add a `TODO` comment here to indicate that this is not the correct logic yet.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
```diff
@@ -568,6 +723,68 @@ mod tests {
...
+        fn new() -> Self {
+            Self {
+                is_aligned: false,
+                latitude: Arinc429Word::new(Angle::default(), SignStatus::FailureWarning),
+                longitude: Arinc429Word::new(Angle::default(), SignStatus::FailureWarning),
+                heading: Arinc429Word::new(Angle::default(), SignStatus::FailureWarning),
+                vertical_speed: Arinc429Word::new(Velocity::default(), SignStatus::FailureWarning),
+                altitude: Arinc429Word::new(Length::default(), SignStatus::FailureWarning),
+                angle_of_attack: Arinc429Word::new(
+                    Angle::new::<degree>(0.0),
```
Gurgel100, line 745:
```suggestion
                    Angle::default(),
```

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
flogross89: Thanks! Done

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
flogross89: Thanks! Done

Gurgel100 (approved):
Great work.

PR #8714, opened 2024-06-30
feat(a380x): PFD THS indicator

flogross89:
**A380X only, no QA needed**

## Summary of Changes
Implements the THS indicator found in the lower right corner of the PFD.
Active during preflight, take-off and after landing. 

Also adds the spoiler indicator in the lower left of the PFD. First iteration, not too accurate. 

## Screenshots (if necessary)
image
image

## References
https://www.youtube.com/watch?v=AI8yFhuEa4Q&t=3632s
FCOM
image
image

## Additional context


fbw-a380x/src/systems/instruments/src/PFD/PFD.tsx
```diff
@@ -57,6 +61,15 @@ export class PFDComponent extends DisplayComponent<PFDProps> {
...
   private failuresConsumer;
 
+  private readonly fwcFlightPhase = ConsumerSubject.create(this.sub.on('fwcFlightPhase').whenChanged(), 0);
+
+  private readonly groundSpeed = ConsumerSubject.create(this.sub.on('groundSpeed').whenChanged(), 0);
+
+  private readonly raHeight = ConsumerSubject.create(this.sub.on('chosenRa').whenChanged(), Arinc429Word.empty());
+
+  // FIXME add touch&go condition
+  private pitchTrimIndicatorVisible = MappedSubject.create(([flightPhase, gs, raHeight]) => (flightPhase >= 9 && gs < 30) || (flightPhase < 6 && raHeight.valueOr(0) < 50), this.fwcFlightPhase, this.groundSpeed, this.raHeight);
```
lukecologne, line 71: If I had to guess, I'd say that this probably doesn't use the FWC flight phase IRL. More likely I think it's a simple flip-flop with the RA height and the GS. Of course this is still just guessing, but experience shows that the flight phase is rarely used outside of the FWS itself

beheh:
Just from the photo I think the left and right padding of the THS % need to be reduced slightly. The left digit and % char are almost touching the border.

fbw-a380x/src/systems/instruments/src/PFD/PFD.tsx
flogross89: But the flip-flop would then reset, if I were to re-load the instrument in flight, right? 
I think for maintainability using the FWC flight phase would have some advantages.

We would anyway have to improve the display conditions in another PR, this doesn't take touch&gos into account right now.

lukecologne: Not necessarily, for example, take the following conditions:
```
if (ra > 50) {
    showTrim = false;
} else if (gs < 30) {
    showTrim = true;
}
```
and showTrim would be initialized to false. If I'm not totally overlooking something I think this should behave as expected. But in any case, after looking at some more references I'm doubting that any of the two solutions are correct

lukecologne: No it would not, the variable would need to be a field of the class of course to be persistent. Otherwise it would not be a flip-flop. So if you stay below 50ft and go above 30kts, it will retain the last value which would be `true`. Maybe my example didn't make that clear enough

flogross89: > No it would not, the variable would need to be a field of the class of course to be persistent. Otherwise it would not be a flip-flop. So if you stay below 50ft and go above 30kts, it will retain the last value which would be `true`. Maybe my example didn't make that clear enough

Ah, you're right, didn't think about the persistent state. Sounds like a good solution, which is easily extensible for touch&go cases. There's one corner case (take-off roll above 30kts, but still on ground), where an instrument reload would lead to the indicator not showing, but that's negligible.

flogross89:
> Just from the photo I think the left and right padding of the THS % need to be reduced slightly. The left digit and % char are almost touching the border.
> 
> The reference photo is clearly padded more in the vertical than in the horizontal. In the sim screenshot the padding looks very square.

image
updated

fbw-a380x/src/systems/instruments/src/PFD/PFD.tsx
flogross89: I updated the conditions, and also added a condition for touch&gos

2hwk (approved):

PR #8729, opened 2024-07-02
chore(a380x): adjusted and added scrape points

Steveveepee:
## Summary of Changes
Added multiple scrape points on external extremities of the aircraft model as people will try and crash this into things.

## Additional context
Currently there are only 4 wheels listed in the file. Should this be adjusted to include the fwd and rear wheel pair in each main gear to allow correct bogy animation on takeoff and landing?

All scrape points have Impact damage threshold crash velocity set to 100 ft/min at the moment, perhaps the points on the pods and tailscrape point could have higher values to allow sparking/dust/smoke but not 'crash' the a/c. Appropriate value to be determined.

Discord username : Shrike


flogross89:
When trying it out, the sim seems to crash when selecting an airport to load. I discovered that you didn't update the max_number_of_points parameter, maybe that causes the crash? 
https://docs.flightsimulator.com/html/Content_Configuration/SimObjects/Aircraft_SimO/flight_model/contact_points.htm#max_number_of_points

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
Steveveepee, line 1: Line 105 is probably the culprit of a crash as I dont think that the comment can be on the end of the line like that.

PR #8789, opened 2024-07-31
fix: path.join consistency

Creative-Difficulty:
Fixes #[issue_no]

## Summary of Changes

My goal with this PR is to choose either only using `x/y/z` wherever possible (Strings with slashes as paths) or individual segments `path.join(x, y, z)` and I've chosen the latter as it is used more widely and seeing / in a path for the windows file system could cause confusion.

See https://ptb.discord.com/channels/738864299392630914/747622836381810739

## Additional context

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -52,7 +52,7 @@ module.exports = {
   settings: {
     tailwindcss: {
       groupByResponsive: true,
-      config: path.join(__dirname, 'fbw-common/src/systems/instruments/src/EFB/tailwind.config.js'),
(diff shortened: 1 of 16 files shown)


Creative-Difficulty:
@FoxtrotSierra6829 ESlint violations have been fixed

2hwk (approved):

tracernz:
Regular paths seems much nicer than a tuple of params as you can a) parse them easier as a human, and b) you can copy-paste them into/out of the scripts.

PR #8814, opened 2024-08-20
feat(gear): A380 body gear steering animation

crocket63:
Fixes #[issue_no]

## Summary of Changes

Animation implementation for the rear steering and differential wheel speeds.
Note that the actual steering system is not implemented, it's only visually based on the nose steering angle.

## Additional context


fbw-a380x/src/wasm/systems/a380_systems_wasm/src/rear_wheel_steering.rs
Gurgel100, line 1: Why not call it `body_wheel_steering`?

fbw-a380x/src/wasm/systems/a380_systems_wasm/src/body_wheel_steering.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn normalise_angle(angle: f64) -> f64 {
+    let raw = angle % 360.;
+
+    if raw > 0. {
+        raw
+    } else {
+        raw + 360.
+    }
+}
```
2hwk, line 77:
```suggestion
fn normalise_angle(angle: f64) -> f64 {
    let raw = angle % 360.;

    if raw >= 0. {
        raw
    } else {
        raw + 360.
    }
}
```

Feels more intuitive to normalise to 0-359 than have 0 always be 360 and 1-360, but maybe there's a reason for this?

2hwk (dismissed):

Gurgel100 (approved):

PR #8816, opened 2024-08-20
ref(sentry): Update Sentry SDK to version 8

Lms24:
Fixes: no issue (if you want I can open an issue if you want to discuss the change first)

## Summary of Changes

Hi Sentry JS SDK engineer and avid FBW fan here 👋 I learnt a couple of months ago that y'all are using Sentry so I thought I'd try to contribute a tiny PR to this awesome project when I get some time. If you're not interested in this PR, feel free to close it, no hard feelings :)

This PR updates the `@sentry/browser` SDK from version 6 to version 8.

From inspecting the FBW repo code, the Sentry setup looks fairly simple (good!) so there shouldn't be any problematic changes with jumping two major versions. So this update should reduce your bundle size footprint from the Sentry SDK as well as remove a couple of dated dependencies which we still relied on in v6.

I do have a question though, see PR comment

## Additional context

 If you're interested in specific changes, feel free to check out the Sentry JS SDK migration guides:

- [v6 -> v7](https://docs.sentry.io/platforms/javascript/migration/v6-to-v7/) (we mainly reduced bundle size and removed unnecessary abstractions in this major release)
- [v7 -> v8](https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/) (the main breaking changes are in our tracing and performance APIs which don't seem to be used in the FBW repo).

Discord username (if different from GitHub): thunderstrack

## Testing instructions
Based on what I've seen, your CI setup should test the SDK build well. I'm not sure if y'all have any test procedures that should apply here.


fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
```diff
@@ -216,10 +213,10 @@ export class FbwAircraftSentryClient {
       );
     }
 
-    const integrations: Integration[] = [new CaptureConsoleIntegration({ levels: ['error'] })];
+    const integrations = [Sentry.captureConsoleIntegration({ levels: ['error'] })];
```
Lms24, line 216: Integrations were converted from classes to functions. Other than that, neither this, nor the browserTracingIntegration changed functionality-wise.

fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
```diff
@@ -216,10 +213,10 @@ export class FbwAircraftSentryClient {
...
 
     if (config.enableTracing) {
-      integrations.push(new BrowserTracing());
+      integrations.push(Sentry.browserTracingIntegration());
     }
```
Lms24, line 220: question: Just adding `browserTracingIntegration` doesn't enable Sentry's full tracing capabilities.
With just the integration, you get distributed tracing between services to connect errors in Sentry between individual services.

If you want to measure and record performance data, you also need to add `tracesSampleRate > 0` to your `Sentry.init` options. Let me know if you want to change this. Can also open a separate PR to keep things atomic :)

fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
```diff
@@ -5,9 +5,6 @@
 /* eslint-disable no-console */
 
 import * as Sentry from '@sentry/browser';
-import { BrowserTracing } from '@sentry/tracing';
-import { CaptureConsole as CaptureConsoleIntegration } from '@sentry/integrations';
-import { Integration } from '@sentry/types';
```
Lms24, line 10: All of these integrations are now directly part of `@sentry/browser` so we can remove a couple of direct dependencies from the repo :)

Benjozork (approved):

fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
Benjozork: It's been a while since I set this up - does the performance data measurement depend on React? We're in the process of rewriting the EFB to use an msfs-specific JS framework, so it may not be worth to keep this it if the integration only works with React or other web frameworks.

Benjozork:
Thanks a lot for the PR!

fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
Lms24: > does the performance data measurement depend on React?

Since you're using `@sentry/browser` there's nothing framework-specific in the package. What you'll get from performance monitoring is essentially a span tree of how long e.g. browser resource loading takes, web vitals like LCP. More [information here](https://docs.sentry.io/product/performance/). To be honest, I don't know how well this translates to the MSFS renderer, so take what's in these docs with a grain of salt :)

If you want, we can give it a try but I suggest we open a separate PR to revert more easily if we need to

(there's also `@sentry/react` with some React-specific instrumentation but considering you're moving away from React I think it's better to stay with `@sentry/browser).

tylerclt:
Quality AssuranceTrainee Report

Discord Username: Tylerb780#3594 
Object of testing : #8816  
Aircraft: A32NX
Tier of Testing: 1
Date : 07/09/2024 

Testing Process: 
1. Loaded into the sim 
2. Powered on Aircraft as usual
3. Made sure nothing was broken with the switches
4. Verified displays powered up correctly

Testing Results: 
Not Passed  

Negatives:       
When the aircraft was fully powered on with EXT PWR connected no displays powered up including the FlyPad

Media:

tracernz:
Something goes wrong in SystemsHost and ExtrasHost.
image
image
image

Lms24:
huh that's interesting! this likely means that the JS engine (is this coherent.js?) doesn't support ES2018 fully. Which is what we took as a baseline for [version 8](https://docs.sentry.io/platforms/javascript/troubleshooting/supported-browsers/) of the SDK. How would you usually access the global object?

tracernz:
Yeah. I think is supports most or all of ES2017, but definitely not ES2018. `window` works okay, but I'm not sure about `global` (haven't tried).

Lms24:
yup, that should work! I wonder though if this will be the only incompatibility. Do you have a preferred way of setting polyfills? Specifically, I'm wondering _where_ to set it best so that it is available everywhere the Sentry SDK is active. I could of course directly set it before the `Sentry.init` call and give this a try.

tracernz:
In the absence of anything else that's probably the thing to do at this point. @Benjozork might have a better idea.

Lms24:
Hey, sorry that this got a bit stale. I'll try to take a look at this again in the next days. Been a bit busy lately...

Maximilian-Reuter:
Hey, is this still being worked on ?

PR #8845, opened 2024-08-26
feat(pfd): A380 improvements

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Changes the FLX FMA layout to match A350/A380.
Changes sidestick indicator to only appear if left or right gear are compressed while engines are running.
Slightly increase font size of pitch attitude scale numbers (not 100% sure about this one or if my eyes are playing tricks)
Adds tailstrike indicator during takeoff
## Screenshots (if necessary)

left = new, right = old
FLX

attitude

## References

YT vids & A380 PFD pictures. Example of flex: https://youtu.be/sOhDDj5agvo?t=332

## Additional context

bruno_pt99


flogross89 (approved):
LGTM, thanks! In the future, we can consider replacing the GENERAL ENG COMBUSTION SimVars with something more realistic, but let's do this when we have better references on the actual data source

BravoMike99:
> LGTM, thanks! In the future, we can consider replacing the GENERAL ENG COMBUSTION SimVars with something more realistic, but let's do this when we have better references on the actual data source

Yeah, I guess its probably some FADEC/EEC signal like on the A320. I just copied it from the existing logic since it has been working so far on both planes 😅

BravoMike99:
In the mean time I have noticed that the pitch tailstrike indicator was missing during takeoff and implemented some basic logic for it to show (hence the review quest)

flogross89:
> In the mean time I have noticed that the pitch tailstrike indicator was missing during takeoff and implemented some basic logic for it to show (hence the review quest)

Which pitch angle does it currently show at? It seems this angle is dependent on MLG compressed or not. 
On newer models, there also seems to be an aural "PITCH PITCH" warning when in excessive pitch angle close to ground

BravoMike99:
> > In the mean time I have noticed that the pitch tailstrike indicator was missing during takeoff and implemented some basic logic for it to show (hence the review quest)
> 
> Which pitch angle does it currently show at? It seems this angle is dependent on MLG compressed or not. On newer models, there also seems to be an aural "PITCH PITCH" warning when in excessive pitch angle close to ground

At 12.5º degrees. I'm not sure if its MLG compression related as in this video it seems to move before it is compressed as the sidestick indicator is still visible, unless it has to do with different update rates for each indication https://youtu.be/5PfIW-Ar4Fo?t=25 (I didn't implement the 3 second condition after liftoff as of now anyways). I assume that the "pitch pitch" has been available on the A380 since launch as it has been available since day one on the A346 and A345.

flogross89:
> > > In the mean time I have noticed that the pitch tailstrike indicator was missing during takeoff and implemented some basic logic for it to show (hence the review quest)
> > 
> > 
> > Which pitch angle does it currently show at? It seems this angle is dependent on MLG compressed or not. On newer models, there also seems to be an aural "PITCH PITCH" warning when in excessive pitch angle close to ground
> 
> At 12.5º. I'm not sure if its MLG compression related as in this video it seems to move while it is compressed as the sidestick indicator is still visible, unless it has to do with different update rates for each indication https://youtu.be/5PfIW-Ar4Fo?t=25 (I didn't implement the 3 second condition after liftoff as of now anyways). I assume that the "pitch pitch" has been available on the A380 since launch as it has been available since day one on the A346 and A345.

I'm seeing two things in the video, thanks for the link:

1. When rotating, the landing gear goes from compressed to extended (but not compressed), which changes the max. pitch attitude from 12.5° to ~13-13.5°. So there seem to be two different limits, depending on whether the MLG is compressed or extended
2. The indication disappears 3s after liftoff.

Without having the 3s delay after take-off it doesn't make sense though to implement the different max. pitch attitude, as it would disappear immediately when MLG is un-compressed :)
Do you see a chance that you could implement the 3s timer and the two different max. pitch attitudes in this PR? Or should we wait for a future PR. 


You're right about the "PITCH PITCH" alert, seems to be there from the beginning. I was mixing it up with "BANK BANK" close to ground, which was only added recently.

flogross89 (approved):
LGTM to me overall, if you can still add the proposed changes that would be amazing, without it it's still a good improvement

PR #8907, opened 2024-09-14
feat(a380/mfd): add automatic perf page switching depending on fms phase

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Adds automatic page switching on PERF page depending on FMS flightphase.
## References

https://youtu.be/pHui34_ODnc?t=105

## Additional context

bruno_pt99


tracernz:
Should this only switch if you're on the active phase page?

BravoMike99:
> Should this only switch if you're on the active phase page?

Sounds logical if it works that way but I couldn't find any refs other than the video above sadly. If you can confirm that's the expected behavior ill add that logic (is that also how the 320 works?).

tracernz:
According to my info the A320 pages should only sequence to the next flight phase from the current one (but keep in mind "next flight phase" doesn't mean the next in numerical sequence, rather the next one activated). I don't know about the A380.

BravoMike99:
I have added that logic in the meantime if its ok by you, unless they implemented that in some recent software update I don't see why it shouldn't be similar on the A380.

tracernz (approved):

PR #8883, opened 2024-09-07
feat(a380x): ata21 failures

mjuhe:
## Summary of Changes

Adds 50 failures to the air conditioning, ventilation and pressurization systems (a380x). It also changes the architecture of the system a bit so data is sent from each of the 4 CPIOM B computers independently. List of failures:

- Rapid Decompression
- Cabin Fan (x4)
- Hot Air / Trim Air Pressure Regulating Valve (x2)
- Cargo Isolation Valves (x2 - fwd and bulk)
- Cargo Extraction Fans (x2 - fwd and bulk)
- Bulk Cargo Heater
- Local controllers:
      - FDAC (2 controllers, 2 channels each)
      - TADD (2 channels)
      - VCM (2 controllers, 2 channels each)
      - OCSM Automatic Partition (x4)
      - OCSM (4 controllers, 2 channels each)
- CPIOM B Applications:
      - AGS (x4)
      - TCS (x4)
      - VCS (x4)
      - CPCS (x4)

## Additional context

None of these are connected to the FWC yet, only to their respective SD pages. More failures can be added in the future but I wanted to keep this PR somewhat small.

Discord username (if different from GitHub): Commander Ga

## Testing instructions

A380 only - no testing required.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo** or **flybywire-aircraft-a380-842** download link at the bottom of the page

PR #8922, opened 2024-09-15
feat(a380/pfd): Add ARS indication on speed tape

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Adds ARS speed indication on PFD speed tape.
Updates target speed bug
## Screenshots (if necessary)

image
image

## References

FCOM & https://youtu.be/yGlS2QBuX7s?t=1664

## Additional context

bruno_pt99
## Testing instructions
A32NX regression testing, Perform a full flight:
Check that indications on lower SD page show valid data.
Check that normal law indications show on PFD (green " on top of the attitude indicator) and that the sidestick indication is available after one engine running.
No erroneous ecam/PFD messages or GPWS warnings are triggered during flight.


flogross89:
Nice one! Just to understand the behavior: After the a/c passed 212kt with a positive speed trend, the ARS will become active and move the VFE to 263kt?

BravoMike99:
> Nice one! Just to understand the behavior: After the a/c passed 212kt with a positive speed trend, the ARS will become active and move the VFE to 263kt? As a consequence (and that's why I'm asking), as soon as the current speed reference line passes 212kt, the small green ARS line should disappear and the red barber pole should move to 263kt, correct?

Yes exactly.

fbw-a380x/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -957,6 +957,59 @@ class V1Offtape extends DisplayComponent<{ bus: EventBus }> {
   }
 }
 
+class ArsBar extends DisplayComponent<{ bus: ArincEventBus }> {
+  private arsPath = Subject.create<string>('');
+
+  private arsVisiblity = Subject.create<string>('hidden');
```
flogross89, line 963: For our newer components, we try to move towards working with ConsumerSubjects and MappedSubjects, instead of using setXxx() functions called after subject updates.
Could you try implementing this? 
Examples are in e.g. https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/EWD/elements/EGT.tsx#L25

For example:
```
private readonly slatsFlapsStatus = Arinc429ConsumerSubject.create(this.sub.on('slatsFlapsStatus').whenChanged());
private readonly conf1FSelected = this.slatsFlapsStatus.map((s) => s.getBitValueOr(28, false) && s.getBitValue(29) && s.getBitValue(18) && !s.getBitValue(26));
private readonly arsVisibility = this.conf1FSelected.map((c) => c ? 'visible' : 'hidden');
```

and so on, feel free to DM me for questions

flogross89 (changes requested):

fbw-a380x/src/systems/instruments/src/PFD/SpeedIndicator.tsx
tracernz: Typescript can also infer the type just fine here, so you don't need to specify it with <string>.

fbw-a380x/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -957,6 +957,59 @@ class V1Offtape extends DisplayComponent<{ bus: EventBus }> {
...
+
+  private airSpeed = new Arinc429Word(0);
+
+  private conf1FSelected = false;
+
+  private readonly ARS1FSpeed = 212;
+
+  private readonly Conf1FVFE = 222;
```
tracernz, line 971: These should probably be `private static readonly ARS_1_F_SPEED`. If they're instance members they should be camel case rather than pascal case.

fbw-a380x/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -957,6 +957,59 @@ class V1Offtape extends DisplayComponent<{ bus: EventBus }> {
...
+
+  private readonly offset = ((this.ARS1FSpeed - this.Conf1FVFE) * DistanceSpacing) / ValueSpacing;
+
+  private setArspath() {
+    if (this.conf1FSelected) {
+      this.arsVisiblity.set('visible');
+      const arsPos = ((this.airSpeed.value - this.ARS1FSpeed) * DistanceSpacing) / ValueSpacing + 80.818;
+      this.arsPath.set(`m19.031 ${arsPos}h 1.9748v${this.offset}`);
```
tracernz, line 979: It would be much more efficient to transform this rather than changing the path constantly, but I think we really have that problem all over the PFD so I guess it's okay for now.

BravoMike99:
Thanks both for the reviews. I changed the subject logic with floridude's help and changed the vertical offset to use translate3D. I did, however, need to apply one patch provided by holland from the ATA 21 failures PR per floridude suggestion as the getBitValue methods are not exposed in the arinc interface currently.

flogross89 (approved):
LGTM now :)

fbw-common/src/systems/shared/src/arinc429.ts
```diff
@@ -154,14 +160,24 @@ export class Arinc429Register implements Arinc429WordData {
...
+  getBitValueOr(bit: number, defaultValue: boolean | undefined | null): boolean {
+    return this.isNormalOperation() ? ((this.value >> (bit - 1)) & 1) !== 0 : defaultValue;
+  }
+
+  /** @deprecated use {@link getBitValue} */
   bitValue(bit: number): boolean {
     return ((this.value >> (bit - 1)) & 1) !== 0;
   }
 
+  /** @deprecated use {@link getBitValueOr} */
```
tracernz, line 180: What's the reason for this change? `bitValue` and `bitValueOr` match up with `value` and `valueOr`. These new functions you've introduced do not match up with the rest of the API.

tracernz (changes requested):

fbw-common/src/systems/shared/src/arinc429.ts
BravoMike99: Initially I was trying to use Arinc429ConsumerSubjects on the PFD side and as the bitValueOr methods weren't being exposed in the interface I picked the changes from here https://github.com/flybywiresim/aircraft/pull/8883#discussion_r1760085768. In any case, I've changed it so it all uses btValue & bitValueOr & removed the get functions so its consistent across the board.

tylerclt:

Discord Username: tylerb780 
Object of testing : #8922  
Aircraft: A32NX 
Tier of Testing: 2
Date: 22/09/2024 

Testing Process: 
1. Loaded into the A32NX at KLAS for a flight to KSAN
2. Powered Aircraft On
3. Conducted a normal preflight setup
4. Checked that normal law indications show on PFD (green " on top of the attitude indicator) and that the sidestick indication is available after one engine running.
5. Taxied out and Departed KLAS
6. Check that indications on the lower SD page show valid data.
7. Climbed to cruise
8. Verified no erroneous ecam/PFD messages or GPWS warnings are triggered during flight and descent

Testing Results: 
Passed  

Conclusions: 
Works as intended

Media:

PR #8932, opened 2024-09-17
fix(nd): Remove ARC mask for TERR map from A380X ND

flogross89:
## Summary of Changes

- Keeps the round mask at the top of the ARC rose only for A32NX, not for A380X. Terrain mask is different on the A380

## Additional context

## Testing instructions

Use new SimBridge version: https://www.swisstransfer.com/d/a0233bb1-53bc-438b-a0aa-bdf54f460a23

A32NX: 
Everything in the ARC TERR display should look as before, TERR should be cut off/masked at the top by the ARC ring

A380X:
Terrain mask should be square at the top when in ARC mode, and not cut off/masked by the ARC ring


fbw-a32nx/src/systems/instruments/src/ND/style.scss
```diff
@@ -236,3 +236,11 @@ path.rounded {
   stroke-linecap: round;
   stroke-linejoin: round;
 }
+
+.hidden {
+  display: none;
+}
+
+.a380xInvisible {
+  display: block;
```
tracernz, line 245: I don't think this should force block. Do we even need any style for it?

flogross89: empty rulesets are not allowed, or do you mean just not defining it in a32nx's style.scss? Don't know how that behaves. I would do `display: inherit` now

tracernz: That would also modify the behaviour. Why do we need a style in the A32NX CSS for this at all?

flogross89: Hmm yeah you're right. There's no a32nx/a380x prop passed down or defined anywhere, so I found this the best way to have an aircraft-specific style/visibility, but I'm open to other ideas. 
I could leave the style empty, although then lint complains

tracernz: You can just not have it in the a32nx stylesheet at all.

flogross89: Fixed that now

PR #8938, opened 2024-09-19
fix(a380x): arrivals page approach data

tracernz:
Fixes #[issue_no]

## Summary of Changes

- Adds approach suffix to the Approach navdata type,
- Fixes a bug that caused all non-RNAV approaches to be marked as AR/SAAR,
- Fixes a bug where 0 was returned instead of undefined for a blank runway LS, causing the A380 departure page to show wrong info,
- Fixes the display of approach frequency/channel and ident on the A380 arrivals page,
- Fixes the naming of approaches in the A380 arrivals page,
- Filters out invalid and unsupported approach types on the arrivals page.

## Additional context

## Testing instructions

Check that the arrivals page in the A32NX still works properly in general (no change expected, but some code touched).


fbw-a380x/src/systems/instruments/src/MFD/shared/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  [ApproachType.NdbDme]: 'NDB',
+  [ApproachType.Fms]: 'RNAV',
+  [ApproachType.Mls]: 'MLS', // not actually supported
+  [ApproachType.MlsTypeA]: 'MLS', // not actually supported
+  [ApproachType.MlsTypeBC]: 'MLS', // not actually supported
+  [ApproachType.Tacan]: 'TAC', // not actually supported
+  [ApproachType.Unknown]: '',
+};
+
+export function getApproachName(approach: Approach, withRnpSuffix = true): string {
```
BlueberryKing, line 42: It would be cool to include this in `ApproachUtils.ts`, but I understand that unifying them is hard because of the many different formats.

tracernz: Yeah, in the end the two planes are actually quite different with this, so the other approach utils probably needs to move into fbw-a32nx/.

BravoMike99:
Perhaps you might want to fix the identifier on the APPR PERF page too? That way I change #8902 to only add the approach phase confirmation button.

BlueberryKing (approved):

tracernz:
> Perhaps you might want to fix the identifier on the APPR PERF page too? That way I change #8902 to only add the approach phase confirmation button.

All done, thanks! :)

tracernz:
Dev Team Report

Discord Username : tracernz
Object of testing : https://github.com/flybywiresim/aircraft/pull/8938
Aircraft : A32NX
Tier of Testing : 1
Date : 28/09/2024

Testing Process:
Test various approaches, STARs, and transitions in the A32NX. No regressions found.

Testing Results:
Passed

PR #8950, opened 2024-09-21
feat(a380x/fms): nav db swap to reset fms

tracernz:
Fixes #[issue_no]

## Summary of Changes

This implements equivalent nav db swap function to the A32NX, for quickly resetting the FMS state. It also fixes the database effectivity dates and db p/n.

## Additional context


flogross89:
Thank you for the PR! We still have some data which is outside of flightPlanService (since we don’t have a/c specific performance data yet, and possibly need other modifications), so a bunch of data is stored in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.tsx#L52

We should also reset this data somehow, otherwise e.g the PERF/APPR page would remain filled

flogross89 (changes requested):
See previous comment

tracernz:
> Thank you for the PR! We still have some data which is outside of flightPlanService (since we don’t have a/c specific performance data yet, and possibly need other modifications), so a bunch of data is stored in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.tsx#L52
> 
> We should also reset this data somehow, otherwise e.g the PERF/APPR page would remain filled

Ah, that is reset by the initVariables function in the A32NX. Probably makes sense for the A380 to be the same.

flogross89:
> > Thank you for the PR! We still have some data which is outside of flightPlanService (since we don’t have a/c specific performance data yet, and possibly need other modifications), so a bunch of data is stored in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.tsx#L52
> > We should also reset this data somehow, otherwise e.g the PERF/APPR page would remain filled
> 
> Ah, that is reset by the initVariables function in the A32NX. Probably makes sense for the A380 to be the same.

Yep. FmgcData could just get a reset() function and that’s it, all performance/FP data not in flightPlanService has to be stored in FmgcData. Parts of the SWAP nav db code could maybe also be called when transitioning to the FMGC DONE phase? Since we’re also trying to reset the FMC there

tracernz:
> > > Thank you for the PR! We still have some data which is outside of flightPlanService (since we don’t have a/c specific performance data yet, and possibly need other modifications), so a bunch of data is stored in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.tsx#L52
> > > We should also reset this data somehow, otherwise e.g the PERF/APPR page would remain filled
> > 
> > 
> > Ah, that is reset by the initVariables function in the A32NX. Probably makes sense for the A380 to be the same.
> 
> Yep. FmgcData could just get a reset() function and that’s it, all performance/FP data not in flightPlanService has to be stored in FmgcData. Parts of the SWAP nav db code could maybe also be called when transitioning to the FMGC DONE phase? Since we’re also trying to reset the FMC there

It would be nice if the "unset" value for every field in the performance data was `null`. Then we could have a reset method that just loops through and sets everything to null, rather than an error prone reset method that you have to remember to add any new props to.

flogross89:
> > > > Thank you for the PR! We still have some data which is outside of flightPlanService (since we don’t have a/c specific performance data yet, and possibly need other modifications), so a bunch of data is stored in https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.tsx#L52
> > > > We should also reset this data somehow, otherwise e.g the PERF/APPR page would remain filled
> > > 
> > > 
> > > Ah, that is reset by the initVariables function in the A32NX. Probably makes sense for the A380 to be the same.
> > 
> > 
> > Yep. FmgcData could just get a reset() function and that’s it, all performance/FP data not in flightPlanService has to be stored in FmgcData. Parts of the SWAP nav db code could maybe also be called when transitioning to the FMGC DONE phase? Since we’re also trying to reset the FMC there
> 
> It would be nice if the "unset" value for every field in the performance data was `null`. Then we could have a reset method that just loops through and sets everything to null, rather than an error prone reset method that you have to remember to add any new props to. I think fixing this is well outside the scope of this PR, and also exists for the DONE phase, no? Also, that file is .tsx but contains no JSX..

Overlooked this comment, sorry. Yeah we can also fix this outside of this PR, fine for me.

flogross89 (approved):

tracernz:
> Overlooked this comment, sorry. Yeah we can also fix this outside of this PR, fine for me.

To track it: https://github.com/flybywiresim/aircraft/issues/9019

PR #8973, opened 2024-09-22
fix(a380x): prevented clickspots being activated behind L/G GVTY master guard

2hwk:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue with clickspots on the L/G GVTY lever.

## Additional context

## Testing instructions

- Check the L/G GVTY extension switch can NOT be moved through the L/G GVTY guard
- Check that L/G GVTY extension switch can be moved once the guard is lifted
- Check that manual L/G GVTY extension guards (two red) still operate as per normal.


tracernz:
Can't we just use a collision mesh on the lid to keep it simple?

becas22:
Quality Assurance Report

Discord Username  : bernardor96
Object of testing : #8973  
Aircraft          : A380X 
Date              : 24/09/2024 

Testing Process: 
Check the L/G GVTY extension switch can NOT be moved through the L/G GVTY guard
Check that L/G GVTY extension switch can be moved once the guard is lifted
Check that manual L/G GVTY extension guards (two red) still operate as per normal. 


Testing Results: 
Passed  


Negatives:       
Clickspots visible without interaction, as visible in media below.
Red guards clickspots could be changed to hand cursor instead of arrows

Conclusions: 
The PR achieves its scope. Not displaying the clickspots with the guard down would be better imo

Media:
https://discord.com/channels/738864299392630914/1193615390861951086/1288094710807658506

2hwk:
> Can't we just use a collision mesh on the lid to keep it simple?

We could, but then this would still work either way as a backup if that doesn't work. Also this seems to work nicer with the non-legacy interaction system.

alepouna (approved):
xml wise lgtm

PR #9007, opened 2024-09-30
fix(a380x/eng): Reduce thrust on ground at low speeds

flogross89:
## Summary of Changes

Taxi thrust is too high, this PR reduces the thrust for low machs and N1 < 30%
Ideally this would be fixed by lowering the N1, but that's a more invasive change and prone to errors. Should be improved with next-gen engine model only

## Additional context

## Testing instructions

Needs verification from type rated pilots


2hwk (approved):

BlueberryKing:
Do we also need to update these parameters in the performance tables used for predictions?

flogross89:
> Do we also need to update these parameters in the performance tables used for predictions?

This change is only active below 0.1 mach, so it shouldn't be relevant for the VNAV predictions in flight. Could be relevant for taxi fuel prediction, but I think we just use a default value there anyways.

becas22:
> This change is only active below 0.1 mach, so it shouldn't be relevant for the VNAV predictions in flight. Could be relevant for taxi fuel prediction, but I think we just use a default value there anyways.

I think we are using A320 taxi fuel figure, as I remember it displaying 200kgs there

pilotseyea350:
> > This change is only active below 0.1 mach, so it shouldn't be relevant for the VNAV predictions in flight. Could be relevant for taxi fuel prediction, but I think we just use a default value there anyways.
> 
> I think we are using A320 taxi fuel figure, as I remember it displaying 200kgs there

Yeah I can confirm that.

PR #9025, opened 2024-10-04
feat(a380): Change touchdown sound to use radio vertical speed as reference

aguther:
## Summary of Changes
This PR changes the vertical speed reference for the touchdown sounds to use a derived signal from radio altitude (as it is in the A32NX). This improves correctness in the case of sloped runways.

## Testing instructions
Land with different vertical speeds also on sloped runways. Ensure sounds are still working.


hotshotp (approved):
LGTM!

PR #9045, opened 2024-10-08
fix(a380x): Various fuel system fixes

donstim:
Fixes various fuel system issues. See summary of changes.

## Summary of Changes
Fixes the following issues with the fuel system:
1) Fixed issue with transfers into feed tanks 2 and 3 being lower than fuel flow during cruise and very unbalanced with transfers into feed tanks 1 and 2
2) Fixed issues with trim and outer tanks not transferring fuel
3) Added missing 2nd trim tank inlet tank valve

## Screenshots (if necessary)

Visualisation of the MSFS Fuelsystem according to the flightmodel.cfg
Fuelsystem A380(7)

## Additional context
ECAM fuel system page may need changing due to the addition of the 2nd trim tank inlet valve . If any overhead fuel system buttons are operational, these may also need to be checked.

Be aware that not all automatic transfers are modeled. Specifically, the load alleviation transfers are not included because MSFS does not have the necessary triggering conditions to accurately model these.

Valves, lines, and junctions are included to model fuel transfers beyond what is used by the automatic transfers that are modeled. These will need to be checked once the overhead push buttons for all the pumps, valves, etc. are all fully modeled.

Fuel transfer rates have not yet been tuned. This first implementation is just to ensure that the fuel transfer occur according to the logic detailed in the testing instructions and at sufficient rates such that the feed tanks will not empty before the transfer tanks.

## Testing instructions
**Most immediate testing need is for a short test as follows:
1. First change the values of the FBWA380X.ini file in the work file to: 
Screenshot 2024-10-18 101853

The work folder can be found using these instructions: https://docs.flybywiresim.com/fbw-a32nx/feature-guides/flypados3/throttle-calibration/?h=work+folder#work-folder

2. Load in cold/dark. Start airplane, load 63,000 kg of fuel using the EFB. Take off, then using either the MSFS Weight and Balance UI or the ECAM Fuel page
Screenshot 2024-10-18 101401
 , note if fuel is being transferred out of the inner tanks (fuel quantity decreasing).  Report back (either here or on Discord)**

For testing of all transfers, here is the transfer logic:

The feed tanks feed the engines. A certain amount of fuel is kept in the feed tanks with that amount varying depending on the amount of fuel in the transfer tanks (inners, mids, outers, and trim). The trim tank is also used for cg control transfers.

Starting with fuel in each tank (including more than 4000 kg in the left mid tank), the normal transfer sequence should be:  inners, then mids, then trim, then outers. In addition, trim tank fuel may be transferred at any time to keep the cg from going too far aft. (Normal loading is for an aft cg, so there is no transferring back to the trim tank to move the cg further aft.)

Until the left mid tank has less than 4000 kg of fuel, the inner and mid tank transfers should work like this:
1) Transfers to feed 1 and feed 4 should begin when the one with the lowest fuel amount (in the A380X they will usually be the same since we haven't yet incorporated fuel burn differences between the engines) reaches 19560 kg, the inner tanks (and mids once the inners are empty) will start transferring fuel to it. When the fuel levels of feed 1 and 4 become equal (if they were unequal at the start of the transfer), fuel will begin transferring to the other feed (the one that was not the lowest of feed 1 and 4). Since the feed tanks will normally be equal, the transfers to both feed 1 and feed 4 should start pretty much simultaneously. The transfers to feed 1 and feed 4 will each end when each tank reaches 20560 kg. A balancing technique is used to try to keep the fuel level in the pairs of fuel tanks 1/4 and 2/3 balanced. If there is fuel left in the inner tanks, it will start to transfer again when the level in the feed tanks reach the 19650 kg activation threshold again. (Note: These thresholds change when the remaining time to the destination is less than 90 minutes, but a trigger for this is not available in the MSFS fuel system.)

2 The transfers to feed tanks 2 and 3 work the same way, except they begin when the lower one reaches 20840 kg and end when they each reach 21840 kg. Note that feed tanks 2 and 3 are higher capacity than 1 and 4, are normally fueled initially to a higher level, and retain a higher level until made symmetrical by the transfer process detailed below.

3. When there is less than 8000 kg in the mid tanks (the MSFS trigger actually uses <4000 kg in the left mid), the begin and end transfer levels for feed tanks 2 and 3 are changed to be the same as those used for feed tanks 1 and 4. The following transfer process is then used to balance the amount of fuel in the tanks:  When the tank with the least amount of fuel reaches the transfer activation threshold, fuel begins transferring in. (This will be both feed tanks 1 and 4 since they will be approximately equal, and they start at a lower level than 2 and 3.) As the level in feed tanks 1 and 4 increase, and 2 and 3 decrease, when they become equal, fuel will start transferring into 2 and 3. (Although this approximately equalizes the fuel level in all the feed tanks, it will differ by a bit since the transfers don't start immediately and the transfer rates into 1 and 4 are just a tiny bit different than 2 and 3.)

4. When the mid tanks are empty, fuel will be transferred from the trim tank to the feed tanks when the fuel level in any feed tank is less than 6000 kg. The same balancing process is used as described above. Once started, the trim tank will transfer fuel until it is empty.

5. When the trim tank is empty, fuel wil be transferred from the outer tanks to the feed tanks when the fuel level in any feed tank is less than 4000 kg. The transfer process uses a similar balancing technique, but only between the pairs or symmetric fuel tanks (1/4 and 2/3). The transfer stops when the fuel level in each pair of feed tanks is greater than 4500 kg and starts again when the fuel level  in any feed tank is less than 4000 kg.

6. CG fuel transfers:  These may happen independently and simultaneously with the main transfers described above. The transfer begins when the CG is higher (more aft) than the schedule vs gross weight shown below and ends when the CG is 1% lower (more forward) than than target. (Note: the CG shown in the CRUISE page of the ECAM is rounded off, so the transfer will start a little while after the ECAM shows a CG equal to the target CG.) The fuel will be transferred to the inner tanks, or to the mid tanks if the inner tanks are empty, or to the feed tanks if  both the inner and mid tanks are empty.
Screenshot 2024-10-17 095923

For the testing, please observe whether the fuel transfers are following the description given above. You can use the MSFS fuel and payload UI to see the transfer activation/deactivation points without the round off used in the cockpit display. There should not be any fuel transfers taking place while the airplane is on the ground.

Additional notes:
It would be helpful to have flights done starting with different initial fuel loads: 18 T, 26 T, 36 T, 47 T, 103 T, and >158 T. The test can be terminated whenever one of the following occur: fuel transfer not occurring per above schedule

Note if there are any discrepancies and describe the circumstances as much as possible. 


donstim:
Looks like some changes were made post the master code I used for this PR. I will need to back these out.

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -77,7 +79,7 @@ static_pitch = -0.23 ; degrees, pitch when at rest on the ground (+=Up, -=Dn)
 static_cg_height = 8.89 ; feet, altitude of CG when at rest on the ground
 gear_system_type = 3 ; gear system type (betweeen 0 and 4) 0 = electrical, 1 = hydraulic, 2 = pneumatic, 3 = manual, 4 = none, 5 = undefined
 tailwheel_lock = 0 ; Is tailwheel lock available TRUE/FALSE
-max_number_of_points = 23 ; Number of contact points
+max_number_of_points = 11 ; Number of contact points
```
flogross89, line 82: I assume this change was not intended? This was changed as part of another PR recently, most likely you worked on an out-of-date flight_model.cfg

flogross89 (changes requested):
One inadvertent change still, you already saw the other ones :)

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
donstim: Was still working on them. Think I got them all now.

flogross89: looks good now, thanks!

flogross89 (approved):
Looks good to me syntax-wise. However I can't verify the fuel system changes, need @Maximilian-Reuter or someone else for that

flogross89:
@BlueberryKing FYI, maybe we need to change something on the SD FUEL page?

2hwk (approved):

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -213,39 +215,39 @@ Line.50 = Name:RightMidTankPumpAftToAftGalleryJunction#Source:RightMidTankPumpAf
...
-Line.79 = Name:FeedTank3FwdXferValve1_2ToFeedTank3#Source:FeedTank3FwdTransferValve1_2#Destination:Feed3
-Line.80 = Name:FeedTank4FwdXferValve1ToFeedTank4#Source:FeedTank4FwdTransferValve1#Destination:Feed4
-Line.81 = Name:FeedTank1FwdXferValve2ToFeedTank1#Source:FeedTank1FwdTransferValve2#Destination:Feed1
-Line.82 = Name:FeedTank2FwdXferValve2_1ToFeedTank2#Source:FeedTank2FwdTransferValve2_1#Destination:Feed2
-Line.83 = Name:FeedTank3FwdXferValve2_1ToFeedTank3#Source:FeedTank3FwdTransferValve2_1#Destination:Feed3
-Line.84 = Name:FeedTank2FwdXferValve2_2ToFeedTank2#Source:FeedTank2FwdTransferValve2_2#Destination:Feed2
-Line.85 = Name:FeedTank3FwdXferValve2_2ToFeedTank3#Source:FeedTank3FwdTransferValve2_2#Destination:Feed3
-Line.86 = Name:FeedTank4FwdXferValve2ToFeedTank4#Source:FeedTank4FwdTransferValve2#Destination:Feed4
+Line.75 = Name:FeedTank1FwdXferValve1ToFeedTank1#Source:FeedTank1FwdTransferValve1#Destination:Feed1#FuelFlowAt1PSI:0.001
+Line.79 = Name:FeedTank2FwdXferValve1_1ToFeedTank2#Source:FeedTank2FwdTransferValve1_1#Destination:Feed2#FuelFlowAt1PSI:100.0
```
Maximilian-Reuter, line 240: should be 76

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -143,23 +145,23 @@ point.22 = 2,79.979874,0.0,21.776155,100,0,0.0001,0,1,0.0001,0,0,0,4,0,0,1 ; Bod
...
+Tank.8 = Name:RightMid#Title:FUEL RIGHT MID#Capacity:9632#UnusableCapacity:0#Position:7.1,45.9,5.9#Priority:3#InputOnlylines:RightMidFwdXferValveToRightMidTank,RightMidAftXferValve2ToRightMidTank#OutputOnlyLines:RightMidTankToRightMidTankPumpFwd,RightMidTankToRightMidTankPumpAft
+Tank.9 = Name:Feed4#Title:FUEL FEED FOUR#Capacity:7299.6#UnusableCapacity:0#Position:-7.45,71,7.3#Priority:5#InputOnlyLines:FeedTank4FwdXferValve1ToFeedTank4,FeedTank4FwdXferValve2ToFeedTank4,FeedTank4AftXferValve2ToFeedTank4#OutputOnlyLines:FeedTank4ToFeed4TankPump1,FeedTank4ToFeed4TankPump2
+Tank.10 = Name:RightOuter#Title:FUEL RIGHT OUTER#Capacity:2731.5#UnusableCapacity:0#Position:-25.0,100,8.5#Priority:2#InputOnlyLines:RightOuterFwdXferValveToRightOuterTank,RightOuterAftXferValve2ToRightOuterTank#OutputOnlyLines:RightOuterTankToRightOuterTankPump,RightOuterTankToRightOuterEmerXferValve
+Tank.11 = Name:Trim#Title:FUEL TRIM#Capacity:6260.3#UnusableCapacity:0#Position:-87.14,0,12.1#Priority:1#InputOnlyLines:TrimTankInletValveToTrimTank#OutputOnlyLines:TrimTankToTrimTankPumpLeft,TrimTankToTrimTankPumpRight
```
Maximilian-Reuter, line 164: The inputline doesnt exist anymore and probably should refer to the two new lines

Maximilian-Reuter (changes requested):

Maximilian-Reuter (approved):

frankkopp:
> Not for this PR, but @frankkopp will this affect your EFB load states preset (ready to taxi, ready to takeoff, etc.), since we are changing the .flt spawn states? May need a patch for that after this critical fix.

Not sure - if numbers for pumps/valves did not change its not likely. 

See here:
https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/base/flybywire-aircraft-a380-842/config/a380x/a380-842/aircraft_preset_procedures.xml#L284

Not sure I find time before the weekend to check.

donstim:
Fixed a few more issues that I found when I was able to do a long flight. Will have to repeat that flight to confirm it is indeed fixed.

donstim:
Added missing triggers to start equalizing fuel in the tanks during fuel transfers when total fuel in the mid-tanks is less than 8000 kg (<4000 kg in the left tank is used as the trigger for this in MSFS). Previously, the equalizing process was only started when fuel is transferred from the trim tank (but not the trim tank cg transfers).

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -399,44 +415,50 @@ Pump.18 = Name:RightInnerTankPumpAft#Pressure:20#DestinationLine:RightInnerTankP
...
 Trigger.6 = Name:Equalize2and3#Target:Feed2,Feed3#Threshold:2#Condition:TankAbsImbalanceBelow#EffectTrue:OpenValve.FeedTank2FwdTransferValve2_1,OpenValve.FeedTank3FwdTransferValve2_1,OpenValve.FeedTank2FwdTransferValve2_2,OpenValve.FeedTank3FwdTransferValve2_2,OpenValve.FeedTank2AftTransferValve1,OpenValve.FeedTank3AftTransferValve1
-Trigger.7 = Name:Innerand MidTanksXferFeed1and4End#Target:Feed1#Threshold:6765#Condition:TankQuantityAbove#EffectTrue:CloseValve.FeedTank1FwdTransferValve1,CloseValve.FeedTank1FwdTransferValve2,CloseValve.FeedTank4FwdTransferValve1,CloseValve.FeedTank4FwdTransferValve2,SetJunction.FwdGalleryJunction.1,SetJunction.FeedTanks2_3Junction1.1,SetJunction.FeedTanks2_3Junction1.1
+Trigger.7 = Name:Innerand MidTanksXferFeed1and4End#Target:Feed1#Threshold:6765#Condition:TankQuantityAbove#EffectTrue:CloseValve.FeedTank1FwdTransferValve1,CloseValve.FeedTank1FwdTransferValve2,CloseValve.FeedTank4FwdTransferValve1,CloseValve.FeedTank4FwdTransferValve2,SetJunction.FwdGalleryJunction.1  ;
```
Maximilian-Reuter, line 425: no comment after comment symbol so probably better to remove it

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -399,44 +415,50 @@ Pump.18 = Name:RightInnerTankPumpAft#Pressure:20#DestinationLine:RightInnerTankP
...
-Trigger.11 = Name:InnerTankLeftEmpty#Target:LeftInner#Threshold:0.1#Condition:TankQuantityBelow#EffectTrue:StopPump.LeftInnerTankPumpFwd,StartPump.LeftMidTankPumpFwd,CloseValve.LeftInnerAftTransferValve2,OpenValve.LeftMidAftTransferValve2#EffectFalse:OpenValve.LeftInnerAftTransferValve2#EffectFalse:StopPump.LeftMidTankPumpFwd,StartPump.LeftInnerTankPumpFwd
-Trigger.12 = Name:InnerTankRightEmpty#Target:RightInner#Threshold:0.1#Condition:TankQuantityBelow#EffectTrue:StopPump.RightInnerTankPumpFwd,StartPump.RightMidTankPumpFwd,CloseValve.RightInnerAftTransferValve2,OpenValve.RightMidAftTransferValve2#EffectFalse:OpenValve.RightInnerAftTransferValve2#EffectFalse:StopPump.RightMidTankPumpFwd,StartPump.RightInnerTankPumpFwd
+Trigger.10 = Name:InnerandMidTanksXferFeed1and4EndRedundant#Target:Feed4#Threshold:6766#Condition:TankQuantityAbove#EffectTrue:CloseValve.FeedTank1FwdTransferValve1,CloseValve.FeedTank1FwdTransferValve2,CloseValve.FeedTank4FwdTransferValve1,CloseValve.FeedTank4FwdTransferValve2,SetJunction.FwdGalleryJunction.1  ;
```
Maximilian-Reuter, line 428: same as above

Maximilian-Reuter (changes requested):

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -341,42 +354,45 @@ Valve.19 = Name:LeftOuterFwdTransferValve#Circuit:19
...
 Valve.39 = Name:LeftOuterAftTransferValve2#Circuit:39
-Valve.40 = Name:RightInnerAftTransferValve2#Circuit:40
-Valve.41 = Name:RightMidAftTransferValve2#Circuit:41
+Valve.40 = Name:RightInnerAftTransferValve2#Circuit:40#DestinationLine:RightInnerAftXferValve2ToRightInnerTank
+Valve.41 = Name:RightMidAftTransferValve2#Circuit:41#DestinationLine:RightMidAftXferValve2ToRightMidTank
 Valve.42 = Name:RightOuterAftTransferValve2#Circuit:42
-Valve.43 = Name:TrimTankInletValve#Circuit:43
-Valve.44 = Name:TrimLineIsolationValveFwd#Circuit:44
-Valve.45 = Name:TrimLineIsolationValveAft#Circuit:45
+Valve.43 = Name:TrimTankInletValve1#Circuit:43#DestinationLine:TrimTankInletValve1ToTrimTank#Source
```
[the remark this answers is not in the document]
donstim: Thanks, fixed

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
donstim: Thanks, fixed

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
donstim: Thanks, fixed

Maximilian-Reuter (approved):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -0,0 +1,431 @@
...
+import { FuelSystemEvents } from 'systems-host/systems/FuelSystemPublisher';
+
+/**
+ * This is needed to initialize the desired fuel L:Var on load to sync with the fuel quantity as per the fuel state management per ATC ID
+ * This is a temporary solution until all fuel state related ops are contained in the same module.
+ */
+
+/* TODO: remove this file after proper FQMS is implemented in Rust */
+export class LegacyFuel implements Instrument {
+  private readonly NUMBER_OF_TRIGGERS = 42;
```
Saschl, line 20: Nitpick: I believe it's better to make those static and then reference them like `NUMBER_OF_TRIGGERS` (wthout `this`).

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -0,0 +1,431 @@
...
+    if (
+      (this.trimTankQty.get() < 0.1 && !this.triggerStates.get(32).get()) ||
+      (this.trimTankQty.get() >= 1 && this.triggerStates.get(32).get())
+    ) {
+      this.toggleTrigger(32);
+    }
+  }
+
+  public onUpdate(): void {
+    const onGround = SimVar.GetSimVarValue('SIM ON GROUND', 'bool');
```
Saschl, line 130: 2 Questions:

1. Do we not need to wait for GameStart here?
2. You mentioned you wanted to throttle the update rate here, is that still something that's missing or did you decide against it?

Saschl, line 130: Might consider moving this to a publisher as well(SIM_ON_GROUND is also used elsewhere on systems-host), but maybe not for this PR

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -0,0 +1,431 @@
...
+    if (this.refuelStarted.get()) {
+      this.refuelInProgress = true;
```
Saschl, line 132: NItpick: maybe call it something like `refuelWasInProgress`, had to think a bit what the reasoning of this field is

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Maximilian-Reuter: I honestly forgot I could do that :D

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Maximilian-Reuter: In my experience, onUpdate is already not called before gamestart. As for throttling, the actions are throttle already in, that they only activate once a change is necessary. I also didnt see any real performance impact as the real work is getting all the updated simvar values.

tracernz: Even more nitpick, anything using SIM ON GROUND should really be in extras-host because it's not an aircraft system (they use other aircraft systems to determine on ground/in-flight status).

tracernz: onUpdate is definitely called before the gamestate is ingame.

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Maximilian-Reuter: Yes the idea was to notice when fuelling has ended. But the var itself really contains the information whether fueling is happening right now. As I turn it to false once I detect that fuelling isnt happening anymore. Hence I felt that this name is more appropriate.

2hwk:
Other than cleaning up the review comments as above, these changes seem to have a positive impact on ensuring fuel is not imbalanced and doesn't get "stuck" anywhere.

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Maximilian-Reuter: I was kinda going back and forth on that and ended up just using the existing file. In the end this code isnt designed to stay around for long (hopefully) and be replaced by our own fqms.

Maximilian-Reuter: fair enough, though the code in on update doesnt really execute anyways until off the ground. So I wager the only situation where you might get issues is during an airspawn, probably why I havent seen any yet

PR #9071, opened 2024-10-15
fix(a380x): Flight model update by heckup and engine performance boost

flogross89:
## Summary of Changes

- Modify elevator properties to remove auto-take off / early rotation effect (heckup's optimization)
- Modify engine thrust to better match known perf. figures: Increase thrust for N1 above 80% and mach above 0.5 by 35%

## Screenshots (if necessary)
Method: Compare N1 at FL300 with two different gross weights to known performance figures
Before: 
image

After: 
image
image

## References
RL pilot

## Additional context


2hwk:
* Modify elevator properties to remove auto-take off / early rotation effect (heckup's optimization)

This will also need to be accounted for in the flight computers, as it will compromise normal law protections at low speed during and after rotation (before gear up). In general the autoflight is very fragile at the moment.

flogross89:
> * Modify elevator properties to remove auto-take off / early rotation effect (heckup's optimization)
> 
> This will also need to be accounted for in the flight computers, as it will compromise normal law protections at low speed during and after rotation (before gear up). In general the autoflight is very fragile at the moment.

Can you elaborate, why auto flight is very fragile? After the cobra fix, that’s not my impression. During flight, you have to try quite hard to move it outside the protected envelope. During takeoff, you’re in ground mode until a certain time/height

flogross89:
> > > * Modify elevator properties to remove auto-take off / early rotation effect (heckup's optimization)
> > > 
> > > This will also need to be accounted for in the flight computers, as it will compromise normal law protections at low speed during and after rotation (before gear up). In general the autoflight is very fragile at the moment.
> > 
> > 
> > Can you elaborate, why auto flight is very fragile? After the cobra fix, that’s not my impression. During flight, you have to try quite hard to move it outside the protected envelope. During takeoff, you’re in ground mode until a certain time/height
> 
> I'll try this again under more controlled conditions, wasn't a very scientific test.

luke reproduced this on master, so it isn't an issue which was introduced by this PR, but maybe exacerbated

2hwk:
> > > > * Modify elevator properties to remove auto-take off / early rotation effect (heckup's optimization)
> > > > 
> > > > This will also need to be accounted for in the flight computers, as it will compromise normal law protections at low speed during and after rotation (before gear up). In general the autoflight is very fragile at the moment.
> > > 
> > > 
> > > Can you elaborate, why auto flight is very fragile? After the cobra fix, that’s not my impression. During flight, you have to try quite hard to move it outside the protected envelope. During takeoff, you’re in ground mode until a certain time/height
> > 
> > 
> > I'll try this again under more controlled conditions, wasn't a very scientific test.
> 
> luke reproduced this on master, so it isn't an issue which was introduced by this PR, but maybe exacerbated

Yes it was something I did notice occasionally before, but this PR has exacerbated the issue imo. Just something to note.

2hwk (approved):

PR #9102, opened 2024-10-19
fix(a380x/fuel): Fix fuel system update rate and balancing issues

Maximilian-Reuter:
Fixes #[issue_no]

## Summary of Changes

This is a small fix to make sure the update loop in the fuelsystem code isn't too fast with high glasscockpit refreshrates.
I also deleted an duplicate class.
Also fixes an inadvertent change to the FWSCore instantiation in SystemHost

Thanks to further investigation by @donstim we were also able to fix an imbalance during the outer tank transfer.
Cleaned up an left over OUTR TK FUEL XFER memo from the 320 too.

## Additional context

## Testing instructions

1. set cockpit glass refreshrate to HIGH
2. do a testflight and make sure fuel is properly transfered to the feed tanks


fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -127,6 +138,11 @@ export class LegacyFuel implements Instrument {
   }
 
   public onUpdate(): void {
+    const dt = this.sysHost.deltaTime;
+    if (!this.hasInit || !this.throttler.canUpdate(dt)) {
```
Saschl, line 142:
```suggestion
    if (!this.hasInit || !this.throttler.canUpdate(dt) !== -1) {
```

flogross89, line 142: canUpdate returns -1 if you can't update, and a cycle time > 0 if you can.

```suggestion
    if (!this.hasInit || this.throttler.canUpdate(dt) <= 0) {
```

flogross89 (changes requested):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Saschl: I like your suggestion better :D

flogross89: Haha :D I recently implemented your proposal, but forgot the minus, so I checked for ``=== 1``, which of course didn't work. That's why I'm now moving to >0 or <= 0

Maximilian-Reuter: good catch, that is me being unattentive and assuming :D the name made me think it was boolean. Though I wager it would still work :D

flogross89: Afaik negative numbers evaluate to true, so no it wouldn’t have worked 😄

flogross89 (approved):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Saschl: Yep :D https://developer.mozilla.org/en-US/docs/Glossary/Truthy

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -1,4 +1,4 @@
-import { MathUtils } from '@flybywiresim/fbw-sdk';
+import { MathUtils, UpdateThrottler } from '@flybywiresim/fbw-sdk';
```
2hwk, line 1: UpdateThrottler should be deprecated in favour of using ``ClockEvents`` i.e. 

```

    const sub = this.bus.getSubscriber<ClockEvents>();
    let lastUpdateTime: number;
    sub
      .on('realTime')
      .onlyAfter(1000)
      .handle((now) => {
        const dt = lastUpdateTime === undefined ? 0 : now - lastUpdateTime;
        lastUpdateTime = now;
        this.update(dt);
      });
```

Add a TODO FIXME: for this

2hwk (changes requested):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -127,6 +135,11 @@ export class LegacyFuel implements Instrument {
   }
 
   public onUpdate(): void {
+    const dt = this.sysHost.deltaTime;
+    if (!this.hasInit || this.throttler.canUpdate(dt) <= 0) {
+      return;
+    }
```
2hwk, line 141: TODO FIXME: Use ClockEvents instead of deprecated UpdateThrottler

2hwk (dismissed):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
Maximilian-Reuter: I am more with @tracernz on this one, adding some sort of deltatime or update rate to the actual backplane instrument cause doing this in each on yourself seems counterintuitive to the original idea

Saschl: I agree, maybe let's just put the FIXME inside index.ts of the systemshost to have it noted down somewhere.

flogross89: Yes, I also wouldn't know why UpdateThrottler itself is deprecated

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
flogross89: This is not relevant anymore, is it?

PR #9105, opened 2024-10-20
fix(a380x): fixed strange behaviors in FBW computers

aguther:
## Summary of Changes
Fixes some weird behaviors in the FBW computers (i.e. double activation of PRIM with different trim target values, middle aileron drooping more than the others).

## Testing instructions
Fly the airplane, ensure no cobra is happening.


flogross89 (approved):
I'm getting the same diff now when rebuilding the models, don't know what happened there.

PR #9119, opened 2024-10-22
feat(a380x/mfd): remap comma to dot

beheh:
## Summary of Changes

Many countries in the world do not use dot as their decimal separator (anything that isn't blue):
World Map with countries colored by decimal separator

In those countries it is common for the numpad to have the appropriate decimal separator on the numpad (e.g. most of Europe).
When focussing an input field in the MFD and using the numpad to enter decimals on such a keyboard layout there is two common issues, explained e.g. on a German keyboard layout where the decimal separator is a comma:
- the comma is not available in the MFD font and a whitespace is rendered
- the decimal is not parsed after blurring the input (due to a FORMAT ERROR)

This is very annoying when one is used to entering e.g. frequencies via the numpad. So for now, this PR just replaces any entered comma with a dot. This should not be a problem, as the comma is not a valid key on the KCCU.

Whether you *should* be able to enter keys that are not on the KCCU at all (basically anything that isn'z A-Z, Space, Dot, plus or minus) is another question, and one that we could tackle in future. We could decide to just outright ignore all of those (probably a future PR).

*Sidenote*: if this were a real browser we could detect the numpad decimal key by checking for `KeyboardEvent.keyCode == KeyCode.KEY_DECIMAL`. That would then not interfere if you actually wanted to enter the comma explicitly. Unfortunately Coherent doesn't expose this an just exposes both as a generic comma.

## Screenshots (if necessary)

Before:
image

After:
image

## Additional context

## Testing instructions

1. Set your Windows keyboard layout to one where the decimal separator is a comma (e.g. German).
2. Focus any input on the MFD (e.g. FMS 1/2 > POSITION > NAVAIDS > FREQ
3. Enter a decimal number entirely via the numpad on your keyboard. You should notice the decimal key being consistenly entered as a dot.

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/common/InputField.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/common/InputField.tsx
@@ -190,10 +190,14 @@ export class InputField<T> extends DisplayComponent<InputFieldProps<T>> {
     // ev.key is undefined, so we have to use the deprecated keyCode here
     const key = String.fromCharCode(ev.keyCode).toUpperCase();
 
-    if (ev.keyCode !== KeyCode.KEY_ENTER) {
-      this.handleKeyInput(key);
-    } else {
+    if (ev.keyCode === KeyCode.KEY_ENTER) {
       this.handleEnter();
+    } else if (key === ',') {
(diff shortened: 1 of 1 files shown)


beheh:
Replaced by #9130

PR #9137, opened 2024-10-24
fix(A380x): Turned off PB lights for INOP systems on upper OVHD 

Steveveepee:
Turned off emissives for no simulated systems with BPs on upper overhead.

Decoupled the overhead NSS AVNCS switch from the Cockpit Door Video Switch (some previous copypasta).

## Screenshots (if necessary)

overhead

Discord username (if different from GitHub): Shrike


flogross89 (approved):

flogross89:
Better than before. GND HF DATALINK doesn't seem to have SEQ1 or SEQ2, so it's still lit. 
image

Testing Results:
Passed

PR #9138, opened 2024-10-24
fix(380/fuel): reset valvestates when refueling

Maximilian-Reuter:
Fixes #[issue_no]

## Summary of Changes

This PR tries to resolve an issue where after landing the fuel system is in a state where it cannot be refueled properly as the valves aren't in the correct position.

The current solution is to simply reset the valves to their default position and then have the fuel system recalculate the way they should be triggered.

## Additional context

## Testing instructions

1. Spawn c/d
2. load some fuel for a flight, in either fast or real
3. monitor fuel to see if it loads what is requested
4. check no fuel gets transfered after fueling ends
5. takeoff
6. check that fuel transfers are happening if required
7. monitor fuel until landing to make sure nothing weird happens
8. land
9. after landing and parking up try loading fuel again in either fast or real speed
10. monitor fuel to make sure it all happens as expected

Bonuspoints: Repeat test but spawn on the runway instead


donstim:
LGTM

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -136,22 +142,30 @@ export class LegacyFuel implements Instrument {
...
       for (let index = 1; index <= LegacyFuel.NUMBER_OF_TRIGGERS; index++) {
         if (this.triggerStates.get(index).get()) {
           this.keyEventManager.triggerKey('FUELSYSTEM_TRIGGER_OFF', true, index);
         }
       }
-    } else if (this.refuelInProgress) {
+      // starts at 5 since 1-4 are the engine valves controlled by the engine masters
+      // valve 37, 40, 50, 51 should remain unchanged
+      for (let index = 5; index <= LegacyFuel.NUMBER_OF_VALVES; index++) {
+        if (![37, 40, 50, 51].includes(index)) {
```
flogross89, line 161: This should only execute once, if refueling just started, right? If it would run every frame, we should avoid the array allocation and rather should check for the inequality with these 4 valves in four statements

flogross89: Could you also explain in the comment why it's these four valves?

flogross89 (approved):

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
tracernz: Yeah, don't allocate objects in onUpdate if at all avoidable. You can make this array a private static readonly member of the class.

Maximilian-Reuter: I was wondering how it would do it internally, I can add it statically ofc

fbw-a380x/src/systems/systems-host/systems/LegacyFuel.ts
```diff
@@ -19,6 +24,10 @@ import { FuelSystemEvents } from 'systems-host/systems/FuelSystemPublisher';
 export class LegacyFuel implements Instrument {
   private static NUMBER_OF_TRIGGERS = 42;
   private static NUMBER_OF_JUNCTIONS = 13;
+  private static NUMBER_OF_VALVES = 60;
+
+  //these Valves are set to true in the FLT files so we dont want to set them to false
```
tracernz, line 29:
```suggestion
  /** These Valves are set to true in the FLT files so we dont want to set them to false. */
```
Better to make it a JSDoc so people will see it when they mouse over the property anywhere else.

tracernz (approved):

donstim:
Latest build tested spawning at KLAX, short flight with 25 T of FOB and one transfer from the outer tanks, return to KLAX. Shut down, left fuel pumps and APU on, refueling on either fast or real worked fine.

PR #9146, opened 2024-10-26
fix: undo change impacting performance in cpp logging

frankkopp:
## Summary of Changes
A previous change in the logging library of the cpp framework caused a 10 fold increase in run time.
The expectation that the compiler optimizes away empty functions was not working out. 

Reverted to the previous version of logging code which also works well. 

Also added the {fmt] library and made logging.h use this instead (avoids issue with multiple lines in console when using << stream)

## Testing instructions
Should not have any functional impact.

Framework is used by the FADEC, Pushback, Presets code. 


Saschl (approved):
LGTM, thanks for the quick fix!

tracernz (approved):
We should move fmtlib up to fbw-common/src when we can figure out how to convince cmake about it, but it's not a blocker. LGTM for me.

tracernz:
Needs 0.13 changelog.

Saschl:

Discord Username : saschl
Object of testing : https://github.com/flybywiresim/aircraft/pull/9146
Aircraft : A380 and A320
Tier of Testing : 1
Date : 27/10/2024

Testing Process:

Spawned C&D. Did the setup and started the engines. Performed a pushback  without issues. Also runtime of the wasm modules is back to normal.

Negatives: N/A
Testing Results:
Passed

PR #9166, opened 2024-10-28
docs: Fix a380x inop page wording

flogross89:
## Summary of Changes

Improves wording on the INOP page. Previous version lead to misunderstanding. 
Removes ATA chapter headings for which no entries exist. 

## Additional context

### DIFF
--- a/fbw-a380x/README.md
+++ b/fbw-a380x/README.md
@@ -15,29 +15,28 @@ This is essentially a clean build.
 
 These systems are currently not operational / these features are missing and shall thus not lead to bug reports.
 
-### ATA 20 Aircraft General / Exterior 3D Model
+This list is divided into standardized ATA chapters, if there are no noteworthy systems missing from a chapter, the chapter heading is hidden (this was different in a previous version of this list). This list is not exhaustive, there might be details missing in other systems which are not stated here.
 
-### ATA 21 Air Con / Pressurization / Ventilation
+<!-- ### ATA 20 Aircraft General / Exterior 3D Model -->
 
-### ATA 22 Flight Envelope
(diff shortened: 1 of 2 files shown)


fbw-a380x/README.md
```diff
@@ -15,48 +15,48 @@ This is essentially a clean build.
...
-- Complete SLAT/FLAP CTL implementation (adapted A32NX computers atm)
-- Automatic pitch trim setting
+- Completely migrated SLAT/FLAP CTL implementation (adapted A32NX computers at the moment)
+- Automatic THS setting on ground
 
 ### ATA 28 Fuel
+
 - Refuel Driver/Refuel Application placeholder pseudo-FQMS implementation (pre-cursor for realistic FQMS implementation while supporting instant load)
   - CPIOM etc. not implemented yet.
 - Automatic fuel transfer system - Not fully implemented yet i.e. AGT off when 2 engines on
```
donstim, line 62: This particular item has led someone to believe that automatic fuel transfers are not implemented at all. In any case, I believe it is unclear as it refers to an acronym that is not explained as part of only one case that doesn't fully explain what is not implemented vs what is.

I suggest something like - "Portions of the automatic fuel transfer system -- main transfers from the inner, mid, trim, and outer tanks are implemented with the correct scheduling as are CG control transfers from the trim tank to the appropriate inner, mid, or feed tanks. Not implemented yet are load alleviation transfers, transfers based on remaining time to destination, cold fuel transfers, and automatic transfers that occur on the ground."

donstim:
For the manual fueling being inop entry, I think it would be clearer to say that "Manually controlling how much fuel each tank receives" is inop.

PR #9306, opened 2024-11-02
fix(a380x): FMS, ECL, BTV fallback fixes

flogross89:
## Summary of Changes

- Fix color of mandatory FMS fields: Not amber anymore when set to 0
- Automatically check DEPARTURE CHANGE ECL as completed, until deferred checklists are implemented
- Don't allow AIRWAYS vertical revisions to be called for airports and runway legs
- Disable CMS vertical revision menu entry for now
- Lower FMS input field auto-blur time from 30s to 20s
- Don't allow FMS input field keystroke if not focused

## Additional context

## Testing instructions

1. Enter "0" into FUEL&LOAD -> PAX NBR field, value should be cyan not amber
2. Open up ECL in EWD, should show DEPARTURE CHANGE as completed already
3. Try opening up AIRWAYS revisions for airport or runway legs, shouldn't work
4. Click on ALTN RTE field on INIT page and press keys on physical keyboard, shouldn't work


fbw-common/src/systems/instruments/src/OANC/BrakeToVacateUtils.ts
```diff
@@ -328,7 +328,8 @@ export class BrakeToVacateUtils<T extends number> {
     this.btvExitPosition = btvExitPosition;
 
     // Account for touchdown zone distance
-    const correctedStoppingDistance = reqStoppingDistance - MIN_TOUCHDOWN_ZONE_DISTANCE;
+    // Don't subtract touchdown zone distance for now, creates confusion for users
+    const correctedStoppingDistance = reqStoppingDistance;
```
tracernz, line 332: This should be either correct or not. Doesn't really matter if people are confused when it works like the real thing.

flogross89: This is for the Navigraph-less BTV fallback, where users enter a braking distance manually. So there's no real thing anyways, this line is just about how we want to define the LDA or rather where the BTV stop target is.

Before: braking distance was measured from 400m, i.e. static TDZ distance
After: Braking distance now measured from threshold, far easier to read from charts from my POV

I'd say after the change we're more correct, since it resembles the definition of LDA (beginning at the threshold)

flogross89: improved the comment now

flogross89: Btv, I removed this change from here, I refactored most of the code here: https://github.com/flybywiresim/aircraft/pull/9355

Saschl:
Dev Team Testing Report

Object of testing : https://github.com/flybywiresim/aircraft/pull/9306
Aircraft : A380X
Tier of Testing : 1
Date : 03/11/2024

Testing Process:

Enter "0" into FUEL&LOAD -> PAX NBR field, value should be cyan not amber. Check
Open up ECL in EWD, should show DEPARTURE CHANGE as completed already. Check
Try opening up AIRWAYS revisions for airport or runway legs, shouldn't work.  Check
Click on ALTN RTE field on INIT page and press keys on physical keyboard, shouldn't work. Check

Tested auto unfocus after 20s. Check


Negatives:

N/A
Testing Results:
QA Passed

PR #9323, opened 2024-11-02
fix(a32nx/lights): Strobe lights flashing when switched from off to auto on ground

### LINKED ISSUE #8647: Strobes flash once when switch from OFF to AUTO
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-04-27T16:40:29+00:00",
    "ref": "refs/heads/master",
    "sha": "7518ab0c54b296f4cb774519889171ed7d8b4df5",
    "actor": "Benjozork",
    "event_name": "manual",
    "pretty_release_name": "master:7518ab0c",
    "version": "a32nx-v0.12.0-dev.7518ab0"
}
```

### Describe the bug

Strobe lights flash once when turned from OFF to AUTO.

### Expected behavior

Strobe lights should only strobe in AUTO when the mean gear strut is not compressed.

### Steps to reproduce

1. Load into the aircraft C&D at night and next to reflective surfaces.
2. Basic power up
3. Move strobe light switch from OFF to AUTO
4. Observe described behavior

### References (optional)

https://github.com/flybywiresim/aircraft/assets/58574351/e27aff07-dc2b-4780-b268-8829c95c2604

_No response_

examplewastaken

MicahBCode:
Fixes #8647

## Summary of Changes

Fixes an issue where the strobe lights flash once if the aircraft is on the ground (main gear compressed) and the strobe lights are switched from off to auto
In the auto position the Potentiometer 24 is set to 0 if the main gear is compressed or to 100 if it is not compressed. Update rate: 1/s

To fix this the Potentiometer should now also be set to 0 if the strobe light switch is in the off position.

## Additional context

## Testing instructions

1. Load in the aircraft on the ground (a32nx)
2. Power up the aircraft
3. Check that the strobe lights do not flash (in the off position)
4. Switch the strobe light switch from off to auto and monitor the strobe lights (lights should not flash)
5. Switch the strobe light switch to on (Lights should flash now in the expected interval)
6. Switch the strobe light back to auto and takeoff
7. Check that the strobe lights flash as soon as airborne (in the expected interval)
8. Switch the strobe light switch to on and check that the strobe lights are still flashing
9. Switch the strobe light switch to off (strobe lights should not flash anymore)

### DIFF
--- a/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO_INTERIOR.xml
+++ b/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/model/A320_NEO_INTERIOR.xml
@@ -1689,7 +1689,7 @@
                 <!-- EXT LT STROBE -->
                 <UseTemplate Name="FBW_Anim_Interactions">
                     <ANIM_TYPE>SWITCH</ANIM_TYPE>
-                    <ANIM_TEMPLATE>ASOBO_LIGHTING_Switch_Light_Strobe_Template</ANIM_TEMPLATE>
+                    <ANIM_TEMPLATE>FBW_Switch_Light_Strobe</ANIM_TEMPLATE>
                     <NODE_ID>SWITCH_OVHD_EXTLT_STROBE</NODE_ID>
                     <PART_ID>EXTLT_STROBE</PART_ID>
                     <ANIM_NAME>SWITCH_OVHD_EXTLT_STROBE</ANIM_NAME>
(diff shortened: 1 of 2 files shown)


MicahBCode:
If anyone knows why this change causes the strobe lights to flash only once instead of twice (as it should) let me know. Was not able to figure that out as of now.

MicahBCode:
> I've been looking into this too, and I thought it was strange that the existing LIGHT_UPDATE_CODE sets the potentiometer to 100 as long as the switch is not in auto

Exactly, that was my thought as well. Due to this LIGHT_UPDATE_CODE only being triggered once per second, there is a small time gap of < 1s in which the potentiometer is not set to the correct value, which causes this issue.

> With your fix, I'm guessing that the update code might cut off the animation too fast.

I don't think this is the case since the LIGHT_UPDATE_CODE only then changes the Potentiometers value if the switch is actually being switched but not If you just do nothing and leave it in the ON position for example.

Seems strange to me that this change to the LIGHT_UPDATE_CODE changes the behaviour of the animation itself (or its emitters).

PR #9324, opened 2024-11-02
fix(fdr2csv): refactor to unify fdr2csv, fix conversion bug

lukecologne:
## Summary of Changes

This PR fixes a bug in the new FDR2CSV, where a wrong type was used in the FdrData struct was used, leading to garbled data when decoded. It also refactors the FDR2CSV to autodetect the aircraft type and decode accordingly, improving usability.
Also adds an option `--get_raw_input_file_version`/`-r` to print out only the file version, for automation purposes, and makes `-o` optional when `-r` or `-g` are present
Further, fix a bug in the FDR itself, where a float value was written to the INI instead of a string of int.

## Additional context


tools/fdr2csv/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+    buf_writer.write(header.as_bytes())?;
+
+    // Create the CSV writer, and serialize the file.
+    let mut writer = WriterBuilder::new()
+        .delimiter(args.delimiter as u8)
+        .has_headers(false)
+        .from_writer(buf_writer);
+
+    match aircraft_type {
+        AircraftType::A320 => {
```
lukecologne, line 145: If any rustaceans have feedback on how to do this better, I'd love to hear it

pwnorbitals: I think you might be able to define a `read_record_function` function pointer 
```
let read_record_function = match aircraft_type { AircraftType::A320 =>  {a320::read_record}, AircraftType::A380 => { a380::read_record } }
```
Then use that in your common workflow. Another approach could be to define a `read_record` function in your AircraftType that returns the right value depending on the variant.

lukecologne: I tried defining the function as a variable, but it doesn't work as the return types of the functions are different

lukecologne: And defining the return type as Box<dyn Serialize> also doesn't work:
```
    |
54  | pub fn read_record(reader: &mut impl Read) -> Result<Box<dyn Serialize>, Error> {
    |                                                          ^^^^^^^^^^^^^ `a320::_::_serde::Serialize` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
```

aguther (approved):
- loaded up A32NX and A380X and created FDR files
- converted them using the fdr2csv
- checked the data for plausibility

Test succeeded. Approved.

PR #9351, opened 2024-11-03
fix(nd): reuse the terrain NanoVG texture and fix memory leak

Nufflee:
## Summary of Changes

Previously, the terrain NanoVG texture was recreated on every terrain update (several times per second) but this is not necessary and we can instead reuse the same texture all the time as the size of the terrain image never changes (there is a failsafe in place in case this assumption is broken). In order to do this, I had to use the `stbi` library to manually parse the PNG data and then upload it to the existing texture.

This also seems to fix the reported memory leak within MSFS (not within SimBridge) when using TERR display, but to confirm this further, long haul testing would be needed. It appears that creating and then deleting a texture causes a memory leak within the MSFS NanoVG backend.

## Additional context

## Testing instructions

1. Start a several hour long flight in the A380X (the longer the better)
2. Note the `CPU Mem` value in the FPS display (`DevMode -> Debug -> Display FPS`)
3. Turn the TERR display on and keep it on for the rest of the flight, ideally on both NDs
4. Observe the `CPU Mem` usage throughout the flight, it should not rise rapidly or increase significantly compared to the value noted at the start (during the cruise at least)
5. Ensure that the TERR display looks the same as before, without any visual glitches or similar

image

### DIFF
--- a/fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
+++ b/fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
@@ -139,6 +139,7 @@ class Display : public DisplayBase {
         if (decodedWidth != width || decodedHeight != height) {
           // This should never happen, but bail just in case
           fprintf(stderr, "TERR ON ND: The image size does not match the expected size. Expected: %dx%d, actual: %dx%d\n", width, height, decodedWidth, decodedHeight);
+          stbi_image_free(decodedImage);
           return;
         }
 
--- a/fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
+++ b/fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
@@ -6,6 +6,7 @@
 #pragma clang diagnostic ignored "-Wsign-conversion"
 #include <MSFS/Legacy/gauges.h>
 #include <MSFS/Render/nanovg.h>
+#include <MSFS/Render/stb_image.h>
 #pragma clang diagnostic pop
 #include <cstdint>
 #include <string_view>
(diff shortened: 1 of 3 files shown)


2hwk (approved):

frankkopp (approved):

fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
```diff
@@ -113,14 +114,37 @@ class Display : public DisplayBase {
...
-
       if (!this->_ignoreNextFrame && this->_configuration.terrainActive) {
-        this->_nanovgImage =
-            nvgCreateImageMem(this->_context, 0, this->_frameData->data().data(), static_cast<int>(this->_frameBufferSize));
         if (this->_nanovgImage == 0) {
-          std::cerr << "TERR ON ND: Unable to decode the image from the stream" << std::endl;
+          // If we don't have an image yet, create one
+          this->_nanovgImage = nvgCreateImageMem(this->_context, 0, this->_frameData->data().data(), static_cast<int>(this->_frameBufferSize));
+          if (this->_nanovgImage == 0) {
+            fprintf(stderr, "TERR ON ND: Unable to decode the image from the stream. Reason: %s\n", stbi_failure_reason());
```
Saschl, line 122: Any reason not to use a stream for printing the error like before?

fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
```diff
@@ -113,14 +114,37 @@ class Display : public DisplayBase {
...
+        uint8_t* decodedImage = stbi_load_from_memory(this->_frameData->data().data(), static_cast<int>(this->_frameBufferSize), &decodedWidth, &decodedHeight, nullptr, 4);
+        if (decodedImage == nullptr) {
+          fprintf(stderr, "TERR ON ND: Unable to decode the image from the stream. Reason: %s\n", stbi_failure_reason());
+          return;
         }
+
+        int width, height;
+        nvgImageSize(this->_context, this->_nanovgImage, &width, &height);
+
+        if (decodedWidth != width || decodedHeight != height) {
```
Saschl, line 139: Does the image not need a cleanup in this case as well?

Saschl (commented):
Just some questions from my side

fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
Nufflee: That's because streams add a new line on every `<<` in MSFS which makes the log hard to read so this seems like the easiest way to get around that.

fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
Nufflee: Good catch, fixed. If only we had defer..

fbw-common/src/wasm/terronnd/src/navigationdisplay/display.h
frankkopp: We have the fmt lib now available. Check it out.

frankkopp: > That's because streams add a new line on every `<<` in MSFS which makes the log hard to read so this seems like the easiest way to get around that.

You can use this as IU did in the cpp framework:
fbw-common/src/wasm/cpp-msfs-framework/lib/fmt
Example:
fbw-common/src/wasm/cpp-msfs-framework/lib/logging.h

Saschl:
Dev Team Testing Report

Object of testing : https://github.com/flybywiresim/aircraft/pull/9351
Aircraft : A380X
Tier of Testing : 1
Date : 03/11/2024

Testing Process:

Took off and flew for 4 hours with Terrain Display on
Observed CPU usage at random intervals, usage was stable
Could not notice any visible changes or glitches in the display.

MSFS RAM usage after 4 hours

Screenshot 2024-11-03 190025


vs 30 minutes with the current state without this PR:

Screenshot 2024-11-03 212414


Negatives:

N/A
Testing Results:
QA Passed

frankkopp:
Dev Team Testing Report

Object of testing : https://github.com/flybywiresim/aircraft/pull/9351
Aircraft : A380X
Tier of Testing : 1
Date : 03/11/2024

Testing Process:

Below a video from before when noticed the issue:

https://github.com/user-attachments/assets/1b285bcd-a0ad-4212-a445-04d8dff9bd39

This is after your PR:

https://github.com/user-attachments/assets/2cc31bbd-145b-4037-b3f2-c5e1775bbe8c

Negatives:

N/A
Testing Results:
QA Passed

alepouna:
This PR is milestone 0.12.2, maybe move the changelog or good to merge?

alepouna:
nvm confirmed on disc its cherry picked

PR #9414, opened 2024-11-05
feat(A380X/BAT): Move BAT to avionics framework

ngoerlitz:
## Summary of Changes

Converts the BAT instrument component from React to the MSFS avionics framework. 

## Screenshots (if necessary)

Visually the component is identical to the previous implementation. 

bat_sw

## References

- Previous implementation
- Video of IRL usage

## Additional context

When moving the BAT-SW from OFF to one of the other positions, the value resets to '00.0V', before showing the actual voltage. Currently, the '00.0V' are set directly, as soon as the BAT-SW is turned to the OFF position. 

Discord username (if different from GitHub): nikolas_g

## Testing instructions

1. Load the A380X
2. Navigate to the upper overhead and move the switch position
3. Compare with LValues `L:A32NX_ELEC_BAT_{1,2,3,4}_POTENTIAL`

- Simply build the BAT component (for example using `mach build -f BAT`


flogross89:
Thanks! Please make sure to keep the architecture we had until now, with common parts being in fbw-common, and only the aircraft-specifics being in fbw-a32nx and fbw-a380x respectively. The displays themselves should behave pretty much the same, except different colors.

Benjozork:
@flogross89 Do you have a comment on the above?

tracernz:
> @flogross89 Do you have a comment on the above?

I would suggest a data provider for each aircraft that supplies the subscribables for shared code like this. That way the aircraft can have different interfaces with other systems, and even some different logic in the data provider if needed. The shared code should be more generic. Another approach is an abstract shared class containing generic code, and each aircraft implements the subscribables and some aircraft-specific bits in an aircraft-specific subclass if there are quite a few differences.

flogross89:
@tracernz Thank you for the guidance! 
@ngoerlitz Apologies for not replying! Feel free to ping me on Discord if I’m slow with replying.

PR #9425, opened 2024-11-06
fix(a380x/mfd): Accept 0°C temperature values

FDM-Tech:
Fixes #[issue_no]

## Additional context

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/common/DataEntryFormats.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/common/DataEntryFormats.tsx
@@ -435,7 +435,16 @@ export class TemperatureFormat implements DataEntryFormat<number> {
       return null;
     }
 
-    const nbr = Number(input);
+    let nbr = Number(input);
+    
+    if (nbr > 0 && input.substring(0, 1) !== '+') {
+      nbr *= -1;
+    }
+    
+    if (nbr > 0 && input.substring(0, 1) === '+') {
+      nbr = input.substring(1);
+    }
+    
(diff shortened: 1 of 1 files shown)


fbw-a380x/src/systems/instruments/src/MFD/pages/common/DataEntryFormats.tsx
```diff
@@ -435,7 +435,16 @@ export class TemperatureFormat implements DataEntryFormat<number> {
       return null;
     }
 
-    const nbr = Number(input);
+    let nbr = Number(input);
+    
+    if (nbr > 0 && input.substring(0, 1) !== '+') {
```
beheh, line 440: This doesn't seem to make much sense. This will make everything negative unless you explicitly prefix with +.

fbw-a380x/src/systems/instruments/src/MFD/pages/common/DataEntryFormats.tsx
```diff
@@ -435,7 +435,16 @@ export class TemperatureFormat implements DataEntryFormat<number> {
...
+      nbr *= -1;
+    }
+    
+    if (nbr > 0 && input.substring(0, 1) === '+') {
+      nbr = input.substring(1);
```
beheh, line 445: I don't understand what this is trying to do. At this point nbr has already successfully been parsed, and you're now replacing it with a string (without parsing).

beheh (changes requested):
Thanks for your contribution. 

However, have you tested this? The changes didn't really make much sense to me as they only apply when the parsed number is larger than 0.

As part of the PR please also fill out the template, especially with proper test instructions.

alepouna:
Please fill the PR template as BehEh mentioned - remove Invalid after filled.

PR #9438, opened 2024-11-07
fix(a380x/mfd): Fixed DIRECT TO selection of active waypoint does nothing

sognodelx:
## Summary of Changes

DIRECT TO selection of active waypoint results in index 0 and thus onModified leads to no action. Changed the condition to check for null in the onModified action to allow selection of active waypoint.

## Screenshots (if necessary)

image

## References

image

## Additional context

## Testing instructions

In flight with an active flight plan select the active waypoint as DIRECT TO.

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnDirectTo.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnDirectTo.tsx
@@ -164,7 +164,7 @@ export class MfdFmsFplnDirectTo extends FmsPage<MfdFmsFplnDirectToProps> {
                     containerStyle="width: 175px;"
                     alignLabels="flex-start"
                     onModified={(i, text) => {
-                      if (i) {
+                      if (i != null) {
                         this.onDropdownModified(i, text);
                       }
                     }}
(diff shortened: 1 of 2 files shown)


Benjozork (approved):

sognodelx:
Sorry, had to force push the commit again so that the correct e-mail address is set (no changes in content).

fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnDirectTo.tsx
```diff
@@ -164,7 +164,7 @@ export class MfdFmsFplnDirectTo extends FmsPage<MfdFmsFplnDirectToProps> {
                     containerStyle="width: 175px;"
                     alignLabels="flex-start"
                     onModified={(i, text) => {
-                      if (i) {
+                      if (i != null) {
```
flogross89, line 167:
```suggestion
                      if (i !== null && i !== undefined) {
```
Thanks for the fix, well spotted! We should avoid juggling checks, as they could be perceived as unintentional. I'd rather check for null and undefined explicitly

flogross89 (changes requested):

nebj00la:

Discord Username  : nebj00la
Object of testing : #9438
Aircraft          : A380X
Tier of Testing   : 1
Date              : 07/11/2024

Testing Process: 

1. Loaded various waypoints for approach to KBOS.
2. Flew HDG after a new waypoint was set, then loaded DIRECT to the same waypoint.

Testing Results: 
Passed

Media:

flogross89 (approved):

fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnDirectTo.tsx
sognodelx: Saw it in the live stream with capta320 and riiiva, in which you were also present. I asked myself exactly whether I should do an explicit null check. Thx for the feedback, even if it is a small change.

flogross89: Sure, thanks for the fix. We will include it in version 0.12.2, which we're going to release soon.

PR #9453, opened 2024-11-08
feat(A380X/FADEC): Basic METOTS-Implementation

### LINKED ISSUE #9283: Implement METOTS system for A380X
### Aircraft Version

Stable

### Description

When taking off, the FADEC limits the maximum thrust of the engines to 78% N1 until speed is above 35 knots to prevent flutter on the engine's front fan blades.

### References (optional)

https://www.youtube.com/watch?v=8d5q-J6ZEwo

I would like to help work on and implement such a system into the aircraft. Please contact me on discord if you're working on the engine model, I'd love to collaborate.

tankman69

cptnuss-ops:
Fixes #9283 

## Summary of Changes

Implemented METOTS into the FADEC-Logic of the A380X. Till 35kts ground speed, 78% N1 should not be exceeded, even with thrust lever in toga.

## Additional context

Besides the video in the issue, there are no other reference I could find. I'm quite sure, the limits on the ENG should display the "normal" N1-limits. This is an issue right now, because the limit for the display seems the be read from the variable i set to apply the N1-Limit in the FADEC. I currently working on a solution for that.

Discord username (if different from GitHub): cpt.nuss

## Testing instructions

Load plane and apply toga, observe N1 beeing held at 78% until GS > 35kts.


fbw-a380x/src/wasm/fadec_a380x/src/Fadec/FadecSimData_A380X.hpp
```diff
@@ -136,45 +136,47 @@ class FadecSimData_A380X {
...
-      {"TURB ENG N1",                  2, UNITS.Percent  }, //
-      {"TURB ENG N1",                  3, UNITS.Percent  }, //
-      {"TURB ENG N1",                  4, UNITS.Percent  }, //
-      {"TURB ENG N2",                  1, UNITS.Percent  }, //
-      {"TURB ENG N2",                  2, UNITS.Percent  }, //
-      {"TURB ENG N2",                  3, UNITS.Percent  }, //
-      {"TURB ENG N2",                  4, UNITS.Percent  }, //
+      {"ANIMATION DELTA TIME",          0, UNITS.Seconds  }, //
+      {"AIRSPEED MACH",                 0, UNITS.Mach     }, //
+      {"SURFACE RELATIVE GROUND SPEED", 0, UNITS.knots    }, //
```
cptnuss-ops, line 155: same here: added "surfaceRelativeGroundSpeed", whole block changed due to format change

flogross89:
Thanks for the proposal! This solution might be undesirable, two reasons I can think of:
- It seems the THR limit is changed below a certain speed? This would lead to a false THR limit being displayed in the EWD
- The transition of the FMC flight phase (needed for flight plan sequencing) depends on the engine N1, we would have to check whether that properly activates (PERF/TO page becoming ineditable). Should work though

I would rather propose to intercept the commanded N1 and modify that with the given conditions. Btw, there are two conditions:
Keep Out Zone: N1 doesn't stay between 64% and 72% when below 60kts. It's OK to transition through that area though
METOTS: THR below 78% when below 35kts. Keep in mind that this seems to be w.r.t. to thrust, which is not estimated in the FADEC right now. So METOTS would also only be very rudimentary right now.

lukecologne:
A few comments:
* METOTS is indeed related to N1 (as it's to prevent fan instability, so N1 is the relevant parameter).
* METOTS is related to CAS, not GS. GS would not really be a sensible thing to use, as again, it's to prevent fan instability so the aerodynamic speed is relevant, not ground speed.
* The thrust limit is unaffected by METOTS, as you already suspected.

The way to implement this would be in our Autothrust Simulink model, as that already contains a very basic N1 control law, no additional intercepting is needed.

Tankman61:
I've also checked out the implementation in sim and it seems to not work with in FLEX or climb detent. So it's definitely not as simple as an if statement
Screenshot 2024-11-08 224128

beheh (commented):
Instead of depending on the magical `SURFACE RELATIVE GROUND SPEED`, would it be possible to use the ADRs for this? They have A429 simvars you should be able to access, not unlike some other SimVars such as `A32NX_APU_N_RAW` below.

Issue #9436, opened 2024-11-07
ND Airpot display filter


ksleungac:
### Aircraft Version

Stable

### Build info

```json
{
    "built": "2024-11-02T19:03:34+00:00",
    "ref": "refs/tags/v0.12.1",
    "sha": "2776100c286d3723826c62df36dc48207657d937",
    "actor": "Benjozork",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.1",
    "version": "a380x-v0.12.1-rel.2776100"
}
```

### Describe the bug

When displaying airports on the ND while using **'ARPT'** button on the EFIS, airports with no suitable runways for the A380 are shown.

### Expected behavior

Normally, airports with shorter runways is filtered out from the ND display.

IRL, this runway length threshold is a value set by the operator. (Thanks @BravoMike99 for the information!)

### Steps to reproduce

1. Select ARPT on the EFIS for the ND.
2. Observe so many small airports with no suitable runway shown.

### Discord Username (optional)

pzb-85

ksleungac:
I've been reading through the source of which I've found in `fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts` 
Where, for the ND ARPT display, the function `hasSuitableRunway` has hardcoded runway length of 1500 meters for filtering (likely an value for the A320).
```Typescript
    const hasSuitableRunway = (airport: Airport): boolean =>
      airport.longestRunwayLength >= 1500 && airport.longestRunwaySurfaceType === RunwaySurfaceType.Hard;
```

As consulted with the fellow development team member, the src for the A380x fmgc is located and shared with A32nx, and we would need to watch for the regression while changing sections as such.

Therefore, I think it would be best to not hard-code a value for a specific aircraft, rather to just let the user enter a value for the runway threshold in the Flypad - Settings - AircraftOptionsPinPrograms page, and take that value as the value for determining if an airport has a suitable runway. In this way I can keep this code working even if the refactor happens nor if affects a32nx.

so, 
1) I will be looking into EFB code to add such option to take in that threshold value
2) I will take that value as the filter in `hasSuitableRunway`
3) I will try setting some default value for both a32nx/a380x.

I am new to here, and anyway having any kind of advice is welcome :)

BravoMike99:
[...]

PR #9458, opened 2024-11-09
feat(efb/efis): minimum runway length option to filter airports on the EFIS/ND.

ksleungac:
Fixes [#9436](https://github.com/flybywiresim/aircraft/issues/9436)

## Summary of Changes

Added an Minimum Runway Length option in the EFB (Aircraft PIN Programs). Then take that value to filter on airport displayed on the ND when selection ARPT on the EFIS.

#### EFB Setting
6E7A257C90FA5E5DFD80E3816FB11D72

#### Minimum Runway Length = 1500
7E6C7289A9CE65D479E0C9326491D7A6

#### Minimum Runway Length = 1500
ADF8ED89184ABE6D2379DC592B16B828

## Additional context

I will try to work on the default values for this setting to be separate for A32NX/A380X. It's my first PR on the opensource community, feel free to comment on the coding styles, implementations or anything :)

Discord username (if different from GitHub): pzb-85

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
@@ -18,6 +18,7 @@ import {
   EfisSide,
   Arinc429SignStatusMatrix,
   Arinc429OutputWord,
+  NXDataStore,
 } from '@flybywiresim/fbw-sdk';
 
 import { Coordinates } from '@fmgc/flightplanning/data/geo';
@@ -187,8 +188,13 @@ export class EfisSymbols<T extends number> {
     const vnavPredictionsChanged = this.lastVnavDriverVersion !== this.guidanceController.vnavDriver.version;
     this.lastVnavDriverVersion = this.guidanceController.vnavDriver.version;
 
-    const hasSuitableRunway = (airport: Airport): boolean =>
-      airport.longestRunwayLength >= 1500 && airport.longestRunwaySurfaceType === RunwaySurfaceType.Hard;
+    const hasSuitableRunway = (airport: Airport): boolean => {
+      const minRwyLengthDisplay = parseInt(NXDataStore.get('CONFIG_MIN_RWY_LENGTH', '1500'));
+      return (
+        airport.longestRunwayLength >= minRwyLengthDisplay &&
+        airport.longestRunwaySurfaceType === RunwaySurfaceType.Hard
+      );
+    };
 
     for (const side of EfisSymbols.sides) {
       const range = this.rangeValues[SimVar.GetSimVarValue(`L:A32NX_EFIS_${side}_ND_RANGE`, 'number')];
--- a/fbw-common/src/systems/instruments/src/EFB/Localization/data/en.json
+++ b/fbw-common/src/systems/instruments/src/EFB/Localization/data/en.json
@@ -573,6 +573,7 @@
     "AircraftOptionsPinPrograms": {
       "AccelerationHeight": "Acceleration Height (ft)",
       "EngineOutAccelerationHeight": "Engine-Out Acceleration Height (ft)",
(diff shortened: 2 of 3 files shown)


fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -187,8 +188,13 @@ export class EfisSymbols<T extends number> {
     const vnavPredictionsChanged = this.lastVnavDriverVersion !== this.guidanceController.vnavDriver.version;
     this.lastVnavDriverVersion = this.guidanceController.vnavDriver.version;
 
-    const hasSuitableRunway = (airport: Airport): boolean =>
-      airport.longestRunwayLength >= 1500 && airport.longestRunwaySurfaceType === RunwaySurfaceType.Hard;
+    const hasSuitableRunway = (airport: Airport): boolean => {
+      const minRwyLengthDisplay = parseInt(NXDataStore.get('CONFIG_MIN_RWY_LENGTH', '1500'));
```
tracernz, line 192: It would be better to store this in a class property and use NXDataStore.getAndSubscribe rather than parsing a string every time we check a runway (a lot!).

tracernz (commented):
We discussed this quite a bit back when this feature was first implemented and decided we didn't want to add yet another setting for something like this (we already have too many settings for things that aren't options on the real aircraft!). I think the only thing that might change that is if there's evidence this is settable on the real aircraft FMS (e.g. AMI or OPC database)

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
ksleungac: Thank you for your suggestion. Definitely trying that to save some resources :)

ksleungac:
> we already have too many settings for things that aren't options on the real aircraft!.

I can understand that no one would want to pack the EFB with dozens of options pages like PMDG did with their FMC (Though maybe there are more things customizable on a Boeing).

> I think the only thing that might change that is if there's evidence this is settable on the real aircraft FMS (e.g. AMI or OPC database)

Bit of speculation here but as Bruno mentioned in the [comment](https://github.com/flybywiresim/aircraft/issues/9436#issuecomment-2461985819). The filtering is probably done on the database, of which we don't have yet. So we can't do much before that is implemented.

It's just that I want a temporary solution. Probably rather than exposing this as an EFB option, we just hard-code another value, let's say 2500/3000 metres for the A380 specifically.

2hwk:
I think it would be better to leave it as an EFB option than hardcode this. Maybe something that could be improved in the future to match reality closer.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -78,6 +79,8 @@ export class EfisSymbols<T extends number> {
 
   private lastEfisInterfaceVersions: Record<EfisSide, number> = { L: -1, R: -1 };
 
+  private minRwyLengthDisplay: number = parseInt(NXDataStore.get('CONFIG_MIN_RWY_LENGTH', '1500'));
```
2hwk, line 82: Here the default value is 1500, but in [fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AircraftOptionsPinProgramsPage.tsx](https://github.com/flybywiresim/aircraft/pull/9458/files#diff-d316475b8413cbbdc58316f48ae0b80eb70cd558c4f46152264a46c8ccab5b81) the default value is 3000.

2hwk (changes requested):

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
ksleungac: I aligned them to be 1500 now.

PR #9470, opened 2024-11-09
feat(GPU): Implemented more realistic GPU handling

Maximilian-Reuter:
Fixes #[issue_no]

## Summary of Changes

This PR intends to make the GPU handling on both aircraft more realistic and immersive. 
It mainly does that by decoupling the GPU connection state from the MSFS vars.

As a result you will now be able to disconnect the GPU from the plane even when on a powered stand.

I also fixed an issue in the version checker for the a32nx.

In order to facilitate communication from the EFB via the eventbus to the extrashost, I integrated an eventbus instance into the EFB Wrapper

## Additional context

## Testing instructions

Since this PR implements the same function for both planes the tests will be the same:

1. spawn C/D at a gate that you know has power 
2. after spawning check that the EXT Power buttons show AVAIL
3. check that the EFB Ground page shows GPU connected
4. click the GPU button on the EFB page
5. AVAIL lights should turn off
6. click it again
7. AVAIL lights should come back on
8. request pushback through some means
9. pushback
10. AVAIL lights should turn off the moment you start moving
11. Go back to the main menu and start a new flight at a stand that you know has no power
12. after spawning check that the EXT Power buttons dont show AVAIL
13. check that the EFB Ground page shows GPU disconnected
14. click the GPU button on the EFB page
15. button should turn orange and the msfs GPU cart should be called
16. if airport has a GPU cart then it should evenetually arrive
17. once it is hooked up the button should turn green and the EXT Power buttons should show AVAIL
18. click GPU button on EFB again
19. GPU cart should be sent away and AVAIL lights should turn off


fbw-common/src/systems/shared/src/publishers/Extras/ExtrasSimVarPublisher.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2023 FlyByWire Simulations
```
tracernz, line 1: I don't think we should add any more of these "publish all sorts of random stuff" publishers. Each publisher should just publish the topics for one subsystem.

Maximilian-Reuter: I understand that and was wondering about that, in this case though unless I misunderstand I would end up with like 3 single var publishers. Or do you mean putting all the vars that I use for the GPU stuff into one publisher ?

alepouna:
Are you able to also connect the plane to power if the gate doesn't have gp available? (from the efb gpu btn)

Maximilian-Reuter:
> Are you able to also connect the plane to power if the gate doesn't have gp available? (from the efb gpu btn)

Yes, if that airport supplies a GPU cart then clicking the EFB button will call it. If not then that call wont be answered. That is the current behavior as well so I didnt change that. 

Btw the testinstructions should tell you that ;)

alepouna:
> > Are you able to also connect the plane to power if the gate doesn't have gp available? (from the efb gpu btn)
> 
> Yes, if that airport supplies a GPU cart then clicking the EFB button will call it. If not then that call wont be answered. That is the current behavior as well so I didnt change that.
> 
> Btw the testinstructions should tell you that ;)

I asked for confirmation because specifically there isn't a mention if there is "phantom power" if you like but good to know now 👍

fbw-common/src/systems/shared/src/publishers/Extras/ExtrasSimVarPublisher.ts
tracernz: They grow over time as code gets refactored to improve quality and new code written. There's no problem with having one var at the start. It doesn't make sense to duplicate simvar definitions over and over in every instrument.

For the MSFS ones there's https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/systems/shared/src/publishers/Msfs/. Pay careful attention to the JSDocs in the existing ones for how they should be setup... everything should exactly match the SDK docs, including naming, units, and which publisher things go in. That way it's easy to find things, and there's no room for debate, and with the msfs_ prefix no chance of collisions with our events.
```ts
/**
 * Events for simvars listed on https://docs.flightsimulator.com/html/Programming_Tools/SimVars/Aircraft_SimVars/Aircraft_AutopilotAssistant_Variables.htm.
 * Event names are the same as the simvar names, with msfs_ prefix, and index as suffix for indexed simvars.
 */
 ```

File: fbw-a380x/src/wasm/systems/a380_systems_wasm/src/lib.rs [hunk 2]
```diff
@@ -582,13 +581,9 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
             );
 
             builder.copy(
-                Variable::aircraft("EXTERNAL POWER AVAILABLE", "Bool", i),
+                Variable::named(&format!("A32NX_EXT_PWR_AVAIL:{i}")),
```
2hwk, line 584: Might want to define "A32NX_" as key_prefix i.e.

```rs
let key_prefix = "A32NX_";
...
Variable::named(&format!("{}_EXT_PWR_AVAIL:{}", key_prefix, i)),
```

2hwk: I'm not even sure if the `A32NX_` should be necessary in rs, typically it should not be needed unless I'm missing something.

2hwk (changes requested):

tracernz:
Why do we need to use the event bus to toggle the GPU? Can't we just intercept the MSFS K event (with no passthrough so it doesn't go to the sim), and handle it our way? This means it will work for external clients as well e.g. Mobiflight, SPAD etc.

fbw-a380x/src/wasm/systems/a380_systems_wasm/src/lib.rs
Maximilian-Reuter: I am honestly unclear how this prefix thing works, so i am just doing what works and cant wait for the day the prefix finally dies

Maximilian-Reuter:
> Why do we need to use the event bus to toggle the GPU? Can't we just intercept the MSFS K event (with no passthrough so it doesn't go to the sim), and handle it our way? This means it will work for external clients as well e.g. Mobiflight, SPAD etc.

Hmm, not sure why I didnt think of that. Guess I was really deep into the eventbus and just wanted to use it for it :D 
I guess I didnt expect people to want to bind stuff that happens on the EFB

fbw-a380x/src/wasm/systems/a380_systems_wasm/src/lib.rs
Gurgel100: For a `named` variable the prefix is automatically added. So correct would be:
```suggestion
                Variable::named(&format!("EXT_PWR_AVAIL:{i}")),
```

Maximilian-Reuter: yep i noticed that now and changed it accordingly

fbw-a380x/src/systems/extras-host/index.ts
```diff
@@ -88,8 +111,15 @@ class ExtrasHost extends BaseInstrument {
...
 
-    this.backplane.addInstrument('Clock', this.clock);
+    this.backplane.addPublisher('SimvarPublisher', this.simVarPublisher);
+    this.backplane.addPublisher('MsfsElectricsPublisher', this.msfsElectricsPublisher);
+    this.backplane.addPublisher('MsfsFlightModelPublisher', this.msfsFlightModelPublisher);
+    this.backplane.addPublisher('MsfsMiscPublisher', this.msfsMiscPublisher);
+    this.backplane.addPublisher('GroundSupportPublisher', this.groundSupportPublisher);
+
     this.backplane.addInstrument('PilotSeatManager', this.pilotSeatManager);
+    this.backplane.addInstrument('GPUManagement', this.gpuManagement);
```
2hwk, line 121:
```suggestion
    this.backplane.addInstrument('GpuManagement', this.gpuManagement);
```

Minor correction for camel case (i.e. MSFS is also an abbreviation but MsfsElectricsPublisher is correct)

2hwk (approved):

fbw-a380x/src/systems/instruments/src/SD/Pages/ElecAc/elements/ExternalPower.tsx
```diff
@@ -25,10 +25,10 @@ export const ExternalPower: FC<ExternalPowerProps> = ({ x, y, bus }) => {
 
   const [lineContactorClosed] = useSimVar(`L:A32NX_ELEC_CONTACTOR_990XG${bus}_IS_CLOSED`, 'number', 500);
 
-  const [extPwrConnected] = useSimVar(`EXTERNAL POWER AVAILABLE:${bus}`, 'number', 500);
+  const [extPwrConnected] = useSimVar(`EXT_PWR_AVAIL:${bus}`, 'number', 500);
```
Saschl, line 28: Shouldn't those be L:A32NX prefixed?

Maximilian-Reuter: you are totally right! Good catch!

LeDerpLegend:

Discord Username : LeDerpLegend
Object of testing : https://github.com/flybywiresim/aircraft/pull/9470
Aircraft : A32NX
Simulator: 2020
Tier of Testing : 1
Date : 24/11/2024
Testing Process:
1.     spawn C/D at a gate that you know has power
2.     after spawning check that the EXT Power buttons show AVAIL
3.     check that the EFB Ground page shows GPU connected
4.     click the GPU button on the EFB page
5.     AVAIL lights should turn off
6.     click it again
7.     AVAIL lights should come back on
8.     request pushback through some means
9.     pushback
10.     AVAIL lights should turn off the moment you start moving
11.     Go back to the main menu and start a new flight at a stand that you know has no power
12.     after spawning check that the EXT Power buttons dont show AVAIL
13.     check that the EFB Ground page shows GPU disconnected
14.     click the GPU button on the EFB page
15.     button should turn orange and the msfs GPU cart should be called
16.     if airport has a GPU cart then it should evenetually arrive
17.     once it is hooked up the button should turn green and the EXT Power buttons should show AVAIL
18.     click GPU button on EFB again
19.     GPU cart should be sent away and AVAIL lights should turn off


Testing Results:
Steps worked as intended, AVAIL was off and overhead left light was off when GPU was unavailable. AVAIL and overhead left light turn on as expected when GPU cart is connected. GPU disconnects the moment pushback starts to move. 

Testing Results:
PASSED

Negatives:
N/A

Notes:
When using GSX, some airports may require use of GSX over MSFS GPU at airports that do not provide a GPU cart by MSFS. 
Some airports provide GPU by default at a parking spot that shouldn't have one. (Sim issue?)

Issue #9561, opened 2024-11-20
A32NX: pending airway legs do not display on the ND


robertxing2004:
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-11-19T23:32:19+00:00",
    "ref": "HEAD\n2f774e9202935c8b4616b9fe66166a36889d7d13 refs/remotes/upstream/HEAD",
    "sha": "2f774e9202935c8b4616b9fe66166a36889d7d13\n2f774e9202935c8b4616b9fe66166a36889d7d13",
    "actor": "Robert Xing <73909958+robertxing2004@users.noreply.github.com>",
    "event_name": "manual",
    "pretty_release_name": "master:2f774e92",
    "version": "a32nx-v0.13.0-dev.2f774e9"
}
```

### Describe the bug

Pending airway legs do not display on the ND (top image), however it's worth noting that pending DIRs (bottom left) and SIDs/STARS (bottom right) do display correctly.

<p float="left">
  <img src="https://github.com/user-attachments/assets/dee9e1ac-653f-4c88-acbd-b36c61d9d425" width="800" />
  <img src="https://github.com/user-attachments/assets/6d54e3c1-b1f7-4ebe-9d74-d4c94363f2b4" width="400" /> 
  <img src="https://github.com/user-attachments/assets/46466386-9461-44ab-90b1-4d7af9c061e5" width="400" />
</p>

### Expected behavior

After a terminating waypoint is entered onto the end of a pending airway entry, the segment should be displayed on the ND.

### Steps to reproduce

1. Spawn anywhere and start a fpln from the MCDU
2. Enter an airway segment with a terminating WPT and observe the ND

### Discord Username (optional)

robeet

tracernz:
This report is confirmed correct on a Honeywell H4 FMS.

Jonny23787:
I will fix this issue

Issue #9560, opened 2024-11-19
AWY/WPT mismatch does not trigger in airways page


robertxing2004:
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-11-19T23:32:19+00:00",
    "ref": "HEAD\n2f774e9202935c8b4616b9fe66166a36889d7d13 refs/remotes/upstream/HEAD",
    "sha": "2f774e9202935c8b4616b9fe66166a36889d7d13\n2f774e9202935c8b4616b9fe66166a36889d7d13",
    "actor": "Robert Xing <73909958+robertxing2004@users.noreply.github.com>",
    "event_name": "manual",
    "pretty_release_name": "master:2f774e92",
    "version": "a32nx-v0.13.0-dev.2f774e9"
}
```

### Describe the bug

If an invalid waypoint is entered in the TO field after entering an airway, the MCDU displays the create new waypoint page and gets stuck on said page (the return LSK does not work).

<p float="left">
  <img src="https://github.com/user-attachments/assets/19265179-62fa-4baf-b134-f682088fb5d8" width="400" />
  <img src="https://github.com/user-attachments/assets/09daae15-979b-451e-8e51-997788ee8bd7" width="400" />
</p>

### Expected behavior

When an invalid waypoint is entered in the TO field, the scratchpad should display AWY/WPT mismatch.

### Steps to reproduce

1. Spawn anywhere, start fpln from MCDU
2. In airways page, after entering an airway, deliberately enter an incorrect terminating waypoint

### References (optional)

![Screenshot 2024-11-19 184322](https://github.com/user-attachments/assets/705eb4ef-926c-4189-9a6a-5b5cd99932c3)

### Discord Username (optional)

robeet

tracernz:
Duplicate of https://github.com/flybywiresim/aircraft/issues/9040

PR #9148, opened 2024-10-26
fix(CDU): suppress TMPY FPLN when no modification made in airways page

robertxing2004:
Fixes #8910 

## Summary of Changes

Corrected FMS logic to prevent TMPY FPLN from appearing after leaving airways page without making any modifications.

## Screenshots (if necessary)

Refer to #8910 for previous incorrect behaviour

Corrected:
https://github.com/user-attachments/assets/76c8f6e1-591a-43c0-9f33-521591564ba4

## Additional context

Discord username (if different from GitHub):
robeet

## Testing instructions

1. Load anywhere, and start a flight plan from the MCDU
2. Open lateral revision page for a waypoint -> open airways page
3. Without making any modifications, press LSK6L (RETURN)
4. Should return to lat rev page -> press LSK6L (RETURN) again
5. Should return to flight plan page -> verify no TMPY FPLN state

Secondary:
1. Following step 3 above, return to airways page (LSK5R) and make a modification, then insert
2. Observe TMPY FPLN and confirm insert
3. Verify new modifications inserted correctly

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo** or **flybywire-aircraft-a380-842** download link at the bottom of the page

PR #9562, opened 2024-11-20
fix(cdu): correct tmply fpln activation logic in airways page

robertxing2004:
## Summary of Changes

Extension of #9148

The airways page now only creates a tmpy fpln after a valid segment is entered; either an airway and a terminating waypoint, or an airway followed by an intersecting airway. Prior to either of these two conditions being met, there is no pending flight plan, and pressing return will not result in a false tmpy state on the fpln page.

## Additional context

While troubleshooting this issue, I discovered bugs #9560 and #9561, but as it stands this PR only fixes the tmpy fpln activation logic. Those bugs can be addressed in a separate PR.

robeet

## Testing instructions

1. Spawn anywhere and start a fpln from the mcdu
2. After entering a STAR and transition, enter the airways page and make modifications
3. Verify that there is no option to erase or insert until a terminating waypoint or intersecting airway has been entered
4. Returning any time before the above two conditions should not create a tmpy fpln

### DIFF
--- a/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_AirwaysFromWaypointPage.js
+++ b/fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_AirwaysFromWaypointPage.js
@@ -98,11 +98,17 @@ class A320_Neo_CDU_AirwaysFromWaypointPage {
                     rows[i] = [`${pendingAirway.ident}[color]cyan`, "[\xa0\xa0\xa0][color]cyan"];
 
                     mcdu.onRightInput[i] = (value, scratchpadCallback) => {
-                        const targetPlan = mcdu.flightPlan(forPlan, inAlternate);
+                        const currentPlan = mcdu.flightPlan(forPlan, inAlternate);
 
                         if (value.length > 0) {
                             Fmgc.WaypointEntryUtils.getOrCreateWaypoint(mcdu, value, false).then(/** @param wp {import('msfs-navdata').Fix | undefined} */ (wp) => {
                                 if (wp) {
+                                    if (!currentPlan.pendingAirways) {
+                                        mcdu.flightPlanService.startAirwayEntry(prevFpIndex, forPlan, inAlternate);
+                                    }
+
+                                    const targetPlan = mcdu.flightPlan(forPlan, inAlternate);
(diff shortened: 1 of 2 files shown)


fbw-a32nx/src/base/flybywire-aircraft-a320-neo/html_ui/Pages/VCockpit/Instruments/Airliners/FlyByWire_A320_Neo/CDU/A320_Neo_CDU_AirwaysFromWaypointPage.js
```diff
@@ -102,11 +98,17 @@ class A320_Neo_CDU_AirwaysFromWaypointPage {
...
+                        const currentPlan = mcdu.flightPlan(forPlan, inAlternate);
 
                         if (value.length > 0) {
                             Fmgc.WaypointEntryUtils.getOrCreateWaypoint(mcdu, value, false).then(/** @param wp {import('msfs-navdata').Fix | undefined} */ (wp) => {
                                 if (wp) {
+                                    if (!currentPlan.pendingAirways) {
+                                        mcdu.flightPlanService.startAirwayEntry(prevFpIndex, forPlan, inAlternate);
+                                    }
+
+                                    const targetPlan = mcdu.flightPlan(forPlan, inAlternate);
```
BlueberryKing, line 110: How is `targetPlan` different from `currentPlan`? Will `targetPlan` be a temporary plan?

robertxing2004: They're the same thing; I originally was not declaring `currentPlan`, and only declaring `targetPlan`.

I ran into an issue where if I only assigned `const targetPlan = mcdu.flightPlan...` once, either at the top or after the `startAirwayEntry` then the respective results would be that either `targetPlan.pendingAirways.thenAirway` would be undefined or `if (!targetPlan.pendingAirways) {...}` would be undefined.

I'm pretty sure it's because `const targetPlan = mcdu.flightPlan...` needs to be assigned after `mcdu.flightPlanService.startAirwayEntry(prevFpIndex, forPlan, inAlternate);` or it won't pick up the new pending airway entry. At the same time though, if I don't assign `mcdu.flightPlan` before doing `if (!targetPlan.pendingAirways) {...}` then that if statement is undefined because `targetPlan` is shadowed under `mcdu.onLeftInput` and out of scope of `mcdu.onRightInput`.

Of course, I can't use the same variable name twice here if I'm reassigning, hence the two different names, and I'm admittedly not seeing a way of preserving the logic without redeclaring the variable.

Issue #9578, opened 2024-11-23
[A380X] Outflow valves not animated


flogross89:
### Aircraft Version

Stable

### Build info

```json
dev
```

### Describe the bug

Behaviors code reads in `<ANIM_CODE>(L:A32NX_PRESS_OUTFLOW_VALVE_OPEN_PERCENTAGE)</ANIM_CODE>`, which doesn't seem to be populated anymore after a recent PRESS change

### Expected behavior

Proposal: Publish `L:A32NX_PRESS_OUTFLOW_VALVE_1_OPEN_PERCENTAGE_ANIM` from the pressurization code to deliver a non-Arinc value to the xml code

### Steps to reproduce

-

### Discord Username (optional)

floridude

PR #9579, opened 2024-11-23
fix(a380x): Pressurization & AMU SimVar fixes 

flogross89:
## Summary of Changes

Two fixes for SimVars, gets rid of error messages in MSFS2024. Won't have a huge functional impact I guess though. 

Out of scope: ``L:A32NX_PRESS_OUTFLOW_VALVE_x_OPEN_PERCENTAGE_ANIM`` won't be populated, i.e. #9578 needs to be fixed with another PR

## Additional context


tracernz:
Can you please split these into PRs that do one thing? It's better for git blame, and also a lot better if one needs reverted.

Saschl (approved):

PR #9586, opened 2024-11-24
fix(a380x): Consider .flt file loading state for WASM startup

flogross89:
## Summary of Changes

Changes the startup and initialization phase for A32NX and A380X, which hopefully solves some problems for MSFS2024.
Now, A32NX_IS_READY is only set when the FLT file has been fully loaded. 

## References

FBW Startup (4)

## Additional context

## Testing instructions

MSFS2020 and MSFS2024: Load aircraft in different scenarios (apron, runway, in flight) and verify that the plane spawns as expected. 


tracernz:
I think it's a bit dangerous to ignore SimConnect messages like that, and will lead to confusing bugs. Loading .flt files will not work correctly for instance if we've intercepted/blocked a key event that's used to load the .flt file, and then ignored the event message.

flogross89:
> I think it's a bit dangerous to ignore SimConnect messages like that, and will lead to confusing bugs. Loading .flt files will not work correctly for instance if we've intercepted/blocked a key event that's used to load the .flt file, and then ignored the event message.

Which change is this comment referring to?

tracernz:
This one https://github.com/flybywiresim/aircraft/pull/9586/commits/7b714de32dce64028aa9c7861b46589744c1e3db

flogross89:
> This one [7b714de](https://github.com/flybywiresim/aircraft/commit/7b714de32dce64028aa9c7861b46589744c1e3db)

Ah ok, didn't find another way to do it, and it's also hugely inefficient for me to spend time on solving these Rust environment issues. Happy to have some of the Rust folks have a look at it

PR #9587, opened 2024-11-25
fix: harden airways code to not cause crashes

Saschl:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue that caused an endless loop (and sim crashes) in case of navdata issues.

## Additional context

### DIFF
--- a/fbw-common/src/systems/navdata/client/backends/Msfs/Mapping.ts
+++ b/fbw-common/src/systems/navdata/client/backends/Msfs/Mapping.ts
@@ -1449,7 +1449,12 @@ export class MsfsMapping {
       // eslint-disable-next-line prefer-destructuring
       let nextIcao: string | undefined = route.nextIcao;
 
+      let iterationCount = 0;
+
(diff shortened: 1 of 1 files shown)


fbw-common/src/systems/navdata/client/backends/Msfs/Mapping.ts
```diff
@@ -1449,7 +1449,12 @@ export class MsfsMapping {
       // eslint-disable-next-line prefer-destructuring
       let nextIcao: string | undefined = route.nextIcao;
 
+      let iterationCount = 0;
+
       while (previousIcao?.trim() || nextIcao?.trim()) {
+        if (++iterationCount > 100) {
```
tracernz, line 1455: I did a quick check; we should make this 250 to be on the safe side for all cases.

tracernz (approved):
No user facing change, no changelog 👍.

MJ-AmJay:
Footage: QA Passed for fix #9587
https://youtu.be/N_O5D6VBd8M

2hwk (approved):

PR #9528, opened 2024-11-15
feat(GSX): 380 Refuel/Door Sync, Harmonize GSX Code and External Power-Sync for 320+380

Fragtality:
Fixes #9383 [maybe 9287 too]

## Summary of Changes

Well I just wanted to Port PR 8407 to the 380 and get the Cargo Doors synched and it grew a bit bigger 😅 
- 380: Refuel/Defuel automatically starts when GSX Hose is connected
- 380: Instant Fuel allowed with GSX Fuel Sync
- 380: Cargo Doors are synced with GSX (being automatically opened and closed while Boarding/Deboarding)
- 380: FOB (Total) and ETA added to Fuel Page
- 380/320: Harmonized GSX Code (except Payload Synch) in shared Host/Module
- 380/320: Better handling for FOB >= planned (Defuel / No Action) Situations with GSX
- 380/320: Bettor Error-Handling while refueling (e.g. Couatl restarted)
- 380/320: Override GSX Progressive Refuel Setting to disabled
- 380/320: EFB Changes Label "Call GSX Refuel" to "GSX Refuel requested" when the GSX Refuel Service is requested (overall rework of format class/label logic)
- 380/320: Import SimBrief Button on Fuel Page is hidden when planned SB Value is set or refueling in progress (general)
- 380/320: Disable Slider and Input Box when (FBW) Refuel/Defuel Process is active
- 380/320: Play/Start Button now also hidden/disabled when Rate is Instant and GSX Refuel is requested
- 380/320 - already incorporating PR 9470 - New GSX Power Synch Feature (+EFB Option) to synchronize External Power Available with GSX

## Screenshots (if necessary)

image
image

## Additional context

Note for Cases with FOB >= planned/desired:
GSX can't detect if the Aircraft is defueling and has no Means to send the truck away - we need an Increase of Fuel.
So the proposed PR sets the desired Fuel temporarily to a lower Amount and starts Defuel. After that an instant Refuel to the actual desired Value is started. => GSX in any Case gets a Signal that Fuel is finished and the Truck will leave.
<br/>
Note for Error-Handling:
If the Fuel Hose gets disconnected during De/Refueling (GSX Crash, Simconnect Issues, User restarts Couatl), the Integration will reset the desired Fuel on the original Value. De/Refuel then still continues - the User can instantly finish De/Refuel by selecting Instant as Rate.
<br/>
Note on Power-Synch:
[...]

PR #9602, opened 2024-11-29
fix: implemented brute force closing of doors when tug is attached

### LINKED ISSUE #9405: Baggage & Catering cargo doors on aircraft failed to close
### Aircraft Version

Stable

### Build info

```json
{
    "built": "2024-11-02T19:03:13+00:00",
    "ref": "refs/tags/v0.12.1",
    "sha": "2776100c286d3723826c62df36dc48207657d937",
    "actor": "Benjozork",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.1",
    "version": "a380x-v0.12.1-rel.2776100"
}
```

### Describe the bug

I used the default baggage and catering vehicles at heathrow - I only watched the first 10secs of the animation, but couldnt get a clean T.O. readout for taxi - but I took off anyway.
5 mins into the flight I looked externally and saw the cargo doors were still open lol
Do I have to do the C/L to search/remember for this and how do I close the doors?
Baggage and Catering buttons were greyed out in flight

### Expected behavior

Doors to auto close - as I think they do in the a320.

### Steps to reproduce

1. Use Baggage in sim 
2. Use Catering in sim
Check both door statuses after ground vehicles leave. Did they auto close?

### References (optional)

{sorry passengers - it would seem your suitcase is now in the English channnel} :-)

_No response_

_No response_

frankkopp:
Fixes #9405 

## Summary of Changes
Ensures doors are closed when the tug is attached, the aircraft moves or is not on ground. 

TODO: 
- [ ] Check if the aircraft stationary check and effect has too much performance impact when taxiing
- [ ] Check if this clashed with GSX support

## Additional context
Check if this clashes with #9528 or the GSX support in general (although it addresses the issue that GSX is not able to handle higher door indexes).

## Testing instructions
Try to get into a state that doors stay open once pushback is attached or when in the air. 


beheh:
Is this explicitly when the tug is moving? Some people might want to attach the tug a few minutes ahead of push back while boarding is still running, just so it's ready to go later.

frankkopp:
> Is this explicitly when the tug is moving? Some people might want to attach the tug a few minutes ahead of push back while boarding is still running, just so it's ready to go later.

It is explicitly when the tug var is showing attached:
```ts
[pushBackAttached] = useSimVar('Pushback Attached', 'enum', 250);
```

Unfortunately this is set to true the moment you call the tug - not when it is actually visually attached. 

This is actually how it has been since the beginning of the flyPad Ground pages - I never changed this.
But there was an issue that you had to be on the Ground services page for it to work and also it only closed the doors controlled by the Services page. This is what this PR fixes - it closes all doors no matter on which flyPad page you are.

 I do understand you point but I think using this as trigger is the best solution.

fbw-common/src/systems/instruments/src/EFB/Efb.tsx
```diff
@@ -399,6 +399,25 @@ export const Efb: React.FC<EfbProps> = ({ aircraftChecklistsProp }) => {
     }
   }, [nwStrgDisc]);
 
+  // Check if the aircraft is on the ground, tug not attached and stationary if not close all doors
+  // This uses all available interactive point indices to close all doors - on the A80X there are 20 doors.
+  // Triggering an interaction point that is not defined is ignored therefore we can also use this for the A32NX.
```
tracernz, line 404: I don't think we should do things that generate errors inside the sim intentionally.

.github/CHANGELOG.md
```diff
@@ -98,6 +98,7 @@
 1. [ELEC] Improved elec system startup behaviour - @Gurgel100 (Pascal) - @saschl
 1. [A380X] Improve pilot and copilot camera positions - @heclak (Heclak)
 1. [A380X/EFIS] Illuminate ND range and mode selectors during light test - @BravoMike99 (bruno_pt99)
+1. [EFB] Added fix to make sure all doors are closed when pushing back - @frankkopp (cdr_maverick)
```
tracernz, line 101: Please move to 0.14.

frankkopp: I don't think this PR is even required anymore. Haven't been watching but this was meant as a workaround as we had some users report the not all doors closed when releasing tug - not sure if this is still an issue.

PR #9605, opened 2024-11-29
feat(A380X/mfd): ATCCOM request page sub header

EA888-3B:
## Summary of Changes
Added ATCCOM REQUEST page menu sub headers(Its a header? bc its alwayse there). Using logic from FMS header. Also added a left pointing triangle. Please let me know if you guys like this format and logic.  **Styling and positioning has not been done yet.**
If you guys like it. I will continue with this format and create every page under ATCCOM Request. (there is a lot)

## Screenshots (if necessary)
[](url)

## References
From A350 MFD manual. PM me for pictures

## Additional context
Have not tested yet. Im trying to figure out how to open the whole MFD in ACEv3. Any help would be appreciated. Pardon if there is any mistakes or error, Im new to msfs here and still learning.

Olympus593

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -80,6 +80,7 @@
 1. [FMS] Fixed issue with airport loading timing out on MSFS2024 - @tracernz (Mike)
 1. [A32NX] Fixed APU fire detection - @tracernz (Mike)
 1. [A380X/COND] Fix wasm crash during rapid decompression - @mjuhe (Miquel Juhe)
+1. [A32NX/EWD] Corrected fuel flow step to 20kg/40lbs per hour - @BravoMike99 (bruno_pt99)
 1. [A380X] Fix EWD avail. thrust fill area & PFD rudder trim visibility on ground - @flogross89 (floridude)
 1. [A380X/LIGHTS] Fix function of FCU brightness knobs - @heclak (Heclak)
 1. [A380X/FWS] Fix "NO ZFW OR ZFWCG DATA" ECAM alert after landing - @flogross89 (floridude)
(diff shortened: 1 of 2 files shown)


EA888-3B:
atccomrequest
screenshot of the ref

EA888-3B:
> Hi! As mentioned on Discord, a few of us are working on the ATCCOM pages and some coordination probably needs to take place so we're not pushing duplicate files. eg. the dropdown menu format you're using for the request header is going to be a duplicate of the person working on the report and modify page, etc. It would be a problem when trying to merge all the PRs. 
> 
> 
> 
> Maybe this should be a standardised UI object as a standalone PR first, or maybe an extension of the DropDownMenu? We can look into finalising the styling and functional requirements of this across the A380 MFD so individuals don't need to edit this function later on which might break someone else's PR. This should include styling too.
> 
> 
> 
> Should also be using the A380 manual instead of the A350? From the A380 docs, there is an ADD TEXT feature that isn't in your header. The header is also dynamic based on the ATC Center function. A623, Fans A, Fans B.
> 
> 
> 
> I'm not sure if having a new page for every single request would be the most manageable. Looking at the existing ATCCOM code, which the original request was to look at porting, my initial thoughts was to look at mimicking a similar technique based on how the ATSU logic works later on (which is why the request page was delayed). I believe our last discussion with the dev working on it is progress is coming along so additional coordination needs to happen as more people are starting to work on the ATCCOM section. I don't yet know the final solution to this, I've only been drafting the departure request page for the initial build of the ATCCOM system. But definitely something to discuss?

>Yeah I didn't noticed the other ATCCOM folder until yesterday. I agree with making a single page for every option would not be efficient, but put it all together might be challenging as well. There's lot of different pages with different type of inputs😅😅. However there's also OTHER REPORTS from REPORT&MODIFY tab. Maybe some of them will share same components or values? But yeah right now definitely wait for the dev updates and go from there,as well as more discussions👍

PR #9617, opened 2024-12-01
fix(fcu-api): fix KOHLSMANN_SET and custom baro set events

lukecologne:
## Summary of Changes

Adds missing /16 factor for KOHLSMANN_SET, and also add the same factor to the custom event, to facilitate setting inHg with four significant digits.

## Additional context


Saschl (approved):

Saschl:
Did a quick test and both events react correctly (i.e. providing 16319 as input results in 30.12inHG and 1020hPA).

lukecologne:
The event string is correct, only the enum has a different spelling

PR #9625, opened 2024-12-03
fix(fcu): set QNH bit to false when in STD

lukecologne:
## Summary of Changes

Set the bits in the FCU output bus of QNH and QFE to false, if STD mode is active. Changed to conform to new refs.

## Additional context


tracernz (approved):
💯

PR #9637, opened 2024-12-08
feat(a380x/fcu): add correct QFE label using baro preselect

heclak:
Fixes #[issue_no]

## Summary of Changes

* add QFE label using baro preselect
* add characters q,f,e and test q to FCU font
* clean up old QFE label
* adjust the alignment and size of elements in the baro window
* edit the 7-segment digits/character in the FCU font to sit on the correct baseline
* edit styling of all FCU digits to match changes in new font scale

## Screenshots (if necessary)

image
image

## References

image
image

## Additional context

## Testing instructions

1. Turn on ANN LT to test and compare to screenshot/reference above for light test mode.
2. Testing the QFE label would require the use of behaviour debug tool to trigger as there is no in-cockpit method to switch to QFE mode yet. You can set XMLVAR_Baro1_Mode to 0 to switch baro mode to 0.
3. Turn all the FCU controls to create all the various 7-segment digit combinations, TRK/FPA mode, light test, etc to check that the edited 7-segment font is correctly implemented.

image


tracernz (approved):
Nice!

BravoMike99:
Quality Assurance TesterReport

Object of testing : #9637
Aircraft          : A380X 
Tier of Testing   : 1
Date              : 09/12/2024 

Testing Process: 
1) Performed a light test and verified the preselect text matches the screenshot and ref picture ✔️ 
2) Changed the barometric mode to QFE through the dev menu and noticed the QFE font seems slightly cutout vs the screenshot above & ref pic ❌ 

Testing Results: 
Not Passed

Media:
LT Test
QFE

heclak:
@BravoMike99 ah darn. It looks like my workaround is flawed. I guess redrawing the font file to font standards might be needed. I was trying to avoid that as that means redoing all the font sizes for all the FCU fonts. I'll work on this and come back with an update.

heclak:
> 
> Weirdly enough when I opened the debugger and "hovered" over the FCU display, it would show the QFE properly just like your screenshot

@BravoMike99 It's a legitimate bug because it's an issue with the font file. When the font was originally created, it appear to be made with the baseline in the wrong location. So there is no room for the descender (eg the bottom of a q, g, etc). I worked around by exporting a non-validated font and it seemed to have worked initially. However, because part of that tail is outside of the font's HTML area, the tail doesn't get refreshed correctly all the time. Sometimes the text is hidden but the tail remains, sometimes the text is drawn but the tail is forgotten. I didn't notice this issue until you pointed it out. I researched about extending the render area but it seems the only way to have it work reliably is to remake the font with the correct specifications.

heclak:
I've redone the 7-segment fonts to the correct font baseline. This font file is now validated and will fix the issue with the cutoff Q in the previous QA test. However, this fix required me to retouch all the 7-segment digits in the FCU and will require a more careful test of all the 7-segment digit functions in QA (baro window, SPD, HDG, ALT, VS). Please turn all the possible FCU digits to check that alignment and digit combinations all work.

PR #9358, opened 2024-11-03
fix(a380x/flightmodel): adjust climb performance

BlueberryKing:
Fixes #9042

## Summary of Changes

Tune engine thrust and drag model to be more accurate.

- [x] Adjust performance in .cfg files
- [x] Update engine performance tables for vertical predictions
- [x] Update performance tables for flight model

## Additional context

Discord username (if different from GitHub):

## Testing instructions

- Perform a full flight, ensure you get (close to) accurate predictions throughout the flight.
- Ensure that the climb, cruise, and descent performance is believable.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo** or **flybywire-aircraft-a380-842** download link at the bottom of the page

Issue #9311, opened 2024-11-02
Unable to pitch down after takeoff when pausing to main menu and then unpausing


1tradyy:
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-10-31T14:33:02+00:00",
    "ref": "refs/tags/v0.12.0",
    "sha": "f26029363ad3edeeafbed20aa79c13b32c8090f8",
    "actor": "FoxtrotSierra6829",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.0",
    "version": "a380x-v0.12.0-HEAD.f260293"
}
```

### Describe the bug

After takeoff, I paused in order to tweak some sensitivity settings on my joystick. After exiting the menu and returning to the flight, the plane abruptly pitched up and did not respond to any commands to pitch down. I was able to reproduce this several times to confirm it was not a one time thing

### Expected behavior

Abrupt pitch up when returning from pause menu and unable to pitch down with joystick causing the plane to stall and crash (similar to a pitch trim runaway)

### Steps to reproduce

Takeoff
Enter pause menu at 100 feet (roughly)
wait 10ish seconds
unpause 

### Discord Username (optional)

yungtrady

Issue #9235, opened 2024-10-31
FlyByWire A380X randomly giving inputs and ignoring Sidestick Inputs


byFalon:
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-10-31T15:54:41+00:00",
    "ref": "refs/tags/v0.12.0",
    "sha": "2fe89113291f2478347dde0dc3228b55dc924f74",
    "actor": "FoxtrotSierra6829",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.0",
    "version": "a380x-v0.12.0-rel.2fe8911"
}
```

### Describe the bug

Issue appears with the Thrustmaster TCA Airbus Pack. 

Aircraft seems to be ignoring inputs from my sidestick as seen in the video below, giving itself too agressive inputs (i assume by the simulated flybywire system, sometimes 100% deflection in the opposite direction) making it rather hard (almost impossible) to land the plane because it cannot be predicted where the aircraft is going.

The video in the link below tries to show what appears to be the issue. MSFS AP Graphs show the Aileron deflection, PFD shows angle of flight and the sidestick animation tries to represent my given inputs.

Hardware tested with other aircraft. All working fine. Just the A380X seems to be having this issue with the TCA (and probably other hardware aswell as friends reported on discord)

### Expected behavior

/

### Steps to reproduce

/

### References (optional)

https://youtu.be/vsfq6kpFlsc

### Discord Username (optional)

byFalon

ungenickt-de:
I can confirm this is happening on my end as well.

AirPlaneBoy506:
Same on my end idk what is going on i had to end the flight

san0760:
I have the same Problem

MSVF:
Can confirm with TCA Quadrant "but" Winwing Ursa airliner stick.

IntegrierterKurde:
Same problem for me (Winwing Ursa Minor). After Take Off it just pitched up to 30-40° and stalled. It didn’t react to any inputs.

Consti091:
same porblem here

semartinezmo:
same for me, elevator just gets stuck on the full up position and it is visible from the outside view

JerryFhuurr:
same to me, the aircraft suddenly pitch up to more than 40 degrees and ignore my side stick input until stall (even th A/P is on)

exocetfalling:
[...]

PR #9645, opened 2024-12-10
fix(a380x/flight model): Fix pitchup and unrecoverable stall

### LINKED ISSUE #9269: Unrecoverable stall + flight controls frozen during takeoff
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-10-31T15:54:41+00:00",
    "ref": "refs/tags/v0.12.0",
    "sha": "2fe89113291f2478347dde0dc3228b55dc924f74",
    "actor": "FoxtrotSierra6829",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.0",
    "version": "a380x-v0.12.0-rel.2fe8911"
}
```

### Describe the bug

Takeoff at any gross weight and if you pitch up to anything above 10-12 degrees, it will start to pitch up by itself. You cannot stop this behavior as it seems like the flight controls are frozen once you enter the amber speed protection range or sometimes it pops up if you exceed 17 degrees pitch up. 

Because the flight controls are frozen. you cannot use standard stall recovery to save yourself. The plane will not respond to nose down inputs nor roll inputs. 

The stall at low altitude just means you have no option but to crash. 

### Expected behavior

Expected behavior is before a stall, your flight controls should still respond to the inputs as there is enough airflow going past them  given your airspeed isn't terribly low. The plane should pitch down and start to gain airspeed but that does not happen. 

### Steps to reproduce

1. Takeoff at any gross weight 
2. Pitch up to 13-15 degrees and observe 
3. The plane will keep pitching up by itself and eventually put you in a stall. 

### References (optional)

image

I have found a workaround for this, given you have enough altitude to recover from the insane nose up attitude. 
1. Full throttle
2. Turn off PRIM 1
3. Turn off SEC 1
4. Turn off PRIM 3
5. Turn off SEC 3
6. Pitch down

I am not a master at Airbus theory, but the turned off Flight control computers should drop you down to alternate law. This change to alternate law seems to reset the elevator, in the sense that it recovers from its jammed pitch state, and the aircraft will respond to any nose down inputs that you give it. After you have recovered and gained enough airspeed, you can turn all the flight computers back on. 

banjigaming99

donstim:
Fixes #9269  and probably #9311 

## Summary of Changes
- Fixes the undesired pitchup and unrecoverable stall behavior
- Improves pitch response in manual flying
- Potentially improves roll control associated with some of the comments on #9235 (rolling back against the initial input)
- Partial fix for autorotation (Nose wheel can still unweight early, but airplane should not lift off before rotation speed)
- Reduced the gear drag, which I felt was much too high
- Tried to match the flaps up drag level of the previous flight model for lift coefficients between 0.3 and 0.7 so as not to impact PR #9358. May need to iterate further on this.
## Additional context
The stall angle-of-attack has been increased from 15 degrees to 18 degrees. At some point, the alpha-stall related envelope protection and PFD indications should be updated. 

The takeoff rotation law needs further adaptation for the A380X and these/future flight model changes. At heavier weights with the current rotation law, be aware that more back stick will be needed to allow the airplane to fly away. At mid to heavier weights, the stick should not be released for 3-4 seconds (or until 100 ft RA is reached) or the airplane's pitch angle and flight path may drop. This is more pronounced on CONF 1+F than on CONF 2 or 3. (Testing included an anticipated MTOW of 575 T. The takeoff rotation law should be updated before the MTOW is increased from the current 510 T.)

## Testing instructions
Takeoffs, full flights, and landings at light, medium, and heavy weights (within the A380X's takeoff and landing weight limits). In flight max pitch ups and recoveries.


aadee9940:
Tested at various gross weights ranging from 350-575tons, it is pretty consistent all round, and there is enough elevator authority for me to push down on the stick and the aircraft reacts accordingly. A.Floor works as it should as well. 

No more of the continuous pitch up moment, I can recover safely while in the amber speed band range.

Saschl (approved):

mattking247:
Quality Assurance Tester Trainee Report

Discord Username : Matt
Object of testing : https://github.com/flybywiresim/aircraft/pull/9645
Aircraft : A380X
Tier of Testing : 1
Date : 12/12/2024

Testing Process:
Full flight OMDB - EGLL, full departure from OMDB Rwy 30R with calm weather.

Notes:
500 tonnes at take off.

Flight notes:
Test flight complete and then several descents, climbs, turns and configuration changes were performed.

The aircraft flies pretty much as expected in roll. However, when returning the stick to neutral, there was a slight roll back in the opposite direction. For example, after completing a 25 degree bank to the right, I rolled out to the left to return to a wings level attitude at a normal roll rate. On reaching wings level, I released the stick and the aircraft rolled slightly back to the right. I think this feels like it may just be a slight over correction to stopping the role rate. Overall though, not bad at all.

The elevator and pitch is where most work needs to be completed. The aircraft nose does not stop moving in the direction of travel once the stick is returned to neutral. An example of this, aircraft level at 5,000 ft AGL, over the sea on a calm day. Nose raised to 15 degrees nose up at 250 kts. Side stick was then moved forward to lower the nose at a similar rate to that on take off, but in the opposite direction. On releasing the stick at 5 degrees nose up, the  nose continues to track down until roughly 10 degrees nose down. This still happens at higher speed, but is less pronounced. When configured for landing and on approach the same thing happens. I watched the flight control page during these manoeuvres and it seems like the elevator is slow to return to the neutral position when released. The elevator still be in the down position is then driving the nose to continue downwards. It does affect nose pitch up also, but not as much. Could not replicate on the ground due to different flight control logic. Connected to this, it feels on approach like the nose always wants to pitch down subtly, and I felt like I was constantly trying to raise the nose.

Overall, I think the aileron response was pretty close to expected, elevator still not so much.  However, I was able to use the deflection authority available to counter the nose up pitching tendency.

donstim:
> Quality Assurance Tester Trainee Report
> 
> Discord Username : Matt Object of testing : #9645 Aircraft : A380X Tier of Testing : 1 Date : 12/12/2024
> 
Thank you for your test and detailed report. Keep in mind that this PR is intended to fix the unrecoverable stall issue. It does not fix existing issues with the FBW control system, for example, the roll "rebound."  I tried to reduce that effect but it is not caused by and cannot be eliminated by the flight model. 

Having said that, I cannot reproduce the pitch issues you described. Would you be able to provide videos of these issues?  Here is my attempt at reproducing the pitch up to 15 degrees at 5000 feet, 500 T, 250 KCAS, then pitching down to 5 deg NU and releasing the stick. You can see that in my video, although there is some oscillation and the final settled out pitch value is less than 5 degrees NU, it comes nowhere close to 10 deg ND.  https://1drv.ms/v/s!Ajsy7Ey4_8uFgZdBf70oLghQNRY2NA?e=caMmRA

For the approach, I flew the same maneuver, then performed several small pitch adjustments in CONF FULL, starting at approx -3 deg FPA, VAPP at around 320 T. No issues with the pitch up, then down maneuver, and it was very stable in pitch after making any pitch adjustments. https://1drv.ms/v/s!Ajsy7Ey4_8uFgZdCxx2OWj7FSiGbDQ?e=8jBEA1  

I've flown a lot of approaches and landings while developing this PR and never did I feel that I was constantly trying to raise the nose. One example: https://1drv.ms/v/s!Ajsy7Ey4_8uFgZdDWInA5c5L0vKPjg?e=SIko8u. Pitch is rock steady on this mock approach over the water (other than the one adjustment I made when I realized the descent rate was too high). Are you maybe seeing wind changes or turbulence?

As a result of your comment, though, I did go through the PR again and found I had neglected to include some changes that were in my test version as well as correcting one (important from a stall avoidance standpoint) typo. None of these should have had any impact on what you reported, and the videos above were made without those corrections.

mattking247:
Thanks for looking at this. I completely get the slightly out of scope test comments, but thought I would link it in as the tet mentioned flight controls and improving them. The only video I have to hand  are these. You will see in the first one, the nose falling around the 18 second mark and my hand in the screen to show it is not still on the stick. 


https://github.com/user-attachments/assets/afad4aa0-3621-420a-958b-abce9c5ba91f

The second movie shows how long it takes the elevator to go back to neutral once the stick is released. It is a large flight control I know, but this seems way to long for some of the surface to still be moving, even with dampening.


https://github.com/user-attachments/assets/027aa395-efc2-4d9f-887f-deff4dc30fef

Let me know your thoughts. Thank you.

donstim:
Hey, I was finally able to reproduce your issue. It took "harder" inputs and releases on my controller without trying to come to a smooth release. Anyway, once I found out how to do it, I could do it on both the pre-PR and PR versions. This confirms to me that the PR did not cause this.

mattking247:
Hi, thank you for looking into this. Apologies for the late reply, just got back home. I can confirm now this was not just this PR. I am, however, glad you could reproduce it. I thought I was going crazy, as no one else seems to have seen it before! Apologies for getting the wrong end of the stick and holding up your PR.

Matt.

donstim:
No problem. It didn't hold anything up, and it resulted in not only a good discussion on a current issue with the airplane, but also led me to find and correct some mistakes I'd made in this PR. Thanks again for the testing and detailed descriptions of what you encountered.

flogross89:
@donstim Thanks a lot for these changes. Gave her a quick spin, feels lovely so far. 

- Uncontrollable pitch-up and stall behavior seems to be fixed or at least largely improved, I couldn't provoke it in any way
- Auto-rotation on takeoff also seems improved, weight on nose wheel definitely larger for speeds until V2
- Opposite aileron movement when ending turn also seems to be gone, but at the same time the roll behavior seems a bit sluggish --> This is something which can and should be fixed in the FBW laws after this PR. 

The flight envelope protections can be easily adapted, it's just important to know in which manner the angles should be changed (fixed scale on all angles, or only adaption of the upper third, ...)

donstim:
> 
> The flight envelope protections can be easily adapted, it's just important to know in which manner the angles should be changed (fixed scale on all angles, or only adaption of the upper third, ...)

I'm unsure what you mean by all angles vs upper third...  May be my lack of understanding of how it's coded. The aoa associated with both the alpha max and alpha protection functions can be increased (which would lower the associated speeds shown on the PFD) for those functions. It would take tuning to ensure that the stall aoa is not exceeded during a high rate full back stick maneuver, but I would think that a 1-2 degree increase in aoa would be reasonable.

I think a bigger issue with the PFD indications of the alpha protection and alpha max speeds is that in the A380X they seem to be based on the FMS input ZFW+FOB. That is not the case on the real airplane where they are based on an aerodynamically determined weight. We should be using the real weight of the airplane rather than FMS input ZFW+FOB (although I have to admit the way it is done currently is nice for my testing in that I can vary the alph prot/alpha max PFD speeds by simply changing the FMS input ZFW.)

flogross89:
@donstim 
alpha_max and alpha_prot are not a constant value, but different for each flap config and mach speed (I suppose this allows for a more fine-grained control of protection behavior for different regimes).

For example, alpha_max has the following lookup table: 
image
Dimension 1 = mach 
Dimension 2 = flap config (0, 1, 1+F, 2, 3, FULL) 

alpha_0 is only depending on flap conf: 
image

The alpha protection speeds are not based on any weights, where did you get that impression? VLS and the characteristic speeds (green dot, F+S speeds) on the other hand are depending on the FMS ZFW + FQMS FOB, as the real example

donstim:
@flogross89 How about we park this conversation here and continue on Discord. I'm going to look at simplifying this PR and take out the stall AOA change so the protection values can be changed at the same time if we decide to go that route at all.

flogross89:
@donstim Sure!

alexr4339:

Discord Username  : alexdnzdmb 
Object of testing : #9645 
Aircraft          : A380X 
Tier of Testing   : 2
Date              : 24/12/2024 

Testing Process: 
Flight 1:
ZFW: 300T, GW 323,6T
Takeoff with Conf1, then stalling and going back for landing

Flight 2:
ZFW: 310T, GW 418T
Takeoff with Conf1, then stalling and going back for landing

Testing Results: 
Flight 1
Takeoff as described, feels good and FD suggestion gives around 3°/s pitchup. To fly away, little more back pressure needed as described in description. Stall is recoverable in all confs, even spins are not possible/recoverable. Landing feels good.

Flight 2
Takeoff as described, feels good and FD suggestion gives a little lower pitchup than 3°/s which however is easily compensated by more pressure. Rotation as a result feels good and logical. To fly away, little more back pressure needed as described in description. Stall is recoverable in all confs, even spins are not possible/recoverable. Landing felt direct but more back stick was needed as anticipated, but probably my fault as I was still used to the light landing.

Still speeds close to green dot whil pitchipg up doe to accel lead speeds close to alpha prot
image
Dont know whether it is that extrem irl. However makes sense as it'd be an accelerated stall. Question here is whether accelerated influence is that great

Recovery however works normal

Negatives:       


Conclusions: 


Media:

flogross89:
@alexr4339 One note about „stall speeds close to green dot“: The stall alpha for the PFD speeds is too low in some cases, I will start a PR asap to fix that. The problem is that the PFD stall speed doesn’t fully represent the actual stall speed (there’s always a margin on top, ours is too big)

PR #9702, opened 2025-01-02
feat(elec): fuel pump power consumption

Gurgel100:
Fixes #[issue_no]

## Summary of Changes

Adds electric power consumption to fuel pumps. This PR is a preparation for moving some logic of the fuel system into the Rust system (especially for the A380 as it is a more complex system than MSFS can handle).
Also adds the correct condition for the fail lights on the fuel pumps for the A320 even though they will never illuminate for now.
It also adds the open indication of the cross feed valve.

## Additional context

## Testing instructions

Turn on the fuel pumps individually and check if they consume power from some bus. You can check this by observing the generator loads for example. Be aware that on the A380 not all pumps are running all the time.
Check that the "open" indication on the cross feed valve push button appears when opening it.


tracernz (approved):
I am not the rust expert, but the code seems correct to me.

Jonny23787:

Discord Username  : jonny_23
Object of testing : #9702 
Tier of Testing   : 1
Date              : 09/03/2025 

Testing Process: 

1. Loaded A380 at gate and started APU
2. Toggled each fuel pump on and off and observe increase and decrease in APU generator load
3. Repeated steps 1-2 in A320
4. Checked that the cross bleed open light was on when opening the cross bleed value (A320 only)

Testing Results: 
Passed

Negatives:       


Conclusions: 


Media:

PR #9743, opened 2025-01-10
fix: Use SERVER Callsign for the Hoppie is Station Available Check

### LINKED ISSUE #9704: Hoppie ACARS will require removal of 'ALL-CALLSIGNS' call when checking the flight number
### Aircraft Version

Stable, Development

### Describe the bug

As of last week, Hoppie has introduced an error when calling the Hoppie ACARS for an 'ALL-CALLSIGNS' ping to the 'ALL-CALLSIGNS' station. This added load to their platform hence the error. As of today, this was downgraded to a warning as multiple clients 'abuse' this system and the errors should be no more, however we will need to update the functionality to be handled better and not cause issues in the future once the call returns an error again. 

The culprit from a quick glance seems to be: https://github.com/flybywiresim/aircraft/blob/a0c9ddd95133343f7b5413a834fe207ba747691c/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts#L85

which performs the call `ALL-CALLSIGNS` and prompted the proxy error https://github.com/flybywiresim/aircraft/blob/a0c9ddd95133343f7b5413a834fe207ba747691c/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts#L103

According to Hoppie and the docs, we can just check for an error for "duplicate callsign" whenever we perform any action in the network as that callsign, including pings (which are not logged in the network's messages and do not reserve the callsign). 
Solution 1 would be to do a 'precheck' and send a ping as described, or alternatively solution 2 would be to simply check for if the callsign is in use when registering later down the ACARS API pipeline. Both would have the same result, but one less call. 

### Steps to reproduce

*Can no longer reproduce the error as it's been changed to a warning*

1. Begin a flight 
2. Assign a flight number
3. See for the API return from Hoppie ACARS

### References (optional)

_No response_

_No response_

_No response_

### LINKED ISSUE #9722: problem with ATC MENU
### Aircraft Version

Development

### Build info

```json
{
    "built": "2025-01-06T04:36:43+00:00",
    "ref": "refs/heads/master",
    "sha": "074143e35920894457b81b11170880de0adace94",
    "actor": "flogross89",
    "event_name": "manual",
    "pretty_release_name": "master:074143e3",
    "version": "a32nx-v0.13.0-dev.074143e"
}
```

### Describe the bug

I am unable to interact with the ATC Menu. When I try to enter the Departure Clearance Station, it gives an error saying "No active ATC" (even though PDC/DCL is supported at the airport where I am loading). Alternatively, it shows an error saying "No Hoppie Connection" (even though I have entered the Hoppie code and enabled Telex). Additionally, there are issues with the Notification and Flight Req menus. In the EFB menu under the ATIS/AOC section, I have VATSIM selected.

### Expected behavior

I expected that if I entered the Hoppie code, everything would work.
Or I expected that after changing the airport, everything would start working as well.

### Steps to reproduce

1.I start the aircraft at an airport where PDC/DCL is supported.
2.I submit the flight plan in VATSIM.
3.I start the aircraft, fill out the MCDU, and when it’s time to complete the DCL, I can’t enter the Station.

### References (optional)

_No response_

_No response_

_No response_

CronixZero:
fix: Use SERVER Callsign as to in the isStationAvailable Method for the HoppieConnector

Fixes #9704 
Fixes #9722 

## Summary of Changes

Changed the `to` Callsign when checking whether a station is available on Hoppie to `SERVER` as that field gets ignored and `ALL-CALLSIGNS` is the only special fake callsign

## Screenshots (if necessary)

image

## References

https://www.hoppie.nl/acars/system/tech.html
https://github.com/flybywiresim/aircraft/issues/9704

## Additional context

## Testing instructions

1. Power up Aircraft as normal
2. Enter INIT Page
3. Open ATC Communications page -> Ground Requests -> Departure
4. Enter online Station Code
5. Should work instead of displaying "NO ACTIVE ATC"

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,7 +7,6 @@
 
 ## 0.13.0
 
-1. [ATSU] Fixed issues with the ALL-CALLSIGNS recipient on Hoppie - @CronixZero (CronixZero)
 1. [GENERAL] Fixed issue in C++ WASM Framework that caused performance degradation in some WASM modules - @frankkopp (cdr_maverick)
 1. [A32NX/FCU] Fixed auto-initialisation of baro unit - @tracernz (Mike)
 1. [A380X/FCU] Fix baro-preselect not recognising baro unit changes - @tracernz (Mike)
@@ -117,6 +116,7 @@
 1. [A380X/ND] Remove leading zeros from terrain elevation display - @BravoMike99 (bruno_pt99)
 1. [A32NX/FWS] Fix autopilot instinctive disconnect button logic for 3D model - @flogross89 (floridude)
 1. [A380X/EFIS] Fix VV pb indicator not turning on when TRK-FPA mode is selected - @heclak (Heclak)
+1. [ATSU] Fixed issues with the ALL-CALLSIGNS recipient on Hoppie - @CronixZero (CronixZero)
 
 ## 0.12.0
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## 0.13.0
 
+1. [ATSU] Fixed issues with the ALL-CALLSIGNS recipient on Hoppie - @CronixZero (CronixZero)
 1. [GENERAL] Fixed issue in C++ WASM Framework that caused performance degradation in some WASM modules - @frankkopp (cdr_maverick)
 1. [A32NX/FCU] Fixed auto-initialisation of baro unit - @tracernz (Mike)
 1. [A380X/FCU] Fix baro-preselect not recognising baro unit changes - @tracernz (Mike)
--- a/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
+++ b/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
@@ -96,7 +96,7 @@ export class HoppieConnector {
     };
     const text = await Hoppie.sendRequest(body).then((resp) => resp.response);
 
-    if (text === 'error {callsign already in use}' || text === `ok {${station}}`) {
+    if (text === 'error {callsign already in use}' || text.includes(station)) {
       return AtsuStatusCodes.CallsignInUse;
     }
     if (text.includes('error')) {
--- a/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
+++ b/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
@@ -49,7 +49,7 @@ export class HoppieConnector {
     };
 
     Hoppie.sendRequest(body).then((resp) => {
-      if (resp.response !== 'error {illegal logon code}') {
+      if (resp.response !== 'error {invalid logon code}') {
         SimVar.SetSimVarValue('L:A32NX_HOPPIE_ACTIVE', 'number', 1);
         console.log('Activated Hoppie ID');
       } else {
@@ -96,7 +96,7 @@ export class HoppieConnector {
     };
     const text = await Hoppie.sendRequest(body).then((resp) => resp.response);
 
-    if (text === 'error {callsign already in use}' || text === `ok {${station}`) {
+    if (text === 'error {callsign already in use}' || text === `ok {${station}}`) {
       return AtsuStatusCodes.CallsignInUse;
     }
     if (text.includes('error')) {
--- a/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
+++ b/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
@@ -121,7 +121,7 @@ export class HoppieConnector {
     const body = {
       logon: NXDataStore.get('CONFIG_HOPPIE_USERID', ''),
       from: HoppieConnector.flightNumber,
-      to: 'ALL-CALLSIGNS',
+      to: 'SERVER', // Not needed as Hoppie ignores this field usually
       type: 'ping',
       packet: station,
     };
--- a/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
+++ b/fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
@@ -43,7 +43,7 @@ export class HoppieConnector {
     const body = {
       logon: NXDataStore.get('CONFIG_HOPPIE_USERID', ''),
       from: 'FBWA32NX',
-      to: 'ALL-CALLSIGNS',
+      to: 'SERVER',
       type: 'ping',
       packet: '',
     };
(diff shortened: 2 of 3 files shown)


CronixZero:
Further explanation on why this is needed: https://github.com/flybywiresim/aircraft/issues/9704#issuecomment-2584147777

CronixZero:
Yeah, the other main use is to check whether a call sign is already on the network but I'm not quite sure how to check for that yet. 
The issue for that one is https://github.com/flybywiresim/aircraft/issues/9704 which isn't an issue currently as Hoppie changed it to a warning instead of returning an error. This gives us plenty time to fix it. The CPDLC is currently however fully broken due to the isStationAvailable function returning noAtc which blocks the use of it

CronixZero:
And as far as I understand the `isCallsignInUse` method doesn't care about anything that comes after `ok` so the warning does not affect its operation

fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
```diff
@@ -121,7 +121,7 @@ export class HoppieConnector {
     const body = {
       logon: NXDataStore.get('CONFIG_HOPPIE_USERID', ''),
       from: HoppieConnector.flightNumber,
-      to: 'ALL-CALLSIGNS',
+      to: 'SERVER', // Not needed as Hoppie ignores this field usually
```
alepouna, line 124:
```suggestion
```
According to the docs, we can just remove the `to` and just offer it in the `packet` alone or vice versa, so it should fine if we keep this line, but as the comment says we can just get rid of it :)

alepouna (commented):
See notes for a comment, otherwise LGTM

BravoMike99:
> And as far as I understand the `isCallsignInUse` method doesn't care about anything that comes after `ok` so the warning does not affect its operation

Indeed but from quick testing it looks like hoppie will always return the "ok {warning,,,,}" if you send to ALL-CALLSIGNS even if such callsign is in use. The same happens for the logon code check (`activateHoppie`) which makes both misleading and prone to errors later on as they will always succeed. It should just be a direct replacement everywhere like the one you did here.

alepouna:
> It should just be a direct replacement everywhere like the one you did here.
+1

fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
CronixZero:
```
{
  "statusCode": 400,
  "message": [
    "to should not be empty"
  ],
  "error": "Bad Request"
}
```

alepouna: okay yeah I am legally blind, see this is why I don't code at 1am 🤣

BravoMike99:
Thanks looks good now, please add a changelog entry too & testing instructions 👍

CronixZero:
Which area would that be?

CronixZero:
> Which area would that be?

Used `ATSU` now

CronixZero:
(Needs new Workflow Approval now)

alepouna (approved):

.github/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ## 0.13.0
 
+1. [ATSU] Fixed issues with the ALL-CALLSIGNS recipient on Hoppie - @CronixZero (CronixZero)
```
BravoMike99, line 10: At the bottom please. New entries go to the bottom so they are numbered by order of creation 
image

BravoMike99 (changes requested):

.github/CHANGELOG.md
CronixZero: Changed!

BravoMike99 (approved):

CronixZero:
Sorry, please approve again

fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
```diff
@@ -90,13 +90,13 @@ export class HoppieConnector {
...
       from: station,
-      to: 'ALL-CALLSIGNS',
+      to: 'SERVER',
       type: 'ping',
       packet: station,
     };
     const text = await Hoppie.sendRequest(body).then((resp) => resp.response);
 
-    if (text === 'error {callsign already in use}') {
+    if (text === 'error {callsign already in use}' || text === `ok {${station}`) {
```
BravoMike99, line 99: During testing I noticed that the addition wont match if its already in use it (due to the missing `}`, however,  I struggle to understand in general how this hoppie endpoint works. Seems like it may be doing a matching/starts with of some sorts. Maybe it might be a good idea to just replace it with .includes(station)

image

image

image

BravoMike99 (changes requested):

fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
BravoMike99: In addition could you perhaps change line 52 of `HoppieConnector.ts` so  it checks for `"invalid"` instead of `"illegal"` while we are at it? (Unrelated to your changes but just something I noticed which is a straight replacement)
image

CronixZero: That should do it, I think. Good catch with the matching. Didn't notice that

BravoMike99 (approved):

fbw-common/src/systems/datalink/router/src/webinterfaces/HoppieConnector.ts
BravoMike99: Thanks, in the meantime I noticed that the AtsuAocPage.tsx in the EFB is also using illegal instead of invalid. If you could also change that it would be appreciated  👍

CronixZero: Done. Any ETA on when QA is likely to try this out?

BravoMike99:

Object of testing : #9743  
Aircraft          : A32NX 
Tier of Testing   : 1
Date              : 11/01/2025 

Testing Process: 
Setup hoppie on EFB.
Entered duplicate callsign on INIT page, FLT NBR in use scratchpad message appeared.
Entering a callsign not in use does not trigger the above scratchpad message.
Went to ground request page and inserted an active hoppie station, "NO ACTIVE ATC" message did not appear ✔️ 
Inserted a station which is not online on hoppie, "NO ACTIVE ATC" message appeared on mcdu ✔️ 

Testing Results: 
Passed


Conclusions: 
LGTM

Media:

https://github.com/user-attachments/assets/a1ac0fc5-2e14-4ce0-9c58-30953b20efcb

CronixZero:
Thanks :)

PR #9559, opened 2024-11-19
feat(fms): improved nav database for msfs2024

tracernz:
Fixes #[issue_no]

## Summary of Changes

- Add some new facility typings for MSFS2024 (all new stuff is optional and marked as "FS2024 only").
- Add a hacky util method to determine if we're in 2024 so we can disable MSFS2020-specific hax.
- Load the transition altitude and level from the nav data in MSFS2024.
- Load the leg RNP from the nav data in MSFS2024.
- Load the ILS category from the nav data in MSFS2024.
- Handle/discard speed constraints appropriately in MSFS2024.
- Load the RNP-AR flags from the nav data in MSFS2024.

More to come in future...

## Additional context

Discord username (if different from GitHub):

## Testing instructions

Perform full flight in MSFS2020 and MSFS2024. Check that transition alt/level are auto-filled appropriately in 2024 (including being not filled from airports with "by ATC" transition level. Load an RNP AR procedure like NZWN UPLU1R SID and check the RNP on the progress page matches the procedure. Select a departure runway with ILS, and on the selected navaids page click the LSK next to the ILS; check the category is shown on MSFS2024 and not on MSFS2020.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #9747, opened 2025-01-11
feat(efb): troubleshooting page

tracernz:
Fixes #[issue_no]

## Summary of Changes

Adds a troubleshooting page under the about page in the EFB for advanced user support. The idea here is to log things that have user visible errors that don't provide enough info to determine the cause, e.g. "INTERNAL ERROR" in the FMS. It is not intended to be a general-purpose error log as it would lose it's usefulness when filled with lots of random messages like the debugger.

Hopefully we can figure out the recent spike in these INTERNAL ERRORs.

Needs https://github.com/flybywiresim/aircraft/pull/9559 merged first (for isMsfs2024 function).

## Screenshots (if necessary)

image
image

## Additional context

## Testing instructions

Go to the about page on the EFB, and then click Troubleshooting.. check that the info displayed is accurate and the layout reasonable. If you know of a repro to cause an INTERNAL ERROR on the A32NX, you could also try it to see the message in the troubleshooting log.

### DIFF
--- a/fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AboutPage.tsx
+++ b/fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AboutPage.tsx
@@ -17,7 +17,7 @@ import { SettingsPage } from '../Settings';
 import FbwTail from '../../Assets/FBW-Tail.svg';
 import { useViewListenerEvent } from '../../Utils/listener';
 import { Link, Route, Switch } from 'react-router-dom';
-import { TroubleshootingPage } from 'instruments/src/EFB/Settings/Pages/TroubleshootingPage';
+import { TroubleshootingPage } from './TroubleshootingPage';
 
 const baseAboutRoute = `/settings/${pathify('About')}`;
 
--- a/fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AboutPage.tsx
+++ b/fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AboutPage.tsx
@@ -1,4 +1,4 @@
-// Copyright (c) 2023-2024 FlyByWire Simulations
+// Copyright (c) 2023-2025 FlyByWire Simulations
 // SPDX-License-Identifier: GPL-3.0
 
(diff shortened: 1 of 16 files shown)


fbw-common/src/systems/instruments/src/EFB/TroubleshootingContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+const TroubleshootingContext = React.createContext<string[]>(undefined as any);
+
+interface TroubleshootingContextProps {
+  eventBus: EventBus;
+}
+
+export const TroubleshootingContextProvider: React.FC<TroubleshootingContextProps> = ({ eventBus, children }) => {
+  const [errorLog, setErrorLog] = useState([] as string[]);
+
+  eventBus
```
Benjozork, line 17: This will leak subscriptions every time it renders.

You should place this in a `useEffect` (with an empty dependency array) whose cleanup function destroys the subscription.

Benjozork (dismissed):

fbw-common/src/systems/instruments/src/EFB/Settings/Pages/AboutPage.tsx
```diff
@@ -71,44 +79,61 @@ export const AboutPage = () => {
     AircraftGithubVersionChecker.getBuildInfo(process.env.AIRCRAFT_PROJECT_PREFIX).then((info) => setBuildInfo(info));
   }, [process.env.AIRCRAFT_PROJECT_PREFIX]);
 
+  useEventBus().pub('troubleshooting_log_error', 'Rendered troubleshooting page!', true, false);
```
Benjozork, line 82: Is this meant to be kept?

flogross89 (changes requested):
1) There are a few try...catch in GuidanceController.ts and VnavDriver.ts; Do you think it makes sense to also include them? Most of the guidance errors also remain hidden. Although there's a danger of cluttering the log
2) We could also consider saving these logs to a file, but you can also do this in another PR, your call.

flogross89 (approved):
Let’s tackle my comments in another PR, if needed.

PR #9770, opened 2025-01-17
fix(a380x/pfd): Fix precision of pitch trim indicator

### LINKED ISSUE #9762: CG position shown below PFD is incorrect
### Simulator

2020

### Aircraft Version

Development (4k)

### Build info

```json
"built": "2025-01-01T10:39:42+00:00",
    "ref": "refs/heads/master",
    "sha": "a0c9ddd95133343f7b5413a834fe207ba747691c",
    "actor": "tracernz",
    "event_name": "manual",
    "pretty_release_name": "master:a0c9ddd9",
    "version": "a380x-v0.13.0-dev.a0c9ddd"
```

### Describe the bug

The CG position shown below the PFD is different than what is shown on the SD. The one on the SD is correct. I don't know where the one shown below the PFD comes from, but this is the one that is used to set the pitch trim position for takeoff. When it is incorrect in the nose up direction, it makes the current auto-rotation issue on takeoff worse.
Screenshot 2025-01-15 130153

### Expected behavior

The CG positions on both displays should be the same.

### Steps to reproduce

Load the airplane for takeoff. Observe the CG positions on the two screens compared to the CG position shown either in the MFSFS fuel and payload UI (or the simvar). (The one shown on the SD uses the simvar value.)

### References (optional)

_No response_

_No response_

donbikes

flogross89:
Fixes #9762 

## Summary of Changes

Additional clean-up: Remove references to failures which were removed (PFD failure without reconf logic doesn't make sense)

## Additional context

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -124,6 +124,7 @@
 1. [A380X/FCU] Add TRUE indication on FCU when TRUE North reference is selected on AFS CP - @heclak (Heclak)
 1. [A380X/MFD] Add airport data page into the MFD (DATA > AIRPORT) - @bulenteroglu (senolitam)
 1. [A380X/EFB] Adds PRIM/SEC/FCDC failures to EFB - @flogross89 (floridude)
+1. [A380X/PFD] Fix precision of pitch trim indicator - @flogross89 (floridude)
 
 
 ## 0.12.0
--- a/fbw-a380x/src/systems/instruments/src/PFD/PFD.tsx
+++ b/fbw-a380x/src/systems/instruments/src/PFD/PFD.tsx
@@ -1,4 +1,3 @@
-import { A380Failure } from '@flybywiresim/failures';
 import {
   ClockEvents,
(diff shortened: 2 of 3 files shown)


tracernz:
> (PFD failure without reconf logic doesn't make sense)

I'm not sure I agree with that. We'd never get anywhere if we have to build the whole world all the time.

flogross89:
> > (PFD failure without reconf logic doesn't make sense)
> 
> I'm not sure I agree with that. We'd never get anywhere if we have to build the whole world all the time.

You know I'm a fan of an iterative approach, but IMO it doesn't make sense to enable failures if we're not able to at least simulate the immediate pilot-noticeable behavior when a display is failing. For the PFD failure, this would mean automatic reconfiguration on the onside ND.

We could however enable failure of the ND, SD or MDF from my POV, since they don't require automatic reconf, it's more explainable that we don't have manual reconf simulated yet.

tracernz (approved):

PR #9786, opened 2025-01-19
feat(a380x/mfd): Add DATA/WAYPOINT page

Slightlyclueless:
## Summary of Changes

Add the waypoint page under the data header
## Additional context


fbw-a32nx/src/systems/fmgc/src/flightplanning/DataManager.ts
```diff
@@ -366,4 +366,12 @@ export class DataManager {
...
     return this.storedWaypoints.filter((wp) => wp && wp.waypoint.ident === ident);
   }
+
+  getAllStoredWaypoints(): PilotWaypoint[] {
+    if (this.storedWaypoints === undefined) {
+      return [];
+    } else {
+      return this.storedWaypoints;
+    }
+  }
```
tracernz, line 376:
```suggestion
  getAllStoredWaypoints(): PilotWaypoint[] {
    return this.storedWaypoints ?? [];
  }
```

Slightlyclueless: Hmmm, I dislike returning undefined for elements like this. This *shouldn't* be called that often so I feel like the allocation is not going to be that bad (this may be C/Rust brain thinking where an allocation for that slice would be tiny)

tracernz: Then why not have this.storedWaypoints always an array instead? If this  is being called every time the page updates it's allocating quite a lot of unneeded objects across a flight. It's death by a thousand cuts with GC pauses = microstutter.

Slightlyclueless: > Then why not have this.storedWaypoints always an array instead
just checked, it already is (Line 61). So this function doesn't need any nullish coaliescing at all, it can just return the array

heclak:
is this still being worked on?

Slightlyclueless:
> is this still being worked on?

Kinda, I'd got bored and moved on to other projects. But intended to revisit all the MSFS branches very soon, probably not this week as I've got to fix my bike

flogross89:
Closing this as long as there's no further activity. Feel free to re-open once work picks up again.

PR #9683, opened 2024-12-28
feat(elec): a380 electrical improvements

Gurgel100:
Fixes #9758

## Summary of Changes

Fixes a couple of issues in the current A380 electrical system:
* Improves the behavior of the BCRUs when to open or close the associated battery line contactor
* Fixes AC GND FLT SERVICE BUS
* Fixes the battery voltage indicator. Previously the APU and ESS battery voltage were exchanged.

## Additional context

Discord username (if different from GitHub):

## Testing instructions

1. Check that the electrical system behaves as expected
2. Check that the batteries disconnect 5 seconds after it's TR doesn't get power from the AC side anymore
3. Check that nav lights are working without ground power connected and with a fully powered aircraft
4. Check that the battery voltage indicator shows the voltage of the corresponding battery (for example check with SD page by disconnecting one battery at a time)

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #9803, opened 2025-01-25
fix(elec): fix A380 ground servicing mode

Gurgel100:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue which was introduced with #9683: If battery 2 was not turned off and ground power was available but not active the battery was connected to the DC BUS 2 even though it shouldn't.

## Additional context

## Testing instructions

**A380 only**
* Spawn cold & dark.
* Attach ground power if not attached via the EFB (so that the green avail lights are on). Check (via LVars) that DC 2 is not powered.
* Turn on the batteries. Check (via LVars) that DC 2 is not powered.
* Turn on ground power. Check (via LVars) that DC 2 is powered.
* Turn off ground power. Check (via LVars) that only DC 2 is not powered.


beheh (approved):

heclak:

Discord Username  : Heclak 
Object of testing : #9803 
Aircraft          : A380X 
Tier of Testing   : 1
Date              : 15/02/2025 

Testing Process: 
 

- Spawn cold & dark.
- Attach ground power if not attached via the EFB (so that the green avail lights are on). Check (via LVars) that DC 2 is not powered. ✅

image

- Turn on the batteries. Check (via LVars) that DC 2 is not powered.✅
 
image

- Turn on ground power. Check (via LVars) that DC 2 is powered.✅
 
image

- Turn off ground power. Check (via LVars) that only DC 2 is not powered.✅

image


Testing Results: 
Passed

fbw-common/src/wasm/systems/systems/src/electrical/battery_charge_rectifier_unit.rs
```diff
@@ -72,7 +76,7 @@ impl BatteryChargeRectifierUnit {
         self.battery_pb_is_auto = battery_push_buttons.bat_is_auto(self.number);
         let ac_bus_powered = electricity.is_powered(self);
 
-        self.loss_of_ac_duration = if ac_bus_powered {
+        self.loss_of_ac_duration = if ac_bus_powered || self.ground_servicing && !ground_servicing {
```
tracernz, line 79: The naming here makes it a little hard to read. Maybe an expression like `let ground_servicing_turned_off = self.ground_servicing && !ground_servicing;` would improve it.

Gurgel100: Thanks for the suggestion.
`ground_servicing` means that ground servicing mode is enabled. Currently this means that the contactor connecting the AC GND SVC bus with TR2 is closed.

About the logic: Good point. I can't remember exactly why it's needed. I'll do some more tests to be sure it's needed.

Gurgel100: Since it's working I'll merge it as is. I'll have a look at the logic at a later point.

PR #9829, opened 2025-02-05
fix(nd): wind and speed display update rate for more frequent updates

bulenteroglu:
Fixes #[issue_no]

## Summary of Changes

Increased event subscription frequency from 2Hz to 10Hz for wind/speed updates

## Video Demo

**Demo of braking**
https://youtu.be/HNmfSFqwQlA

**Demo of taking off**
https://youtu.be/2K3j6WIwcJw

## References

A380 VIDEO 
https://youtu.be/hDh_9MEz-hg?si=lJ4N-CDYFLmz8xo1&t=93

A320 VIDEOS
https://youtu.be/TEi-WX08ov8?si=xCeeyAiIY2FiGs_7&t=36
https://youtu.be/WF7XWAUmOKY?si=SozzSBCR8cKFhM7_&t=539

Look at the ND and watch the wind direction and speed values change

## Additional context

Discord username (if different from GitHub): senolitam

### A32NX & A380X
- **During flight:**
 - Observe wind direction and speed values on Navigation Display (ND)
 - Values should update more frequently compared to previous behaviour

- **During landing/braking:**
 - Monitor ground speed values
 - Speed changes should be more responsive and less delayed during braking


fbw-common/src/systems/instruments/src/ND/ND.tsx
```diff
@@ -532,7 +532,7 @@ class SpeedIndicator extends DisplayComponent<{ bus: EventBus }> {
 
     sub
       .on('groundSpeed')
-      .atFrequency(2)
+      .atFrequency(10)
```
bulenteroglu, line 535: Just a question to the maintainers, I had a gander around the codebase to see other values being used with `atFrequency` - I went with 10, it does the job but also wanted to make sure this value is sensible

lukecologne: I just looked it up and it seems the wind speed and direction are indeed transmitted every 100ms, so 10Hz is correct. Ground speed and true heading is transmitted every 40ms, and true airspeed every 125ms.
Although this is done at the level of transmissing in the transmitting unit (i.e. IR/ADR) in the real aircraft

flogross89:
Before changing this, we should really separate the high-frequency SVG parts from the lower frequency SVG parts, meaning the wind indicator (or the complete info section at the top) should get its own <svg> tag. Otherwise this would lead to considerable performance degradation

flogross89:
Sorry, clicked "Close with comment" instead of just "Comment" :)

flogross89:
@bulenteroglu Sure!
Currently, `<WindIndicator>` and `<SpeedIndicator>` (fbw-common\src\systems\instruments\src\ND\ND.tsx, lines 428-429) are both part of one SVG element spanning all ND elements, when the ND is displayed. The SVG element for that is defined in line 371 (on current master). One of the elements inside the SVG changing is leading to the whole SVG being redrawn (I'm fuzzy on the details, but that's how I understand it). So with your PR, the whole ND SVG would be redrawn 10 times per second, which degrades performance. 

In the end we should have separate SVG element for the Wind and SpeedIndicator, and leave the rest of the ND as is for now. Look at line 352, it's done similarly there.

lukecologne:
Would it actually redraw at that frequency? Because the values are rounded, I would expect a significantly slower redraw rate

flogross89:
> Would it actually redraw at that frequency? Because the values are rounded, I would expect a significantly slower redraw rate (unless the values are actually changing very quickly, which I'd say is rare)

I wouldn't expect it to be much lower, as we can see in the videos above the wind direction changes quite a lot relative to the aircraft, even more so if it's gusty. It's still a change we need to do in our instruments, so better start doing it now :)

tracernz:
> In the end we should have separate SVG element for the Wind and SpeedIndicator, and leave the rest of the ND as is for now. Look at line 352, it's done similarly there.

If we are doing that it would be better to just do them in HTML for much better perf. SVGs should generally be static only for best perf in Coherent.

2hwk:
PR seems stale, marking as draft. Please mark as ready for review when ready.

PR #9659, opened 2024-12-14
feat(a32nx): accurate altimetry and split baro corrections

tracernz:
Fixes #6717
Fixes #8397
Fixes #9649
Fixes #9507

## Summary of Changes

- Implement accurate altimetry computations from static pressure.
- Split EIS baro corrections between captain and first officer side.
- Increased available baro correction range to 745-1100 hPa.
- Add baro ref and altitude discrepancy FWC warnings.
- Add CHECK ALT message on PFD.
- Fixed the A380X to the extent necessary for everything to still work, and extended the baro range to 745-1100 hPa. We cannot split the baro corrections here due to the 3D model, but plenty of prep work is done.

## Additional context

Discord username (if different from GitHub):

## Testing instructions

Perform full flights and look out for any altitude errors. Cross-check with the ISIS.
Perform failure scenarios of FCUs, ADRs.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #9831, opened 2025-02-06
feat(a380x/vd): Vertical display basic implementation

### LINKED ISSUE #8728: [A380X] Implement VD
### Aircraft Version

Development

### Description

Implement vertical display, and replace current placeholder. 

### References (optional)

_No response_

_No response_

_No response_

### LINKED ISSUE #9236: pink block/missing data on VD
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-10-31T15:54:41+00:00",
    "ref": "refs/tags/v0.12.0",
    "sha": "2fe89113291f2478347dde0dc3228b55dc924f74",
    "actor": "FoxtrotSierra6829",
    "event_name": "manual",
    "pretty_release_name": "stable/v0.12.0",
    "version": "a380x-v0.12.0-rel.2fe8911"
}
```

### Describe the bug

VD showing pink "block" on 640nm range. Looks like missing data to me.

### Expected behavior

Showing the usual terrain hight for every distance.

### Steps to reproduce

1. Activate TERR on ND/VD
2. Select 640NM range

### References (optional)

image

Pink "block" is not moving while rest of the terrain data moves as indented. 

maoam1123

flogross89:
Fixes #8728 
Fixes #9236 

## Summary of Changes

Adds an implementation of the vertical display which can be expanded by other contributors
Scope of this PR:

- [x] Transmit vertical path from FMS to VD
- [x] Set up instrument with canvas for drawing vertical map, altitude band, range markers etc.
- [x] Add range change message, "VD RANGE CHANGE"
- [x] Add constraints
- [x] Add plane symbol, orientation depending on pitch
- [x] Missed constraint symbology
- [x] Add AFS CP selected altitude including ALT SEL marker, different colors cyan/magenta/white
- [x] Vertical selected mode path
- [x] Fix distance along currently active leg
- [x] "End of VD" marker in ND
- [x] Grey area if path deviates more than 3° from current track
- [x] Update SimBridge: Altitude band, range, flight plan path
- [x] VD: TAWS failure messages
- [x] TERR on VD as long as TAWS is not INOP
- [x] Change MSFS -> SimBridge communication for TAWS parts from WASM to TS simbridge-client
- [x] Hook up TERR SYS on SURV page
- [x] TERR/GPWS failure via EFB
- [x] ECAM faults for TERR/GPWS
- [x] Level-off segments / dashed line when AFS CP altitude limiting climbs / descents

Out of scope (ideas for future PRs): 

- Integration of descent VNAV profile
- Radio altimeter altitude visualization
- Safety altitudes
- Manual AZIM mode
- Different TAWS RNP values, using 1nm for now
- Symbol for "not valid pitch attitude" (yellow dot, ATT marker)
- Holding patterns
- FLS beams

## Screenshots (if necessary)

vd1_kl

## Additional context

## Testing instructions

You will need a special SimBridge build to test the terrain functionality of the VD: https://www.swisstransfer.com/d/2776f28c-4610-4ba7-bc8e-c8cf665c07b6

A380X: 
Tier 2 tests (normal A-B flights) please.
After starting SimBridge, make sure it's properly connected in the EFB (pull-down menu at top, green SimBridge icon)

**In scope/things you can expect:** 
- Correct indications for not-aligned IRs
- Range change messages
- Airplane symbol in yellow
- FCU alt in cyan, magenta or grey
- Constraints in magenta, missed constraints in amber with dashed line
- TAWS failures in EFB, correct TAWS/WX failure indications; TAWS selectable through MFD SURV
- Vertical path for planned FMS path and actual flown path (also in selected modes)
- Terrain only shown along FMS path, if available and managed lateral mode armed or active. Along acft track is selected lateral mode

**Out of scope/things which still don't work as intended for this basic version, but also won't be fixed with this basic PR:**
- Blue bracket on ND (end of VD marker) might be off
- Grey area (for more than 3° course change) might be off
- Sometimes the VD doesn't update as expected, change range to fix
- Airport+runway symbols in VD are off, wrong distance #10021 
- Things from the "out of scope" list at the top

A32NX: 
Verify that TERR ON ND via SimBridge works as before
Verify that pseudo waypoints and symbols in the ND are displayed as before. 


Ueeew:
@flogross89 im Sorry for asking all the time but could you make this PR public. so that I can download and test it thanks 🙂

Ueeew:
@flogross89 it seams that the 6 checks are always skipped and that why I can’t download it. Is there any way you could disable that these checks aren’t skipped so I could use and test those PR.

SpaceTechAviation:
> @Ueeew  im Sorry for asking all the time but could you make this PR public. so that I can download and test it thanks 🙂

Hello, how are you? As you can see, it hasn't been merged yet and is not available for testing at the moment.

I will send you a message similar to what flogross89 told someone else, answering some questions that are similar to yours and also clarifying a few things.

Is this something everybody can test, or is it just for a few members?

The test builds are public, so theoretically, everyone could test them. However, we only encourage members of our QA team to do so, since bugs may occur, and we need to know when and where they happen. If you're interested in joining our QA team, feel free to reach out to alepouna on Discord. :)

I hope this helps!

flogross89:
@Ueeew Ah yeah, the checks and builds were skipped because the PR is still in draft. With good reason, it just wasn't ready for testing yet. It doesn't help having testing feedback before the PR is ready for review and test, it rather just consumes time.
But now it's ready, I'll push details in a few minutes in the test instructions.

fbw-a380x/src/systems/instruments/src/ND/VdSimvarPublisher.tsx
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2024 FlyByWire Simulations
+// SPDX-License-Identifier: GPL-3.0
+
+import { EventBus, SimVarDefinition, SimVarValueType } from '@microsoft/msfs-sdk';
+import { UpdatableSimVarPublisher } from '../MsfsAvionicsCommon/UpdatableSimVarPublisher';
+
+export type VdSimvars = {
```
Benjozork, line 7: When making new publishers I think we should organise them by system, not by the receiving display.

@tracernz might have thoughts

fbw-a32nx/src/systems/instruments/src/ND/NDControlEvents.ts
```diff
@@ -38,6 +38,11 @@ export interface NDControlEvents {
    */
   set_map_recomputing: boolean;
 
+  /**
+   * Set the reason for recomputing (RANGE CHANGE or MODE CHANGE)
+   */
+  set_map_recomputing_reason: EfisRecomputingReason;
```
tracernz, line 44: What if it's both? This would need to be a bitmask.

fbw-a380x/src/systems/instruments/src/ND/VdSimvarPublisher.tsx
flogross89: That's right, I believe we already started that somewhere. I'll look into restructuring it.

fbw-a32nx/src/systems/instruments/src/ND/NDControlEvents.ts
flogross89: I stayed with an enum for now and added a "rangeAndMode" state

tracernz: An enum can have flags in it, e.g.
```ts
enum EfisRecomputingReason {
  RangeChange = 1 << 0,
  ModeChange = 1 << 1,
}
```

fbw-a380x/src/systems/instruments/src/ND/VdSimvarPublisher.tsx
tracernz: > When making new publishers I think we should organise them by system, not by the receiving display.
> 
> @tracernz might have thoughts

100% agree. The topics should also be decently namespaced and in snake_case.

fbw-a32nx/src/systems/instruments/src/ND/NDControlEvents.ts
flogross89: But the A380 won't show both messages at the same time, so somewhere we'd need some sort of priority

fbw-a380x/src/systems/instruments/src/ND/VdSimvarPublisher.tsx
tracernz: Example https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/shared/src/publishers/A32NXFcuBusPublisher.ts

fbw-a32nx/src/systems/instruments/src/ND/NDControlEvents.ts
tracernz: There was a bunch of discussion and findings on how that works on the recent PR for the range/mode change messages.

flogross89: > There was a bunch of discussion and findings on how that works on the recent PR for the range/mode change messages.

There was no information about any precedence of messages or both messages being shown at the same time from what I can see. MODE CHANGE will have priority over RANGE CHANGE (since some mode change will also automatically trigger a range change), but apart from that I don't see any conflict with my current implementation

fbw-a380x/src/systems/instruments/src/ND/VdSimvarPublisher.tsx
flogross89: > Example https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/shared/src/publishers/A32NXFcuBusPublisher.ts

Any rules on when to use https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/instruments/src/MsfsAvionicsCommon/providers/FcuBusPublisher.ts and when to use /shared/src/publishers? We seem to have some duplication there in the a32nx

flogross89: restructured the publishers now

tracernz: > > Example https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/shared/src/publishers/A32NXFcuBusPublisher.ts
> 
> Any rules on when to use https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/instruments/src/MsfsAvionicsCommon/providers/FcuBusPublisher.ts and when to use /shared/src/publishers? We seem to have some duplication there in the a32nx

The other one should be deprecated. It was written a very long time ago in old style, even if only recently merged.

fbw-a32nx/src/systems/instruments/src/ND/NDControlEvents.ts
tracernz: Maybe it was on discord. The A380 does not have a priority other than the order the events occur, whereas the A320 does have a defined priority (there were some videos demonstrating it on the 380). We should use the same data and logic for both to avoid annoying bugs like desync or differences in logic.

flogross89: > Maybe it was on discord. The A380 does not have a priority other than the order the events occur, whereas the A320 does have a defined priority (there were some videos demonstrating it on the 380). We should use the same data and logic for both to avoid annoying bugs like desync or differences in logic.

So if I understood you right we'll be using the same logic for ND and VD, to ease maintainability? I'm fine with that, just pushed a commit implementing that.

alexr4339:
Gave it an initial test.
Departed KFLL 10L with a GW of 400T. Right after departure VNAV/VD recaculated the profile way above the aircraft. I think the reason may be that the profile is drawn from the runway beginning but you do lift of at the end which will give a vertical diffeence
image
Another point I noticed is that the terrain map does not really update when flying many circles:
image
After turning off and on
image
And for some reason druing vectoring VD has drawn the runway but only temporarily, around half a minute later it disappeared
image

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplay.tsx
```diff
@@ -0,0 +1,838 @@
...
+    this.targetAltitude,
+    this.verticalRange,
+  );
+
+  private readonly altitudeTargetColor = MappedSubject.create(
+    ([mode, altCstr]) => {
+      if (mode === VerticalMode.ALT_CST || mode === VerticalMode.ALT_CST_CPT || altCstr) {
+        return '#ff94ff';
+      } else if (
+        mode === VerticalMode.FINAL ||
```
BravoMike99, line 270: There's no FINAL on the A380 although I guess it could replace APP-DES

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplay.tsx
```diff
@@ -0,0 +1,838 @@
...
+    this.vdRange.map((value) => (value / 4) * 2),
+    this.vdRange.map((value) => (value / 4) * 3),
+    this.vdRange.map((value) => (value / 4) * 4),
+  ];
+
+  private readonly activeLateralMode = ConsumerSubject.create(this.sub.on('fg.fma.lateralMode'), 0);
+  private readonly armedLateralMode = ConsumerSubject.create(this.sub.on('fg.fma.lateralArmedBitmask'), 0);
+  private readonly shouldShowTrackLine = MappedSubject.create(
+    ([active, armed]) =>
+      (active === LateralMode.NONE ||
```
BravoMike99, line 215: Perhaps this could be moved to an util/static method? Just so its not duplicated here and in the trackline.tsx

fbw-common/src/systems/shared/src/publishers/AdiruBusPublisher.ts
```diff
@@ -0,0 +1,52 @@
...
+  IndexedEventType,
+  PublishPacer,
+  SimVarPublisher,
+  SimVarPublisherEntry,
+  SimVarValueType,
+} from '@microsoft/msfs-sdk';
+
+interface AdiruBusBaseEvents {
+  /** Raw Arinc 429 word. */
+  a32nx_adiru_true_track: number;
```
tracernz, line 15:
```suggestion
  a32nx_ir_true_track: number;
```
IR and ADR have separate bus outputs. The publishers should be arranged around output busses. Thew JSDoc should also specify what data is contained within.

fbw-common/src/systems/shared/src/publishers/AdiruBusPublisher.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type IndexedTopics = 'a32nx_adiru_true_track';
+
+type AdiruBusIndexedEvents = {
+  [P in keyof Pick<AdiruBusBaseEvents, IndexedTopics> as IndexedEventType<P>]: AdiruBusBaseEvents[P];
```
tracernz, line 21: The ADR and IR bus publishers should use a more restrained type that retricts the indices to 1 | 2 | 3. The ADR bus publisher in https://github.com/flybywiresim/aircraft/pull/9659/files#diff-a52ba40e0f76f9fd77751641600f0b1b599c203bc4c018fa4c2515db0065a3ef might be a good reference.

tracernz (commented):
The VD terr should go inop in polar regions.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/PathVector.ts
```diff
@@ -44,6 +45,13 @@ export interface DebugPointPathVector {
 
 export type PathVector = LinePathVector | ArcPathVector | DebugPointPathVector;
 
+export interface VerticalPathCheckpoint {
```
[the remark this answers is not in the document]
flogross89: It doesn't belong to the a32nx specific defs (hence I'll remove it here), but I think it should be present in fbw-common's PathVector. It's relevant for the vertical path transmission, IMO PathVector.ts could also cover vertical aspects, not only lateral. What's your proposal?

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplay.tsx
flogross89: Good idea

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/PathVector.ts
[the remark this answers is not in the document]
Benjozork: In my branch, I had made the vertical path sending completely separate. And enables via an FMS config option

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplay.tsx
flogross89: Yep, but I didn't want to touch this enum here, should be done as part of a bigger FG PR. How about keeping it here to have a reminder on what to change if someone decides to implement F-GS and F-LOC (we don't have APP-DES in the a380)

fbw-a380x/src/systems/systems-host/systems/EfisTawsBridge.ts
```diff
@@ -0,0 +1,450 @@
...
+            .map((p) => {
+              const waypoint: WaypointDto = {
+                latitude: p.startPoint.lat,
+                longitude: p.startPoint.long,
+              };
+              return waypoint;
+            });
+
+      const data: ElevationSamplePathDto = {
+        pathWidth: 1,
```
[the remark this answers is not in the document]
flogross89: Yep, listed it in the "out of scope" list as "Different TAWS RNP values, using 1nm for now". I'll add a FIXME, needs more verification in SimBridge.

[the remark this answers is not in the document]
flogross89: .. And the logic is a lot more complicated, depends on EPU, FMS RNP, predicted RNP, standard TAWS RNP, depending on what we're sampling

fbw-common/src/systems/shared/src/publishers/AdiruBusPublisher.ts
flogross89: Ok, I renamed the file to IrBusPublisher.ts and changed everything accordingly. Didn't all of the IR outputs though, only the one I need for now. The PR is massive as it is right now...

fbw-common/src/systems/shared/src/publishers/AdiruBusPublisher.ts
flogross89: added as well

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/PathVector.ts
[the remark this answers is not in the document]
flogross89: I think there are a lot of nice synergies when combining the lateral and vertical path transmission. Maybe in the future it makes sense to de-couple this, when we want to improve the prediction of vertical paths for selected modes, amongst other necessary improvements.

fbw-common/src/systems/shared/src/simbridge/Taws/taws.ts
```diff
@@ -0,0 +1,67 @@
+export interface TawsEfisDataDto {
```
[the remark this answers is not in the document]
flogross89: added

flogross89:
> The VD terr should go inop in polar regions.

Ah yeah totally, also made that check in the VD for displaying the flag, but it should be at the place where the failure conditions are checked, which is EfisTawsBridge. Moved it there.

flogross89:
@alexr4339 About "behind FPLN line straight down": Couldn't reproduce it
image

But I'm very sure that we are going to have problems with matching our position to existing trajctories, it's a very tricky problem doing this right. We might have to keep improvements on that for later.

alexr4339:
Test run 2 using 6982c8b:

Issues:
Colors still seem to dark
I would suggest HEX: 626272 for now. Should be much closer to the following ref:
image

Grey area incorrectly drawn:
image
In this case there should be no grey area, as fplan and track match up.
Interesting was, that this issue only happened on the capt side, FO side displayed grey area correctly
image

Another note: See that simbridge has still drawn TERR on ND, even though I just recently turned TERR off

VD didnt match VNAV
image
See that VD started a descent from PPOS, while VNAV predicted the decent later on. Would expect VD to draw a horizontal line and start the descent at the point displayed by the arrow.

What did I do?
Fplan:
CYOW/14 N0477F380 DCT TUKIR IMEBA9 CYYZ/33R
Did a pretty much routine flight, departed, went direct TUKIR, climbed to my cruise alt of FL200, initiated the descent at TOD marker and flew a standard ILS and select a DTO towards the waypoint before the FAF in cruise

flogross89:
@alexr4339  Why are you saying it's "still" too dark? I thought it was too bright before? In your ref image the exposure was pretty low due to a lot of ambient light. In other refs the area looks way brighter. Let's go with #4e4e61 for now, and re-evaluate after merging or when we have more data points. 

About the grey area: I might have to debug that further. 

TERR still on ND even if TERR is off: Most likely also a SimBridge cycle refresh issue, can't/won't fix that in this PR.

VD didn't match VNAV: Seems like the T/D pseudo waypoint didn't get considered, I'll see whether I'm missing something.

flogross89:
@flogross89 Just updated the SimBridge build with the new colors

alexr4339:
@flogross89 apologies, meant too bright for sure. I'll re-evaluate once with the new simbridge build!

flogross89:
@alexr4339 New SimBridge build and code update, should fix some issues with the end of VD marker, the grey area, and possibly some other stuff. 

Didn't fix the terrain color in the grey area yet, didn't notice that until now. 

If you see an erroneous grey area, change range to reset the SimBridge image to make sure it really is a wrong grey area extent, and not some sort of ghosting.

alexr4339:
Tested with latest build from 08.03.2025

- [ ] Grey area first showed entire area, then after cycling TERR there was none at all
image
image

Positives:
- Grey area now has a much more reasonable color tone!

flogross89:
@alexr4339 Thanks a lot for the test!
1) See discord, my latest commit should fix that
2) I agree, this is baked into our VNAV code at the moment though, so not easy to fix. Will leave this for later
3) Hmm, this might be due to us using the same data as the ND for the lateral path. Don't know how the ND receives the previous segment. Might have to switch to transmitting my own data, will look at it after next week
4) Was this persistent or did it disappear after a while? Really interesting, maybe invalid fpln leg coordinates were transmitted? 
5) I will have to check what our VNAV transmits for ILS approaches, don't know actually. I assume this was persistent, right? So it didn't disappear.

alexr4339:
@flogross89 

4. The magenta area changed as the track of the aircraft varied but while being on that track the magenta area seemed to be rather constant. Will try to gather more info there

5. Correct, that was persistent, the ILS/approach never appeared on the VD. The waypoints appeared and it seemed as if they appeared to be placed at the correct altitude. However there was no connecting line

flogross89:
@alexr4339 Had another debug session and found a few things:
Concerning your 3): I'm now also drawing the previous leg, seems it was systematically not drawn
Concerning your 4): The TERR module drew a pink area whenever the FMS FPLN was finished. I modified that so that we now continue fetching the terrain after FMS FPLN has ended, with the same track as the last segment
Concerning your 5): There was a systematic error in the association of altitude, which meant that the last segment also was not drawn. This should be fixed now

The VD path should look a lot better now, I'd appreciate if you could give it another spin. The distance where the grey area starts won't always be precise though, that will be a systematic shortcoming of this PR's implementation. I would need to change a lot more things in the backend for this to work more precisely.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -189,6 +191,10 @@ export class EfisSymbols<T extends number> {
     if (mrpChanged) {
       this.lastMrp = mapReferencePoint;
     }
+    const pposChanged = distanceTo(this.lastPpos, ppos) > 2;
+    if (pposChanged) {
+      this.lastPpos = ppos;
+    }
```
tracernz, line 197: This should not be needed, as the map only depends on ppos when it is the MRP.

flogross89: Would be a pity if we would have to use a different method for a very similar, if not the same task. When you talk about A320 map symbol generation, are you also excluding usage in the A380 ND? 
How is it not usable for the VD? One pitfall I see is that if the symbol is not within the edit area it won't get transmitted, but nothing else. 
lastPpos is only used to derive the distance to the aircraft which is an additional attribute used in the VD.

tracernz: > lastPpos is only used to derive the distance to the aircraft which is an additional attribute used in the VD.

But it causes all the symbols to be re-generated and transmitted when not needed, especially on the A32NX that doesn't have a VD. It's also a bunch of extra data to be serialised/de-serialised which has a perf cost. Does the VD even need this data when the MRP is not AC PPOS?

alexr4339:
New test:
Roue EDDH23 AMLU3B dep, when airborne inserted EDDH, then inserted EDHI ILS23 via HAM, inserted DTO HAM
My previous points 1 and 3-5 seem to be fixed
6. For some reason TERR wasnt displayed on VD at a certain distance
image

7. DECEL not drawn correctly
8.  Approach RWY missing
image
image
9. Last segment drawn but VNAV calculates shallower profile than 3 deg
image




General: 
Grey area was not super precise but worked way better than before

image
Jojo displaed below profile but VD displayed above profile

And have we already decided whether we want to merge the new display colors #9945? because if so it may make sense to integrate the colors in this PR as the VD grey area color is defined in simbridge and not affected by the color definitions, correct?

fbw-common/src/systems/fmgc/constraint.ts
```diff
@@ -30,7 +30,10 @@ export interface SpeedConstraint {
 }
 
 export class ConstraintUtils {
-  static minimumAltitude(constraint: AltitudeConstraint): Feet {
+  /**
+   * @returns minimum altitude in feet
+   */
+  static minimumAltitude(constraint: AltitudeConstraint): number | undefined {
```
BlueberryKing, line 36: Can this really return `undefined`?

BlueberryKing (changes requested):
Looks good for the most part. The big thing to look at for me would be whether we can treat the VD marker as a pseudowaypoint and think about moving the VD path generation logic to the FMGC.

alexr4339:
Tested 26f103

Route EDDH23 AMLU3B, EDHII23 via HAM, deleted the disco in between. CI 40, FL070

Probably an edge case: Should the runway of the arrival be drawn on the 10 NM marker as its actually 10 NM away or should it be drawn at the end as acccording to the fplan its still 52 away?
image

Started Simbridge while in flight. When toggling it off, TERR is still drawn. An ND range change fixes that
image

Seems as if runway is only drawn on VD when aircraft is on track towards rwy. I think it should always be drawn at the end but thats just what I think would make sense, dont have any refs by now
image

TERR ( brown ground) on VD should always be drawn regardless of TERR mode
image

Plane lost VNAV profil while on LOC intercept
image

Incorrectly drawn ILS and TERR display on VD does not math with altitude it should be (should be 0 ft but is drawn at roughly -500)
image

Go around: Evn on touchdown caluclates profil for same alt
image
Calculates profile only once in GA mode
image
I think the calculated GA path should be shown regardless
image



Overall its in a much better state and pretty usable! Needs a bit more polishing but feels much closer now to "getting there"

I think the approach/ILS segment drawing need some attention still. As well as the handling of the GA segment. Another note: With this build I have not seen the grey area at all.

Old points:
1: (have not seen grey area at all)
2: better
3: fixed
4: fixed
5: fixed
6: havent found that issue
7: fixed
8: fixed
9: still present but better

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplayCanvasMap.tsx
```diff
@@ -137,7 +136,7 @@ export class VerticalDisplayCanvasMap extends DisplayComponent<VerticalDisplayCa
     context.strokeStyle = '#0f0';
     context.lineWidth = 2;
 
-    if (
+    /* if (
```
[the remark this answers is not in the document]
flogross89: That's the rest of the special selected modes logic, haven't cleaned it up yet in case ND profile doesn't work out :) 
But it seems it works, so I'll remove it.

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplayCanvasMap.tsx
```diff
@@ -0,0 +1,369 @@
...
+    this.pposLon,
+    this.baroCorrectedAltitude,
+  );
+
+  private readonly offsetDistance = this.props.fmsTargetVdProfile.map((_path) => 0);
+
+  private readonly activeVerticalMode = ConsumerSubject.create(this.sub.on('fg.fma.verticalMode'), 0);
+  /* private readonly selectedVs = ConsumerSubject.create(this.sub.on('a380x_fcu_selected_vertical_speed'), 0);
+  private readonly selectedFpa = ConsumerSubject.create(this.sub.on('a380x_fcu_selected_fpa'), 0);
+  private readonly groundSpeed = Arinc429LocalVarConsumerSubject.create(this.sub.on('groundSpeed'), 0); // FIXME ADIRS selection for ND not implemented yet */
```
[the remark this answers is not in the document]
flogross89: Not needed as of now, I'll remove the publisher as well.

fbw-a380x/src/systems/instruments/src/MsfsAvionicsCommon/providers/MfdSurvPublisher.ts
```diff
@@ -14,4 +22,14 @@ export interface MfdSurvEvents {
   mfd_tcas_alert_level: number;
   /** (MFD SURV -> TCAS) TCAS Alt Select 0 - NORM | 1 - ABV | 2 - BLW */
   mfd_tcas_alt_select: number;
+  /** (FMS -> TERR SYS) FMS active flight plan managed profile for vertical display */
+  a32nx_fms_vertical_target_profile: VerticalPathCheckpoint[];
+  /** (FMS -> TERR SYS) FMS active flight plan descent profile for vertical display */
```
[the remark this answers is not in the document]
flogross89: fixed, thanks

fbw-a380x/src/systems/instruments/src/ND/VerticalDisplay/VerticalDisplay.tsx
```diff
@@ -0,0 +1,825 @@
...
+        break;
+      case 160:
+        altitudePerDash = 10000;
+        break;
+    }
+    return Math.max(0, Math.ceil(verticalRange[0] / altitudePerDash) * altitudePerDash) + altitudePerDash * index;
+  }
+
+  calculateAndTransmitEndOfVdMarker() {
+    const isInVdMapMode = this.ndMode.get() === EfisNdMode.ARC || this.ndMode.get() === EfisNdMode.ROSE_NAV;
```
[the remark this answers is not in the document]
flogross89: Indeed, ideally this end of VD marker and the start of the grey area should be PWP, and use the same logic. I'll have a look how to accomplish that.

fbw-common/src/systems/instruments/src/ND/shared/map/CanvasMap.tsx
```diff
@@ -141,6 +144,7 @@ export class CanvasMap extends DisplayComponent<CanvasMapProps> {
     sub.on('set_map_up_course').handle((v) => this.mapRotation.set(v));
     sub.on('set_map_pixel_radius').handle((v) => this.mapPixelRadius.set(v));
     sub.on('set_map_range_radius').handle((v) => this.mapRangeRadius.set(v));
+    sub.on('endOfVdMarker').handle((symbol) => (this.endOfVdMarkerData = symbol));
```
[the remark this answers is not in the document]
flogross89: see above, ideally it can be a PWP

fbw-common/src/systems/fmgc/constraint.ts
flogross89: Yes, both `constraint.altitude1` and `constraint.altitude2` can be `undefined`. Shouldn't happen as `constraint?.altitudeDescriptor` shouldn't resolve to these cases, but strict mode complains anyway

fbw-common/src/wasm/terronnd/src/navigationdisplay/collection.cpp
```diff
@@ -34,10 +42,10 @@ Collection::Collection(simconnect::Connection& connection)
     this->_egpwcData.presentLongitude =
         types::Arinc429Word<types::Angle>::fromSimVar(this->_aircraftStatus->value<EgpwcPresentLong>(), types::degree);
     this->_egpwcData.altitude = types::Arinc429Word<types::Length>::fromSimVar(this->_aircraftStatus->value<EgpwcAltitude>(), types::feet);
-    this->_egpwcData.heading = types::Arinc429Word<types::Angle>::fromSimVar(this->_aircraftStatus->value<EgpwcHeading>(), types::degree);
+    this->_egpwcData.heading  = types::Arinc429Word<types::Angle>::fromSimVar(this->_aircraftStatus->value<EgpwcHeading>(), types::degree);
     this->_egpwcData.verticalSpeed =
         types::Arinc429Word<types::Velocity>::fromSimVar(this->_aircraftStatus->value<EgpwcVerticalSpeed>(), types::ftpmin);
-    this->_egpwcData.gearIsDown = static_cast<std::uint8_t>(this->_aircraftStatus->value<EgpwcGearIsDown>()) != 0;
+    this->_egpwcData.gearIsDown            = static_cast<std::uint8_t>(this->_aircraftStatus->value<EgpwcGearIsDown>()) != 0;
```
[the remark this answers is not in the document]
flogross89: I swear I'm innocent, the auto formatting keeps doing it :) Can't change it.

flogross89:
@alexr4339 Thanks.
End of VD marker implementation changed in the last commit, could be worth a try whether it still works.
Runway display also has been improved, should be consistent now and always displayed.

TERR shown when TERR disabled and SimBridge started & VD not showing: Don't know where this comes from, I'm increasingly frustrated from all the SimBridge issues. Will probably leave this for after the PR.

VNAV issues (not calculating for GA, lost during LOC intercept): VNAV bugs, out of scope for this PR. The VD will likely uncover come VNAV bugs, we can create issues for them after merge.

Incorrectly drawn symbol/TERR altitude: Was the QNH set properly? I've seen this before with an improper QNH. Still interesting why it's plotting negative TERR altitudes there. But out of scope for now. 

Grey area not at all: That's certainly weird, was SimBridge properly connected in the EFB?

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/PseudoWaypoints.ts
```diff
@@ -172,6 +179,37 @@ export class PseudoWaypoints implements GuidanceComponent {
...
+      ) {
+        const vdRangeLeft =
+          this.guidanceController.efisStateForSide.L.mode === EfisNdMode.ARC
+            ? Math.max(10, Math.min(this.guidanceController.efisStateForSide.L.range, 160))
+            : Math.max(5, Math.min(this.guidanceController.efisStateForSide.L.range / 2, 160));
+        const endOfVdLeft = this.createEndOfVdMarker(geometry, 'L', vdRangeLeft);
+        if (endOfVdLeft) {
+          newPseudoWaypoints.push(endOfVdLeft);
+        }
+      }
```
BlueberryKing, line 196: Maybe you can move these blocks of logic into their own function or add them into `createEndOfVdMarker`. That protects us from bugs introduced by only updating one of the blocks.

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/VnavDriver.ts
```diff
@@ -52,7 +52,7 @@ export class VnavDriver implements GuidanceComponent {
 
   private headingProfile: NavHeadingProfile;
 
-  private profileManager: VerticalProfileManager;
+  public profileManager: VerticalProfileManager;
```
BlueberryKing, line 55: Instead of making this public, I suggest introducing a wrapper function `computeTacticalToGuidanceProfileOffset` in `VnavDriver` that just calls to `computeTacticalToGuidanceProfileOffset` on `profileManager`. That avoids exposing the entire API of `VerticalProfileManager` to the public.

fbw-common/src/systems/instruments/src/ND/shared/map/PseudoWaypointLayer.ts
```diff
@@ -91,16 +94,17 @@ export class PseudoWaypointLayer implements MapLayer<NdSymbol> {
             (this.groundSpeed.value * (Date.now() - this.lastUpdateTime)) / 1000 / 60 / 60) *
           mapParameters.nmToPx;
         const rotate = MathUtils.diffAngle(this.headingWord.get().value, this.trackWord.get().value);
-        context.translate(384, 620);
+        context.translate(384, 378 + mapParameters.centerYBias); // bug for ROSE NAV modes, didn't consider centerYBias
```
BlueberryKing, line 97: Very nice bug to have fixed actually 👍 However, I think you can remove the comment as it will not be clear which you are referring to later on.

fbw-common/src/systems/instruments/src/NavigationDisplay.ts
```diff
@@ -78,19 +79,40 @@ export enum NdSymbolTypeFlags {
   MagentaColor = 1 << 31,
 }
 
-export interface NdSymbol {
+export enum NdSymbolTypeFlags2 {
```
BlueberryKing, line 82: Maybe you can add that earlier comment describing why this exists here as well.

BlueberryKing (changes requested):
Much cleaner to me now! Just a few small things.

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -542,13 +557,62 @@ export class EfisSymbols<T extends number> {
...
+        (this.side === 'L' && pwp.efisSymbolFlag2 & NdSymbolTypeFlags2.RightSideOnly) ||
+        (this.side === 'R' && pwp.efisSymbolFlag2 & NdSymbolTypeFlags2.LeftSideOnly)
+      ) {
+        continue;
+      }
+
+      // End of VD marker needs direction
+      let direction = undefined;
+
+      if (pwp.efisSymbolFlag2 & NdSymbolTypeFlags2.PwpEndOfVdMarker) {
```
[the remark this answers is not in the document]
flogross89: Indeed, done

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -781,13 +850,30 @@ export class EfisSymbols<T extends number> {
...
         type,
         constraints: constraints.length > 0 ? constraints : undefined,
+        altConstraint: leg.altitudeConstraint,
+        isAltitudeConstraintMet:
+          predictions && predictions.waypointPredictions.has(i)
+            ? predictions.waypointPredictions.get(i).isAltitudeConstraintMet
+            : true,
         direction,
+        distanceFromAirplane: distanceFromAirplane,
+        predictedAltitude: predictedAltitude,
```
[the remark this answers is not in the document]
flogross89: Indeed, I'm sometimes very verbose :)

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
```diff
@@ -985,6 +1080,45 @@ export class EfisSymbols<T extends number> {
 
     return matchingGeometryLeg.terminationWaypoint.location;
   }
+
+  private transmitNdSymbols(symbols: InternalFmsSymbol[]) {
+    const ndSymbols: NdSymbol[] = symbols.map((s): NdSymbol => {
```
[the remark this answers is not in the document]
flogross89: NdSymbol and VdSymbol are a subset of InternalFmsSymbol, don't exactly know what happens if we pass InternalFmsSymbol to these methods. tracer requested that I keep the NdSymbol type as is.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/PseudoWaypoints.ts
flogross89: totally, makes sense. done.

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/VnavDriver.ts
flogross89: done

fbw-common/src/systems/instruments/src/NavigationDisplay.ts
flogross89: done

BlueberryKing (approved):
LGTM 💯

flogross89:
> LGTM 💯

Wohoo, thanks! :)

Jonny23787:

Discord Username  : jonny_23
Object of testing : #9831 
Aircraft          : A380X
Tier of Testing   : 1
Date              : 17/04/2025 

Testing Process: 

- Completed full flight, tested terrain function using sim brief.
- Used the VSD with various functions, testing range and various vertical changes.

Testing Results: 
Not an official test pass or fail test

Negatives:       
Not all necessarily from this PR however some issues I noticed (I'm not an expert on the VSD).

Probably not because of this PR but just an overlap I found
Screenshot 2025-04-17 105953

Flight path line drawn on the bottom of the VSD
Screenshot 2025-04-17 110419

VSD not showing flight path
Screenshot 2025-04-17 114054

VSD not showing flight path correctly with a step climb (from FL380 to 400 at VALEK down to FL390 at LIPNI just after)
Screenshot 2025-04-17 114459

Same issue just smaller range
Screenshot 2025-04-17 114609

fbw-a32nx/src/systems/fmgc/src/efis/EfisSymbols.ts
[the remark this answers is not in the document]
tracernz: I would not want to copy them like this, but rather have the VD data as sideband data for the A380 only.

flogross89:
@Jonny23787 Image 2&3 fixed, thanks! 
I couldn't reproduce the first image (RANGE CHANGE in green), how did you get that? 
Concerning step climbs, most likely a VNAV issue, so I'm not investigating that right now.

Jonny23787:
@flogross89 all I did was set the ADIRS to NAV mode to start aligning them, inserted my route, went into PLAN mode to have a look at the route then went back to ARC mode, cycling between the two caused both to appear.

alexr4339:
New Test: OTHH/34R N0378F210 ALSE2E ALSEM L305 ITBUL M677 VUTEB VUTE3C OMDB/30L, FL210 with build downloaded on 18/04/25

In SRS mode VD draws two expected routes 
image

flogross89:
@alexr4339 Thanks again! The majority of these findings seem to be related to VNAV or how we compute data, some of these even behave as expected. Let's go through all of them: 

1. SRS has two paths, one solid, one dashed: The predicted FMS path is dashed, and the actual flown path is solid. That's expected, VNAV predicts a very steep climb after t/o atm.
2. The grey area behavior is hard to explain or debug sometimes atm, clearly a bug but will be further investigated after merge
3. Course mismatch: Could you elaborate? The last leg seems to be on 75° on both the FMS and ND
4. Sea at FL40: The left part hasn't been redrawn, some sort of SimBridge issue again. Clearly a bug, but also will be investigated further after merge
5. Destination airport and runway: #10021 tracked here
6. DES before TOD, two lines visible: More refined line selection logic comes after merge of basic implementation
7. ILS mismatch: Looks like VNAV issue, combined with #10021 
Overall I still see some bugs, but no showstopper for merge as of now. Let's continue with T2 tests

heclak:
is anyone getting a drawing delay with the latest VD PR? I was getting huge CoherentGT spikes upon takeoff and the VD started acting up. So I tried turning off the Simbridge connection in the EFB to try and recover. The spikes have improved but now the doesn't seem to draw correctly now. It looks like it's still trying to draw and update the VD? Is there some form of recovery fallback when Simbridge is lost/disconnected?

https://www.youtube.com/watch?v=qWBknLOn7R8

flogross89:
@heclak Oh wow, amazing find. Currently I don’t check whether points are inside the drawing canvas, so for extreme distances it seems like the canvas size gets expanded, leading to these spikes? 
Before merge I should really add a check for that

alexr4339:
When performing a turn around, VD displays routing as soon as weights are entered
image

SpaceTechAviation:

Discord Username : christophercr
Object of testing : https://github.com/flybywiresim/aircraft/pull/9831
Aircraft : A380X
Tier of Testing : 1
Date : 21/04/2025

Testing Process:
Cold&Dark
LFPB/07 N0276F080 DCT ASVOK/N0272F070 DCT LFFC DCT LFXU DCT LFPG/08R

Testing Results:
Not Passed

**Pull Request – A380X: Detailed Report on Vertical Display (VD) System Issues** 

During a series of test flights departing from **LFPB (Paris–Le Bourget Airport)** with the destination set to **LFPG (Charles de Gaulle)**, multiple inconsistencies were observed in the Vertical Display (VD) system and its interaction with the Navigation Displays (NDs) on both the **Captain and First Officer** sides. These inconsistencies affect situational awareness and system reliability, especially during approach phases.

### 🧪 **Observed Behaviors & Issues**

#### 1. 🛫 **Departure from LFPB – Distance Offset on Captain ND**
While positioned on **Runway 07 at LFPB**, I noticed that the **Captain’s ND** (with the range set to 10NM on the EFIS panel) was displaying the runway as if it were significantly farther away from the aircraft's actual position. This misrepresentation did not align with the real location of the aircraft on the runway.

In contrast, the **First Officer’s ND**, also set to a 10NM range, showed a different zoom and behavior: it displayed both LFPB and LFPG runways more centered and seemingly closer. However, even this display was **not accurate**, as it still didn’t reflect the correct ground position of the aircraft.
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 15 45 84
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 19 38 94

#### 2. 🔄 **Switching Behavior Between Captain and FO NDs**
After takeoff, I observed that the **runway remained visible** on the Captain’s ND but in an **offset and misleading position**, suggesting we had departed from a location far from the actual position.

Interestingly, the **First Officer’s ND** was displaying more complete and seemingly higher-priority data, even though both EFIS panels were set identically. This issue **persisted at different stages** of the flight. Later on, it appeared the displays would **randomly swap behavior** — with the Captain ND now showing more data, and the First Officer's becoming less informative.
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 21 41 47
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 21 50 35
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 22 40 47
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 26 18 39

#### 3. ⚠️ **EFIS Filters Not Functioning Correctly on Captain Side**
When selecting the **WPT, VORD, NDB, and ARPT filters** via the EFIS control panel on the Captain's side, **no icons or data appeared on the ND**. On the other hand, using the same filter options on the **First Officer’s side** worked correctly, showing all relevant information on the ND.

I discovered a **temporary workaround** during flight: by **switching the display mode from ARC to NAV or VOR**, and then switching **back to ARC or PLAN**, the Captain's ND would refresh and correctly display the information. This confirms a potential **data refresh or syncing issue** between display modes.

Even after stabilizing at 5,000 feet, the green descent profile line remained above the 5,000-foot level on the VD, showing no adjustment or correction to reflect the actual aircraft altitude. The system did not acknowledge the stable altitude, further confirming that the vertical data being presented is unreliable.
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 27 08 56
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 27 19 19
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 27 27 00
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 30 53 66
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 31 00 62

#### 4. ⬇️ **Incorrect Altitude Indication on Glide Slope**
During descent for an ILS approach into **LFPG Runway 08R**, the VD on the Captain ND indicated that the aircraft was **above the ideal descent path**, even though we were at the correct published altitude for the glide slope.

This persisted until **passing the waypoint FF08R**, at which point the VD finally updated and **matched the actual aircraft altitude**. This delay in correction can cause **false glide path awareness**, especially during manual approaches.

Additionally, the **green vertical line** representing the descent profile stretched **across both airports (LFPB and LFPG)** with no clear endpoint, making it visually difficult to trust or follow.
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 33 43 58
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 35 02 96
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 37 10 91
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 37 17 50
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 39 23 33

#### 5. 🛬 **Final Approach – Green Line Glitch**
While on final approach to LFPG 08R, the **green VD guidance line** was expected to terminate at the runway threshold. However, it **continued past the airport**, becoming **progressively longer and detached from the runway** as the aircraft approached.

Eventually, the **runway itself disappeared** from the ND, as if the system assumed the aircraft had already passed it, even though we were still in final approach phase. This can create major confusion when trying to visually verify landing alignment using the display.
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 39 23 33
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 39 49 40
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 40 53 19
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 41 39 76
Microsoft Flight Simulator Screenshot 2025 04 20 - 22 41 55 37

#### 6. ❓ **Unverified TERR Mode**
I want to note that **I did not test** the **TERR mode** to observe how it behaves under these VD conditions or with the map inconsistencies described above.

Jonny23787:

Discord Username  : jonny_23
Object of testing : #9831 
Aircraft          : A32NX
Tier of Testing   : 1
Date              : 03/05/2025 

Testing Process: 

1. Toggled terrain overlay and confirmed it worked as normal and as before.
2. Inserted and flew FPLN, confirmed all waypoints including pseudo waypoints appeared as normal and as before.

Testing Results: 
Passed

Media:
Screenshot 2025-05-03 135722
Screenshot 2025-05-03 135039
Screenshot 2025-05-03 135924
Screenshot 2025-05-03 140233
Screenshot 2025-05-03 140357

alexr4339:
image
image

flogross89:
@alexr4339 Thank you! For the case of the missing path during approach, when we compare the FMS page to the VD path it seems it's displaying the same information, which is what we want. So seems like a temporary VNAV bug during approach, i.e. out of scope for this PR.

PR #9840, opened 2025-02-08
feat(a32nx): ecam control panel (ecp)

tracernz:
Fixes #[issue_no]

## Summary of Changes

Implemented a full simulation of the ECAM control panel with all bus IO and hardwired signals, excluding the ECAM DU brightness pots. I don't think there's any value in adding a layer of indirection for the pots. All of the buttons have input events and proper tooltips.

This is really setup for future PRs (FWC improvements, and EIS display switching).

Removed some cruft from the model behaviours files, and the SD index.tsx.

Known issue: The EMER CANC button cover only works on MSFS2024. There seems to be a bug in MSFS2020 where it can't find the cover node. Our FWC doesn't have emergency cancel functionality yet so it's not such a problem.

## Screenshots (if necessary)

image

## Additional context

## Testing instructions

A32NX only.
Test that the ECAM system buttons work correctly.
Test that the CLR and RCL buttons work correctly (you will need to create some warnings).
Test that the T.O CONF button works correctly.
Fail the ECP (in the EFB failures under ATA31), and ensure the CLR/RCL and ALL buttons work (holding all buttons causes the ECAM SD to cycle through all the system pages), and the T.O CONF button does not.


2hwk (approved):

heclak:
@tracernz Did a test on this build. I've noticed the ECAM buttons are less responsive than the previous implementation. Short clicks sometimes do not register, and the only way to get them to trigger reliably is to click and hold, waiting for the button animation to bottom out. 

Particularly when trying to press the same button again to hide the page. I've noticed that the less responsive behavior might be more prevalent when trying to hide a page.

1. Press ENG to show ENG page
2. Press ENG again to hide ENG page

Could this latency be tracked down? I think it should be at least equally responsive to the previous build. Maybe it has to do with model behavior timings?

And can I clarify this part of the testing instructions?

>Fail the ECP (in the EFB failures under ATA31), and ensure the CLR/RCL and ALL buttons work (holding all buttons causes the ECAM SD to cycle through all the system pages), and the T.O CONF button does not.

1. All ECAM CP buttons EXCEPT 'CLR', 'RCL', and 'ALL' should be INOP
2. Holding the 'ALL' button when the ECP is failed should still cycle through all the system pages regardless of the failure.
3. TO CONF button should be INOP if ECP is failed.

tracernz:
> Particularly when trying to press the same button again to hide the page. I've noticed that the less responsive behavior might be more prevalent when trying to hide a page.

The button press status is held for a short time, exactly as in the real thing, so that seems accurate.

heclak:
@tracernz I see. Now i understand the reason for the delay. I could be wrong but it appears that repeated clicks of the button might be retriggering the timeout? Is that accurate to how it should work? Or should it be a lock out of the button for a fixed duration so repeated spamming of the button will eventually register after the lockout time?

I've also noticed the animation speed of the button is much slower than before. It would make the buttons feel more responsive visually if the anim speed could be doubled. I've recorded a comparison of the current dev build vs the PR button speed.

Current ECP - https://youtu.be/K2BZtIPS9MM
New ECP - https://youtu.be/6BtAO4owF0o

The button sounds are missing in the PR version.

flogross89:
Didn’t we conclude somewhere else, that we shouldn’t rely on a minimal time of button pressed for simple button interactions, even if the real thing would behave like that? 
Because we are on a PC setup, users expect stuff to work when only clicking buttons and not holding them for some time. 
The above discussion sounds like this is not the case?

tracernz:
> Didn’t we conclude somewhere else, that we shouldn’t rely on a minimal time of button pressed for simple button interactions, even if the real thing would behave like that? Because we are on a PC setup, users expect stuff to work when only clicking buttons and not holding them for some time. The above discussion sounds like this is not the case?

This doesn’t rely on that to detect buttons presses. It does what we discussed with an event for rising edge and an event for falling edge from the XML. This is about the ARINC bus output from the ECP, and it’s the opposite… it sets the level to high for a minimum time to ensure the users detect every press.

tracernz:
@heclak I've adjusted the logic so the hold time on the ARINC bus outputs is not retriggered if you push the button again before the previous press reset, and additionally sped up the bus transmission time so the hold time is lower. It seems very reliable to me. I actually tried the master branch accidentally (wrong git workspace symlinked to community dir), and the TO CONF button is very unreliable for short presses without this PR!

alepouna:

Discord Username  : alepouna 
Object of testing : #9840 
Aircraft          : A32NX 
Tier of Testing   : 1
Date              : 02/06/2025 

Testing Process: 
- Test all  ECP buttons work correctly
- Test CLR and RCL behaves as it should
- Test T.O. CONFIG button behaves as it should
- Fail the ECP and ensure only ALL, CLR and RCL work, while the rest do not

Testing Results: 
Passed

Negatives:       
**Consider this a non pass if the following is important:** 
On the current master version, pressing the STS button takes you to a static status normal, which then takes you back to whatever page was previously selected. 
Video of the behavior on master:

https://github.com/user-attachments/assets/cfb9fd83-699b-4776-896c-401d3d9b9b7f

With this PR, the STS button does nothing. In my opinion, I consider this a non issue as STS has no value (as its not implemented), however it might be weird to some users who pay attention if the page doesn't show up at all.

PR #9866, opened 2025-02-19
fix: invalid bing map calls

tracernz:
Fixes #[issue_no]

## Summary of Changes

Fix invalid bing map calls (wrong param type), ensure only one map listener per vcockpit, document all that, and fix up the A32NX TcasComputer to actually be a singleton rather than a dangerous half/half state where you'd end up with multiple instances if you call .instance.

## Additional context

## Testing instructions

Spawn at a busy airport and ensure TCAS shows some traffic on the ND.


fbw-a32nx/src/systems/tcas/src/loop.ts
```diff
@@ -1,16 +0,0 @@
-// Copyright (c) 2021-2023 FlyByWire Simulations
-//
-// SPDX-License-Identifier: GPL-3.0
-
-import { TcasComponent } from './lib/TcasComponent';
-import { TcasComputer } from './components/TcasComputer';
-
-const components: TcasComponent[] = [TcasComputer.instance];
```
tracernz, line 8: Yikes, given Atc/index.tsx creates an instance through the constructor/new!

fbw-a32nx/src/systems/tcas/src/components/TcasSoundManager.ts
```diff
@@ -19,7 +17,7 @@ class PeriodicSound {
   }
 }
 
-export class TcasSoundManager implements TcasComponent {
+export class TcasSoundManager {
```
Saschl, line 20: Shouldn't this constructor be private as well as it also looks like a Singleton?

tracernz: Heh, good catch, I didn't notice that one is also like that.

Saschl (approved):

heclak:

Discord Username  : Heclak 
Object of testing : #9866 
Aircraft          : A32NX 
Tier of Testing   : 1
Date              : 22/02/2025

Testing Process: 
 

- Spawned at LSZH and tested with BATC traffic ✔️ 
- Spawned at EDDF and tested with VATSIM traffic ✔️ 

Testing Results: 
Passed  

Media:

image
image
image

PR #9761, opened 2025-01-15
feat(fms): implement dir to with radial in/out

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Work in progress.

## Additional context

Discord username (if different from GitHub):

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #9898, opened 2025-03-08
feat(fms): compute capture condition in fms

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Implement the computation of the NAV capture zone in the flight management system. Previously, NAV mode would engage whenever it was armed and XTE < 1NM. Now, the capture zone is computed dynamically based on leg type, ground speed, etc.

This is a prerequisite for #9761.

## Additional context

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/flightplanning/legs/FlightPlanLeg.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/flightplanning/legs/FlightPlanLeg.ts
@@ -216,6 +216,26 @@ export class FlightPlanLeg implements ReadonlyFlightPlanLeg {
     return legType === LegType.CA || legType === LegType.FA || legType === LegType.HA || legType === LegType.VA;
   }
 
+  isCx() {
+    return (
+      this.definition.type === LegType.CA ||
+      this.definition.type === LegType.CD ||
+      this.definition.type === LegType.CF ||
+      this.definition.type === LegType.CI ||
+      this.definition.type === LegType.CR
+    );
+  }
+
+  isVx() {
+    return (
+      this.definition.type === LegType.VA ||
+      this.definition.type === LegType.VD ||
+      this.definition.type === LegType.VI ||
+      this.definition.type === LegType.VM ||
+      this.definition.type === LegType.VR
+    );
+  }
+
   isVectors() {
     const legType = this.definition.type;
 
--- a/fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
@@ -3,8 +3,8 @@
 //
 // SPDX-License-Identifier: GPL-3.0
 
-import { ControlLaw, LateralMode, VerticalMode } from '@shared/autopilot';
-import { MathUtils, TurnDirection } from '@flybywiresim/fbw-sdk';
+import { ControlLaw } from '@shared/autopilot';
+import { Arinc429Register, Constants, LegType, MathUtils, TurnDirection } from '@flybywiresim/fbw-sdk';
 import { Geometry } from '@fmgc/guidance/Geometry';
 import { Leg } from '@fmgc/guidance/lnav/legs/Leg';
 import { LnavConfig } from '@fmgc/guidance/LnavConfig';
@@ -23,6 +23,7 @@ import { VMLeg } from '@fmgc/guidance/lnav/legs/VM';
 import { FMLeg } from '@fmgc/guidance/lnav/legs/FM';
 import { GuidanceController } from '../GuidanceController';
 import { GuidanceComponent } from '../GuidanceComponent';
+import { FlightPlanLeg, isDiscontinuity } from '../../flightplanning/legs/FlightPlanLeg';
 
 /**
  * Represents the current turn state of the LNAV driver
@@ -45,6 +46,8 @@ export enum LnavTurnState {
 }
 
 export class LnavDriver implements GuidanceComponent {
+  private static readonly NavActiveCaptureZone = 1.0;
+
   private guidanceController: GuidanceController;
 
   private lastAvail: boolean;
@@ -57,9 +60,13 @@ export class LnavDriver implements GuidanceComponent {
 
   private lastPhi: number;
 
+  private lastNavCaptureCondition: boolean;
+
   public turnState = LnavTurnState.Normal;
 
   public ppos: LatLongAlt = new LatLongAlt();
+
+  private register = Arinc429Register.empty();
 
   private listener = RegisterViewListener('JS_LISTENER_SIMVARS', null, true);
 
@@ -91,6 +98,10 @@ export class LnavDriver implements GuidanceComponent {
 
     const geometry = this.guidanceController.activeGeometry;
     const activeLegIdx = this.guidanceController.activeLegIndex;
+
+    // this is not the correct groundspeed to use, but it will suffice for now
+    const tas = SimVar.GetSimVarValue('AIRSPEED TRUE', 'Knots');
+    const gs = SimVar.GetSimVarValue('GPS GROUND SPEED', 'knots');
 
     if (geometry && geometry.legs.size > 0) {
       const dtg = geometry.getDistanceToGo(this.guidanceController.activeLegIndex, this.ppos);
@@ -179,10 +190,6 @@ export class LnavDriver implements GuidanceComponent {
       }
 
       // Leg sequencing
-
-      // this is not the correct groundspeed to use, but it will suffice for now
-      const tas = SimVar.GetSimVarValue('AIRSPEED TRUE', 'Knots');
-      const gs = SimVar.GetSimVarValue('GPS GROUND SPEED', 'knots');
 
       const params = geometry.getGuidanceParameters(activeLegIdx, this.ppos, trueTrack, gs, tas);
 
@@ -393,6 +400,8 @@ export class LnavDriver implements GuidanceComponent {
       }
     }
 
+    this.updateNavCaptureCondition(tas, gs);
+
     /* Set FG parameters */
 
     if (!available && this.lastAvail !== false) {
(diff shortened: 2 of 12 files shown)


fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -94,6 +105,10 @@ export class LnavDriver implements GuidanceComponent {
 
     const activeLegIdx = this.guidanceController.activeLegIndex;
 
+    // this is not the correct groundspeed to use, but it will suffice for now
+    const tas = SimVar.GetSimVarValue('AIRSPEED TRUE', 'Knots');
+    const gs = SimVar.GetSimVarValue('GPS GROUND SPEED', 'knots');
```
BravoMike99: Perhaps we could add some registered simvars here?

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -58,10 +63,16 @@ export class LnavDriver implements GuidanceComponent {
...
 
+  private lastNavCaptureCondition: boolean;
+
+  private isNavCaptureInhibited: boolean = false;
+
   public turnState = LnavTurnState.Normal;
 
   public ppos: LatLongAlt = new LatLongAlt();
 
+  private register = Arinc429Register.empty();
```
BravoMike99:
```suggestion
  private readonly fmDiscreteWord2 = Arinc429Register.empty();
```

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -585,4 +571,99 @@ export class LnavDriver implements GuidanceComponent {
...
+  private cannotCaptureLegType(el: FlightPlanLeg): boolean {
+    // FIXME We probably should not prevent capture of IF legs like this. Instead, we should not even generate any guidance parameters
+    // for IF legs, and then the capture condition will not be met.
+    return el.type === LegType.IF;
+  }
+
+  private isNavModeEngaged(): boolean {
+    // TODO use correct FM
+    return this.register
+      .setFromSimVar('L:A32NX_FMGC_1_DISCRETE_WORD_2')
```
BravoMike99: And on this function too?

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -71,10 +78,26 @@ export class LnavDriver implements GuidanceComponent {
...
+    'L:A32NX_FMGC_1_DISCRETE_WORD_2',
+    'number',
+  );
+
+  private static readonly fmgc2DiscreteWord2Var = RegisteredSimVar.create<number>(
+    'L:A32NX_FMGC_1_DISCRETE_WORD_2',
+    'number',
+  );
+
+  private static readonly latitudeVar = RegisteredSimVar.create<number>('PLANE LATITUDE', 'degree latitude');
```
tracernz: Note, you can omit the `<number>`.

fbw-a32nx/src/systems/fmgc/src/flightplanning/legs/FlightPlanLeg.ts
```diff
@@ -243,6 +243,26 @@ export class FlightPlanLeg implements ReadonlyFlightPlanLeg {
     return legType === LegType.CA || legType === LegType.FA || legType === LegType.HA || legType === LegType.VA;
   }
 
+  isCx() {
+    return (
+      this.definition.type === LegType.CA ||
+      this.definition.type === LegType.CD ||
+      this.definition.type === LegType.CF ||
```
tracernz: Not sure the general usefulness of this function, as all the Cx are floating legs except CF, which is actually a ground fixed xF leg. It's existence might lead to bugs. Maybe something like isFloatingCourseLeg that excludes CF would be better.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -585,4 +594,99 @@ export class LnavDriver implements GuidanceComponent {
...
+    this.setNavCaptureCondition(this.computeNavCaptureCondition(tas, gs));
+  }
+
+  private setNavCaptureCondition(condition: boolean) {
+    if (condition !== this.lastNavCaptureCondition) {
+      SimVar.SetSimVarValue('L:A32NX_FM1_NAV_CAPTURE_CONDITION', 'Bool', condition);
+      SimVar.SetSimVarValue('L:A32NX_FM2_NAV_CAPTURE_CONDITION', 'Bool', condition);
+
+      this.lastNavCaptureCondition = condition;
+    }
```
tracernz: Suggestion: you could use a subscribable rather than manually tracking the mutation. RegisteredSimVar is also a good idea.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -585,4 +594,99 @@ export class LnavDriver implements GuidanceComponent {
...
+  }
+
+  private canAlwaysCaptureLeg(leg: FlightPlanLeg) {
+    return leg.isVx() || (leg.isCx() && leg.type !== LegType.CF);
+  }
+
+  private cannotCaptureLegType(el: FlightPlanLeg): boolean {
+    // FIXME We probably should not prevent capture of IF legs like this. Instead, we should not even generate any guidance parameters
```
tracernz, line 618: That's true. null guidance params would perhaps lead to a more natural disengagement of NAV mode when an IF leg is the TO waypoint as well.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
```diff
@@ -585,4 +594,99 @@ export class LnavDriver implements GuidanceComponent {
...
+
+class HpathLaw {
+  static readonly Tau = 3; // seconds
+  static readonly Zeta = 0.8; // 1
+  static readonly G = Constants.G * 6997.84; // kts/h
+  static readonly T = this.Tau / 3600; // hours
+  static readonly K1 = 180 / 4 / Math.PI ** 2 / this.Zeta / this.T; // 1 / h
+  static readonly K2 = this.Zeta / Math.PI / this.G / this.T; // 1 / kts
+  static readonly InterceptAngle = 45;
 }
```
tracernz: We already have some of these constants in other places like the roll anticipation calculation I believe, although that's stretching my memory back a long way!

tracernz (approved):
I really think the isCx existing is quite dangerous, but otherwise just some minor suggestions. Since GitHub doesn't have an "Approve With Suggestions", I just approve 😁 .

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
BlueberryKing: I found this: https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/fmgc/src/guidance/lnav/CommonGeometry.ts#L332. Somehow it seems like `k2` is not the same as I would expect. It should be `0.0045` but it's `0.0038`. I'll leave it as it is.

fbw-a32nx/src/systems/fmgc/src/guidance/lnav/LnavDriver.ts
BlueberryKing: I will leave this for future work. We can enable strict mode on the touched files to make sure we handle the null guidance params case properly everywhere.

BravoMike99:
Dev team QA report

Object of testing : #9898 
Simulator         : 2020
Tier of Testing   : 1
Date              : 14/02/2026 

Testing Process: 
Flew LPPT-EGLL and played around with flying an offset to the route and adding discontinuities mid flight.
NAV capture worked ok.
Bonus: NAV now disengages when flying through a discontinuity

Testing Results: 
Passed

Media:
DISCO

BravoMike99:
Dev team QA report

Object of testing : #9898 
Aircraft          :  A380X
Simulator         : 2020
Tier of Testing   : 1
Date              : 17/02/2026

Testing Process: 
Performed a quick pattern around LPPD.
Flew through discontinuity and checked mode reversion logic. Takes a few seconds before it occurs.
Flew away from procedure, checked that NAV only activates when flying close to the required leg.
Testing Results: 
Not Passed

Negatives:       
After flying through a discontinuity it takes a few seconds for NAV to disconnect. In addition, there is no triple click (in master this is immediate and there is a triple click)
Nav cannot be armed while in HDG but that may be the case aswell on master currently. Instead it goes managed for a bit and then reverts to heading.
Nav capture logic is a bit wonky when flying 180º away from the active leg.

lukecologne:
Hm, this is probably because of these https://github.com/flybywiresim/aircraft/pull/9898/changes#diff-73a6284ff493749b447e1f0de3a9c42ad5eac5fb097dc17ec16523b93c471b94L549-L576 deleted lines, and the fact that the A380X AP does not have the same mode logic with reversions as the A32NX. We might need to do a conditional in the FMS to handle this until the A380X AP is updated

lukecologne:
I have now fixed it by adding a capture condition input to the A380 AP state machine. This should make it functionally closer to the A320 FMGC, and not require any special handling from the FMS. @BravoMike99 could you re-test for the A380? This should fix your issue

BravoMike99:
> I have now fixed it by adding a capture condition input to the A380 AP state machine. This should make it functionally closer to the A320 FMGC, and not require any special handling from the FMS. @BravoMike99 could you re-test for the A380? This should fix your issue

Nav immediately  activated on the ground 😭

lukecologne:
Oops, my bad, should be fixed now

BravoMike99:
Dev team QA report

Object of testing : #9898
Aircraft          : A380X 
Simulator         : 2020
Tier of Testing   : 1
Date              : 21/02/2026 

Testing Process: 
Performed a quick pattern as before flying through discontinuities and away from the active leg.
Mode reversion now triggers immediately alongside a triple click.
NAV now arms in blue when the capture condition is not met and activates automatically when it is satisifed.

Testing Results: 
Passed

Negatives:       
Capture logic when the active leg behind as before.

lukecologne:
Is this good to go from your side @BlueberryKing?

BlueberryKing:
> Is this good to go from your side @BlueberryKing?

Sorry, yes, ready to go 👍

PR #9933, opened 2025-03-19
feat: adr and ir bus publishers

tracernz:
Fixes #[issue_no]

## Summary of Changes

Pulled the new style ADR bus publisher out of #9659 and also added an equivalent IR bus publisher. This is for use by all instruments and systems, and should gradually replace any instrument-specific publishers that duplicated these definitions before.

## Additional context

## Testing instructions

No change, so no test.


fbw-common/src/systems/shared/src/publishers/ArincBus/IrBusPublisher.ts
```diff
@@ -0,0 +1,326 @@
...
+          name: 'L:A32NX_ADIRS_IR_#index#_ROLL',
+          type: SimVarValueType.Enum,
+          indexed: true,
+        },
+      ],
+      [
+        'ir_heading',
+        {
+          name: 'L:A32NX_ADIRS_IR_#index#_HEADING',
+          type: SimVarValueType.Enum,
```
Saschl, line 134: Any reason all of those are type "Enum"?

Saschl (approved):

fbw-common/src/systems/shared/src/publishers/ArincBus/IrBusPublisher.ts
tracernz: We want integers only, no decimals.

PR #9935, opened 2025-03-19
fix: only close simconnect when connection was established

Saschl:
Fixes #[issue_no]

## Summary of Changes

I missed to check if the simconnect handle is actually valid (meaning != 0). This can cause issues in 2024 as one can abort the flight loading.

## Additional context


tracernz:
Is it documented somewhere that 0 is not a valid SimConnect handle? SimConnect_Open returns an HRESULT to tell you if it succeeded or not.

Saschl:
> Is it documented somewhere that 0 is not a valid SimConnect handle? SimConnect_Open returns an HRESULT to tell you if it succeeded or not.

Good point, I just copied that behaviour from another gauge.. :) But as the handle is a pointer, I would assume 0 is simply not valid. But I agree, it would be cleaner to use the acutal result. Will adapt 👍

fbw-common/src/wasm/cpp-msfs-framework/MsfsHandler/MsfsHandler.cpp
```diff
@@ -190,10 +192,8 @@ bool MsfsHandler::shutdown() {
   modules.clear();
   unregister_key_event_handler_EX1(reinterpret_cast<GAUGE_KEY_EVENT_HANDLER_EX1>(keyEventHandlerEx1), nullptr);
   unregister_all_named_vars();
-  if (hSimConnect != 0) {
+  if (simConnectInitialized) {
```
tracernz, line 195:
```suggestion
  if (simConnectInitialized) {
    simConnectInitialized = false;
```

Saschl: *sigh* yup it's getting late here :P

tracernz (approved):

PR #9954, opened 2025-03-28
fix(a380x/efb): add separate button for m4r door, fix interaction with catering

### LINKED ISSUE #9850: EFB door status
### Simulator

2020

### Aircraft Version

Stable (8k)

### Build info

```json
{
    "built": "2025-02-10T22:09:03+00:00",
    "ref": "refs/heads/master",
    "sha": "9fabd5740b0e2640af199033a81a717dddd32862",
    "actor": "BravoMike99",
    "event_name": "manual",
    "pretty_release_name": "master:9fabd574",
    "version": "a380x-v0.13.0-dev.9fabd57"
}
```

### Describe the bug

Door's "Open" status display on EFB does not match the aircraft's door opening animation in External view

### Expected behavior

The rear second compartment door of the fuselage is open, but the "Open" mark is not visible in the EFB's door state

### Steps to reproduce

I called the Catering Truck in the EFB service item and the rear second door's opening

### References (optional)

Image

_No response_

_No response_

dzoeteman:
Fixes #9850

## Summary of Changes

- Added separate button for M4R door (existing aft button was incorrectly labeled in code as being M4R, it was M5R)
- Catering service only blocks M4R door button, not M5R
- Added 'OPEN' tag for M4R door

## Screenshots (if necessary)

Before only one aft door button, which is blocked by catering, and does not show 'OPEN' tag on M4R:

After, two aft door buttons, independently operable, with 'OPEN' tag shown for M4R too:

Catering correctly blocks M4R door button:

## Additional context

The services page probably needs a redesign at some point to include all doors, doesn't really fit at the moment. Maybe a separate doors page.

## Testing instructions

- ensure GSX compatibility is not broken (cannot test myself)
- ensure changes in common code do not break existing a320nx efb services functionality

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -183,6 +183,7 @@
 1. [A32NX/FMS] Removed insert/erase prompts on approach via page - @tracernz (Mike)
 1. [ND] Fixed alignment of approach message - @tracernz (Mike)
 1. [A380X/OANS] Fix RWY AHEAD not disappearing when OANS Performance Mode is active - @flogross89 (floridude)
+1. [A380X/EFB] Add separate button for M4R door with open tag, don't block M5R button on catering - @dzoeteman (Deniz)
 
 ## 0.12.0
 
--- a/fbw-common/src/systems/instruments/src/EFB/Assets/GroundServiceOutline.tsx
+++ b/fbw-common/src/systems/instruments/src/EFB/Assets/GroundServiceOutline.tsx
@@ -426,12 +426,14 @@ export const A380GroundServiceOutline = ({
   main1LeftStatus,
   main2LeftStatus,
   main4RightStatus,
+  main5RightStatus,
   upper1LeftStatus,
 }: {
   className: string;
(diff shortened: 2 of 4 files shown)


matze-tech:
I don't know if it's a good idea to have two buttons labeled "Door AFT" and "Door FWD." Maybe we should use "Door M4R" and "Door M5R" instead.

2hwk:
Looking at it from a UX and usability standpoint, I think this is better done as part of a wider services page redesign re: doors, as this is looking cluttered and I don't think either having it as 2x door aft or door m4r/m5r is really going to make it any more usable (and less distracting) for people using this page. 

I know this is useful for GSX, however, the page was originally designed around default ground services and as such rather than trying to patch piecemeal, this is better left to a redesign.

Another option I was considering was designing a pop-up modal (replacing all the individual door buttons with one single Doors button) that would control all the doors. However just having it as radio buttons or a multi-select is also a bad design.

In short, this is an unsolved problem. I'll write more in the issue when I have time.

PR #9963, opened 2025-03-29
fix(a380x/mfd): Show dist on DIR TO page, no access to DIR TO when TMPY plan

### LINKED ISSUE #9118: DIST information on DIR TO page not showing
### Aircraft Version

Development

### Build info

```json
{
    "built": "2024-10-21T13:15:57+00:00",
    "ref": "HEAD",
    "sha": "1e433bf5a49ba39a13ff868ed117dd0e03bb14d2",
    "actor": "BBK <22713769+BlueberryKing@users.noreply.github.com>",
    "event_name": "manual",
    "pretty_release_name": "upstream-master:1e433bf5",
    "version": "a380x-v0.12.0-upstream-master.1e433bf"
}
```

### Describe the bug

When using DIR TO the distance to the waypoint should be shown.
(Only way currently to get the distance to plan the descent)

Unfortunately it is not shown directly:
image

To make the distance info available one has to move away from the DIR TO page and come back - then it is populated. 
image

### Expected behavior

DIST info should be shown as soon as a waypoint it selected in the DIR TO page. 

### Steps to reproduce

Enter a DIR TO waypoint (on the ground) and observe that there is now dist information populated. 
Open another MFD page and then open the DIR TO again - then it is populated. 

### References (optional)

_No response_

_No response_

cdr_maverick

dzoeteman:
Fixes #9118

## Summary of Changes
- Shows distance on DIR TO page when waypoint is selected
- Disallows access to DIR TO page when a temporary flight-plan is set, and the temp plan did not come from a DIR TO

## Screenshots (if necessary)

Distance shown on fpln waypoint:
https://github.com/user-attachments/assets/e0b9bccd-430a-4d52-8712-f16ccb38b424

Distance shown on custom waypoint:
https://github.com/user-attachments/assets/c2c44ef0-4d4f-46b7-a297-3d4d7c45a0c1

Message pops up when trying to go to dir to page on temp plan (that's not dir to):
https://github.com/user-attachments/assets/33ba1f59-74db-47c5-a276-fb4e1e909317

## References

FCOM

dirto-noaccess

dirto-msg

## Additional context

Would've liked something nicer than continuously updating the distance, but then again, it does continuously update the distance while you're moving. I'm just not sure if that's accurate with IRL - would be great to get some input on that.

Not 100% on the checking navigation in the MFD component, but it's the best place I could find for now - open for other suggestions.

## Testing instructions

Distance shown scenario:
1. Have active flight plan
2. Try to direct to waypoint
3. Notice distance is now shown on DIR TO page

Navigation scenario:
1. Have active flight plan
2. Try to make some other change to the plan, such as a departure set or deletion of waypoint
3. When the plan is in temporary mode, try to switch to DIR TO page
4. Notice message shown and MFD does not navigate to DIR TO page

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/MFD.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/MFD.tsx
@@ -62,6 +62,8 @@ export interface MfdDisplayInterface {
   interactionMode: Subscribable<InteractionMode>;
 
   openMessageList(): void;
+
+  navigateTo(uri: string): void;
 }
 
 export class MfdComponent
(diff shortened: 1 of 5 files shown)


fbw-a380x/src/systems/instruments/src/MFD/MFD.tsx
```diff
@@ -278,6 +281,20 @@ export class MfdComponent
     this.subs.push(this.fmsDataKnob, this.fmcAIsHealthy, this.fmcBIsHealthy, this.activeFmsSource);
   }
 
+  public navigateTo(uri: string) {
```
flogross89, line 284: In order to stay consistent, we should then replace every call to `this.props.mfd.uiService.navigateTo` with `this.props.mfd.navigateTo`. Or did you find a way to check this within uiService?

flogross89 (changes requested):
Looks good so far, thank you! Just one remark/question

tracernz:
> temp plan (that's not dir to):

If it's like the A320 FMS, DIR TOs don't actually go in the TMPY, hence this restriction. Putting them in the TMPY was a quick hack by us to easily get the route to draw etc.

fbw-a380x/src/systems/instruments/src/MFD/MFD.tsx
dzoeteman: @flogross89 I didn't because it's used in many places - that would increase scope of this PR by quite a bit and make it difficult to test as well.

dzoeteman:
> > temp plan (that's not dir to):
> 
> If it's like the A320 FMS, DIR TOs don't actually go in the TMPY, hence this restriction. Putting them in the TMPY was a quick hack by us to easily get the route to draw etc. but we will want to redo it properly at some point because it has some visible side-effects.

This seems to be actually different between the 320 and 380. 
320 FCOM specifically says
````
No temporary flight plan is created with these functions
````

While the 380 FCOM says 
````
The DIR TO revision creates a temporary flight plan.
No other revisions can be made on this temporary flight plan.
````

PR #9972, opened 2025-04-01
fix(380x/fcu): initial selected speed value not managed target

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Changes initial selected speed value on pull or rotation to managed speed target, if not in SRS.
Changes selected speed preselect timeout to 45 seconds.
## References

https://youtu.be/rvTNYmbes0E?t=194 & FCOM
## Additional context

bruno_pt99
## Testing instructions

Verify managed speed can be engaged when FMS data is inserted.
With managed speed engaged while in SRS, rotate or pull the speed knob on the FCU. Verify the displayed value on the FCU SPD window is the current aircraft speed.
Outside SRS and with managed speed engaged, rotate or pull the speed knob on the FCU. Verify the displayed value is the managed speed target. This will be the managed speed bug displayed on the PFD, or, the short term managed speed if applicable.
Verify the previous behavior but with "MACH" selected on the FCU instead. 


fbw-a380x/src/systems/instruments/src/FCU/Managers/SpeedManager.ts
```diff
@@ -1,17 +1,30 @@
...
 
+  private readonly sub = this.bus.getSubscriber<FGVars>();
+
+  private readonly managedSpeed = ConsumerSubject.create(this.sub.on('fg.speeds.managed'), 0);
+
+  private readonly validManagedSpeed = this.managedSpeed.map((v) => v >= 90);
+
+  private readonly managedSpeedTarget = ConsumerSubject.create(this.sub.on('fg.speeds.managedTarget'), 0);
+
+  private readonly validManagedSpeedTarget = this.managedSpeedTarget.map((v) => v > 90);
```
tracernz, line 26: Do we really need all these subscribables that only subscribe to each other and calculate millions of results that are never used? We could just directly compute from simvars only when actually needed in getInitialSelectedSpeedValue, much like before, which is as few as zero times per flight.

CaptCrystal:
I can test this today.

PR #9976, opened 2025-04-03
fix(a32nx): error reporting popup being stuck

Saschl:
Fixes #[issue_no]

## Summary of Changes

This did not appear in QA as the sentry DSN is not set in QA builds..

MSFS2024 "ingame" status of the GameStateProvider appears to be starting when in walkaround mode. Unfortunately the popup is bugged there and cannot be clicked. In the briefing status it works fine, so I used that for now.

I did not change the MSFS2020 code (it also worked fine with the briefing state but I couldn't test enough).

## Additional context

### DIFF
--- a/fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
+++ b/fbw-common/src/systems/shared/src/FbwAircraftSentryClient.ts
@@ -10,8 +10,7 @@ import { CaptureConsole as CaptureConsoleIntegration } from '@sentry/integration
 import { Integration } from '@sentry/types';
 import { NXDataStore } from './persistence';
 import { PopUpDialog } from './popup';
-import { GameStateProvider, Wait } from '@microsoft/msfs-sdk';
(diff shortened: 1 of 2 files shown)


tracernz:
Weird, for me in my testing it popped up before launching, and worked fine in the loading screen. Maybe it's broken when you go in-game without dismissing it? We could also just pop it up only when you're in the cockpit as we already track that state.

Saschl:
> Weird, for me in my testing it popped up before launching, and worked fine in the loading screen. Maybe it's broken when you go in-game without dismissing it? We could also just pop it up only when you're in the cockpit as we already track that state.

Hm yeah maybe the event is a bit unreliable in 2024? Checking if in cockpit is a good idea, will try to do that later 👍

tracernz (approved):

Saschl:
Popup now just appears when entering the cockpit

https://github.com/user-attachments/assets/fe9bb2ec-1535-419e-8353-80cee93b6638

PR #9998, opened 2025-04-10
fix(efb): split jetway and passenger stairs into separate buttons

### LINKED ISSUE #9812: Jet Bridge Unable To Connect To Aircraft
### Simulator

2024

### Aircraft Version

Development (4k)

### Build info

```json
{
    "built": "2025-01-12T02:20:13+00:00",
    "ref": "refs/heads/master",
    "sha": "4e9c1f87cb859b6f5c1879bda871091fa24bb81e",
    "actor": "Benjozork",
    "event_name": "manual",
    "pretty_release_name": "master:4e9c1f87",
    "version": "a380x-v0.13.0-dev.4e9c1f8"
}
```

### Describe the bug

- When using the jet bridge connect feature in the EFB, the jet bridge is unable to connect to the door successfully
- The door will open and the jet bridge will attempt to connect however will quickly stop and will not connect
- The jet bridge does not get close to the aircraft at all and only moves a very small amount before becoming stationary

### Expected behavior

The jet bridge successfully connects to the aircraft in a realistic manner

### Steps to reproduce

1. Spawn at the gate at either Cairo or Dubai airport that has a jetway
2. Attempt to connect the jet bridge at either airport via EFB
3. Observe the door opening and the jet bridge momentarily moving before stopping

### References (optional)

_No response_

Both airports were the premium deluxe version

_No response_

heclak:
Fixes #9812 

## Summary of Changes

- split jet bridge and passenger stair truck request to separate buttons on EFB
- split buttons fixes jetway not connecting in 2024

There's an existing issue with 2024 where it doesn't handle simultaneous K events. So the way to get the jet bridge connections working again is to send only the request jetway or request ramp truck event. The only implementation sends both which works in 2020.

It was considered to only send the ramp truck event, which would currently connect the jetway too. But I believe it is a bug as the behaviour differs from 2020. Unless there is a way to detect if a jetway is available accurately and to send the appropriate command.

## Screenshots (if necessary)

image
image

## Additional context

## Testing instructions

Tests required in both 2020 and 2024. Total of 4 tests needed.

1. A380 in 2024 ✅ 
2. A380 in 2020
3. A320 in 2024 ✅
4. A320 in 2020

### MSFS2020
- EFB should be unchanged. Jetway button functions like before, triggering either the jetway or passenger stairs based on your current parking position. The new passenger stairs button shouldn't appear.

### MSFS2024
Load onto a stand with a jetway
- Test connecting and disconnecting the jetway using the jetway button
- Note: the passenger stairs button might trigger the jetway too, depending on the specific build of MSFS2024 you're on. It is a known bug in the sim and the reason for this PR. It is expected to function inconsistently until the base sim issue is fixed.

Load on a ramp position without a jetway
- Test calling and sending the passenger stairs with the passenger stairs button. 
- Depending on your sim version, this base sim function could be inconsistently broken (once again is the reason for this PR and why jetways were broken). In SU2 and SU3 up to 1.5.9, the ramp truck should work. However, how well the ramp truck connects is not up to us, the base sim has a number of bugs related to this.


2hwk:
Version and make the change only for the 2024 (compatible) version, and leave 2020 as the old, single button which is working there. Else we'd have to also update our documentation and do more testing (beyond regression) on the 2020 side which for a compatibility hot-fix patch, would take too long.

tracernz:
I think it's a bad idea to have different buttons if we can avoid it. It will just create problems and confusion.

Saschl:
If the issue is that the sim has issues when both events are sent in the same frame, we could maybe add a delay with setTimeout between the two calls for example? That should also work with both sims I guess.

heclak:
> If the issue is that the sim has issues when both events are sent in the same frame, we could maybe add a delay with setTimeout between the two calls for example? That should also work with both sims I guess.

The problem is both events would move the jetway in 2024. The delay would call for the jetway and call again to send it back immediately. Breaking it again.

An unfavourable alternative would be to use the bugged ramp truck event that calls for both in 2024 until the 2024 issues are fixed then we revert the fix.

heclak:
> Version and make the change only for the 2024 (compatible) version, and leave 2020 as the old, single button which is working there. Else we'd have to also update our documentation and do more testing (beyond regression) on the 2020 side which for a compatibility hot-fix patch, would take too long.

I'll look into the versioning for 2024. It makes sense to try and keep the existing implementation for 2020.

heclak:
Code is updated to only show the new passenger stairs button in MSFS2024. EFB appears the same in MSFS2020.

tracernz (approved):

Saschl:

Discord Username  : saschl 
Object of testing : #9998
Aircraft          : A380X/A320 
Simulator         : 2024 SU3 Beta
Tier of Testing   : 1
Date              : 20/07/2025 

Testing Process: 
 

Load onto a stand with a jetway

 - Test connecting and disconnecting the jetway using the jetway button. 
    - Jetway connects properly 

- Test calling and sending the passenger stairs with the passenger stairs button.
  - Stairs appear correctly, connection is a bit random. But this is very likely due to a sim issue and not directly related to this PR.
- Tested rest of the ground services, no change could be observed.

Testing Results: 
Passed  

Negatives:       


Conclusions: 


Media:


Screenshot 2025-07-22 225303
Screenshot 2025-07-22 225246
Screenshot 2025-07-22 225222
image

heclak:
@Saschl thanks for the test. Can I confirm you only tested the A380 and only on 2024 SU3?

heclak:
updated test instructions with additional specific tests.

Saschl:
> @Saschl thanks for the test. Can I confirm you only tested the A380 and only on 2024 SU3?

Yep. Can do the A320 with 2024 SU3 in a bit, unfortunately I do not have 2020 installed anymore

Saschl:
@heclak A320 also looks good, updated the test.

heclak:
> @heclak A320 also looks good, updated the test.

@Saschl thanks!

Still looking for a 2020 set of tests from someone else. If no one else has 2020 installed anymore, I can do it later.

heclak:
Quality Assurance Developer Report

Discord Username  : Heclak 
Object of testing : #9998
Simulator         : 2020 
Tier of Testing   : 1
Date              : 31/07/2025 

Testing Process: 
 

Tested both aircraft in 2020 by checking that the EFB looked the same as the current public builds. No change expected in 2020. Functionality for 2020 is the same. 

Testing Results: 
Passed  

Media:

FlightSimulator_2025-07-31-21-25-25-1469
FlightSimulator_2025-07-31-21-25-36-1470
FlightSimulator_2025-07-31-21-27-42-1471
FlightSimulator_2025-07-31-21-28-08-1472
FlightSimulator_2025-07-31-21-31-45-1473

PR #10006, opened 2025-04-13
feat(a380x/ecam): add FLAPS LEVER NOT ZERO ecam message

Jonny23787:
Fixes #[issue_no]

## Summary of Changes

- Adds FLAPS LEVER NOT ZERO ECAM message when flap lever is not zero at or above 22000ft.

## Screenshots (if necessary)
Screenshot 2025-04-13 194825

## References
image

## Additional context
The second triggering condition has not been added in this PR.

jonny_23

## Testing instructions

1. Take-off and climb with flaps 0 to at or above 22000ft.
2. Slow aircraft down to around 200kt to ensure an overspeed won't occur when flaps deployed.
3. Set flaps lever to position 1, confirm FLAPS LEVER NOT ZERO ecam message appears.
4. Set other flap positions at appropriate speeds while remaining above 22000ft.
5. Descend aircraft below 22000ft with warning present, ensure it disappears below 22000ft.

### DIFF
--- a/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
+++ b/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
@@ -2476,6 +2476,18 @@ export class FwsAbnormalSensed {
       sysPage: SdPages.Fctl,
       inopSysAllPhases: () => [],
     },
+    272800003: {
+      // FLAPS LEVER NOT ZERO
+      flightPhaseInhib: [1, 2, 3, 4, 5, 6, 7, 9, 10, 11, 12],
+      auralWarning: this.fws.flapConfigAural.map((on) => (on ? FwcAuralWarning.Crc : FwcAuralWarning.None)),
+      simVarIsActive: this.fws.flapsLeverNotZeroWarning,
+      notActiveWhenFaults: [],
+      whichItemsToShow: () => [],
+      whichItemsChecked: () => [],
+      failure: 3,
+      sysPage: SdPages.Fctl,
+      inopSysAllPhases: () => [],
+    },
     272800028: {
       // TO FLAPS FMS DISAGREE
       flightPhaseInhib: [1, 4, 5, 6, 7, 8, 9, 10, 12],
--- a/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
+++ b/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
@@ -2479,7 +2479,7 @@ export class FwsAbnormalSensed {
     272800003: {
       // FLAPS LEVER NOT ZERO
       flightPhaseInhib: [1, 2, 3, 4, 5, 6, 7, 9, 10, 11, 12],
-      auralWarning: this.fws.flapConfigAural.map((on) => (on ? FwcAuralWarning.Crc : FwcAuralWarning.None)),
+      auralWarning: this.fws.flapsLeverNotZeroAural.map((on) => (on ? FwcAuralWarning.Crc : FwcAuralWarning.None)),
       simVarIsActive: this.fws.flapsLeverNotZeroWarning,
       notActiveWhenFaults: [],
       whichItemsToShow: () => [],
--- a/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsCore.ts
+++ b/fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsCore.ts
@@ -853,6 +853,8 @@ export class FwsCore {
 
   public readonly flapsLeverNotZeroWarning = Subject.create(false);
 
+  public readonly flapsLeverNotZeroAural = Subject.create(false);
+
   public readonly speedBrakeCommand5sConfirm = new NXLogicConfirmNode(5, true);
 
   public readonly speedBrakeCommand50sConfirm = new NXLogicConfirmNode(50, true);
@@ -3641,6 +3643,7 @@ export class FwsCore {
         this.flightPhase.get() === 8 &&
         !this.slatFlapSelectionS0F0,
     );
+    this.flapsLeverNotZeroAural.set(this.flapsLeverNotZeroWarning.get());
 
     // spd brk still out
     this.speedBrakeCommand5sConfirm.write(this.speedBrakeCommand.get(), deltaTime);
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 1. [A380X/FMS] Fix speed margins being displayed in the wrong place for a Mach target - @BlueberryKing (BlueberryKing)
 1. [A380X/MFD] Visual update to reflect later avionics batches (overlapping tab selectors) - @flogross89 (floridude)
+1. [A380X/ECAM] Add FLAPS LEVER NOT ZERO ECAM message - @Jonny23787 (Jonathan)
 
 ## 0.13.0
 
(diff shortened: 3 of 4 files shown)


fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
```diff
@@ -2476,6 +2476,18 @@ export class FwsAbnormalSensed {
       sysPage: SdPages.Fctl,
       inopSysAllPhases: () => [],
     },
+    272800003: {
+      // FLAPS LEVER NOT ZERO
+      flightPhaseInhib: [1, 2, 3, 4, 5, 6, 7, 9, 10, 11, 12],
+      auralWarning: this.fws.flapsLeverNotZeroAural.map((on) => (on ? FwcAuralWarning.Crc : FwcAuralWarning.None)),
```
BravoMike99, line 2482: The aural warning is not required. It is only required when the audio is to be driven by a different condition other than the activation one.

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsCore.ts
```diff
@@ -853,6 +853,8 @@ export class FwsCore {
 
   public readonly flapsLeverNotZeroWarning = Subject.create(false);
 
+  public readonly flapsLeverNotZeroAural = Subject.create(false);
```
BravoMike99, line 856: Not needed as per the comment above.

BravoMike99 (changes requested):
Some minor comments. In addition to that, is the VFEConf1 + 2.5kt logic implemented?

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
```diff
@@ -2476,6 +2476,18 @@ export class FwsAbnormalSensed {
...
+      simVarIsActive: this.fws.flapsLeverNotZeroWarning,
+      notActiveWhenFaults: [],
+      whichItemsToShow: () => [],
+      whichItemsChecked: () => [],
+      failure: 3,
+      sysPage: SdPages.Fctl,
```
BravoMike99, line 2488: No ECAM page should appear based on the ref you just posted.

Jonny23787:
> Some minor comments. In addition to that, is the VFEConf1 + 2.5kt logic implemented?

No, It was not included in the original logic that was already in place, I can look into implementing though.

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
Jonny23787: Do I completely remove line 2482 or just leave the end part with the .Crc?

BravoMike99: Just remove completely. This abnormal sensed item does not have a separate aural warning logic

Jonny23787: When removing completely, the master warning light and crc does not extinguish when the item is no longer present, for example when descending below 22000ft, the ecam message will disappear but not the master warning or crc.

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
Jonny23787: should it be replaced with `-1`?

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
BravoMike99: That should not be happening unless its a new bug 🤔 We have other L3 warnings without the aural warning bit such as overspeed, L/G not down, CONFIG PRK BRK ON, excess cabin altitude etc. Can you check if the same also happens on those?

Jonny23787: I can confirm that the L/G not down and overspeed warnings don't have this issue, I have changed the name of the simvar to match up with those ones as they don't include "warning" on the end. The warnings still won't cancel after the condition is no longer met.

flogross89: What happens if you add `cancel: true` to this element? Maybe it's somehow regarded as a non-cancellable warning

Jonny23787: Do you mean add `cancel: true` after `auralWarning:`?

flogross89: Yes, somewhere in this object. Shouldn't be necessary since it should default to true, but maybe there's a bug somewhere

BravoMike99: > I can confirm that the L/G not down and overspeed warnings don't have this issue, I have changed the name of the simvar to match up with those ones as they don't include "warning" on the end. The warnings still won't cancel after the condition is no longer met.

I just tested this locally (2020) and the CRC & master warn stop when they should 🤔 Are you able to consistently reproduce this and record a video for example?

Jonny23787: I can also confirm that it is fixed, works as expected

Jonny23787:
Looking in the FCOM, the VFEConf1 + 2.5kt would be 263 + 2.5 = 265.5kt? The VFE seems to be a fixed value for each conf

fbw-a380x/src/systems/systems-host/systems/FlightWarningSystem/FwsAbnormalSensed.ts
flogross89: Does it work without adding `cancel: true`, or did you have to add it?

Jonny23787: I didn't have to change anything, the only thing I did was change the name by removing "warning" off the end of the simvar to match up with the other naming formats

Jonny23787:
Both triggering conditions are now implemented

BravoMike99:
> Both triggering conditions are now implemented

Upon further research, the VFE 1+2.5kt is handled by flaps baulk function being active and not by the FWS unless such feature is failed which is not simulated at the moment. As such the final condition is something like
`phase 8 and ( (altitude > 22000 and flap lever > 0) or (flap lever > 0 and flaps baulk active) )`
In order to check if the baulk is active you can use bit 25 of the SFCC status system word (simvar `A32NX_SFCC_SLAT_FLAP_SYSTEM_STATUS_WORD`). Take a look at any of the arinc429registers in the fws core file as examples on how to read such words.
``

BravoMike99 (approved):

BravoMike99:
Dev team QA report

Object of testing : #10006 
Aircraft          : A380X 
Tier of Testing   : 1
Date              :  25/04/2025

Testing Process: 
With the  flap lever not zero above 22000 feet, the FLAPS LEVER NOT ZERO ecam message appears.
Once below 22000feet, the message disappears and is not trigerred if flaps are not zero.
If flaps lever is set to 1 above VFE1 2.5kt, the FLAPS LEVER NOT ZERO message is triggered regardless of aircraft altitude. Selecting the flaps to other position other than 1 above VFE 2.5kt correctly causes the message to disappear and overspeed appears instead.
 
Testing Results: 
Passed

Media:
F1
F2
Baulk

PR #10049, opened 2025-04-30
feat(a380x): Improvement to direct and alternate law conditions

flogross89:
## Summary of Changes

Modifies PRIM & FWS

- Add third RA to PRIM
- Modify conditions for pitch and lateral law capabilities

TODO: ABN sensed procedures for ALTN and DIRECT LAW

## Additional context

## Testing instructions

TODO

### DIFF
--- a/fbw-a380x/src/wasm/fbw_a380/src/FlyByWireInterface.cpp
+++ b/fbw-a380x/src/wasm/fbw_a380/src/FlyByWireInterface.cpp
@@ -1272,8 +1272,6 @@ bool FlyByWireInterface::updatePrim(double sampleTime, int primIndex) {
   double thsPosition;
   double rudder1Position;
   double rudder2Position;
-  base_ra_bus ra1Bus;
-  base_ra_bus ra2Bus;
   if (primIndex == 0) {
     leftAileron1Position = idLeftAileronInwardPosition->get();
     rightAileron1Position = idRightAileronInwardPosition->get();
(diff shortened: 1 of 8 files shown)


fbw-a380x/src/wasm/fbw_a380/src/FlyByWireInterface.cpp
```diff
@@ -1416,8 +1405,9 @@ bool FlyByWireInterface::updatePrim(double sampleTime, int primIndex) {
   prims[primIndex].modelInputs.in.bus_inputs.rate_gyro_roll_2_bus = {};
   prims[primIndex].modelInputs.in.bus_inputs.rate_gyro_yaw_1_bus = {};
   prims[primIndex].modelInputs.in.bus_inputs.rate_gyro_yaw_2_bus = {};
-  prims[primIndex].modelInputs.in.bus_inputs.ra_1_bus = ra1Bus;
```
lukecologne, line 1419: The fact that the PRIMs only get 2 RAs was intentional, this is how it works IRL. For the third RA value, the inter-FCGU bus would need to be used to share the remaining RA value from the two other PRIMs. I chose not to implement this at the time because the PRIM was already a huge amount of work and I wanted to finish it

lukecologne (changes requested):

fbw-a380x/src/wasm/fbw_a380/src/FlyByWireInterface.cpp
flogross89: Changed now, taking the third value from the other PRIMs

lukecologne (approved):

PR #8841, opened 2024-08-25
fix(fms): remove forced turn on following leg after leg insertion

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes a few issues with flight plan modifications:
1. After deleting a discontinuity, it could happen that the turn of the new leg went the wrong way due to the leg constellation. This is fixes by creating a TF leg from the leg after the discontinuity.
2. After truncating the flight plan by insertion of an existing waypoint or using DIR to, the outbound turn would sometimes go the wrong way because the following leg was coded with a forced turn. This is fixed by removing the forced turn on the following leg.

## Screenshots (if necessary)

Before: 
image
After:
image

Before:
image
After:
image

## Additional context

Discord username (if different from GitHub):

## Testing instructions

- Fly a procedure with a forced turn, try going DIR to the leg with the forced turn (indicated by a left/right arrow in the FPLN page). Make sure the turn goes the wrong way.
- Clear discontinuities on procedures. Make sure all turns go the right way.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo** or **flybywire-aircraft-a380-842** download link at the bottom of the page

PR #10050, opened 2025-04-30
fix(fms): do not remove forced turn on legs that require them

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes a bug that was introduced in #8841, where the coded turn direction was removed on a leg after performing a DIR to. This should not happen for legs that require a coded turn direction, such as hold legs (Hx).

## References

Image of the bug
image

## Additional context

## Testing instructions

Enter a hold in the FMS, perform a DIR to the hold fix, ensure the hold is shown and flown correctly.

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
@@ -23,7 +23,12 @@ import {
   WaypointDescriptor,
 } from '@flybywiresim/fbw-sdk';
 import { OriginSegment } from '@fmgc/flightplanning/segments/OriginSegment';
-import { FlightPlanElement, FlightPlanLeg, FlightPlanLegFlags } from '@fmgc/flightplanning/legs/FlightPlanLeg';
+import {
+  FlightPlanElement,
+  FlightPlanLeg,
+  FlightPlanLegFlags,
+  isDiscontinuity,
+} from '@fmgc/flightplanning/legs/FlightPlanLeg';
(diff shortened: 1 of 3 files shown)


Benjozork (approved):

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -2109,31 +2110,40 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
...
 
-          // Convert TF to CF
-          if (element.type === LegType.TF) {
-            const prevElement = second.allLegs[i - 1];
-            if (!prevElement || prevElement.isDiscontinuity === true) {
-              throw new Error('[FMS/FPM] TF leg without a preceding leg');
-            } else if (!prevElement.terminationWaypoint()) {
-              throw new Error('[FMS/FPM] TF leg without a preceding leg with a termination waypoint');
+            // Convert TF to CF
+            if (element.type === LegType.TF) {
```
tracernz, line 2123: If the leg before is a PI this should *never* be hit, as PI-CF must always and can only ever exist as a combo from an approach transition.

BlueberryKing: Yes, this `if` applies when the leg before is not a PI, for example, when the approach transition ends in a CI

tracernz: It applies if there's a PI, then a CF with the same fix (the or condition)?

tracernz (approved):

PR #10074, opened 2025-05-18
feat(a32nx/fms): add via auto-selection

tracernz:
Fixes #[issue_no]

## Summary of Changes

Adds the approach VIA auto-selection functionality.

## Screenshots (if necessary)

https://github.com/user-attachments/assets/148d7953-be81-4006-a8ce-393b87b43bc9

## References

pylot

## Additional context

## Testing instructions

Selecte some STAR/approach combos that join up with an approach transition. Ensure the via/transition is automatically selected when there is only one available. Insert into the legs are strung correctly.
Select some STAR/approach combos that don't join up with any transition, but with some transitions. Ensure no via/transition is selected, but the VIA page is shown with all the vias under "OTHER VIAS". Insert into the legs are strung correctly.


BlueberryKing (approved):
Nice!

PR #10113, opened 2025-06-05
fix(a32nx/du): Display rework

alexr4339:
Fixes #[issue_no]

## Summary of Changes
Changed the color definitions of all DU's which now more closely match the A380's new colors and match reference as well. Its worth noting that the colors do look different in every reference due to different brightness settings, environment, camera. The proposed new colors should be the best combination of all observable scenarios. Adjusted the FCU (positioning, layouting) to better match real world references.
A380 is untouched

## Screenshots (if necessary)
Screenshot 2025-06-05 210640
Screenshot 2025-06-05 210809

## Additional context

## Testing instructions
Make sure all displays work as expected and look normally

### DIFF
--- a/fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
@@ -103,7 +103,7 @@ export class EisDisplay extends DisplayComponent<EisDisplayProps> {
         <text id="QNH" class={this.qnhLabelSub} x="300" y="61.5" text-anchor="end">
           QNH
         </text>
-        <text id="Value" class="Value" x="28" y="169" letter-spacing="-10px">
+        <text id="Value" class="Value" x="28" y="168" letter-spacing="-12px" font-size="115px">
           {this.baroValueSub}
         </text>
       </g>
--- a/fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
@@ -103,7 +103,7 @@ export class EisDisplay extends DisplayComponent<EisDisplayProps> {
         <text id="QNH" class={this.qnhLabelSub} x="300" y="61.5" text-anchor="end">
           QNH
         </text>
-        <text id="Value" class="Value" x="28" y="165">
+        <text id="Value" class="Value" x="28" y="169" letter-spacing="-10px">
           {this.baroValueSub}
         </text>
       </g>
--- a/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
+++ b/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
@@ -51,10 +51,10 @@ text.Inactive {
 
 text.Value {
   font-family: Digital;
-  font-size: 118px;
+  font-size: 117px;
   text-anchor: start;
   fill: var(--main-display-colour);
-  letter-spacing: -10px;
+  letter-spacing: -12px;
 }
 
 line {
--- a/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
+++ b/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
@@ -51,11 +51,10 @@ text.Inactive {
 
 text.Value {
   font-family: Digital;
-  font-size: 115px;
+  font-size: 118px;
   text-anchor: start;
   fill: var(--main-display-colour);
-  top: 10px;
-  letter-spacing: -2px;
+  letter-spacing: -10px;
 }
 
 line {
--- a/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
+++ b/fbw-a32nx/src/systems/instruments/src/FCU/style.scss
@@ -13,10 +13,10 @@
 }
 
 :root {
-  --main-display-colour: hsl(36, 100%, 45%);
-  --main-display-colour-inactive: hsl(53, 82%, 0%);
+  --main-display-colour: #b4e205;
+  --main-display-colour-inactive: #030204;
   --main-background-opacity: 0;
-  --main-text-shadow-colour: rgba(207, 110, 0, 0);
+  --main-text-shadow-colour: #b4e20500;
 
   // text-shadow places the shadow in the wrong position, seems to be a coherent bug,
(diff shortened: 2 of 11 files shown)


fbw-a32nx/src/systems/instruments/src/FCU/style.scss
```diff
@@ -13,10 +13,10 @@
 }
 
 :root {
-  --main-display-colour: hsl(36, 100%, 45%);
-  --main-display-colour-inactive: hsl(53, 82%, 0%);
+  --main-display-colour: #b4e205;
+  --main-display-colour-inactive: #030204;
```
tracernz: Changing the colour is okay but please keep hsl format as it's easier to gauge the brightness.

fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
```diff
@@ -103,7 +103,7 @@ export class EisDisplay extends DisplayComponent<EisDisplayProps> {
         <text id="QNH" class={this.qnhLabelSub} x="300" y="61.5" text-anchor="end">
           QNH
         </text>
-        <text id="Value" class="Value" x="28" y="165">
+        <text id="Value" class="Value" x="28" y="168" letter-spacing="-12px" font-size="115px">
```
tracernz, line 106: Why would the letter spacing be different in only one place?

fbw-a32nx/src/systems/instruments/src/MCDU/mcdu.scss
```diff
@@ -176,7 +176,7 @@ span .s-text {
 
 .b-text {
   font-family: HoneywellMCDU !important;
-  letter-spacing: normal;
+  letter-spacing: -6px;
```
tracernz: Won't this break the layout since it only applies to large text?

fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
alexr4339: According to my expermentation it seemed as if the Baro was not affected by letter spacing in the style definition, that is why I had to change it in that place. All other values seemed affected by changing letter spacing in the style definition

fbw-a32nx/src/systems/instruments/src/MCDU/mcdu.scss
alexr4339: That is a leftover as I did not want to start a MCDU layouting rework, will remove that change. Question from my side is whether reworking the MCDU layouting even makes sense presently, as I read it will be reworked from the ground up (V2)?

fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
tracernz: Ah, okay. I am curious why though, because the font was made by us to match the real thing already.

fbw-a32nx/src/systems/instruments/src/FCU/style.scss
```diff
@@ -13,10 +13,10 @@
 }
 
 :root {
-  --main-display-colour: hsl(36, 100%, 45%);
-  --main-display-colour-inactive: hsl(53, 82%, 0%);
+  --main-display-colour: hsl(72, 96%, 45%);
+  --main-display-colour-inactive: hsl(270, 33%, 1%);
   --main-background-opacity: 0;
-  --main-text-shadow-colour: rgba(207, 110, 0, 0);
+  --main-text-shadow-colour: #b4e20500;
```
tracernz, line 19: Did you see a difference with these? The following are overridden in code to achieve the ambient +backlight lighting effects:
```
--main-display-colour
--main-text-shadow-colour
--main-background-opacity
```
I would not expect changing anything here to do anything.

tracernz:
I don't think there should be any lockfile changes? The package.json is unchanged..

fbw-a32nx/src/systems/instruments/src/FCU/style.scss
alexr4339: Havent seen any effect either, so it could be reset to the previous values as well

alexr4339:
> I don't think there should be any lockfile changes? The package.json is unchanged..

I have not manually touched the lockfile, dont know where the change comes from, will resolve that. Apparently the lockfile version has somehow changed to 9.0 in the changed file. Present master says 6.0. What would you suggest, copying over the lockfile from master or deleting this one entirely and then uploading the lockfile from master again?

tracernz:
Copy the lockfile from master and commit it, it should then go away from the overall diff for the PR. 👍

fbw-a32nx/src/systems/instruments/src/FCU/style.scss
tracernz: If you wanted to play with it, it's in https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/src/systems/instruments/src/FCU/instrument.tsx#L56. The idea is that with the backlight off during the day, we should just see the grey LCD segments illuminated with ambient light and with 0 amber colour, and at night with the backlight off there will be nothing visible, then the backlight effect on the saturation and brightness of the segments is balanced against the ambient light to get a reasonable look across all times of day. The background colour is also affected by the combination of ambient light and backlight illumination. To do that we collected photographs of various conditions and I tried to adjust it being mindful of camera effects, as photographs are never able to convey the true colour and contrast.

fbw-a32nx/src/systems/instruments/src/MCDU/mcdu.scss
tracernz: Probably not. If you were trying to get the gap down the sides of the text to match photos, this is unfortunately the wrong fix. The problem is the aspect ratio of the display in the 3D model.

fbw-a32nx/src/systems/instruments/src/FCU/Components/EisDisplay.tsx
alexr4339: The font itself is perfect, however I noticed when comparing to irl refs, that the spacing between the letters is a bit too large, thats what I wanted to adjust

tracernz:
Can you please update the title to `fix(a32nx): display colour rework'. It's a bit ambitious at the moment 😁 . It's also more than just the 6 DUs.

PR #10142, opened 2025-06-14
feat(hyd): split slats and flaps in individual surfaces

Eagle941:
## Summary of Changes

This change splits each secondary flight control (flaps/slats) into individual surfaces. Additionally, simulation of the two synchro units `FPPU` and `IPPU` is added for complete representation of the slats/flaps system interface with avionics.

Renamed trait `FeedbackPositionPickoffUnit` to `PositionPickoffUnit` because there are two types of syhcro on each secondary surface: the Feedback Position Pickoff Unit (FPPU) and the Instrumento Position Pickoff Unit (IPPU). The new trait name is more generic and less misleading.

## Screenshots (if necessary)

Not necessary

## References

Flaps/Slats system

## Additional context

The A380 has 8 slats and 3 flaps for each wing. The A320 has 5 slats and 2 flaps for each wing.

## Testing instructions

With plane ready for takeoff, move the flaps lever through all the positions and verify that the flaps/slats move accordingly on the external model and the EWD.


fbw-a32nx/docs/a320-simvars.md
```diff
@@ -1117,6 +1117,26 @@
...
 
+- A32NX_{side}_FLAPS_{number}_POSITION_PERCENT
+    - Percent
+    - Indicates the angle of the flaps out of 40 degrees
+    - Side
+        - LEFT
+        - RIGHT
+    - Number
+        - From 1 (inboard) to 2 for A320
+        - From 1 (inboard) to 3 for A380
```
BravoMike99: Since we also have an a380-simvars.md file, perhaps the descriptions containing the A380 index could also go there? That way it would be documented on both sides with the correct range on each.

Eagle941:
@BravoMike99 I have updated the a380 simvars. Also I have updated the CHANGELOG which I forgot.

BravoMike99:
> @BravoMike99 I have updated the a380 simvars. Also I have updated the CHANGELOG which I forgot.

I think you could split the available index per aircraft type, for example only mention the 1 from 5 on the 320simvars and the 1 to 8 on the A380 for the flaps and then do similar for the slats.

Eagle941:
> > @BravoMike99 I have updated the a380 simvars. Also I have updated the CHANGELOG which I forgot.
> 
> I think you could split the available index per aircraft type, for example only mention the 1 from 5 on the 320simvars and the 1 to 8 on the A380 for the flaps and then do similar for the slats.

Sure, done

Eagle941:
@flogross89 it's both A380 and A320

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
...
+    surface_positions: Vec<Ratio>,
+    surface_angles: Vec<Angle>,
+}
+impl SecondarySurface {
+    // side is LEFT|RIGHT
+    // surface is SLATS|FLAPS
+    pub fn new(
+        context: &mut InitContext,
+        side: &str,
+        surface: &str,
```
Gurgel100: I like using `enum` for such things to not allow other values. You could introduce 2 enums: `SecondarySurfaceSide` and `SecondarySurfacePosition` for example.

Or alternatevily we can have separate constructors. For example `new_slat(...)` and `new_flap(...)`.

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
...
+            surface_angle_ids.push(surface_angle_id);
+
+            surface_positions.push(Ratio::default());
+            surface_angles.push(Angle::default());
+        }
+
+        let surface_position_overall_id =
+            context.get_identifier(format!("{}_{}_POSITION_PERCENT", side, surface));
+        let surface_angle_overall_id =
+            context.get_identifier(format!("{}_{}_ANGLE", side, surface));
```
Gurgel100:
```suggestion
        let surface_position_ids = (1..number_of_surfaces + 1)
            .map(|p| context.get_identifier(format!("{side}_{surface}_{p}_POSITION_PERCENT")))
            .collect();
        let surface_angle_ids = (1..number_of_surfaces + 1)
            .map(|p| context.get_identifier(format!("{side}_{surface}_{p}_ANGLE")))
            .collect();
        let surface_positions = vec![Ratio::default(); number_of_surfaces];
        let surface_angles = vec![Angle::default(); number_of_surfaces];

        let surface_position_overall_id =
            context.get_identifier(format!("{side}_{surface}_POSITION_PERCENT"));
        let surface_angle_overall_id = context.get_identifier(format!("{side}_{surface}_ANGLE"));
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
...
+        }
+    }
+
+    // No failures are considered. All surfaces have the same deflection.
+    fn update(&mut self, position: &Ratio, angle: &Angle) {
+        for idx in 0..self.surface_positions.len() {
+            self.surface_positions[idx] = *position;
+            self.surface_angles[idx] = *angle;
+        }
+    }
```
Gurgel100:
```suggestion
    fn update(&mut self, position: Ratio, angle: Angle) {
        for surface_position in &mut self.surface_positions {
            *surface_position = position;
        }

        for surface_angle in &mut self.surface_angles {
            *surface_angle = angle;
        }
    }
```
## Reasoning
1. uom values are simple `f64` and can easely passed by value.
2. Iterating over the array is less errorprone than using indexes
3. Using the length of array A for the index of array B is normally not a good thing to do (in this case we know that the arrays have the same length but still)

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
...
+}
+impl SimulationElement for SecondarySurface {
+    fn write(&self, writer: &mut SimulatorWriter) {
+        for idx in 0..self.surface_positions.len() {
+            let id = self.surface_position_ids[idx];
+            let position = self.surface_positions[idx];
+            writer.write(&id, position.get::<percent>());
+        }
```
Gurgel100: This is much more idiomatic Rust:
```suggestion
        for (id, position) in self
            .surface_position_ids
            .iter()
            .zip(&self.surface_positions)
        {
            writer.write(id, position.get::<percent>());
        }
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
...
+        let position: Ratio = self.surface_positions.iter().average();
+        writer.write(&self.surface_position_overall_id, position.get::<percent>());
+
+        for idx in 0..self.surface_angles.len() {
+            let id = self.surface_angle_ids[idx];
+            let position = self.surface_angles[idx];
+            writer.write(&id, position.get::<degree>());
+        }
```
Gurgel100:
```suggestion
        for (id, position) in self.surface_angle_ids.iter().zip(&self.surface_angles) {
            writer.write(id, position.get::<degree>());
        }
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -121,17 +123,97 @@ impl Actuator for FlapSlatHydraulicMotor {
     }
 }
 
-pub struct FlapSlatAssembly {
-    position_left_percent_id: VariableIdentifier,
-    position_right_percent_id: VariableIdentifier,
+pub struct SecondarySurface {
```
Gurgel100, line 154: I don't quite get why this is called `SecondarySurface` if it incorporates more than 1 surface? Maybe you find a better name?

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -179,19 +263,16 @@ impl FlapSlatAssembly {
...
-                .get_identifier(format!("RIGHT_{}_POSITION_PERCENT", id)),
-
-            angle_left_id: context.get_identifier(format!("LEFT_{}_ANGLE", id)),
-            angle_right_id: context.get_identifier(format!("RIGHT_{}_ANGLE", id)),
-
             animation_left_id: context.get_identifier(format!("LEFT_{}_ANIMATION_POSITION", id)),
             animation_right_id: context.get_identifier(format!("RIGHT_{}_ANIMATION_POSITION", id)),
 
+            ippu_id: context.get_identifier(format!("{}_IPPU_ANGLE", id)),
+            fppu_id: context.get_identifier(format!("{}_FPPU_ANGLE", id)),
```
Gurgel100:
```suggestion
            ippu_id: context.get_identifier(format!("{id}_IPPU_ANGLE")),
            fppu_id: context.get_identifier(format!("{id}_FPPU_ANGLE")),
```
Yes I know the current code is not doing it this way but I think we still should do it.

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -558,9 +648,11 @@ mod tests {
...
     use std::time::Duration;
     use uom::si::{angle::degree, pressure::psi};
+    use uom::ConstZero;
 
     use crate::shared::update_iterator::MaxStepLoop;
 
+    use crate::simulation::test::ReadByName;
     use crate::simulation::{
         test::{SimulationTestBed, TestBed},
         Aircraft, SimulationElement, SimulationElementVisitor, UpdateContext,
```
Gurgel100:
```suggestion
    use crate::simulation::{
        test::{ReadByName, SimulationTestBed, TestBed},
        Aircraft, SimulationElement, SimulationElementVisitor, UpdateContext,
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -666,472 +758,542 @@ mod tests {
...
+                    ) / interpolation(
+                        &a.flaps_slats.synchro_gear_breakpoints,
+                        &a.flaps_slats.final_surface_angle_carac,
+                        a.flaps_slats.max_synchro_gear_position.get::<degree>(),
+                    ),
+                )
+            });
+
+            let position_percent: f64 = self.read_by_name("LEFT_FLAPS_1_POSITION_PERCENT");
+            assert_eq!(position_percent, expected_position_percent.get::<percent>());
```
Gurgel100: It's probably better to use `assert_about_eq` for floating point comparisons:
```suggestion
            let position_percent: f64 = self.read_by_name("LEFT_FLAPS_1_POSITION_PERCENT");
            assert_about_eq!(position_percent, expected_position_percent.get::<percent>());
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
```diff
@@ -666,472 +758,542 @@ mod tests {
...
+            (current_speed - test_bed.get_max_speed()).abs()
+                <= AngularVelocity::new::<radian_per_second>(0.01)
         );
 
         assert!(
-            test_bed.query(|a| a.flaps_slats.left_motor.speed())
-                >= AngularVelocity::new::<revolution_per_minute>(2000.)
-                && test_bed.query(|a| a.flaps_slats.left_motor.speed())
+            test_bed.left_motor_speed() >= AngularVelocity::new::<revolution_per_minute>(2000.)
+                && test_bed.left_motor_speed()
```
Gurgel100: Can we split this up into 2 asserts? This way in case of failure it's easier to see which condition failed.
(Same for the others below)

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/mod.rs
```diff
@@ -1753,6 +1769,11 @@ impl A380Hydraulic {
             Ratio::new::<ratio>(0.01),
         );
 
+        let left_flaps_factory = A380FlapsFactory::a380_flaps_factory(context, "LEFT");
+        let right_flaps_factory = A380FlapsFactory::a380_flaps_factory(context, "RIGHT");
+        let left_slats_factory = A380SlatsFactory::a380_slats_factory(context, "LEFT");
+        let right_slats_factory = A380SlatsFactory::a380_slats_factory(context, "RIGHT");
```
Gurgel100: The name of the variables is bit misleading here since they actually don't "contain" the factory but the already constructed object.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/mod.rs
```diff
@@ -8123,19 +8148,49 @@ mod tests {
             }
 
             fn get_flaps_left_position_percent(&mut self) -> f64 {
-                self.read_by_name("LEFT_FLAPS_POSITION_PERCENT")
+                let values: Vec<f64> = vec![
```
Gurgel100: I know this is in the tests but I can't hold back myself 😄
Don't need to construct a vector for this (same for the other functions below):
```suggestion
                let values: &[f64] = &[
```

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
Eagle941: I agree. I don't know why many parts of the code use `"LEFT"` and `"RIGHT"`.

fbw-common/src/wasm/systems/systems/src/hydraulic/flap_slat.rs
Eagle941: I wanted to avoid using `FlapsSlatsSurface` and give the struct a more generic name. `SecondarySurface` comes from secondary control surfaces which includes flaps/slats.

Eagle941:
@Gurgel100 Thank you for the review! I agree with all your comments.

Gurgel100 (approved):
LGTM.

becas22:
Any suggestions on how to crosscheck the movement between the model and the EWD? Or just eyeball it?

Eagle941:
> Any suggestions on how to crosscheck the movement between the model and the EWD? Or just eyeball it?

This code change doesn’t touch the model movement of surfaces or EWD indications. Eyeballing it is enough to verify consistency with existing code.

becas22:
Dev Team Quality Assurance Report

Discord Username  : bernardor
Object of testing : #10142
Aircraft          : A32NX+A380X 
Simulator         : 2020 
Tier of Testing   : 1
Date              : 28/06/2025 

Testing Process: 
 
With plane ready for takeoff, move the flaps lever through all the positions and verify that the flaps/slats move accordingly on the external model and the EWD.

Testing Results: 
Passed  
Conclusions: 

Ready to merge if code review completed without further changes

PR #10070, opened 2025-05-15
fix(a32nx/pfd): alt crz fma spacing

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Fixes spacing on ALT CRZ FMA being bigger than it should be (EIS1/widebody style)
## Screenshots (if necessary)

(Left = new, right = old)
image

## References

image 
amongst others 

## Additional context

Discord username (if different from GitHub):
bruno_pt99
## Testing instructions

Fly to your cruise altitude and verify ALT CRZ shows as the picture on the left side.
Engage V/S and FPA modes. Verify V/S and FPA command digits appear as blue on the FMA and that the text fits within the mode change box.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10146, opened 2025-06-15
revert: fix(a32nx/pfd): alt crz fma spacing

BravoMike99:
Reverts flybywiresim/aircraft#10070 due to uncaught bug during second wave of testing the PR.
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,6 @@
 # line endings.
 /fbw-a32nx/src/localization/flypad/*.json text eol=lf
 /fbw-a32nx/src/localization/msfs/*.locPak text eol=lf
+
+# gltfs just make diff messy
+*.gltf binary


tracernz:
If you have a fix in mind already I think it's okay to leave it for now. Also fine to revert too.

tracernz (approved):

BravoMike99:
> If you have a fix in mind already I think it's okay to leave it for now. Also fine to revert too.

Not for now although we already found the culprit and I think it's a somewhat intrusive bug/regression 😅 . Since its affecting some users ill go ahead with the revert unless you have a quick fix in mind.

heclak:
> Not for now although we already found the culprit and I think it's a somewhat intrusive bug/regression 😅 . Since its affecting some users ill go ahead with the revert unless you have a quick fix in mind.

I think it's better to revert first since the fix would still need to wait for QA which would take some time.

PR #10151, opened 2025-06-15
feat(hyd): implemented full CSU simulation

Eagle941:
## Summary of Changes

This change adds a full CSU (Command Sensor Unit) implementation including `Out Of Detent` and `Fault` positions.

## Screenshots (if necessary)

Not necessary

## References

ATA27-50 system

## Additional context

## Testing instructions

With plane ready for takeoff, move the flaps lever through all the positions and verify that the flaps/slats move accordingly on the external model and the EWD.


Eagle941:
@flogross89 Can you add the A320 label as well?

Eagle941:
I have a question: the fault `CSU Misadjust` is triggered when the flaps lever goes from 0 to 4 or viceversa with no intermediate steps. I am worried this messes with the simulation in case of the event `FLAPS_SET` setting flaps to 4. Should I remove it?

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+impl CSUMonitor {
+    pub fn new(context: &mut InitContext) -> Self {
+        Self {
+            handle_position_id: context.get_identifier("FLAPS_HANDLE_INDEX".to_owned()),
+            current_switch_pattern: (0b11000, 0b11000),
+            current_detent: CSU::Conf0,
+            previous_switch_pattern: (0b11000, 0b11000),
+            previous_detent: CSU::Conf0,
+            last_valid_switch_pattern: (0b11000, 0b11000),
```
Gurgel100: Maybe these should be initialized depending on the starting state which you can get with `context.start_state()`.
This would also solve the problem with the `CSU:Misadjust` state.

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+        )
+    }
+
+    // The CSU is made of two parallel switches: this is why there is `switch_pattern_1` and
+    // `switch_pattern_2`.
+    pub fn get_csu_configuration(
+        switch_pattern_1: SwitchPattern,
+        switch_pattern_2: SwitchPattern,
+    ) -> CSU {
+        match (switch_pattern_1, switch_pattern_2) {
```
Gurgel100:
```suggestion
    pub fn get_csu_configuration(
        switch_pattern: (SwitchPattern, SwitchPattern),
    ) -> CSU {
        match switch_pattern {
```
For example then you can just do on line 97:
```rust
CSU::get_csu_configuration(self.current_switch_pattern)
```

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+            match value {
+                0 => (0b11000, 0b11000),
+                1 => (0b01100, 0b01100),
+                2 => (0b00110, 0b00110),
+                3 => (0b00011, 0b00011),
+                4 => (0b10001, 0b10001),
+                _ => (0b00000, 0b00000), // Setting a value that leads to `CSU::Fault` to complete all
+                                         // the alternatives of the match statement.
+            }
+        }
```
Gurgel100: Why not define this function outside of this block?
Another option would be to inline it because it's only used once.

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+    Misadjust, // This is triggered when flaps lever moves from 0 to 4 or viceversa
+               // with no intermediate steps. Not implemented because it can impact the simulation.
+}
+impl CSU {
+    pub fn is_valid(value: CSU) -> bool {
+        matches!(
+            value,
+            CSU::Conf0 | CSU::Conf1 | CSU::Conf2 | CSU::Conf3 | CSU::ConfFull
+        )
+    }
```
Gurgel100:
```suggestion
    fn is_valid(&self) -> bool {
        matches!(
            self,
            CSU::Conf0 | CSU::Conf1 | CSU::Conf2 | CSU::Conf3 | CSU::ConfFull
        )
    }
```
Then you can directly call `.is_valid()` on the type.

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+
+    pub fn update(&mut self, context: &UpdateContext) {
+        self.time_since_last_valid_detent += context.delta();
+    }
+
+    fn get_csu_configuration(&self) -> CSU {
+        CSU::get_csu_configuration(self.current_switch_pattern.0, self.current_switch_pattern.1)
+    }
+
+    pub fn get_last_valid_detent(&self) -> CSU {
```
Gurgel100:
```suggestion
    #[cfg(test)]
    fn get_last_valid_detent(&self) -> CSU {
```
No unnecessary `pub` function.

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+        CSU::get_csu_configuration(
+            self.last_valid_switch_pattern.0,
+            self.last_valid_switch_pattern.1,
+        )
+    }
+
+    pub fn time_since_last_valid_detent(&self) -> Duration {
```
Gurgel100:
```suggestion
    #[cfg(test)]
    fn time_since_last_valid_detent(&self) -> Duration {
```

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::simulation::{
+        test::{SimulationTestBed, TestBed, WriteByName},
+        Aircraft, SimulationElementVisitor,
+    };
+
```
Gurgel100:
```suggestion
```

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,336 @@
...
+        if CSU::is_valid(self.current_detent) {
+            self.last_valid_switch_pattern = new_switch_pattern;
+            self.time_since_last_valid_detent = Duration::ZERO;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Gurgel100, line 154: There are no test cases covering the `CSU::OutOfDetent` case.

Gurgel100 (changes requested):
Great work.
Would be also great if you could add the correct power sources for the sensors and the CSU monitor :). Or at least add a TODO comment.

Eagle941:
> Great work. Would be also great if you could add the correct power sources for the sensors and the CSU monitor :). Or at least add a TODO comment.

The sensor doesn't use any power source because it's just mechanical switches. The CSU monitor could have a power supply but it requires splitting the sfcc first because there are two CSU monitors for each SFCC and each SFCC has a different power source. I add a TODO

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
Eagle941: I created a local function to ensure it's not called elsewhere

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
Eagle941: How would I solve the case of event `FLAPS_DOWN` followed by `FLAPS_UP` or viceversa?

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
Eagle941: Added some tests. However, due to the use of function `u8_to_switch_pattern`, there is no combination which leads to `OutOfDetent` for the time being.

fbw-a380x/src/wasm/systems/a380_systems/src/hydraulic/flaps_computer.rs
```diff
@@ -504,7 +468,7 @@ impl SlatFlapComplex {
         self.sfcc.update(
```
Gurgel100, line 469: I think you need to update the CSU here as well.

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,478 @@
...
+impl CSU {
+    pub fn is_valid(&self) -> bool {
+        matches!(
+            self,
+            CSU::Conf0 | CSU::Conf1 | CSU::Conf2 | CSU::Conf3 | CSU::ConfFull
+        )
+    }
+
+    // The CSU is made of two parallel switches: this is why there is `switch_pattern_1` and
+    // `switch_pattern_2`.
```
Gurgel100: Adjust comment to reflect the new function signature

fbw-common/src/wasm/systems/systems/src/hydraulic/command_sensor_unit.rs
```diff
@@ -0,0 +1,478 @@
...
+    Conf3,
+    ConfFull,
+    OutOfDetent,
+    Fault,
+    Misadjust, // This is triggered when flaps lever moves from 0 to 4 or viceversa
+               // with no intermediate steps. Not implemented because it can impact the simulation
+               // in case of event `FLAPS_DOWN` followed by `FLAPS_UP` or viceversa.
+}
+impl CSU {
+    pub fn is_valid(&self) -> bool {
```
Gurgel100: Does this function need to be public?

Eagle941:
@Gurgel100 Done!

Gurgel100 (approved):
Great work.

becas22:

Discord Username  : bernardor96
Object of testing : #10151
Aircraft          : A32NX+A380X 
Simulator         : 2020
Tier of Testing   : 1
Date              : 15/07/2025 

Testing Process: 
 
With plane ready for takeoff, move the flaps lever through all the positions and verify that the flaps/slats move accordingly on the external model and the EWD.
Testing Results: 
Passed  

Conclusions: 

Ready to merge if code review approved ([6c73a8d](https://github.com/flybywiresim/aircraft/pull/10151/commits/6c73a8dd1dae4c52ad074a07f1017e58ae1dccc0) needs review)

Gurgel100 (approved):

PR #10170, opened 2025-06-21
build: use 4k env var directly

tracernz:
Fixes #[issue_no]

## Summary of Changes

Seems docker doesn't like if the file doesn't exist, so we just use only the one env var directly.

## Additional context


tracernz:
Closing in favour of #10178

PR #10189, opened 2025-06-28
fix(a380x/pfd): Improve visibility logic for tail strike pitch limit

Jonny23787:
Fixes #[issue_no]

## Summary of Changes

- Adds logic for tailstrike indicator to disappear 3 seconds after takeoff and 4 seconds after a go around is initiated. 

## References
FCOM
Screenshot 2025-06-28 172821

## Additional context

jonny_23

## Testing instructions

1. Load aircraft on runway.
2. Set thrust levers to TOGA, confirm tailstrike pitch indicator becomes visible on PFD.
3. After main gears leave the ground during takeoff, confirm indicator is still visible for 3 seconds then becomes hidden.
4. Activate approach phase.
5. Above 400ft RA initiate a go around, ensure indicator is visible for 4 seconds after GA phase is entered, before becoming hidden.
6. Activate approach phase again and prepare aircraft for landing.
7. At 400ft RA confirm indicator becomes visible, land aircraft and confirm it is visible until aircraft is below 50kt GS.


fbw-a380x/src/systems/instruments/src/PFD/AttitudeIndicatorHorizon.tsx
```diff
@@ -340,8 +340,14 @@ class TailstrikeIndicator extends DisplayComponent<{ bus: EventBus }> {
     leftGearCompressed: true,
     rightGearCompressed: true,
     approachPhase: false,
+    goAroundPhase: false,
   };
 
+  private takeoffTimer: number | null = null;
+  private goAroundTimer: number | null = null;
+  private takeoffTimerActive = false;
+  private goAroundTimerActive = false;
```
beheh: These are somewhat redundant. For example, what would it mean if (for any reason) `takeoffTimerActive` is true but takeoff timer is `null`?

Instead, how about using the non-nulliness of takeoffTimer/goAroundTimer respectively as the signal that the timer is active? I'll add an example below for what that could look like.

fbw-a380x/src/systems/instruments/src/PFD/AttitudeIndicatorHorizon.tsx
```diff
@@ -431,7 +486,9 @@ class TailstrikeIndicator extends DisplayComponent<{ bus: EventBus }> {
           this.tailStrikeConditions.rightGearCompressed) ||
         (this.tailStrikeConditions.approachPhase &&
           this.tailStrikeConditions.altitude.value < 400 &&
-          this.tailStrikeConditions.speed > 50)
+          this.tailStrikeConditions.speed > 50) ||
+        this.goAroundTimerActive ||
+        this.takeoffTimerActive
```
beheh: If you get rid of `goAroundTimerActive`/`takeoffTimerActive` and just keep the timers, you could do this as follows:
```suggestion
        this.goAroundTimer !== null ||
        this.takeoffTimer !== null
```

beheh:
Actually now I think about this more, shouldn't this use `realTime` instead of setTimeout? Not sure whether `setTimeout` is an anti-pattern. Maybe someone like @tracernz knows.

tracernz:
You can use DebounceTimer if you want the timeout to not consider simrate, otherwise you’d want to use simTime. 

https://microsoft.github.io/msfs-avionics-mirror/docs/framework/classes/DebounceTimer

.github/CHANGELOG.md
```diff
@@ -132,6 +132,7 @@
 1. [A380X/FG] Fix SRS disengagement conditions for eng 3&4 OEI - @flogross89 (floridude)
 1. [A380X/FMS] Add engine-out indications and behavior to FMS - @flogross89 (floridude)
 1. [A380X/Fuel] Enabled gravityfeeding and fixed crossfeed fuel usage from correct tanks - @Maximilian-Reuter (\_chaoz_)
+1. [A380X/PFD] Improve visiblity of tailstrike pitch limit indicator - @Jonny23787 (Jonathan)
```
tracernz: Needs moved to 0.15.

PR #10200, opened 2025-07-01
Change "packs" default value

### LINKED ISSUE #10199: Default TO Performance calculator to PACKS OFF for A320
### Aircraft Version

Development

### Description

By default, Pack selection is ON.
I would say that the vast majority of airlines takeoff with packs off, so defaulting that to "OFF" would be a nice QoL change

### References (optional)

_No response_

_No response_

_No response_

becas22:
Fixes #10199

## Summary of Changes

Defaults packs selection to "Off" on Takeoff Performance Calculator

## Additional context

## Testing instructions

1. Open EFB Takeoff Performace Calculator
2. Check default value of packs
3. Make sure you can change the value of packs and see if it produces any difference in takeoff performace

### DIFF
--- a/fbw-common/src/systems/instruments/src/EFB/Store/features/performance.ts
+++ b/fbw-common/src/systems/instruments/src/EFB/Store/features/performance.ts
@@ -113,7 +113,7 @@ export const initialState: TPerformanceState = {
     availableRunways: [],
     selectedRunwayIndex: -1,
     antiIce: TakeoffAntiIceSetting.Off,
-    packs: true,
+    packs: false,
     takeoffCg: TakeoffCoGPositions.Standard,
     forceToga: false,
     config: 1,


BravoMike99 (approved):

tracernz:
Why do we want to do this? Packs off takeoff is non-standard?

becas22:
It's a green operating procedure that, while not standard, is used by the vast majority of airlines I would say.

beheh:
IMHO we should be sticking to standard procedures where able. If this is a nuisance, how about we add a setting or memorize it instead?

becas22:
The normal procedure isn't PACKS ON IMO. The SOP calls for PACKS AS RQRD before takeoff and for PACKS ON after takeoff. There's no mention that the normal procedure should be to leave them ON for takeoff. However, in the FCTM green operating procedures, there's this: 
image


Which makes it seem like the SOP calls for PACKS ON but Airbus knows that everyone will depart with PACKS OFF.

becas22:
I agree @beheh , the best would be a setting or to memorize the previous selection, but that is out of my very limited coding abilities :D

tracernz:
The word "can" says it's not the default.

becas22:
In that case I will change the PR to just correct the  `defaultValue={initialState.takeoff.packs}` part and create a feat req for PACKS default option

PR #10250, opened 2025-07-22
fix(fms): handle multiple fixes of same ident for airways

tracernz:
Fixes issue reported by danbw9

## Summary of Changes

If there were multiple waypoints in the same region when we try to find an airway, we'd sometimes pick the wrong one e.g.
image

Now we find the right one. I also refactored the interface to what we actually need, rather than the generic implementation we aimed for when `msfs-navdata` was a separate library, to simplify the logic a little for this fix.

## Additional context

## Testing instructions

Try to import the route: `WSSS/02C N0506F360 VMR6A VMR M771 DUDIS/K0939F360 M771 72PCA/K0941S1095 M771 DOSUT/N0510F360 M771 DULOP/N0507F370 DCT DOTMI/K0939S1130 A470 LJG B221 SHZ W58 BK BK83A ZSPD/16L` in either A32NX or A380X. It should import fine, and especially check the `LJG B221 SHZ` segment.


Jonny23787:

Discord Username  : jonny_23
Object of testing : #10250 
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 22/07/2025

Testing Process: 

1. Imported given route, ensured it imported correctly with extra attention given to the LJG B221 SHZ segment.

Testing Results: 
Passed

fbw-common/src/systems/navdata/client/Database.ts
```diff
@@ -126,11 +126,11 @@ export class Database {
     return this.backend.getAirways(idents);
   }
 
-  public async getAirwaysByFix(fix: Fix, airwayIdent?: string): Promise<Airway[]> {
+  public async getAirwayByFix(fix: Fix, airwayIdent: string): Promise<Airway[]> {
```
Saschl, line 129: Nitpick: you renamed all the methods to ..airway... 
but they still return an array of Airway. Fine to me if it`s only one entry, but then we could get rid of the array?

danbw9:
I have tried with and without navigraph data installed. I have the same result with this bug, as before unfortunately - regarding routing. How it imports from simbrief.

Perhaps I am missing something. Let me know how I can further help.

Screenshot 2025-07-25 011949

fbw-common/src/systems/navdata/client/Database.ts
tracernz: Technically it's possible to have multiple instances of the same airway name, which is why it can still return an array.

tracernz:
@danbw9 Were you using the build artifacts from this PR? It hasn't merged yet so you won't see it on the dev version yet.

danbw9:
> @danbw9 Were you using the build artifacts from this PR? It hasn't merged yet so you won't see it on the dev version yet.

I downloaded the PR yes from build artifacts, not from dev version

tracernz:
Silly question: were you running on MSFS2024 SU2? Airways to/from VORs are broken there due to a sim bug that is fixed in SU3.

tracernz:
I'll merge this as it fixes a real bug that causes issues for people, and we can follow up any other bugs separately.

heclak:
> I'll merge this as it fixes a real bug that causes issues for people, and we can follow up any other bugs separately.

i was going to ask you about that. maybe the issue that I pointed out in my earlier test could be written up as a separate issue? but I'm not sure of the nature of problem anymore.

PR #10307, opened 2025-08-17
fix(a380X/flight model): Flight Model Update for MSFS 2024

donstim:
## Summary of Changes
Update of the A380X flight model for MSFS2024. This PR is a companion to #10546 , except for the following changes that address differences between MSFS 2020 and MSFS 2024:
1. The fuel system needed to be re-tuned for balanced transfer fuel flow from the inner and outer tanks to the feed tanks.
2. The flap lift and drag parameters needed to be re-tuned.
3. The new corrected MOI tensor calculation available only in MSFS 2024 was added.
4. Pitch trim was adjusted so that with full landing flaps at 380T, 36.6% CG, it is 3.4 Up at 1000 ft on glide slope at 141 knots to match simulator data

**This PR has parameters that are only available in MSFS 2024. It cannot be used in MSFS 2020.**

## Additional context

## Testing instructions
1. Confirm good pitch handling characteristics during takeoff and landing, including in crosswinds up to 40 knots.
2. Confirm good hand flying characteristics in pitch.
3. Confirm no auto-rotation before Vr on takeoff.
5. Confirm that the automatic fuel system transfers work correctly. Doing this will entail flights with: a) a fuel load of 26 to 36T of fuel, b) a fuel load of 47 T to 103 T, and a fuel load of 160 to 180 T. Make sure you have the FDR option turned on in the EFB and upload your FDR file here.


NoCabin:
I previewed this by swapping out the installer flightmodel.cfg with the one in this PR (IN 2024)

I noticed a few not so nice things. 

I tried doing a flight OMDB-YSSY with TOW @ 520T 

The auto-rotation bug is back, and it starts lifting the nose after 120kt, I made sure the trim was set properly. 
In terms of handling it was fine, no major changes felt when hand flying 

However, the aircraft barely climbed after takeoff, I missed the Dubai departure ALT constraints by 2-3000ft.
This behavior continued to higher altitudes, ultimately meant that I reached by cruise altitude of 33000 after a long 45 minutes. 

I did another departure test run and the aircraft exhibited the same behavior.

donstim:
@77W9940 Interesting. That certainly sounds like you still had the old flight model active. Please confirm that you replaced the old one with this new one in the correct MSFS 2024 community packages folder, and that you did this before starting MSFS 2024.  I just did two takeoffs from OMDB at your takeoff weight with CONF 1+F and did not experience any auto-rotation. For one takeoff, I just let it go until the nosewheel lifted off by itself, and that did not occur until 180 KCAS.

This flight model change should not significantly affect flaps up performance. Note that although this flight model's autorotation fix has been tested up to a TOW of 575 T, the A380X is still officially limited to a max TOW of 510 T for a number of other reasons. In my test, it took less than 25 minutes to reach FL330.

donstim:
I should add that I did developmental testing at weight-cg combinations comprising the corner points of the gross weight-cg envelope (up to 575 T) to make sure that the nosewheels did not lift off before VR. 

For your OMDB takeoff weight of 520 T, I manipulated the payload sliders to conduct tests at both the forward and aft limits of the cg envelope (as shown on the payload page of the EFB). I enabled the "wheels" data view from the dev menu to observe the weight on the nosewheels during the takeoff roll, and to see when they lifted off. I did not make any inputs on the pitch control during the tests. 

For the forward cg test, the nosewheels left the ground just after VR (164 KCAS) was reached:
https://1drv.ms/v/c/85cbffb84cec323b/ERjeLJHkkU1Do0WescGBS_wBf-1_ZK-GycTaCJsPGbJLlQ?e=HCSWBm

For the aft cg test, the nosewheels did not leave the ground, and the test was terminated when the airplane was at about 195 KCAS near the end of the runway.
https://1drv.ms/v/c/85cbffb84cec323b/ESQ8_0wcKvJLuqh5E8H60lYBQHPd97UUBSDeyvo_TWbkig?e=Pxb8hB

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/flight_model.cfg
```diff
@@ -705,7 +704,7 @@ lift_coef_mach_table = 0:1
 lift_coef_delta_elevator_mach_table = 0:0
 lift_coef_daoa_mach_table = 0:0
 lift_coef_pitch_rate_mach_table = 0:0
-lift_coef_horizontal_incidence_mach_table = 0:0
+lift_coef_horizontal_incidence_mach_table = 0.0
```
Maximilian-Reuter: Was that change intended ? Also I know this isnt really the point of the PR but I feel like when going to 2024 we could clean up this file a bit and remove legacy fsx stuff like this. I know the comment says they must be present but the docs at least for 2024 state that parameters like this one are only required for the legacy fm unless that is false.

donstim: No, it wasn't. Regarding removing the legacy params, see the note in the line labeling the legacy params. Although they are not used by the modern flight model, deleting some of them will cause an MSFS crash. I don't have the patience to try each one in turn to find out specifically which ones can be safely deleted. Instead, I segregated them from the parameters that are actually used so they are out of the way.

Maximilian-Reuter (approved):

stratavelo92:

Discord Username  : spottersim 
Object of testing : #10307  
Aircraft          : A380X 
Simulator         : 2024 
Tier of Testing   : 2
Date              : 16/04/2026 

Testing Process: 
Extensive testing in conjunction with #10607 and #10634
1. Pitch control was adequate in all situations
2. Felt adequate using -30% pitch sensitivity
3. Not observed
4. FDRs uploaded to discord, no abnormalities observed on FUEL SD pages


Testing Results: 
Passed  

Negatives:       
Pitch might feel a little more sensitive than roll
Aircraft might have a tendency to bounce after touchdown

Media:
Uploaded to dedicated discord thread

PR #10292, opened 2025-08-11
feat(a32nx/sfcc): implementation of SAP signals for A320 SFCC

Eagle941:
## Summary of Changes

This PR adds the implementation of the SAP (Slat Actual Position) Discrete Outputs for the A320 SFCC.

## Screenshots (if necessary)

N/A

## References

ATA27-50

## Additional context

To be merged after #10245

I would like to continue the work started in the draft PR https://github.com/flybywiresim/aircraft/pull/7632 which aims to rewrite the SFCC. Instead of making a big PR, I would like to make incremental changes.

Discord username (if different from GitHub): pythonist

## Testing instructions

This change improves the SFCC interface, no changes to the systems. Therefore, I don't see the need for testing instructions (happy to be suggested alternatives!)

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10315, opened 2025-08-22
feat(a32nx/sfcc): added connection to power for SFCC

Eagle941:
## Summary of Changes

This PR adds connection to electrical power for the SFCC.
SFCC1 is connected to DC ESS BUS
SFCC2 is connected to DC BUS 2

Logic is added in the flaps auto command function to initialise correctly upon power recovery.
Logic is implemented to handle long and short power fluctuations in the SFCC.

## Screenshots (if necessary)

N/A

## References

ATA27-50

## Additional context

To be merged after #10292 

## Testing instructions

TBD


fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
```diff
@@ -39,6 +48,11 @@ impl SlatsChannel {
             slats_demanded_angle: Angle::ZERO,
             slats_feedback_angle: Angle::ZERO,
 
+            powered_by,
+            is_powered: false,
+            recovered_power: false,
+            power_duration: Duration::ZERO,
```
Gurgel100: You could just use `systems::shared::DelayedTrueLogicGate` instead of implementing the logic yourself again.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
```diff
@@ -81,13 +104,22 @@ impl SlatsChannel {
         context: &UpdateContext,
         slats_feedback: &impl PositionPickoffUnit,
     ) {
+        self.power_duration += context.delta();
+
+        if self.recovered_power {
+            self.recovered_power = false;
+            self.power_duration = Duration::ZERO;
+        }
```
Gurgel100: When using `systems::shared::DelayedTrueLogicGate` this code could be simplified to:
```suggestion
        self.power_valid.update(context, self.is_powered);
```

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
```diff
@@ -81,13 +104,22 @@ impl SlatsChannel {
...
         self.csu_monitor.update(context);
         self.slats_demanded_angle = self.generate_slat_angle();
 
         self.slats_feedback_angle = slats_feedback.angle();
         self.sap_update();
     }
 
+    // The result of `get_demanded_angle` shall not be used outside of the SFCC.
+    // It returns 0.0 when the SFCC is powered off regardless of the last status.
     pub(super) fn get_demanded_angle(&self) -> Angle {
```
Gurgel100, line 121: Should this function then not be private (i.e. remove the `pub(super)`) if it shouldn't be use outside of the SFCC?

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
```diff
@@ -145,6 +189,15 @@ impl SimulationElement for SlatsChannel {
         visitor.visit(self);
     }
 
+    fn receive_power(&mut self, buses: &impl ElectricalBuses) {
+        if self.is_powered != buses.is_powered(self.powered_by) {
+            // If is_powered returns TRUE and the previous is FALSE,
+            // it means we have just restored the power
+            self.recovered_power = !self.is_powered;
+        }
```
Gurgel100: You can also remove this when using `systems::shared::DelayedTrueLogicGate`.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/mod.rs
```diff
@@ -72,28 +79,49 @@ impl SlatFlapControlComputerMisc {
         Self::in_enlarged_target_range(position, target_position)
             || position >= target_position + tolerance
     }
+
+    pub fn between_0_1f_enlarged_target_range(position: Angle) -> bool {
```
Gurgel100: I think I missed this in my last review but this doesn't need to be marked `pub` since it's only used inside this module. This also applies to the struct `SlatFlapControlComputerMisc` and all other member functions.
```suggestion
    fn between_0_1f_enlarged_target_range(position: Angle) -> bool {
```

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/mod.rs
```diff
@@ -72,28 +79,49 @@ impl SlatFlapControlComputerMisc {
...
             slat_flap_actual_position_word_id: context
                 .get_identifier(format!("SFCC_{num}_SLAT_FLAP_ACTUAL_POSITION_WORD")),
 
-            flaps_channel: FlapsChannel::new(context, num),
-            slats_channel: SlatsChannel::new(context, num),
+            powered_by,
+            consumed_power: Power::new::<watt>(Self::MAX_POWER_CONSUMPTION_WATT),
+            is_powered: false,
+            recovered_power: false,
+            power_duration: Duration::ZERO,
```
Gurgel100: How about using `systems::shared::DelayedTrueLogicGate` here as well?

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -53,6 +63,11 @@ impl FlapsChannel {
             flaps_demanded_angle: Angle::ZERO,
             flaps_feedback_angle: Angle::ZERO,
 
+            powered_by,
+            is_powered: false,
+            recovered_power: false,
+            power_duration: Duration::ZERO,
```
Gurgel100: Use `systems::shared::DelayedTrueLogicGate`.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -95,6 +115,21 @@ impl FlapsChannel {
...
+    ) -> (Option<Velocity>, Option<Velocity>) {
+        let cas1 = adirs.computed_airspeed(1).normal_value();
+        let cas2 = adirs.computed_airspeed(2).normal_value();
+
+        match (cas1, cas2) {
+            (Some(cas1), Some(cas2)) => (Some(cas1), Some(cas2)),
+            (Some(cas1), None) => (Some(cas1), Some(cas1)),
+            (None, Some(cas2)) => (Some(cas2), Some(cas2)),
+            (None, None) => (None, None),
+        }
```
Gurgel100: This can be simplified to:
```suggestion
        (cas1.or(cas2), cas2.or(cas1))
```

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -252,7 +277,119 @@ impl FlapsChannel {
...
     }
 
+    fn powerup_reset(&mut self, adirs: &impl AdirsMeasurementOutputs) {
+        // Auto Command restart
+        if self.csu_monitor.get_last_valid_detent() != CSU::Conf1 {
+            self.flap_auto_command_active = false;
+        } else {
+            // The match can be shortened by a convoluted if statement however
+            // I believe it would make debugging and understanding the state machine harder
+            match self.get_cas(adirs) {
```
Gurgel100, line 293: I didn't realize in the initial review of the SFCC but there are some cases in this `match` block which will never execute.
For example
```rust
(Some(cas1), _) if cas1 > self.kts_100
    && cas1 < self.kts_210
    && SlatFlapControlComputerMisc::between_0_1f_enlarged_target_range(
        self.flaps_feedback_angle,
    )
```
and
```rust
(_, Some(cas2)) if cas2 > self.kts_100
    && cas2 < self.kts_210
    && SlatFlapControlComputerMisc::between_0_1f_enlarged_target_range(
        self.flaps_feedback_angle,
    )
```
have the exact same condition but the latter will never execute because either both (`cas1` and `cas2`) have some value or both have `None` (because you're calling `self.get_cas`).

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -252,7 +277,119 @@ impl FlapsChannel {
...
+            }
+            self.flap_auto_command_active = true;
+        }
+    }
+
     fn generate_flap_angle(&mut self, adirs: &impl AdirsMeasurementOutputs) -> Angle {
+        if !self.is_powered && self.power_duration > Self::TRANSPARENCY_TIME {
+            self.flap_auto_command_active = false;
+            self.flap_auto_command_engaged = false;
+            self.flap_auto_command_angle = Angle::ZERO;
```
Gurgel100: Maybe we can put such resets into a separate function which is called when power is lost.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -307,6 +456,10 @@ impl FlapsChannel {
     }
 
     fn flap_actual_position_word(&self) -> Arinc429Word<f64> {
+        if !self.is_powered && self.power_duration > Self::TRANSPARENCY_TIME {
```
Gurgel100: I think this condition is wrong?
What if `is_powered` is true (because it just got powered). Then it doesn't matter for how long it was powered it will not go into this branch. Or is this the intended behavior?

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
Eagle941: The match statements are a bit convoluted, I am going to add explanatory comments.
I have checked each condition is tested in the unit tests with code coverage.
`(Some(cas1), _)` is hit during execution of line 1769 of `test.rs`.
`(_, Some(cas2))` is hit during execution of line 1797 of `test.rs`.
These two cases are not overlapping in the case of both ADIRUs having valid but different speeds.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
Eagle941: I set it as `pub(super)` to be able to use it in the unit tests, but also within the methods of the struct. I am thinking that reading the variable directly instead of using `get_demanded_angle` would make the code less flexible.
`pub(super)` should cater for it, I wrote the comment to make people think twice in case of future changes.
I have improved the wording of the comment.

Eagle941:
@Gurgel100 Thank you for your comments! Ready for recheck

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/slats_channel.rs
Gurgel100: Yep. You're right.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/mod.rs
```diff
@@ -41,59 +49,44 @@ impl From<u8> for FlapsConf {
...
-            flaps_channel: FlapsChannel::new(context, num),
-            slats_channel: SlatsChannel::new(context, num),
+            powered_by,
+            consumed_power: Power::new::<watt>(Self::MAX_POWER_CONSUMPTION_WATT),
+            is_powered: false,
+            is_powered_delayed: DelayedFalseLogicGate::new(Self::TRANSPARENCY_TIME)
+                .starting_as(false),
+
+            flaps_channel: FlapsChannel::new(context, num, powered_by),
+            slats_channel: SlatsChannel::new(context, num, powered_by),
```
Gurgel100, line 89: I don't quite understand why the channels do their own power management if they are part of the SFCC? IMO the channels themself should not worry about the power state.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/utils.rs
```diff
@@ -0,0 +1,133 @@
+use uom::{
+    si::{angle::degree, f64::*},
+    ConstZero,
+};
+
+pub struct SlatFlapControlComputerMisc {}
```
Gurgel100:
```suggestion
pub(super) struct SlatFlapControlComputerMisc {}
```
As well for the member functions.

fbw-common/src/wasm/systems/systems/src/shared/mod.rs
```diff
@@ -627,6 +627,20 @@ impl DelayedFalseLogicGate {
...
+        self
+    }
+
+    fn set_output(&mut self, state: bool) {
+        self.expression_result = state;
+        if !state {
+            self.false_duration = self.delay;
+        } else {
+            self.false_duration = Duration::from_millis(0)
+        }
```
Gurgel100:
```suggestion
        self.false_duration = if !state {
            self.delay
        } else {
            Duration::from_millis(0)
        };
```
According to coding [guidelines](https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/wasm/systems/guidelines.md#single-assignment-for-simple-if-else).

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/flaps_channel.rs
```diff
@@ -252,7 +288,129 @@ impl FlapsChannel {
...
+
+    fn power_loss_reset(&mut self) {
+        self.flap_auto_command_active = false;
+        self.flap_auto_command_engaged = false;
+        self.flap_auto_command_angle = Angle::ZERO;
+    }
+
     fn generate_flap_angle(&mut self, adirs: &impl AdirsMeasurementOutputs) -> Angle {
+        if !self.is_powered_delayed.output() {
+            self.power_loss_reset();
```
Gurgel100: The reset logic should be placed in the `update` method not at a "random" place. When the calls are switched around later it might affect the behavior unintentionally.

fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/sfcc/mod.rs
Eagle941: Yes, I agree. I did it like this to avoid passing around too many variables. Should I pass the `is_powered_delayed` argument in the update function of each channel?

Gurgel100: It would be definitely better this way. Or you extract the writing of the simvars into the sfcc.

lukecologne: Well,  technically, the two channels have independent powersupplies, even though the supply bus is the same. But the circuit breaker is not, so this would be relevant once we simulate CBs

Eagle941: True... Should I leave as is with a note or simplify?

tracernz: We should make it as easy as possible to connect the CBs as soon as they're ready. We've been ready except for the rust systems for a couple of years.

Gurgel100: I didn't know that the channels are on different circuit breakers. In this case I agree to the previous comments and we can leave it as it is now.

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -225,10 +230,10 @@ export class Slats extends DisplayComponent<SlatsProps> {
           <text class={this.targetClass} x={-3} y={59}>
             {this.targetText}
           </text>
-          <text class="Standard Center" x={-101} y={15}>
+          <text class={this.sfccValid.map((v) => (v ? 'Standard Center' : 'Standard Center Amber'))} x={-101} y={15}>
```
flogross89:
```suggestion
          <text class={{ 'Amber': this.sfccValid.map((v) => !v), 'Standard': true, 'Center': true }} x={-101} y={15}>
```
is the style we mostly use (might need to be formatted properly).

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -225,10 +230,10 @@ export class Slats extends DisplayComponent<SlatsProps> {
...
             S
           </text>
-          <text class="Standard Center" x={105} y={15}>
+          <text class={this.sfccValid.map((v) => (v ? 'Standard Center' : 'Standard Center Amber'))} x={105} y={15}>
```
flogross89: see above

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -252,11 +257,23 @@ export class Slats extends DisplayComponent<SlatsProps> {
           A LOCK
         </text>
 
-        <path class="Slats" d={this.slatsPath} />
-        <line class="GreenLine" x1={-18} y1={0} x2={this.slatsEndX} y2={this.slatsEndY} />
-
-        <path class="Flaps" d={this.flapsPath} />
-        <line class="GreenLine" x1={0} y1={0} x2={this.flapsEndX} y2={this.flapsEndY} />
+        <path class={this.sfccValid.map((v) => (v ? 'SlatsGreen' : 'SlatsAmber'))} d={this.slatsPath} />
```
flogross89: see above, ideally use one base class `Slats` and supplement it with another class `Amber` if needed and use the style proposed above

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -252,11 +257,23 @@ export class Slats extends DisplayComponent<SlatsProps> {
...
+        <line
+          class={this.sfccValid.map((v) => (v ? 'GreenLine' : 'AmberLine'))}
```
flogross89: also applies here

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -252,11 +257,23 @@ export class Slats extends DisplayComponent<SlatsProps> {
...
+          x1={-18}
+          y1={0}
+          x2={this.slatsEndX}
+          y2={this.slatsEndY}
+        />
+
+        <path class={this.sfccValid.map((v) => (v ? 'FlapsGreen' : 'FlapsAmber'))} d={this.flapsPath} />
```
flogross89: and here

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -252,11 +257,23 @@ export class Slats extends DisplayComponent<SlatsProps> {
...
+        <line
+          class={this.sfccValid.map((v) => (v ? 'GreenLine' : 'AmberLine'))}
```
flogross89: same

fbw-a32nx/src/systems/instruments/src/EWD/shared/ArincValueProvider.ts
```diff
@@ -32,19 +39,68 @@ export class ArincValueProvider {
       publisher.pub('sat', this.sat);
     });
 
-    subscriber.on('slatsFlapsStatusRaw').handle((w) => {
+    subscriber.on('sfcc1SlatsFlapsStatusRaw').handle((w) => {
       this.slatsFlapsStatus = new Arinc429Word(w);
-      publisher.pub('slatsFlapsStatus', this.slatsFlapsStatus);
+      this.sfcc1SlatsFlapsStatus = new Arinc429Word(w);
```
flogross89: This is allocating a new object on every update, which is detrimental to performance. Yes, this is the pattern used at the moment but we should avoid it in the future and someone needs to refactor it :) 
Two possible approaches: 
1) My recommendation: Change `this.sfcc1SlatsFlapsStatus` to an `Arinc429RegisterSubject`, and call `this.sfcc1SlatsFlapsStatus.setWord(w);` in this line. See https://github.com/flybywiresim/aircraft/blob/master/fbw-a380x/src/systems/instruments/src/EWD/shared/ArincValueProvider.ts as reference
2) (not recommended here) You use `Arinc429LocalVarConsumerSubject` which will automatically always hold the latest version of the local var. Then you don't need to handle both raw and arinc word object, and can just subscribe to changes of the `Arinc429LocalVarConsumerSubject`. This requires more refactoring though, see

fbw-a32nx/src/systems/instruments/src/EWD/shared/ArincValueProvider.ts
```diff
@@ -32,19 +39,68 @@ export class ArincValueProvider {
...
+      if (this.sfccToUse === 1) {
+        publisher.pub('slatsFlapsStatus', this.slatsFlapsStatus);
+      } else if (this.sfccToUse === 0) {
+        publisher.pub('slatsFlapsStatus', new Arinc429Word(0));
+      }
     });
 
-    subscriber.on('slatsPositionRaw').handle((w) => {
+    subscriber.on('sfcc1SlatsPositionRaw').handle((w) => {
       this.slatsPosition = new Arinc429Word(w);
```
flogross89: Same here and every line in this file with `new Arinc429Word(w)`

.github/CHANGELOG.md
```diff
@@ -133,6 +133,8 @@
 1. [A380X/OIT] Add initial implementation of OIT AVNCS side, including flight log and inbox page for COMPANY COM app - @flogross89 (floridude)
 1. [A380X/FG] Fix SRS disengagement conditions for eng 3&4 OEI - @flogross89 (floridude)
 1. [A380X/FMS] Add engine-out indications and behavior to FMS - @flogross89 (floridude)
+1. [A32NX/SFCC] Added implementation of SAP and FAP in SFCC - @Eagle941 (Joe)
+1. [A32NX/SFCC] Connected SFCC to electrical busses with simulation of power consumption and power restoration behaviour - @Eagle941 (Joe)
```
Gurgel100: Since a new stable was branched of this should be under the 0.15.0 version

Gurgel100 (approved):
From the Rust side LGTM!

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -58,158 +140,176 @@ export class Slats extends DisplayComponent<SlatsProps> {
...
+        this.sfcc1FlapsPositionSub.resume(true);
+        this.sfcc1SlatsPositionSub.resume(true);
+      }
+    }, true);
+
+    this.sfccSlatsFlapsStatus.sub(this.setSfccStatus.bind(this), true);
+    this.sfccFlapsPosition.sub(this.setFlapsPath.bind(this), true);
+    this.sfccSlatsPosition.sub(this.setSlatsPath.bind(this), true);
+
+    // FIXME should not need realtime and alloc strings every frame for this
```
flogross89, line 308: That could have been solved by two MappedSubjects but to reduce the scope of this PR I'd say it's OK to leave as is.

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
```diff
@@ -225,10 +325,10 @@ export class Slats extends DisplayComponent<SlatsProps> {
           <text class={this.targetClass} x={-3} y={59}>
             {this.targetText}
           </text>
-          <text class="Standard Center" x={-101} y={15}>
+          <text class={{ Amber: this.sfccValid.map((v) => !v), Standard: true, Center: true }} x={-101} y={15}>
```
flogross89: My proposal: Introduce another class property `sfccNotValid`, to save these additional four subscriptions. 
`private readonly sfccNotValid = this.sfccValid.map((v) => !v);`

flogross89 (changes requested):
Nice! Only one tiny change, but TS LGTM

fbw-a32nx/src/systems/instruments/src/EWD/Slats.tsx
Eagle941: Yes, there is other refactoring which could be done in the rest of the file. It's best to leave it for a second PR

fbw-a32nx/src/systems/shared/src/publishers/A32NXSfccBusPublisher.ts
```diff
@@ -0,0 +1,74 @@
+// @ts-strict-ignore
```
Benjozork: We should not be committing new files with strict ignore.

fbw-a32nx/src/systems/shared/src/publishers/A32NXSfccBusPublisher.ts
```diff
@@ -0,0 +1,74 @@
...
+  a32nx_sfcc_1_flaps_position: number;
+
+  /** TODO */
+  a32nx_sfcc_2_slats_flaps_status: number;
+
+  /** TODO */
+  a32nx_sfcc_2_slats_position: number;
+
+  /** TODO */
+  a32nx_sfcc_2_flaps_position: number;
```
tracernz: Needs full documentation.

tracernz (changes requested):

Jonny23787:

Discord Username  : jonny_23
Object of testing : #10315 
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 10/11/2025

Testing Process: 

1. Completed all test steps, everything worked fine.

Testing Results: 
Passed

Negatives:       


Conclusions: 


Media:

PR #10339, opened 2025-09-17
fix(fms): use local magvar for direct-to turning point

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes a bug where a direct to would be calculated with an incorrect initial course due to a missing conversion between true and magnetic courses. Legs not part of a procedure would use a magnetic variation of 0. Now, they use the magnetic variation from the magvar database.

This PR does not affect navdata parsing.

## Screenshots (if necessary)

Before:
image
After:
image

## Additional context

Perform direct-to's in different parts of the world, preferably in regions of very low/high latitude. Make sure the trajectory is drawn accurately on the ND, and that the autopilot flies it without any swaying.

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanInterface.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanInterface.ts
@@ -262,7 +262,7 @@ export interface FlightPlanInterface<P extends FlightPlanPerformanceData = Fligh
 
   directToLeg(
     ppos: Coordinates,
-    trueTrack: Degrees,
+    magneticTrack: Degrees,
     targetLegIndex: number,
     withAbeam: boolean,
     planIndex: number,
@@ -271,7 +271,7 @@ export interface FlightPlanInterface<P extends FlightPlanPerformanceData = Fligh
   // TODO do not pass in fix object (rpc)
   directToWaypoint(
     ppos: Coordinates,
-    trueTrack: Degrees,
+    magneticTrack: Degrees,
     waypoint: Fix,
     withAbeam: boolean,
     planIndex: number,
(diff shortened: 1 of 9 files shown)


tracernz:
We have to be careful here, because legs defined in true degrees are supposed to be always shown in true in the MCDU pages. When the FMS and it's MCDU pages properly support polar ops this fix might get messy.

BlueberryKing:
> We have to be careful here, because legs defined in true degrees are supposed to be always shown in true in the MCDU pages. When the FMS and it's MCDU pages properly support polar ops this fix might get messy.

You're right. I guess if we still want the CF to be magnetic-referenced, it should be defined by the measured magnetic track at the time of insertion, but we store the measured magnetic variation with it, i.e (measured true track - measured magnetic track). That way, we are sure to get the outbound course to be exactly the current true track without relying on a magvar database.

I suppose the real FMS computes these from GNSS/inertial measurements? For now, we can use our inertial ones from the ADIRS or even just the sim variables directly until we implement the FMS navigation more completely.

tracernz:
> You're right. I guess if we still want the CF to be magnetic-referenced, it should be defined by the measured magnetic track at the time of insertion

Only if the magnetic heading/track labels are valid, else it should be true referenced. The IR puts out magnetic (NCD with true track when in polar area) and true track.

Why is it necessary to store the magvar with the leg? It's trivially retrievable from the magvar database at any time (we currently use Facilties.getMagVar for that).

BlueberryKing:
> Why is it necessary to store the magvar with the leg? It's trivially retrievable from the magvar database at any time (we currently use Facilties.getMagVar for that).

We want to make sure the T-P gets the actual true course we are flying when we insert the direct, so we store the measured magvar, i.e true track - magnetic track. Using the magvar from the database might not give us our current, actual true course and lead to an awkward direct turn. I have to check whether this makes a significance difference though.

tracernz:
The T-P has a lat/lon like other waypoints, so the magvar can always be retrieved from the database.

BlueberryKing:
I thought about this some more and ended up rewriting it. The actual issue was that enroute legs used a magvar of 0 instead of `Facilities.getMagVar`.

BlueberryKing:
Superseded by https://github.com/flybywiresim/aircraft/pull/10530

PR #10338, opened 2025-09-16
docs: Revisions to documentation, A380X referencing, file structure consolidation, and optimised naming conventions

CameronCarlyon:
Fixes #10337

## Summary of Changes

- Updated file name from Contributing.md to CONTRIBUTING.md and Resources.md to RESOURCES.md, in-line with FlyByWire naming conventions.

- Consolidated general development documentation into to .github folder.

- Updated CONTRIBUTING.md document to more thoroughly include references to the A380X, with minor tweaks to sentence structure throughout document to enhance readability.

- Updated README.md to ensure correct hyperlink attributes.

- Included change in CHANGELOG.md (not sure if this is necessary given the scope of the changes, feel free to remove)

## Additional context

Additional hyperlink references may need to be updated by an admin, such as these notices:

image

Discord username (if different from GitHub):
YouBaconMeCrazy

## Testing instructions

Links may be tested in the README.md file by clicking through and ensuring there is not a NoPage error.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10348, opened 2025-09-21
Fix broken link to dev resources in Contributing.md

evroon:
## Summary of Changes

Fixes a broken link in Contributing.md that used to point to this, but since v10.0, it is broken. It seems like it's supposed to link to the website instead now.

## Additional context

### DIFF
--- a/.github/Contributing.md
+++ b/.github/Contributing.md
@@ -47,7 +47,7 @@ As this is an open source project, anyone is free to contribute as much or as li
 
 If you're comfortable contributing to Open Source projects on GitHub please ensure you read our expectations for issue tracking, feature proposals and pull requests.
 
-If you're looking for tools and tips to help you develop, see [Development Resources](../fbw-a32nx/docs/resources.md).
+If you're looking for tools and tips to help you develop, see [Development Resources](https://docs.flybywiresim.com/dev-corner/dev-guide/).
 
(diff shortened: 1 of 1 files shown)


BlueberryKing:
Thanks for contributing!

The link is not broken for me. It leads to https://github.com/flybywiresim/aircraft/blob/master/fbw-a32nx/docs/resources.md , where we do list development resources. 

However, some of the resources, such as the tools at the top, are outdated and should be removed from that page. I like adding a link to our website, but maybe that could be appended instead of replacing an existing link?

evroon:
Hmm weird, for me the link goes to https://github.com/flybywiresim/aircraft/blob/fbw-a32nx/docs/resources.md. So github interprets the the path to the file as the branch name because there is no branch in the link. Maybe it should just be an absolute markdown link instead of relative?

I can add the website link instead of replacing it, sure, I'll do that tomorrow.

The slightly bigger changes you mention should perhaps take place after https://github.com/flybywiresim/aircraft/pull/10338 to avoid (more) merge conflicts.

evroon:
Ahh I see now, if you click the link here https://github.com/flybywiresim/aircraft?tab=contributing-ov-file, the link is broken. But from https://github.com/flybywiresim/aircraft/blob/master/.github/Contributing.md, it works. So I guess best way to fix that is to make the link absolute?

FoxtrotSierra6829 (approved):

PR #10363, opened 2025-10-12
fix(a32nx/a380x): fix finite knobs wrapping back to start and enable mouse drag

heclak:
Fixes #[issue_no]

## Summary of Changes

 - fix issue where finite knobs would snap back to start position when turned to max value
 - enable mouse drag for knobs in lock mode

Affected knobs are primarily display brightness knobs and lights in both aircraft. A32NX air cond knobs, A380 cockpit temp knob.

The cabin temp knob in the A380 still uses the old behaviour and would require a new behaviour rewrite to find a more appropriate way to implement mouse drag for lock mode.

## Additional context

## Testing instructions

Test affected knobs in both legacy and lock mode. Legacy mode should have no change but lock mode will now enable mouse drag. Issue where knob skips back to the start position should not be present anymore.

Click spot icon issue in 2024 is not resolved in this PR.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,6 +10,8 @@
 <!--  1. [Changed Area] Title of changes - @github username (Name)  -->
 <!-- The changed area should be prefixed with "A32NX/" or "A380X/" if only one of them is affected. -->
 
+## 0.15.0
+
 ## 0.14.0
 
 1. [A380X/FMS] Fix speed margins being displayed in the wrong place for a Mach target - @BlueberryKing (BlueberryKing)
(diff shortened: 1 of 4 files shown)


heclak:
The A380 cabin temp knob is “correct” for this PR. There was a previous hack make it seem infinite by making it three turns instead of 1 (thus causing the wrapping bug in the first place). It’s the one knob that still uses the old knob implementation which is why it can’t use mouse drag in lock mode. We’ll rewrite an entirely new behaviour for it in a separate PR. If that’s the only “issue”, it would be a pass for the PR.

alexr4339:

Discord Username  : alexdnzdmb 
Object of testing : #10363 
Aircraft          : A380X 
Simulator         : 2024 
Tier of Testing   : 1
Date              : 14/10/2025 

Testing Process: 
Test affected knobs in both legacy and lock mode. Legacy mode should have no change but lock mode will now enable mouse drag. Issue where knob skips back to the start position should not be present anymore.

A380X: Works as expected in all modes. However I noticed the A380X Cabin Temp knob can be turned 3 times around until it stops at the colde or hotter PURS SEL position -accepted as "correct" for now
A32NX: Works as expected

Testing Results: 
Passed  

Negatives:       


Conclusions: 


Media:

PR #10369, opened 2025-10-17
fix: Improve self test sequence for PRIM, SEC, ELAC, FAC, FMGC

flogross89:
## Summary of Changes

Implements the flickering of the overhead lights during the long self-test sequence found on ground for both A380X and A32NX. 

## References

available via dm

For the A380, following times: 
PRIM
36 seconds overall
12.29 off
13.42 on
28.67 off
31.04 on
31.63 off
31.71 on
33.13 off
33.21 on
35.75 off

SEC A380 
22 seconds overall
06.46 off
18.50 on
18.63 off
21.13 on
21.33 off
21.42 on
21.54 off
21.58 on
21.96 off

## Additional context

## Testing instructions

Energize aircraft (both a32nx and a380nx) on ground, observe flickering FAULT lights during the self test sequence for:
A380X: PRIM, SEC
A32NX: FAC, SEC, ELAC, FCU (---- <-> numbers)
Also test the same in the air, ensure that the computers do not run a test sequence and boot up in short time.
Also, check runway and especially in air spawns to ensure these continue to function as before


fbw-a32nx/src/wasm/fbw_a320/src/fac/Fac.cpp
```diff
@@ -77,8 +77,22 @@ void Fac::updateSelfTest(double deltaTime) {
   if (selfTestTimer > 0) {
     selfTestTimer -= deltaTime;
     selfTestComplete = false;
+
+    // Hardcoded test light sequence. Between the times (in seconds) in each array, the light is on.
+    selfTestFaultLightVisible = false;
+    // FIXME no references available, best guess
+    double testLightOnTimes[][2] = {{0, 2},       {4.77, 4.87}, {4.9, 5.0},   {5.17, 5.27}, {5.33, 5.43},
```
lukecologne: Maybe it makes sense to extract this function somewhere to fbw-common for deduplication now that it's used so many times all over? Also, I think all the `testLightOnTimes` could be `const` or even `constexpr`

fbw-a32nx/src/wasm/fbw_a320/src/fac/Fac.cpp
```diff
@@ -126,7 +140,7 @@ base_fac_bus Fac::getBusOutputs() {
 base_fac_discrete_outputs Fac::getDiscreteOutputs() {
   base_fac_discrete_outputs output = {};
 
-  output.fac_healthy = facHealthy;
+  output.fac_healthy = (selfTestComplete && facHealthy) || (!selfTestComplete && !selfTestFaultLightVisible);
```
lukecologne, line 142: These discrete outputs are often used also by other computers for healthy status information (FWC, DMC, FAC opp for example for the FAC). IRL, the FAULT light is a separate discrete to the healthy status discrete. Not a showstopper, but maybe something to look out for during testing if it messes with other behaviour during computer startup.

flogross89: Yep I had a look at some potential impact, but I couldn't find a critical one. The FWS suppresses these faults for flight phase 1 anyways (and has its own startup time), and couldn't find another instance where a system toggling between avail/NA would mess with output.
Also, I didn't wann to separate these discretes at the moment, our aircraft is complex enough...

fbw-a32nx/src/wasm/fbw_a320/src/fac/Fac.cpp
flogross89: Sounded like a good idea so I tried it: When passing testLightOnTimes to a function, it decays to a pointer, and I can't use the range-for loop anymore. It would anyways only save around 5-6 lines, since the logic around these lines is not the same everywhere.I'd propose to keep it as is, I couldn't find a nice way to extract it. 
But I will change it to a constexpr everywhere.

fbw-a32nx/src/wasm/fbw_a320/src/fac/Fac.cpp
BravoMike99: > Yep I had a look at some potential impact, but I couldn't find a critical one. The FWS suppresses these faults for flight phase 1 anyways (and has its own startup time), and couldn't find another instance where a system toggling between avail/NA would mess with output. Also, I didn't wann to separate these discretes at the moment, our aircraft is complex enough...

Only if the P/B is off in phase 1 & 10/12 it's inhibited for the PRIM/ELAC/SEC. Meanwhile the FAC has no inhibition at all. The healthy or faulty signal needs to drive the warning without inhib otherwise faults would not appear during those phases which would be misleading. In any case, I have seen PRIM faults on the wide-bodies during the self test sequence which disappears on its own after some seconds so that should be accurate.

lukecologne: I have done some more research, and for SEC and ELAC, the fault discrete to the FCDC is actually the same signal as the FAULT button light. It might be though that the FCDC does some confirm time processing to eliminate nuisance warnings

Jonny23787:

Discord Username  : jonny_23
Object of testing : #10369 
Simulator         : 2020
Tier of Testing   : 1
Date              : 12/11/2025

Testing Process: 

1. Tested each aircraft cold and dark, ensured fault lights/fcu for A32NX went through the self test sequence.
2. Tested on the runway and in the air for both aircraft, ensure only a short self test was conducted.

Testing Results: 
Passed

Negatives:       


Conclusions: 


Media:

PR #10371, opened 2025-10-17
chore: added `FBW_TYPECHECK=1` to `.env` to match build action

Eagle941:
## Summary of Changes

This PR adds `FBW_TYPECHECK=1` to the `.env` file to match the building process in Github actions.
It is also useful to run the typechecker.

## Screenshots (if necessary)

N/A

## References

N/A

## Additional context

N/A

## Testing instructions

No testing required.


Benjozork:
The reason there is an environment variable in the first place is to only enable it by default in CI since it adds quite a bit to the build time.

Eagle941:
> The reason there is an environment variable in the first place is to only enable it by default in CI since it adds quite a bit to the build time.

Then the CI and `build.sh` behave differently.
I feel adding the flag manually before local builds is not a nice "dev experience", also this isn't mentioned in the developer guide either.
Also local builds use cache which saves on time anyway.

tracernz:
If you want it, you can add it to your local .env at the root of the repo. It's too slow for regular dev builds.

flogross89:
How much does it add approximately to the build time? It would be nice to have a consistent build on both environments, but for me (and new contributors also) build speed is also certainly a factor

Eagle941:
> How much does it add approximately to the build time? It would be nice to have a consistent build on both environments, but for me (and new contributors also) build speed is also certainly a factor

`./scripts/dev-env/run.cmd ./scripts/build.sh --no-cache`

With `FBW_TYPECHECK=1`
image

Without `FBW_TYPECHECK=1`
image

The difference in real time is 3 minutes which is reduced when cache is used.

tracernz:
For an actual realistic scenario during development I do this, making a small change to the MCDU code each time so it's a fresh build as it would be when actually working:
```sh
# First did a few runs so any OS level caches are hot

# FBW_TYPECHECK=1

$ .\scripts\dev-env\run.cmd ./scripts/build.sh -r MCDU --no-cache
real    1m22.350s
user    0m23.380s
sys     0m7.980s

# FBW_TYPECHECK=

$ .\scripts\dev-env\run.cmd ./scripts/build.sh -r MCDU --no-cache
real    0m26.165s
user    0m3.752s
sys     0m4.186s

# FBW_TYPECHECK=1

$ .\scripts\dev-env\run.cmd ./scripts/build.sh -r MCDU --no-cache
real    1m18.219s
user    0m24.033s
sys     0m7.397s

# FBW_TYPECHECK=

$ .\scripts\dev-env\run.cmd ./scripts/build.sh -r MCDU --no-cache
real    0m26.068s
user    0m4.152s
sys     0m3.802s
```

So, untenable for normal development.

The change I was making each time was incrementing a number I added in this string https://github.com/flybywiresim/aircraft/blob/b2fffc4edc0078875b198d2a5a4ce61cb8456834/fbw-a32nx/src/systems/fmgc/src/navigation/Navigation.ts#L97.

Eagle941:
Ok I close it then. As we discussed, I will make an update to the dev docs to explain this.

tracernz:
We should revisit this after moving to typescript 7.

PR #10378, opened 2025-10-25
perf(oans): reduce VRAM usage

Benjozork:
Combine and scale down render size of certain layers to save ~50% of VRAM with very little visual impact.

## Summary of Changes
- Upgraded `turf.js` to v7.
- Refactored some OANS code to keep layer data in one place and preserve order of features according to order of style rules in each layer.
- Combine layers so that we have 5 total layers instead of 8.
- Allow individual layers to render at a different scale to save on VRAM. This is used to render layers that are only visible at higher ranges at a lower resolution.
- Add optional union-ing and simplifying of features per style rule. In this case, this allowed fixing some visual artifacts caused by the topology of runway features that were made worse by a lower render scale..

Possible future work: make the BTV layer be a fixed 786x786 layer instead of making it draw on an airport map-sized layer. Could yield an additional ~25% savings.

## Testing instructions

- Ensure all OANS features behaves and look as they did before.
- Ensure all BTV features behave and look as they did before.

### DIFF
--- a/fbw-common/src/systems/instruments/src/OANC/style-data.ts
+++ b/fbw-common/src/systems/instruments/src/OANC/style-data.ts
@@ -3,113 +3,172 @@
 
 import { FeatureType, PolygonalStructureType } from '@flybywiresim/fbw-sdk';
 
-export interface StyleRule {
-  forFeatureTypes?: FeatureType[];
+export interface LayerSpecification {
+  renderScale: number;
 
-  forPolygonStructureTypes?: PolygonalStructureType[];
+  zoomLevelVisibilities: [boolean, boolean, boolean, boolean, boolean];
 
-  dontFetchFromAmdb?: boolean;
-
-  styles: {
-    doStroke: boolean;
-
-    doFill: boolean;
-
-    strokeStyle?: string;
-
-    lineWidth?: number;
-
-    fillStyle?: string;
-  };
+  styleRules: StyleRule[];
 }
 
-export const STYLE_DATA: Record<number, StyleRule[]> = {
-  0: [
-    {
-      forFeatureTypes: [FeatureType.TaxiwayElement],
-      styles: { doStroke: false, doFill: true, fillStyle: '#8f8f8f' },
-    },
-    {
-      forFeatureTypes: [FeatureType.TaxiwayShoulder],
-      styles: { doStroke: false, doFill: true, fillStyle: '#85451d' },
-    },
-    {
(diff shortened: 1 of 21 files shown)


fbw-common/src/systems/instruments/src/OANC/style-data.ts
```diff
@@ -1,115 +1,164 @@
...
+  /** If `true`, the feature won't be fetched from the AMDB */
   dontFetchFromAmdb?: boolean;
 
-  styles: {
-    doStroke: boolean;
+  /** If applicable, how to union features this rule applies to. Only valid for polygons. */
+  unionBy?: keyof AmdbProperties | 'all';
+
+  /** If provided, the tolerance with which to simplify features. If not provided, the feature is not simplified. */
+  simplify?: number;
```
tracernz, line 31: Does this have units? Would be good to know how to choose a value from the JSDoc.

scripts/build.sh
```diff
@@ -45,7 +47,7 @@ else
   if [ "${USE_4K_TEXTURES}" == "true" ]; then
     time npx igniter -r "^(?!.*ci-build)(?!.*8K).*" "${args[@]}"
   else
-    time npx igniter -r "^(?!.*ci-build)(?!.*4K).*" "${args[@]}"
+    FBW_TYPECHECK=1 time npx igniter -r "^(?!.*ci-build)(?!.*4K).*" "${args[@]}"
```
tracernz: Unintended changes?

fbw-common/src/systems/instruments/src/OANC/style-data.ts
Benjozork: Done.

scripts/build.sh
Benjozork: Reverted.

tracernz (approved):

pilotseyea350:

Discord Username  : PilotEyesA350 
Object of testing : #10378 
Aircraft          : A380X 
Tier of Testing   : 1
Date              : 26/10/2025 

Testing Process: 
 
- Startup on the RWY 04R at KJFK, 
- Normal T.O.
- Set BTV on W RWY31R
- Upon touchdown, opened OANS
- Checked Flags and crosses in ARC and PLAN mode
- Erased all flags and crosses
- Used OANS to navigate to gate

Testing Results: 
Passed  

Negatives:       


Conclusions: 


Media:

PR #10388, opened 2025-10-31
feat(a32nx): Implement HW GPWS in rust

lukecologne:
## Summary of Changes

Implement a GPWS (basic mode only, with provisions for EGPWS outputs) simulation in the rust systems.
It utlilizes simulated input peripherals where applicable (RA, ADIRU, LGCIU, SFCC). Any basic GPWS functionality requiring EGPWS functionality (Terrain Clearance Floor function, envelope modulation) is not planned for this PR.
The actual aural message emission is still implemented in the legacy A32NX core, which is now reading an Lvar.
Pin Programming options are partially implemented, like audio declutter deactivation, alternate lamp format (i.e. GPWS/GS vs. PULL UP/GPWS).

- [x] Basic GPWS Modes (1-5)
- [x] Discrete outputs where applicable (provisions for EGPWS outputs, but not planned)
- [x] Bus output (two main warning discrete words added, rest can be added as needed)
- [x] Tests (started)
- [x] Missing Aural Messages @hotshotp
- [ ] Ability to set Pin Programming options in the EFB
- [ ] Momentary audio suppression discrete input functionality (for EMER CANC button)

Note, there is also an "enhanced_gpwc" mod in the systems crate, which is independent of this GPWC, and seems to handle the terrain display output from simbridge in some capacity, but I have not looked at it.

Some remaining issues that annoy me a little but can be solved in another PR:
* Aurals of different warnings can squeeze into each others pauses, so for example mode 2 PULL UP warning could have a mode 4 TOO LOW TERRAIN squeezed in the pauses. I don't think this is how it should work, but could not find any info on this topic.
* Mode 2 radio altimeter closure rate is very finicky, I needed to filter and rate limit it very heavily to compensate for the MSFS typical extremely low update rate for the radio altimeter when the radio altitude is higher than a few hundred feet. This makes the mode 2 warning respond very late to terrain closure. This is also an issue IRL, as the RA is relatively noisy, but it has a constant refresh rate, unlike the MSFS radio altimeter, so it is worse than IRL in this regard.

The current RA filtering setup is as follows: first, the RA is rate limited to 10k fpm, to prevent excessive jumps caused by the aforementioned potentially very low RA update rate. This is an additional step taken which, to my knowledge, is not present in the real GPWC. Then, the derivative is taken, converted to ft/min, then rate limited again to roughly 1.5g, which seems reasonable. Then, it is first-order low-pass filtered with a time constant of 1s. These two steps are also present IRL, but potentially with less restrictive coefficients.

Note: The TERRAIN TERRAIN PULL UP (GPWS basic mode 2) warning might trigger with only one "TERRAIN", as currently the sound has some silence after it that prevents the second TERRAIN from emitting in time. This will be fixed with an update to the sounds

## Additional context

## Testing instructions

- Check that during normal flight, the GPWS does not trigger an alert
- Check that the GPWS is inhibited by RA failures, IR and ADR failures, and that the GPWS SYS FAULT light illuminates
- Check that the GPWS performs a self test at startup (for example via a temporary AC 1 bus fault that turns off the GPWC), associated with GPWS SYS and GPWS TERR FAULT lights
- Check that the basic GPWS modes (1 to 5) work as expected. An overview of the modes can be found online, for example in the EGPWS pilots guide: https://skybrary.aero/sites/default/files/bookshelf/3364.pdf (or of course any other valid source you have)
- Check that the overhead pushbuttons work as expected with the basic modes (G/S mode off, flaps mode off, flaps 3, GPWS SYS off)

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -63,6 +63,8 @@
 1. [FMS] Use localizer station declination for true courses on approach legs - @BlueberryKing (BlueberryKing)
 1. [A380X/FMS] Add FUEL PENALTY function - @BravoMike99 (bruno_pt99)
 1. [A380X/MFD] Update DATA/STATUS page with newer software layout - @BravoMike99 (bruno_pt99)
+1. [A32NX/FMS] Fix pilot entered managed descent mach being ignored - @BlueberryKing (BlueberryKing)
+1. [A32NX/FWS] Add `PARK BRK ON` master caution - @FozzieHi (fozzie)
 1. [A32NX/GPWS] Implement simulated GPWC in rust systems - @lukecologne (luke)
 
 ## 0.14.0
--- a/fbw-a32nx/src/systems/instruments/src/Common/EWDMessages.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/Common/EWDMessages.tsx
@@ -401,6 +401,8 @@ const EWDMessages = {
   '320005001': '\x1b<2m\x1b4mCONFIG\x1bm PARK BRK ON',
   '320007001': '\x1b<4m\x1b4mBRAKES\x1bm A/SKID N/WS OFF',
   '320007002': '\x1b<5m MAX BRK PR......1000 PSI',
+  '320008101': '\x1b<4m\x1b4mBRAKES\x1bm PARK BRK ON',
+  '320008102': '\x1b<5m -PARK BRK...........OFF',
   '320015001': '\x1b<2m\x1b4mL/G\x1bm GEAR NOT DOWN',
   '320015501': '\x1b<2m\x1b4mL/G\x1bm GEAR NOT DOWN',
   '320018001': '\x1b<4m\x1b4mL/G\x1bm LGCIU 1 FAULT',
--- a/fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
+++ b/fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
@@ -4875,7 +4875,7 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
   public getManagedDescentSpeedMach() {
     const plan = this.getFlightPlan(FlightPlanIndex.Active);
 
-    return plan.performanceData.pilotManagedDescentSpeed.get() ?? this.managedSpeedDescendMach;
+    return plan.performanceData.pilotManagedDescentMach.get() ?? this.managedSpeedDescendMach;
   }
 
   // FIXME... ambiguous name that doesn't say if it's Vapp, GSmini, or something else
--- a/fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
+++ b/fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
@@ -855,6 +855,8 @@ export class PseudoFWC {
 
   private readonly flightPhase67 = Subject.create(false);
 
+  private readonly flightPhase678 = Subject.create(false);
+
   private readonly flightPhase78 = Subject.create(false);
 
   private readonly ldgInhibitTimer = new NXLogicConfirmNode(3);
@@ -875,6 +877,8 @@ export class PseudoFWC {
   private readonly toConfigNormalConf = new NXLogicConfirmNode(0.3, false);
 
   private readonly flightPhase3PulseNode = new NXLogicPulseNode();
+
+  private readonly flightPhase7PulseNode = new NXLogicPulseNode();
 
   private readonly flightPhaseEndedPulseNode = new NXLogicPulseNode();
 
@@ -947,6 +951,10 @@ export class PseudoFWC {
   private readonly antiSkidOffPhase2Pulse = new NXLogicPulseNode(true);
 
   private readonly antiSkidOffWarning = Subject.create(false);
+
+  private readonly parkBrakeOnConfirm = new NXLogicConfirmNode(2);
+
+  private readonly parkBrakeOnWarning = Subject.create(false);
 
   private readonly brakeFan = Subject.create(false);
 
@@ -1607,16 +1615,18 @@ export class PseudoFWC {
 
     this.fwcFlightPhase.set(SimVar.GetSimVarValue('L:A32NX_FWC_FLIGHT_PHASE', 'Enum'));
     const flightPhase = this.fwcFlightPhase.get();
-    this.flightPhase3PulseNode.write(this.fwcFlightPhase.get() === 3, deltaTime);
+    this.flightPhase3PulseNode.write(flightPhase === 3, deltaTime);
+    this.flightPhase7PulseNode.write(flightPhase === 7, deltaTime);
     // flight phase convenience vars
-    this.flightPhase126.set([1, 2, 6].includes(this.fwcFlightPhase.get()));
-    this.flightPhase23.set([2, 3].includes(this.fwcFlightPhase.get()));
-    this.flightPhase34.set([3, 4].includes(this.fwcFlightPhase.get()));
-    this.flightPhase345.set(this.flightPhase34.get() || this.fwcFlightPhase.get() === 5);
-    this.flightPhase129.set([1, 2, 9].includes(this.fwcFlightPhase.get()));
-    this.flightPhase67.set([6, 7].includes(this.fwcFlightPhase.get()));
-    this.flightPhase78.set([7, 8].includes(this.fwcFlightPhase.get()));
-    const flightPhase567 = [5, 6, 7].includes(this.fwcFlightPhase.get());
+    this.flightPhase126.set(flightPhase === 1 || flightPhase === 2 || flightPhase === 6);
+    this.flightPhase129.set(flightPhase === 1 || flightPhase === 2 || flightPhase === 9);
+    this.flightPhase23.set(flightPhase === 2 || flightPhase === 3);
+    this.flightPhase34.set(flightPhase === 3 || flightPhase === 4);
+    this.flightPhase345.set(this.flightPhase34.get() || flightPhase === 5);
+    this.flightPhase67.set(flightPhase === 6 || flightPhase === 7);
+    this.flightPhase678.set(this.flightPhase67.get() || flightPhase === 8);
+    this.flightPhase78.set(flightPhase === 7 || flightPhase === 8);
+    const flightPhase567 = flightPhase === 5 || flightPhase === 6 || flightPhase === 7;
 
     // TO Config convenience vars
     const toConfigTest = this.ecpWarningButtonStatus.get().bitValue(18);
@@ -1767,6 +1777,11 @@ export class PseudoFWC {
     const phase2For60Seconds = this.antiSkidOffPhase2Confirm.write(this.fwcFlightPhase.get() === 2, deltaTime);
     const phase2Pulse = this.antiSkidOffPhase2Pulse.write(phase2For60Seconds, deltaTime);
     this.antiSkidOffWarning.set(!antiSkidActive && !acBusOff && !phase2Pulse);
+
+    const parkBrakeFor2Seconds = this.parkBrakeOnConfirm.write(this.parkBrake.get(), deltaTime);
+    this.parkBrakeOnWarning.set(
+      parkBrakeFor2Seconds && this.flightPhase678.get() && !this.flightPhase7PulseNode.read(),
+    );
 
     const leftCompressedHardwireLgciu1 =
       this.dcESSBusPowered.get() && SimVar.GetSimVarValue('L:A32NX_LGCIU_1_LEFT_GEAR_COMPRESSED', 'bool') > 0;
@@ -4477,7 +4492,7 @@ export class PseudoFWC {
       side: 'LEFT',
     },
     3200050: {
-      // PK BRK ON
+      // CONFIG PARK BRK ON
       flightPhaseInhib: [1, 4, 5, 6, 7, 8, 9, 10],
       simVarIsActive: MappedSubject.create(
         ([fwcFlightPhase, parkBrake]) => fwcFlightPhase === 3 && parkBrake,
@@ -4946,6 +4961,17 @@ export class PseudoFWC {
       memoInhibit: () => false,
       failure: 2,
       sysPage: 9,
+      side: 'LEFT',
+    },
+    3200081: {
+      // BRAKES PARK BRK ON
+      flightPhaseInhib: [1, 2, 3, 4, 5, 8, 9, 10],
+      simVarIsActive: this.parkBrakeOnWarning,
+      whichCodeToReturn: () => [0, 1],
+      codesToReturn: ['320008101', '320008102'],
+      memoInhibit: () => false,
+      failure: 2,
+      sysPage: -1,
       side: 'LEFT',
     },
     3200150: {


fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -31,6 +32,95 @@ use uom::si::{
...
+    /// Label 241
+    fn corrected_angle_of_attack(&self) -> Arinc429Word<Angle>;
+}
+
+pub trait InertialReferenceBus {
+    /// Label 052
+    fn pitch_angular_acc(&self) -> Arinc429Word<AngularAcceleration>;
+    /// Label 053
+    fn roll_angular_acc(&self) -> Arinc429Word<AngularAcceleration>;
+    /// Label 053
```
Gurgel100: I think this should be a different label?

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -1395,10 +1485,79 @@ impl AirDataReference {
...
+impl AirDataReferenceBus for AirDataReference {
+    fn standard_altitude(&self) -> Arinc429Word<Length> {
+        self.altitude()
+    }
+
+    fn baro_corrected_altitude_1(&self) -> Arinc429Word<Length> {
+        Arinc429Word::new(
+            self.baro_corrected_altitude_1.value(),
+            self.baro_corrected_altitude_1.ssm(),
+        )
```
Gurgel100, line 1495:
```suggestion
        self.baro_corrected_altitude_1
```
Same for the other functions.

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -2244,20 +2403,139 @@ impl InertialReference {
...
         Arinc429Word::new(
             Velocity::new::<foot_per_minute>(self.vertical_speed.value()),
             self.vertical_speed.ssm(),
         )
     }
+}
+impl InertialReferenceBus for InertialReference {
+    fn pitch_angular_acc(&self) -> Arinc429Word<AngularAcceleration> {
+        Arinc429Word::new(
+            AngularAcceleration::new::<degree_per_second_squared>(0.),
```
Gurgel100:
```suggestion
            AngularAcceleration::ZERO,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -2244,20 +2403,139 @@ impl InertialReference {
...
+            SignStatus::NoComputedData,
+        )
+    }
+
+    fn roll_angular_acc(&self) -> Arinc429Word<AngularAcceleration> {
+        Arinc429Word::new(
+            AngularAcceleration::new::<degree_per_second_squared>(0.),
```
Gurgel100:
```suggestion
            AngularAcceleration::ZERO,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -2244,20 +2403,139 @@ impl InertialReference {
...
+            SignStatus::NoComputedData,
+        )
+    }
+
+    fn yaw_angular_acc(&self) -> Arinc429Word<AngularAcceleration> {
+        Arinc429Word::new(
+            AngularAcceleration::new::<degree_per_second_squared>(0.),
```
Gurgel100:
```suggestion
            AngularAcceleration::ZERO,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -2244,20 +2403,139 @@ impl InertialReference {
...
+    fn ppos_longitude(&self) -> Arinc429Word<Angle> {
+        self.longitude()
+    }
 
     fn ground_speed(&self) -> Arinc429Word<Velocity> {
         Arinc429Word::new(self.ground_speed.value(), self.ground_speed.ssm())
     }
+
+    fn true_heading(&self) -> Arinc429Word<Angle> {
+        Arinc429Word::new(self.true_heading.value(), self.true_heading.ssm())
```
Gurgel100, line 2437: Also here the same as above and the same for the other functions.

```suggestion
        self.true_heading
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
```diff
@@ -2244,20 +2403,139 @@ impl InertialReference {
...
+    fn body_lat_acc(&self) -> Arinc429Word<Ratio> {
+        Arinc429Word::new(self.body_lateral_acc.value(), self.body_lateral_acc.ssm())
+    }
+
+    fn body_normal_acc(&self) -> Arinc429Word<Ratio> {
+        Arinc429Word::new(self.body_normal_acc.value(), self.body_normal_acc.ssm())
+    }
+
+    fn inertial_altitude(&self) -> Arinc429Word<Length> {
+        Arinc429Word::new(Length::new::<foot>(0.), SignStatus::NoComputedData)
```
Gurgel100:
```suggestion
        Arinc429Word::new(Length::ZERO, SignStatus::NoComputedData)
```

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    fn normalize_180(angle: Angle) -> Angle {
+        let mut normalized_360 = Self::normalize_360(angle);
+
+        if normalized_360 >= Angle::new::<degree>(180.0) {
+            normalized_360 -= Angle::new::<degree>(360.0);
+        }
+
+        normalized_360
```
Gurgel100:
```suggestion
        let normalized_360 = Self::normalize_360(angle);

        if normalized_360 >= Angle::new::<degree>(180.0) {
            normalized_360 - Angle::new::<degree>(360.0)
        } else {
            normalized_360
        }
```
This removes the mutable from the variable.

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    fn normalize_360(angle: Angle) -> Angle {
+        let angle = angle % Angle::new::<degree>(360.0);
+
+        (angle + Angle::new::<degree>(360.0)) % Angle::new::<degree>(360.0)
```
Gurgel100, line 98: I don't quite understand the math behind this. I my head this is equivalent to `angle` since:
```
(angle + 360) % 360 == (angle % 360) + (360 % 360) == (angle % 360) + 0 == angle % 360 == angle (because angle is already modulo 360)
```

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
```diff
@@ -0,0 +1,157 @@
...
+                SignStatus::NormalOperation
+            } else {
+                SignStatus::NoComputedData
+            },
+        )
+    }
+    fn localizer_deviation(&self) -> Arinc429Word<Ratio> {
+        Arinc429Word::new(
+            self.get_corrected_localizer_deviation() / Angle::new::<degree>(0.8)
+                * Ratio::new::<ratio>(0.0775),
```
Gurgel100:
```suggestion
            self.get_corrected_localizer_deviation() / Angle::new::<degree>(0.8)
                * 0.0775,
```
Since the division already returns a `Ratio` you can just "scale" the ratio by `0.0775` and you don't have to create a `Ratio` instance. A small optimization would be to put parathesis around the two constants so the compiler can get rid of a multiplication.

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
```diff
@@ -0,0 +1,157 @@
...
+            if self.loc_valid {
+                SignStatus::NormalOperation
+            } else {
+                SignStatus::NoComputedData
+            },
+        )
+    }
+    fn glideslope_deviation(&self) -> Arinc429Word<Ratio> {
+        Arinc429Word::new(
+            self.glideslope_deviation / Angle::new::<degree>(0.4) * Ratio::new::<ratio>(0.0875),
```
Gurgel100: Same argument as above.

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// 0.175DDM is full scale deflection (two dots) => one dot = 0.0875DDM
+    /// One dot = 0.4° on the glideslope scale according to FCOM
+    fn glideslope_deviation(&self) -> Arinc429Word<Ratio>;
+    /// Label 263
+    fn ground_station_ident_1(&self) -> Arinc429Word<u32>;
+    /// Label 264
+    fn ground_station_ident_2(&self) -> Arinc429Word<u32>;
+}
+
+pub struct MultiModeReceiverShim {
```
Gurgel100, line 31: Maybe we can add some documentation stating that this struct is simply representing the MMR which is implemented outside of Rust.

fbw-common/src/wasm/systems/systems/src/shared/derivative.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl<T> DerivativeNode<T>
+where
+    T: Sub<Output = T> + Div<f64, Output = T> + Copy + Default,
+{
+    pub fn new() -> Self {
+        Self {
+            predecessor: None,
+            output: T::default(),
+        }
```
Gurgel100:
```suggestion
        Self::default()
```

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+/// signal will be emitted again.
+pub struct ConfirmationNode {
+    rising_edge: bool,
+    time_delay: Duration,
+    condition_since: Duration,
+    output: bool,
+}
+
+impl ConfirmationNode {
+    pub fn new(rising_edge: bool, time_delay: Duration) -> Self {
```
Gurgel100, line 21:
```suggestion
    fn new(rising_edge: bool, time_delay: Duration) -> Self {
```
Maybe only `new_rising` and `new_falling` should be the public "constructors"?

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    pub fn new_falling(time_delay: Duration) -> Self {
+        Self::new(false, time_delay)
+    }
+
+    pub fn update(&mut self, hi: bool, delta: Duration) -> bool {
+        self.remaining_trigger = self
+            .remaining_trigger
+            .checked_sub(delta)
+            .unwrap_or_default();
```
Gurgel100:
```suggestion
        self.remaining_trigger = self
            .remaining_trigger
            .saturating_sub(delta);
```

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+/// A node that detects a rising or a falling edge and will trigger exactly once. Similar to a
+/// monostable trigger node, except that the signal will immediately return, and T approaches 0.
+pub struct PulseNode {
+    rising_edge: bool,
+    last_hi: Option<bool>,
+    output: bool,
+}
+
+impl PulseNode {
+    pub fn new(rising_edge: bool) -> Self {
```
Gurgel100, line 129:
```suggestion
    fn new(rising_edge: bool) -> Self {
```
Also here: Maybe only `new_rising` and `new_falling` should be the public "constructors"?

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+}
+
+/// A flip-flop or memory circuit that can be used to store a single bit. It has two inputs: Set and
+/// Reset. At first it will always emit a falsy value, until it receives a signal on the set input,
+/// at which point it will start emitting a truthy value. This will continue until a signal is
+/// received on the reset input, at which point it reverts to the original falsy output. If a signal
+/// is received on both set and reset at the same time, the input with a star will have precedence.
+pub struct MemoryNode {
+    has_set_precedence: bool,
+    pub output: bool,
```
Gurgel100:
```suggestion
    output: bool,
```
Instead we should have a function `output` with which one can retrieve the current output.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+    up: T,
+    dn: T,
+    output: bool,
+}
+
+impl<T> HysteresisNode<T>
+where
+    T: PartialOrd,
+{
+    pub fn new(dn: T, up: T) -> Self {
```
Gurgel100: Maybe `lower` instead of `dn` and `upper` instead of `up` are more descriptive names?

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+            if value <= self.dn {
+                false
+            } else {
+                self.output
+            }
+        } else if value >= self.up {
+            true
+        } else {
+            self.output
+        };
```
Gurgel100:
```suggestion
        self.output = if self.output && value <= self.dn {
            false
        } else if value >= self.up {
            true
        } else {
            self.output
        };
```
I think this is equivalent and easier to read.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub fn update(&mut self, value: bool) {
+        self.predecessor = value;
+    }
+}
+
+/// A node that compares the derivative of a value over time to a threshold and returns true if the
+/// derivative is strictly larger. The threshold is interpreted as change per 1 second.
+/// The derivative is calculated over two subsequent .update calls.
+#[derive(Default)]
+pub struct DerivativeThresholdNode {
```
Gurgel100, line 254: Can this node not be implemented with the `DerivativeNode` from `derivative.rs`? Would reduce code duplication.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    pub fn output(&self) -> bool {
+        self.output
+    }
+}
+
+/// A node that detects a rising or a falling edge and will trigger exactly once. Similar to a
+/// monostable trigger node, except that the signal will immediately return, and T approaches 0.
+pub struct PulseNode {
```
Gurgel100, line 122: This sounds very similar to the already existing `DelayedPulseTrueLogicGate`. So I suggest to either reimplement `DelayedPulseTrueLogicGate` using this node or replace it fully.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+
+// Logic nodes, taken from https://github.com/flybywiresim/aircraft/pull/4872, and adapted Confirmation node to better adhere to sensible
+// initial condition behaviour.
+
+/// A confirmation circuit, which only passes a signal once it has been stable for a certain amount
+/// of time. When it detects either a rising or falling edge (depending on its type) it will wait
+/// for a time delay period and emit the incoming signal if it was stable throughout the period.
+/// If at any point during the period the signal reverts, the state is fully reset and the original
+/// signal will be emitted again.
+pub struct ConfirmationNode {
```
Gurgel100, line 13: This sounds very similar to the already existing `DelayedTrueLogicGate` and `DelayedFalseLogicGate`. I suggest to either reimplement them using this node or replacing them fully.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -0,0 +1,624 @@
...
+        self.output
+    }
+}
+
+/// A flip-flop or memory circuit that can be used to store a single bit. It has two inputs: Set and
+/// Reset. At first it will always emit a falsy value, until it receives a signal on the set input,
+/// at which point it will start emitting a truthy value. This will continue until a signal is
+/// received on the reset input, at which point it reverts to the original falsy output. If a signal
+/// is received on both set and reset at the same time, the input with a star will have precedence.
+pub struct MemoryNode {
```
Gurgel100, line 163: This sounds very similar to the already existing `LatchedTrueLogicGate`.

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
Gurgel100, line 1: Typo in the file name.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    alert_light_on_id: VariableIdentifier,
+
+    aural_output_id: VariableIdentifier,
+}
+
+impl EnhancedGroundProximityWarningComputer {
+    const MINIMUM_STARTUP_TIME_MILLIS: u64 = 18_000;
+    const MAXIMUM_STARTUP_TIME_MILLIS: u64 = 20_000;
+    const MINIMUM_POWER_HOLDOVER: u64 = 200;
+    const MAXIMUM_POWER_HOLDOVER: u64 = 300;
```
Gurgel100:
```suggestion
    const MINIMUM_STARTUP_TIME_MILLIS: Duration = Duration::from_millis(18_000);
    const MAXIMUM_STARTUP_TIME_MILLIS: Duration = Duration::from_millis(20_000);
    const MINIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(200);
    const MAXIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(300);
```
**Reasoning**: Gives documentation of what the unit is exactly.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    pub fn new(context: &mut InitContext, powered_by: ElectricalBusType) -> Self {
+        let is_powered = context.has_engines_running();
+        let on_ground = context.is_on_ground();
+        Self {
+            powered_by,
+            is_powered: false,
+            power_holdover: Duration::from_secs_f64(random_from_range(
+                Self::MINIMUM_POWER_HOLDOVER as f64 / 1000.,
+                Self::MAXIMUM_POWER_HOLDOVER as f64 / 1000.,
```
Gurgel100:
```suggestion
                Self::MINIMUM_POWER_HOLDOVER.as_secs_f64(),
                Self::MAXIMUM_POWER_HOLDOVER.as_secs_f64(),
```

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            )),
+            unpowered_for: if is_powered {
+                Duration::ZERO
+            } else {
+                Duration::from_millis(Self::MAXIMUM_POWER_HOLDOVER)
+            },
+            self_check_time: Duration::from_secs_f64(random_from_range(
+                Self::MINIMUM_STARTUP_TIME_MILLIS as f64 / 1000.,
+                Self::MAXIMUM_STARTUP_TIME_MILLIS as f64 / 1000.,
```
Gurgel100:
```suggestion
                Self::MINIMUM_STARTUP_TIME_MILLIS.as_secs_f64(),
                Self::MAXIMUM_STARTUP_TIME_MILLIS.as_secs_f64(),
```

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    pub fn update(&mut self, time_delta: Duration, new_input: T) -> T {
+        if self.limited_output.is_none() {
+            self.limited_output = Some(new_input);
+            return new_input;
+        }
+        // We need to do this stupid shit because rust doesnt implement Ord for floats.
+        // If anyone has a better way to solve this, please let me know.
```
Gurgel100: One way would be to extend the `systems::shared::Clamp` trait so it is implemented by all types which also implement `Ord` and then a manual implementation for `f32` and/or ` f64`.

fbw-common/src/wasm/systems/systems/src/surveillance/taws.rs
```diff
@@ -0,0 +1,50 @@
...
+    // Label 270
+    pub alert_discrete_1: Arinc429Word<u32>,
+    // Label 274
+    pub alert_discrete_2: Arinc429Word<u32>,
+    // Label
+}
+
+pub trait TerrainAwarenessWarningSystemBusOutput {
+    fn bus_outputs(&self) -> &TerrainAwarenessWarningSystemBusOutputs;
+}
```
Gurgel100, line 33: Why not use the same system as for the ADIRUs (have a trait for the gpws bus)?

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/electrical_harness.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub(super) const GPWS_TEST_KEY: &str = "GPWS_TEST";
+    pub(super) const SFCC_1_FAP_1_KEY: &str = "SFCC_1_FAP_1";
+    pub(super) const SFCC_1_FAP_5_KEY: &str = "SFCC_1_FAP_5";
+    pub(super) const ECP_DISCRETE_OUT_EMER_CANC_KEY: &str = "ECP_DISCRETE_OUT_EMER_CANC";
+    pub(super) const IS_SLEW_ACTIVE_KEY: &str = "IS SLEW ACTIVE";
+
+    pub fn new(context: &mut InitContext) -> Self {
+        Self {
+            discrete_inputs: TerrainAwarenessWarningSystemDiscreteInputs::default(),
+            pin_programs: EnhancedGroundProximityWarningComputerPinProgramming::default(),
```
Gurgel100: Just a note that we have in the guidelines that we should [avoid dependencies between child modules](https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/wasm/systems/guidelines.md#avoid-dependencies-between-child-modules).

fbw-a32nx/src/wasm/systems/a320_systems/src/surveillance.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub(super) const GPWS_TERR_OFF_KEY: &str = "GPWS_TERR_OFF";
+    pub(super) const GPWS_SYS_OFF_KEY: &str = "GPWS_SYS_OFF";
+    pub(super) const GPWS_GS_OFF_KEY: &str = "GPWS_GS_OFF";
+    pub(super) const GPWS_FLAP_OFF_KEY: &str = "GPWS_FLAP_OFF";
+    pub(super) const GPWS_FLAPS3_KEY: &str = "GPWS_FLAPS3";
+    pub(super) const GPWS_TEST_KEY: &str = "GPWS_TEST";
+    pub(super) const SFCC_1_FAP_1_KEY: &str = "SFCC_1_FAP_1";
+    pub(super) const SFCC_1_FAP_5_KEY: &str = "SFCC_1_FAP_5";
+    pub(super) const ECP_DISCRETE_OUT_EMER_CANC_KEY: &str = "ECP_DISCRETE_OUT_EMER_CANC";
+    pub(super) const IS_SLEW_ACTIVE_KEY: &str = "IS SLEW ACTIVE";
```
Gurgel100:
```suggestion
    const GPWS_TERR_OFF_KEY: &str = "GPWS_TERR_OFF";
    const GPWS_SYS_OFF_KEY: &str = "GPWS_SYS_OFF";
    const GPWS_GS_OFF_KEY: &str = "GPWS_GS_OFF";
    const GPWS_FLAP_OFF_KEY: &str = "GPWS_FLAP_OFF";
    const GPWS_FLAPS3_KEY: &str = "GPWS_FLAPS3";
    const GPWS_TEST_KEY: &str = "GPWS_TEST";
    const SFCC_1_FAP_1_KEY: &str = "SFCC_1_FAP_1";
    const SFCC_1_FAP_5_KEY: &str = "SFCC_1_FAP_5";
    const ECP_DISCRETE_OUT_EMER_CANC_KEY: &str = "ECP_DISCRETE_OUT_EMER_CANC";
    const IS_SLEW_ACTIVE_KEY: &str = "IS SLEW ACTIVE";
```
Don't need to be public.

fbw-a32nx/src/wasm/systems/a320_systems/src/surveillance.rs
```diff
@@ -0,0 +1,109 @@
...
+            true
+        } else if self.landing_flaps_3 {
+            // TODO the SFCC discrete outputs are not exposed directly, and I can't be bothered to implement this right now
+            reader.read(&self.sfcc_fap_5_id)
+        } else {
+            reader.read(&self.sfcc_fap_1_id)
+        };
+        self.discrete_inputs.gs_cancel = reader.read(&self.gs_cancel_self_test_id);
+        self.discrete_inputs.terrain_awareness_inhibit = reader.read(&self.terr_off_id);
+        self.discrete_inputs.sim_reposition_active = reader.read(&self.slew_active_id);
```
Gurgel100, line 100: Strictly speaking this is not conformant to the [guidelines](https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/wasm/systems/guidelines.md#law-of-demeter). But I think in this case it is ok.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        ra2: &impl RadioAltimeter,
+        adr: &impl AirDataReferenceBus,
+        ir: &impl InertialReferenceBus,
+        ils: &impl InstrumentLandingSystemBus,
+    ) {
+        if self.is_powered {
+            self.unpowered_for = Duration::ZERO;
+        } else {
+            self.unpowered_for += context.delta();
+        }
```
Gurgel100, line 155: The power check can be implemented using the `DelayedFalseLogicGate` or the `ConfirmationNode`.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        // Check if the internal state (runtime) is lost because either the unit failed, or the
+        // power holdover has been exceed
+        if self.failure.is_active() || self.unpowered_for > self.power_holdover {
+            // Throw away the simulated software runtime, but first save to NVM
+            if let Some(runtime) = self.runtime.as_ref() {
+                self.on_ground = runtime.get_on_ground();
+                self.flight_phase = runtime.get_flight_phase();
+            }
+            self.runtime = None;
```
Gurgel100:
```suggestion
            if let Some(runtime) = self.runtime.take() {
                self.on_ground = runtime.get_on_ground();
                self.flight_phase = runtime.get_flight_phase();
            }
```

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            let runtime = self.runtime.get_or_insert_with(|| {
+                EnhancedGroundProximityWarningComputerRuntime::new(
+                    self_check,
+                    self.on_ground,
+                    self.flight_phase,
+                )
+            });
+            runtime.update(
+                context,
+                electrical_harness.discrete_inputs(),
```
Gurgel100, line 188: This is against the [guidelines](https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/wasm/systems/guidelines.md#passing-data-around).

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            self.discrete_output_data.terrain_inop = true;
+            self.discrete_output_data.terrain_not_available = true;
+            self.bus_output_data = TerrainAwarenessWarningSystemBusOutputs::default();
+            return;
+        }
+
+        // As long as we're powered, we can proceed normally. If not, we can't run the runtime, but
+        // it's state will be frozen and if power is restored soon enough, we can proceed
+        // immediately without waiting for the runtime to start up again.
+        if self.is_powered {
```
Gurgel100:
```suggestion
        }

        // As long as we're powered, we can proceed normally. If not, we can't run the runtime, but
        // it's state will be frozen and if power is restored soon enough, we can proceed
        // immediately without waiting for the runtime to start up again.
        else if self.is_powered {
```
Another possibility. I'm not sure which one is better though.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            // Either initialize and run or continue running the existing runtime
+            let self_check = self.self_check_time;
+            let runtime = self.runtime.get_or_insert_with(|| {
+                EnhancedGroundProximityWarningComputerRuntime::new(
+                    self_check,
+                    self.on_ground,
+                    self.flight_phase,
+                )
+            });
```
Gurgel100:
```suggestion
            let runtime = self.runtime.get_or_insert_with(|| {
                EnhancedGroundProximityWarningComputerRuntime::new(
                    self.self_check_time,
                    self.on_ground,
                    self.flight_phase,
                )
            });
```

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            runtime.update(
+                context,
+                electrical_harness.discrete_inputs(),
+                ra1,
+                ra2,
+                adr,
+                ir,
+                ils,
+            );
+            runtime.set_outputs(&mut self.discrete_output_data, &mut self.bus_output_data);
```
Gurgel100, line 195: This is also against the [guidelines](https://github.com/flybywiresim/aircraft/blob/master/fbw-common/src/wasm/systems/guidelines.md#passing-data-around).

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        ir: &impl InertialReferenceBus,
+        ils: &impl InstrumentLandingSystemBus,
+    ) {
+        // First, check if we're still starting up and if so, simulate a wait until all self tests
+        // have completed.
+        if let Some(new_remaining) = self.remaining_startup.checked_sub(context.delta()) {
+            self.remaining_startup = new_remaining;
+        } else {
+            self.remaining_startup = Duration::ZERO;
+        }
```
Gurgel100:
```suggestion
        self.remaining_startup  = self.remaining_startup.saturating_sub(context.delta());
```

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+
+    fn update_general_logic(
+        &mut self,
+        context: &UpdateContext,
+        adr: &impl AirDataReferenceBus,
+        ir: &impl InertialReferenceBus,
+    ) {
+        // General mode logics
+        let ground_to_air_ra_speed_condition =
+            self.ra_ft >= 25. && adr.computed_airspeed().value() >= Velocity::new::<knot>(90.);
```
Gurgel100, line 349: This doesn't check the validity of the ARINC429 value.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        self.ground_to_air_confirm_node
+            .update(ground_to_air_ra_speed_condition, context.delta());
+        let ground_to_air_pitch_condition =
+            ir.pitch_angle().value_or_default() >= Angle::new::<degree>(5.);
+        let air_to_ground_condition =
+            self.ra_ft < 25. && adr.computed_airspeed().value() < Velocity::new::<knot>(60.);
```
Gurgel100, line 355: This doesn't check the validity of the ARINC429 value.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        // println!(
+        //     "EGPWS General Logic: on_ground: {}, flight_phase: {:?}, takeoff_to_approach_integrator: {:.1}, field_elev: {:.1} ft
+        //     mode_4_c_filter: {:.1}, mode_4_a_upper_boundary: {:.1}",
+        //     self.on_ground,
+        //     self.flight_phase,
+        //     self.takeoff_to_approach_condition_integrator,
+        //     self.field_elevation_ft,
+        //     self.mode_4_c_filter_value,
+        //     self.mode_4_a_alert_upper_boundary,
+        // );
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+
+        // RA computation (TODO implement reasonableness check and maybe filtering)
+        let ra_fault =
+            ra1.radio_altitude().is_failure_warning() && ra2.radio_altitude().is_failure_warning();
+        if !ra1.radio_altitude().is_failure_warning() && !ra2.radio_altitude().is_failure_warning()
+        {
+            let ra1_ft = ra1.radio_altitude().value().get::<foot>();
+            let ra2_ft = ra2.radio_altitude().value().get::<foot>();
+
+            self.ra_ft = if ra1_ft < 2500. || ra2_ft < 2500. && (ra1_ft - ra2_ft).abs() < 500. {
```
Gurgel100: Please check if there are missing parathesis.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        // Finally, choose the ADR barometric Vertical speed.
+        self.vs_fault = false;
+        self.chosen_vertical_speed_ft_min =
+            if let Some(vs) = ir.inertial_vertical_speed().normal_value() {
+                vs
+            } else if let Some(vs) = adr.vertical_speed().normal_value() {
+                vs
+            } else {
+                self.vs_fault = true;
+                adr.vertical_speed().value_or_default()
```
Gurgel100: Could be replaced with `Default::default()`.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        //     "Mode 1: ra: {:.1} ft, biased_vs: {:.1} ft/min, alert_boundary_met: {}, warning_boundary_met: {},
+        //     sinkrate_lamp: {}, sinkrate_voice: {}, pull_up_active: {}",
+        //     self.ra_ft,
+        //     biased_vertical_speed,
+        //     mode_1_alert_boundary_met,
+        //     mode_1_warning_boundary_met,
+        //     self.mode_1_sinkrate_lamp_active,
+        //     self.mode_1_sinkrate_voice_active,
+        //     self.mode_1_pull_up_active,
+        // );
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        //    filtered_closure_rate_ft,
+        //    closure_rate_ft_per_min_raw,
+        //    mode_2_lower_boundary_ft,
+        //    mode_2_upper_boundary_ft,
+        //    mode_2_boundary_met,
+        //    self.mode_2_pull_up_preface_voice_emitted,
+        //    self.mode_2_pull_up_preface_active,
+        //    self.mode_2_pull_up_active,
+        //    self.mode_2_terrain_active,
+        //);
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        //     "Mode 3: phase: {:?}, enabled: {}, max achieved alt: {:.1} ft, altitude loss: {:.1} ft, mode 3 boundary: {:.1} ft, lamp active: {}, voice active: {}, declutter increase: {:.2}",
+        //     self.flight_phase,
+        //     mode_3_enabled,
+        //     self.mode_3_max_achieved_alt_ft.unwrap_or_default(),
+        //     altitude_loss_ft.unwrap_or_default(),
+        //     mode_3_boundary,
+        //     self.mode_3_lamp_active,
+        //     self.mode_3_dont_sink_voice_active,
+        //     self.mode_3_declutter_threshold_increase,
+        // );
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        // This is latched and only reverted when on ground or in takeoff phase (i.e. go-around).
+        if !self.on_ground
+            && self.flight_phase == FlightPhase::Approach
+            && (discrete_inputs.landing_gear_downlocked
+                || (discrete_inputs.landing_flaps && mode_4_b_alternate))
+        {
+            self.mode_4_b_active = true;
+        } else if self.on_ground || self.flight_phase == FlightPhase::Takeoff {
+            self.mode_4_b_active = false;
+        }
```
Gurgel100, line 786:
```suggestion
        self.mode_4_b_active = if !self.on_ground
            && self.flight_phase == FlightPhase::Approach
            && (discrete_inputs.landing_gear_downlocked
                || (discrete_inputs.landing_flaps && mode_4_b_alternate))
        {
            true
        } else if self.on_ground || self.flight_phase == FlightPhase::Takeoff {
            false
        };
```

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        //     self.flight_phase == FlightPhase::Approach,
+        //     mode_4_a_boundary,
+        //     mode_4_b_boundary,
+        //     mode_4_ab_lamp_active,
+        //     self.mode_4_too_low_gear_voice_active,
+        //     self.mode_4_too_low_flaps_voice_active,
+        //     mode_4_ab_too_low_terrain_voice_active,
+        //     self.mode_4_ab_declutter_threshold_increase,
+        //     self.number_of_aural_warning_emissions
+        // );
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        //println!("Mode 4: ra: {:.1} ft, filter: {:.1} ft, upper boundary: {:.1} ft, biased ra: {:.1} ft, enabled: {},
+        // lamp active: {}, too low terrain voice active: {}",
+        //         self.ra_ft,
+        //         self.mode_4_c_filter_value,
+        //         self.mode_4_a_alert_upper_boundary,
+        //         biased_ra_ft,
+        //         mode_4_c_enabled,
+        //         mode_4_c_lamp_active,
+        //         mode_4_c_too_low_terrain_voice_active,
+        //     );
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        discrete_inputs: &TerrainAwarenessWarningSystemDiscreteInputs,
+    ) {
+        // Not sure if this is only when a G/S warning is active, or always
+        if discrete_inputs.gs_cancel && self.ra_ft < 2000. {
+            self.mode_5_cancel_active = true;
+        } else if self.ra_ft >= 2000. || self.ra_ft < 30. {
+            self.mode_5_cancel_active = false;
+        }
+
+        // 0.0875 is 1 dot
```
Gurgel100: This comment is incomplete since it is only for the gs deviation but not the loc deviation. Both of them are used after this comment though.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        let loc_deviation_dots = ils.localizer_deviation().value().get::<ratio>() / 0.0775;
+        let gs_deviation_dots_fly_up = -ils.glideslope_deviation().value().get::<ratio>() / 0.0875;
+
+        let fls_selected = false; // TODO implement
+        let mix_loc_fls_selected = false; // TODO implement
+        let envelope_mod_gear_down_override = false; // TODO implement
+        let heading_difference = (ir.magnetic_track().value().get::<degree>()
+            - ils.runway_heading().value().get::<degree>()
+            + 540.)
```
Gurgel100, line 921:
```suggestion
            + 180.)
```
This should be equivalent since 540 % 360 = 180. And it is more clear what is done.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+            //    mode_5_active,
+            //    mode_5_soft_alert_light_boundary_met,
+            //    mode_5_soft_alert_aural_boundary_met,
+            //    mode_5_hard_alert_boundary_met,
+            //    mode_5_soft_alert_upper_boundary_ft,
+            //    self.mode_5_glideslope_soft_voice_active,
+            //    self.mode_5_glideslope_hard_voice_active,
+            //    self.mode_5_declutter_threshold_increase,
+            //    self.mode_5_time_to_next_aural.as_secs_f64(),
+            //);
```
Gurgel100: Remove commented code.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+            self.aural_output = AuralWarning::SinkRate;
+        } else if self.mode_3_dont_sink_voice_active && !basic_gpws_inhibit {
+            self.aural_output = AuralWarning::DontSink;
+        } else if self.mode_5_glideslope_soft_voice_active && !basic_gpws_inhibit {
+            self.aural_output = AuralWarning::GlideslopeSoft;
+        } else if self.mode_5_glideslope_hard_voice_active && !basic_gpws_inhibit {
+            self.aural_output = AuralWarning::GlideslopeHard;
+        } else {
+            self.aural_output = AuralWarning::None;
+        }
```
Gurgel100: What about doing an initial if statement which checks if `basic_gpws_inhibit` is set and if so just returns `AuralWarning::None`? Then you don't have to check it in each following `else if` statement.
Also do `self.aural_output = if ...`.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
```diff
@@ -0,0 +1,1286 @@
...
+        // First, check if we're still starting up and if so, simulate a wait until all self tests
+        // have completed.
+        if let Some(new_remaining) = self.remaining_startup.checked_sub(context.delta()) {
+            self.remaining_startup = new_remaining;
+        } else {
+            self.remaining_startup = Duration::ZERO;
+        }
+
+        // If there's any startup time remaining, do nothing
+        if self.remaining_startup > Duration::ZERO {
```
Gurgel100: How about using the helper function?

```suggestion
        if !self.is_initialized() {
```

fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -379,6 +383,16 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
     .provides_aircraft_variable("PAYLOAD STATION WEIGHT", "Pounds", 6)?
     .provides_aircraft_variable("PAYLOAD STATION WEIGHT", "Pounds", 7)?
     .provides_aircraft_variable("PAYLOAD STATION WEIGHT", "Pounds", 8)?
+    .provides_aircraft_variable("IS SLEW ACTIVE", "Bool", 0)?
+    .provides_aircraft_variable("NAV HAS NAV", "Bool", 3)?
+    .provides_aircraft_variable("NAV HAS LOC", "Bool", 3)?
+    .provides_aircraft_variable("NAV HAS GLIDE SLOPE", "Bool", 3)?
+    //.provides_aircraft_variable("NAV IDENT", "String", 3)?
```
Gurgel100: Remove commented code.

fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -379,6 +383,16 @@ async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
...
+    .provides_aircraft_variable("NAV MAGVAR", "degree", 3)?
+    .provides_aircraft_variable("NAV RADIAL ERROR", "degree", 3)?
+    .provides_aircraft_variable("NAV GLIDE SLOPE ERROR", "degree", 3)?
+    .provides_aircraft_variable("NAV FREQUENCY", "Hz", 3)?
+    .provides_named_variable("A32NX_FM_LS_COURSE")?
```
Gurgel100:
```suggestion
```
Not needed since it's a local simvar prefixed with `A32NX_`.

fbw-a32nx/src/wasm/systems/a320_systems/src/lib.rs
Gurgel100, line 1: So we do have now two different EGPWS? Does it make sense to merge them or remove the old one?

Gurgel100 (commented):
Uff. That's a big one. Nice work. I just have some suggestions/comments/questions.

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
lukecologne: I think the reason that your reasoning does not work, is that the modulo operation is not distributive. This function normalizes an angle from [-180, 180] to [0, 360]

fbw-a32nx/src/wasm/systems/a320_systems/src/lib.rs
lukecologne: Yes, the old one handles some logic of the terrain display on the ND, but I am not too familiar with it. It should be merged into this new EGPWS at some point, but I did not do it in this PR as it's obviously already quite large. It's probably good either in it's own PR, or along with an implementation of the predictive terrain functions

fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
lukecologne: Ah yes, I forgot about this. I added it for the MMR tuned identifier labels, but IIRC the wasm crashes when trying to read a string valued simvar

fbw-a32nx/src/wasm/systems/a320_systems/src/surveillance.rs
lukecologne: They are used in some of the tests, so the pub(super) is indeed needed

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: It's not equivalent, consider the case trk = 0, rwy_hdg > 180 (let's say 200). Then:
```
(trk - rwy + 540.) % 360. - 180. == 160
(trk - rwy + 180.) % 360. - 180. == -200
```
So it must be +540 to get the result we want

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: This does not work, as then it could have return type `()` if both `if`'s are false. In any case, this would not result in the kind of flip-flop behaviour we want

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: Oof, that would have been a nasty bug, thank you!

lukecologne:
Thank you for the extensive review! I have already fixed some comments, I'll address the rest tomorrow.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: The thing is that the enhanced/predictive GPWS functions have their own failure conditions, which I don't think necessarily coincide with the basic GPWS failure, and their aural warning priorities are inbetween the basic warnings, so checking `basic_gpws_inhibit` beforehand wouldn't work then. This is why I kept it explicit for future implementation.
`self.aural_output = if ...` is done though.

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
lukecologne: `self.baro_corrected_altitude_1` has type `AdirsArinc429Data`, I would have to implement `Into Arinc429Word`, which, considering a rework of the ADIRU is also underway, I think is not necessary

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: That's true, however as the whole GPWS basic mode is invalidated if CAS is detected invalid (FW), it should be correct IMO.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: See above

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: I have implemented a different simplicfication that simply doesn't set `self.output` if value is between the two.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: Hm, I could see a situation where you maybe want to conditionally construct either rising or falling nodes, in that case I think this constructor is a bit nicer, so I think it's fine to leave as-is.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: I think the `DelayedPulseTrueLogicGate` is more like a confirm node and a pulse node in series. This is just the rising/falling edge detection alone.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: Indeed they are basically the same, the difference is this combines both into one. Ideally I think these nodes here should replace the other ones, as they are closer to the normal definition we use, but as the `Delayedx` gates are widely used, I'd like to not replace them here if that is ok.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: The same as above with the confirm node, although this does differ in functionality somewhat.

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
lukecologne: Hm, so replace `T: PartialOrd` with `T: Clamp` and implement Clamp like this:
```rs
impl<T: PartialOrd + std::fmt::Debug> Clamp for T {
    fn clamp(mut self, min: Self, max: Self) -> Self {
        assert!(
            min <= max,
            "min > max, or either was NaN. min = {min:?}, max = {max:?}"
        );
        if self < min {
            self = min;
        }
        if self > max {
            self = max;
        }
        self
    }
}
```
It already works with floats as they do implement PartialOrd, I'm not sure why the rust standard lib binds clamp to Ord and not PartialOrd. But it does feel a bit weird to implement this for every struct that implements PartialOrd, would this not cause conflicts for those who also implement Ord?

fbw-common/src/wasm/systems/systems/src/surveillance/taws.rs
lukecologne: I prefer this type of implementation, as the struct needs to be defined anyway for interface between the computer and the runtime, and to hold the bus values. Then implementing every struct field as a trait fn just duplicates everything two more times IMO. I did it for the ADIRS since it has each bus label directly in the "runtime" struct, which unnecessarily increases clutter and also is not feasable for the Computer<->Runtime model and interface. I'm open to suggestions though.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
lukecologne: How should it be done instead? Having `&mut EnhancedGroundProximityWarningComputer` as the function parameter and passing `self` doesn't work, as then `self` would be borrowed mutably twice. Also I think it makes more sense to pass these structs directly, since the runtime should logically only have access to write on these two structs.

fbw-common/src/wasm/systems/systems/src/navigation/adirs.rs
Gurgel100: My bad, didn't see that. Makes sense.

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
Gurgel100: Ok would be good to maybe document what the function does and expects then.
To my knowledge modulo is distributive when the divisor is the same. And I think even in this case this holds that we only need one module at the end. Here some examples (left current, right my suggestion):
**My suggestion**: `(angle + 360) % 360`
* `((-180 % 360) + 360) % 360 = (-180 + 360) % 360 = 180 % 360 = 180 <=> (-180 + 360) % 360 = 180 % 360 = 180`
* `((-1 % 360) + 360) % 360 = (-1 + 360) % 360 = 359 % 360 = 359 <=> (-1 + 360) % 360 = 359 % 360 = 359`
* `((180 % 360) + 360) % 360 = (180 + 360) % 360 = 540 % 360 = 180 <=> (180 + 360) % 360 = 540 % 360 = 180`

tracernz:
Looks okay to me overall. 👍

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
Gurgel100: Indeed I didn't consider this case. Good thought!

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
```diff
@@ -194,38 +197,32 @@ impl MemoryNode {
...
-        } else if value >= self.up {
-            true
-        } else {
-            self.output
-        };
+        if value <= self.lower {
+            self.output = false;
+        } else if value >= self.upper {
+            self.output = true;
+        }
```
Gurgel100, line 225:
```suggestion
        self.output = if value <= self.lower {
            false
        } else if value >= self.upper {
            true
        };
```

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
Gurgel100: That is true. But I meant it can be implemented with the nodes instead. But I also agree to your other comments that this can be handled in a different PR.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
Gurgel100: I agree.

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
Gurgel100: Good point. I guess it would (but I actually don't know). My suggestion was anyway to not implement it for any type which also implements `PartialOrd` but just do it manually for `f64`: `impl Clamp for f64 {...}`.

fbw-common/src/wasm/systems/systems/src/surveillance/taws.rs
Gurgel100: I think I agree with you there 😄

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
Gurgel100: We discussed this in Discord. And I agree with you here.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
Gurgel100: You are right and it actually kind of makes sense. Maybe in the future we find a way of generalising this pattern so it can be used by other computer simulations as well.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
Gurgel100: Well even in the event that CAS is not valid this code is still updating the `on_ground` state. It is not really clear from the code that this doesn't cause issues (I didn't dig through the code in too much detail though).

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
Gurgel100: Well that is solved by adding 180 instead of subtracting it.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
Gurgel100: Makes sense then.

fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
Gurgel100: Yeah string values for simvars are currently not supported in the Rust systems.
Would you mind removing it then?

Gurgel100 (commented):
Uff. That's a big one. Nice work. I just have some suggestions/comments/questions.

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
lukecologne: It's not always an assignment though, inbetween the values, the current output is kept and not assigned to

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: Then it is wrong below 180°:
 ```
    (0 - 20 + 540.) % 360. - 180. == -20
    (0 - 20 + 180.) % 360. + 180. == 340
 ```

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
lukecologne: Done

fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
lukecologne: Removed

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
lukecologne: There is still a slight difference between these two for angles smaller than -360°: for an input of 10° - 2*360° for example, current gives 10°, your suggestion gives -350°. It's not very significant for this application, but I would still like to keep it correct. This was btw. not my code, but ported from a C++ implementation in fbw-common (in MathUtil.h), done by @tracernz. I have still added the doc comments for clarification.

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: Yeah good point. I'm not sure what the real GPWS would do here. I also noticed that I have not implemented the mode inhibition at general GPWS failure, only the FAULT light and fault status bit. I will fix that as well

fbw-common/src/wasm/systems/systems/src/shared/logic_nodes.rs
Gurgel100: You're right. Sorry I somehow missed that.

fbw-common/src/wasm/systems/systems/src/shared/mod.rs
```diff
@@ -1165,6 +1165,22 @@ impl<D: uom::si::Dimension + ?Sized, U: uom::si::Units<f64> + ?Sized> Clamp
...
+            "min > max, or either was NaN. min = {min:?}, max = {max:?}"
+        );
+        if self > max {
+            max
+        } else if self < min {
+            min
+        } else {
+            self
+        }
+    }
```
Gurgel100: You can replace this with the call to `f64::clamp`:
```suggestion
        self.clamp(min, max)
```

fbw-common/src/wasm/systems/systems/src/shared/rate_limiter.rs
Gurgel100: Actually you can have the same function defined multiple times on a type: https://doc.rust-lang.org/book/ch20-02-advanced-traits.html#disambiguating-between-identically-named-methods

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
Gurgel100: Stupid truncation modulo...
I give up 😄

fbw-common/src/wasm/systems/systems/src/navigation/ils.rs
Gurgel100: I give up 😄 
Stupid truncation modulo...

lukecologne: Yeah, I think if it were the mathematical modulo you would be correct.. In fact `((x % 360) + 360) % 360` is exactly the implementation of a mathematical modulo `x mod 360`

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/runtime.rs
lukecologne: I have now implemented logic to inhibit mode transitions if there is a general GPWS fault. I also now inhibit all visible and aural warnings when there is a general fault.

Gurgel100 (approved):
I ran out of comments 😄 . Good work!

Jonny23787:

Discord Username  : jonny_23
Object of testing : #10388 
Aircraft          : A32NX
Simulator         : 2024
Tier of Testing   : 2
Date              : 12/01/2026

Testing Process: 

1. Completed full flight without any GPWS triggers.
2. Confirmed GPWS was inhibited by RA, IR and ADR failures/faults. GPWS SYS fault light illuminated.
3. Confirmed GPWS performed a self test at startup.
4. Confirmed modes 1-5 worked correctly as best as I could test.
5. Confirmed overhead buttons worked correctly.

Testing Results: 
Passed

Conclusions: 
Great PR!

fbw-common/src/wasm/systems/systems/src/surveillance/egpws/mod.rs
lukecologne: I have reverted this change, as the current ConfirmationNode does not allow for init conditions which are necessary to properly initialize as running when initially powered.

PR #10390, opened 2025-11-01
test(rs): improve failure output for test asserts

tracernz:
Fixes #[issue_no]

## Summary of Changes

Adds the more-asserts package as a dev dependency, and uses it to improve the debug output for eq, gt, ge, lt, le assertions, so now both values will be printed in the log output.

NOTE: skips the ADIRU to avoid conflicts with #10454.

## Additional context

## Testing instructions

No QA.


fbw-common/src/wasm/systems/systems/src/hydraulic/mod.rs
```diff
@@ -3351,8 +3357,8 @@ mod tests {
             LeakMeasurementValve::new(ElectricalBusType::DirectCurrentEssential)
         }));
 
-        assert!(test_bed.query_element(|e| e.downstream_pressure == Pressure::new::<psi>(0.)));
-        assert!(test_bed.query_element(|e| e.upstream_pressure == Pressure::new::<psi>(0.)));
+        test_bed.query_element(|e| assert_eq!(e.downstream_pressure, Pressure::new::<psi>(0.)));
```
Gurgel100:
```suggestion
        assert_eq!(test_bed.query_element(|e| e.downstream_pressure), Pressure::new::<psi>(0.));
```
I think this way is a bit more logical and the stack trace is smaller as well.

fbw-common/src/wasm/systems/systems/src/shared/mod.rs
```diff
@@ -2005,14 +2039,14 @@ mod local_acceleration_at_plane_coordinate {
         test_bed.command_element(|e| e.set_point_position(cabin_position));
 
         test_bed.run_with_delta(Duration::from_secs(0));
-        assert!(test_bed.query_element(|e| e.local_accel == Vector3::default()));
+        test_bed.query_element(|e| assert_eq!(e.local_accel, Vector3::default()));
```
Gurgel100:
```suggestion
        assert_eq!(test_bed.query_element(|e| e.local_accel), Vector3::default());
```

Gurgel100 (commented):
A couple of minor things.

Gurgel100 (approved):

PR #10413, opened 2025-11-22
fix: usage of setting subject map

Benjozork:
## Summary of Changes

- Fixes an issue where subjects would be incessantly created when getting a typed setting from NXDataStore
- Additionally, stores the subject map on the window so settings are synced between instruments running in the same VCockpit instance

## Testing instructions

- N/A


Benjozork:
> I don’t think the sync is an actual issue currently, but probably is for the hybrid EFBv4 setup?

Correct, nothing both uses that kind of setup and writes settings.

PR #10418, opened 2025-11-29
feat(a32nx): fs2024 native package

tracernz:
Fixes #[issue_no]

## Summary of Changes

Moves the A32NX to a native FS2024 package (with non-modular simobject). Fixes issues that were papered over by the sim MSFS2020 backward compatibility.

FS2020 livery packages remain compatible (as long as they have correct layout.json).

## Additional context


oliik2013:
This would bring the potential of adding walk around mode items like opening the door from outside or opening the cargo hold manually etc.

FoxtrotSierra6829:
> This would bring the potential of adding walk around mode items like opening the door from outside or opening the cargo hold manually etc.

This would not be possible until a fully custom 3D model for the A32NX is created. Don't expect such features will come in 2026. For the A380X those are much more easily attainable.

PR #10442, opened 2025-12-18
fix(fms): fix both hold legs and course reversal symbol showing

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue where hold legs were displayed on the ND despite not being within 2 legs of the hold

## Screenshots (if necessary)

Before:
image

After: 
Screenshot 2025-12-16 130208

## Additional context

## Testing instructions

Insert a hold into the flight plan, either active or secondary both should work the same. The hold is shown as a course reversal if the hold is not on the currently active leg.

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanUtils.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanUtils.ts
@@ -5,6 +5,7 @@
 import { PathVector } from '@flybywiresim/fbw-sdk';
 import { ReadonlyFlightPlan } from './plans/ReadonlyFlightPlan';
 import { ReadonlyFlightPlanElement } from './legs/ReadonlyFlightPlanLeg';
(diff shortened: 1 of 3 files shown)


fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanUtils.ts
```diff
@@ -5,6 +5,7 @@
 import { PathVector } from '@flybywiresim/fbw-sdk';
 import { ReadonlyFlightPlan } from './plans/ReadonlyFlightPlan';
 import { ReadonlyFlightPlanElement } from './legs/ReadonlyFlightPlanLeg';
+import { LnavConfig } from '../guidance/LnavConfig';
 
 export class FlightPlanUtils {
   public static getAllPathVectorsInFlightPlan(
```
tracernz, line 11: This should be renamed to indicate that it's only returning the EFIS path vectors. But then, does it even belong in "FlightPlanUtils"? It's not really to do with flight planning.

PR #10454, opened 2025-12-28
feat(ADIRS): computerized ADIRS simulation

### LINKED ISSUE #7678: ADIRS Alignment Timer Starts Without Power (BAT or Ext Pwr)
### Aircraft Version

Development

### Build info

```json
All versions
```

### Describe the bug

The ADIRS Alignment timer starts when knobs are turned independent of power supply.

### Expected behavior

ADIRS Alignment should not start before powered.

### Steps to reproduce

Watch the ADIRS Alignment LVAR in Cold&Dark. 
Set ADIRS to NAV.
Timer starts. 

### References (optional)

_No response_

_No response_

Cdr_Maverick#6475

lukecologne:
Fixes #7678

## Summary of Changes

This PR fully reworks the ADIRS. The ADR runtimes (i.e. ADR application software) is now also split between software using ADM installations (for A320) and integrated installations (A380).

ADIRS Mode Selector Unit (MSU) and ADIRU Interface:
- The ADR and IR OFF are now momentary pushbuttons. The state is handled by the ADIRUs, which emit an OFF discrete, which can also be affected by selecting the ADIRU mode to OFF (lights will extinguish).

ADIRU simulation:
- The ADIRUs are now simulated as a computer with it's powersupplies, and separated ADR and IR runtimes
    - This includes simulation of the primary (AC) and backup (DC) supplies, time delay at opening relays to disconnect the DC powersupply 5min after power loss, and power supply switching of ADIRU 3 depending on the ATT/HDG selector knob
    - The OFF mode selector switch position now properly cuts the power supply of the ADIRU internally
    - The IR now receives and selects between the 3 other ADRs (for wind and baro inertial computation) depending on SDI and for ADIRU 3 the position of the ATT/HDG and AIR DATA knob.
- For the A320: the ON BAT indication is now fully simulated, along with the MECH CALL horn after 15s

Air Data Sensors Simulation:
- The air data sensors now have a basic simulation, basically they simply read their respective simvars (static/dynamic pressure, AOA etc.). They also have the capability to have blockages/wind vane stuck failures to be simulated. These simulations could be expanded in further PRs.
- For the A380 the MFP (Multifunction Probe), SSP (Sideslip Angle Probe) and ISP (Integrated Static Probes) are fully simulated with their respective powersupplies.
- For the A320, the ADMs (Air Data Modules) are fully simulated with their powersupply from the associated ADIRU, as well as the AOA probes with the respective excitation powersupply and the TAT probes.

FWS:
Add ADR and IR FAULT warnings to the A32NX. The A380X already has them implemented.

## Additional context


fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
```diff
@@ -0,0 +1,970 @@
...
+
+        let p = -self.measurement_inputs.body_rotation_rate_z;
+        let q = -self.measurement_inputs.body_rotation_rate_x;
+        let r = self.measurement_inputs.body_rotation_rate_y;
+        bus.body_roll_rate.set(p, ssm);
+        bus.body_pitch_rate.set(q, ssm);
+        bus.body_yaw_rate.set(r, ssm);
+
+        // Calculate attitude rates, by applying the inverse body coordinate transformation matrix.
+        bus.track_angle_rate.set(
```
Nufflee, line 590: Wouldn't this be the yaw/heading angle rate, considering it is based on body rates?

6639835 (commented):
Nice work man

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
lukecologne: Yes, that's correct. In the previous IR, this was actually called heading rate, but this does apparently not exist in the IR bus, only track angle rate. Currently I only renamed it without changing the calculation behind it

fbw-a380x/src/systems/failures/src/a380.ts
```diff
@@ -360,6 +376,24 @@ export const A380FailureDefinitions: FailureDefinition[] = [
...
   [34, A380Failure.Transponder2, 'XPDR 2'],
+  [34, A380Failure.Adr1, 'ADR 1'],
+  [34, A380Failure.Adr2, 'ADR 2'],
+  [34, A380Failure.Adr3, 'ADR 3'],
+  [34, A380Failure.Ir1, 'IR 1'],
+  [34, A380Failure.Ir2, 'IR 2'],
+  [34, A380Failure.Ir3, 'IR 3'],
+  [34, A380Failure.Adr1, 'ADR 1'],
+  [34, A380Failure.Adr2, 'ADR 2'],
+  [34, A380Failure.Adr3, 'ADR 3'],
```
Gurgel100: Duplicates.
```suggestion
```

fbw-a32nx/src/systems/instruments/src/PFD/PseudoDmc.ts
```diff
@@ -177,6 +178,20 @@ export class PseudoDmc implements Instrument {
...
         true,
       ),
+      this.irSwitchingKnob.sub(
+        (knobPosition) => {
+          if (this.isRightSide) {
+            this.dmcDiscreteWord272.setBitValue(11, knobPosition === 2);
+            this.dmcDiscreteWord272.setBitValue(12, knobPosition === 1 || knobPosition === 2);
+          } else {
+            this.dmcDiscreteWord272.setBitValue(11, knobPosition === 0 || knobPosition === 1);
+            this.dmcDiscreteWord272.setBitValue(12, knobPosition === 0);
```
tracernz: You will need to import the enum as well.
```suggestion
            this.dmcDiscreteWord272.setBitValue(11, knobPosition === AttHdgSwitchingKnob.Fo);
            this.dmcDiscreteWord272.setBitValue(12, knobPosition === AttHdgSwitchingKnob.Norm || knobPosition === AttHdgSwitchingKnob.Fo);
          } else {
            this.dmcDiscreteWord272.setBitValue(11, knobPosition === AttHdgSwitchingKnob.Capt || knobPosition === AttHdgSwitchingKnob.Norm);
            this.dmcDiscreteWord272.setBitValue(12, knobPosition === AttHdgSwitchingKnob.Capt);
```

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2178,6 +2228,7 @@ export class PseudoFWC {
     this.flightPhase67.set(flightPhase === 6 || flightPhase === 7);
     this.flightPhase678.set(flightPhase === 6 || flightPhase === 7 || flightPhase === 8);
     this.flightPhase78.set(flightPhase === 7 || flightPhase === 8);
+    this.flightPhase1_10.set(flightPhase === 1 || flightPhase === 10);
```
tracernz: We don't really need a subject since nothing is subscribing to it. You can just do `const flightPhase1_10 = flightPhase === 1 || flightPhase === 10;` and use it directly.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4104,6 +4155,120 @@ export class PseudoFWC {
         !this.flightPhase12910.get(),
     );
 
+    // TODO should come from the SDAC
+    const ir1FaultSignal = SimVar.GetSimVarValue('L:A32NX_ADIRS_IR_1_FAULT_WARN_DISCRETE', 'number');
+    const ir2FaultSignal = SimVar.GetSimVarValue('L:A32NX_ADIRS_IR_2_FAULT_WARN_DISCRETE', 'number');
+    const ir3FaultSignal = SimVar.GetSimVarValue('L:A32NX_ADIRS_IR_3_FAULT_WARN_DISCRETE', 'number');
```
tracernz: Use RegisteredSimVar for perf.

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,605 @@
...
+                2 => (&mut self.sensor_complex_2, None),
+                3 => (
+                    &mut self.sensor_complex_3,
+                    Some(
+                        self.sensor_complex_1
+                            .tat_probe()
+                            .expect("Sensor complex 1 did not have TAT probe."),
+                    ),
+                ),
+                _ => panic!("Impossible Installation position encountered"),
```
tracernz: This should really be forced at compile time if possible (e.g. with types if the rust type system is good enough to do that). We never want to panic as it risks completely destroying a flight that would otherwise be salvageable (more so for the other runtime panics below than the init one, but it's better to never see any panics in the code).

tracernz (commented):
Mostly skimmed the rust code. Looks good overall. 👍

fbw-a32nx/src/systems/instruments/src/PFD/PseudoDmc.ts
lukecologne: This was also not done in the ADR part above. I fixed that one at well.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
lukecologne: This signal was actually already included in an SDAC word already. Switched to using that

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
lukecologne: That's a fair point, we could use an enum here. That would enable the match statement to be exhaustive and not require panics. But panics are not entirely avoidable due to the legacy interface traits requesting usize as input, so there will still need to be a conversion from usize to the enum.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: We actually already have a flightPhase110 Subject already. I do agree though, we should probably cut down on Subjects as a whole within the PseudoFWC as they introduce a lot more risk for one cycle delays.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4282,6 +4333,118 @@ export class PseudoFWC {
...
+      !(this.sdac00200Word.bitValue(18) && this.sdac00210Word.bitValue(20) && this.sdac00201Word.bitValue(20)) &&
+        adr1and2and3Faulty &&
+        !this.flightPhase110.get(),
+    );
+
+    this.adr1and2FaultActive.set(
+      !(this.flightPhase110.get() || this.sdac00210Word.bitValue(20) || this.sdac00200Word.bitValue(18)) &&
+        adr1Faulty &&
+        adr2Faulty &&
+        !this.adr1and2and3FaultActive,
```
FozzieHi:
```suggestion
        !this.adr1and2and3FaultActive.get()
```

fbw-a32nx/src/systems/shared/src/EwdMessages.ts
```diff
@@ -227,6 +240,42 @@ const EwdMessages = new Map<string, { group?: string; text: string }>([
...
+  ['340006502', { text: '\x1b<5m -ATT HDG SWTG....CAPT 3' }],
+  ['340003001', { group: 'NAV$9', text: ' ADR 1+3 FAULT' }],
+  ['340003002', { text: '\x1b<5m -AIR DATA SWTG.....NORM' }],
+  ['340003003', { text: '\x1b<5m -ATC/XPDR.........SYS 2' }],
+  ['340003004', { text: '\x1b<5m -ADR 1 P/B..........OFF' }],
+  ['340003005', { text: '\x1b<5m -ADR 3 P/B..........OFF' }],
+  ['340007001', { group: 'NAV$9', text: ' IR 1+3 FAULT' }],
+  ['340007002', { text: '\x1b<5m -ATT HDG SWTG......NORM' }],
+  ['340003501', { group: 'NAV$9', text: ' ADR 2+3 FAULT' }],
+  ['340003503', { text: '\x1b<5m -ATC/XPDR.........SYS 1' }],
```
FozzieHi, line 259: You're skipping `340003502` above

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4282,6 +4333,118 @@ export class PseudoFWC {
...
+
+    this.adr2FaultActive.set(
+      adr2Faulty &&
+        !(this.adr1and2FaultActive.get() || this.adr2and3FaultActive.get()) &&
+        !this.flightPhase110.get() &&
+        !this.sdac00210Word.bitValue(20) &&
+        !this.adr1and2and3FaultActive.get(),
+    );
+
+    this.adr3FaultFlipFlop.write(adr3Faulty && this.fwcFlightPhase.get() == 2, !ir3FaultDetected);
```
FozzieHi:
```suggestion
    this.adr3FaultFlipFlop.write(adr3Faulty && this.fwcFlightPhase.get() == 2, !adr3Faulty);
```

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1396,6 +1410,42 @@ export class PseudoFWC {
   private readonly ir2NotAlignedPulse = new NXLogicPulseNode(true);
   private readonly ir3NotAlignedPulse = new NXLogicPulseNode(true);
 
+  private readonly ir1FaultConf = new NXLogicConfirmNode(0.5, true);
+  private readonly ir1FaultMtrig = new NXLogicTriggeredMonostableNode(10, true);
```
FozzieHi: These IR monostables should be retriggerable

FozzieHi (changes requested):
Haven't looked through all of the logic yet, just spotted a few things below

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
lukecologne: For some reason, my implementation causes a sim crash/hangup during loading. Not sure why, the tests run fine. I'll revert for now

fbw-common/src/wasm/systems/systems/src/auto_flight/mod.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{
+    shared::arinc429::Arinc429Word,
+    simulation::{InitContext, Read, SimulationElement, SimulatorReader, VariableIdentifier},
+};
+
+#[derive(Default)]
+pub struct FlightControlUnitBusOutputs {
+    pub baro_correction_1_hpa: Arinc429Word<f64>,
+    pub baro_correction_2_hpa: Arinc429Word<f64>,
```
Gurgel100, line 9:
```suggestion
    pub baro_correction_1: Arinc429Word<Pressure>,
    pub baro_correction_2: Arinc429Word<Pressure>,
```
We should use the uom `Pressure` type.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
```diff
@@ -0,0 +1,260 @@
...
+    simulation::{InitContext, UpdateContext},
+};
+
+#[derive(Default)]
+pub struct AirDataModuleBusOutputs {
+    pub left_static_pressure_hpa: Arinc429Word<f64>,
+    pub right_static_pressure_hpa: Arinc429Word<f64>,
+    pub averaged_static_pressure_uncorrected_hpa: Arinc429Word<f64>,
+    pub total_pressure_hpa: Arinc429Word<f64>,
+}
```
Gurgel100, line 20: We should use the uom `Pressure` type instead of `f64`.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
```diff
@@ -0,0 +1,260 @@
...
+    const MINIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(200);
+    const MAXIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(300);
+
+    pub fn new(
+        context: &mut InitContext,
+        installation_position: AirDataModuleInstallationPosition,
+    ) -> Self {
+        let is_powered = context.has_engines_running();
+        Self {
+            is_powered: false,
```
Gurgel100: I think this was the intention?
```suggestion
            is_powered,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        self.installation_position = installation_position;
+
+        self.initialized = true;
+    }
+
+    fn update(
+        &mut self,
+        context: &UpdateContext,
+        pressure_transucer_value: f64,
```
Gurgel100, line 189:
```suggestion
        pressure_transucer_value: Pressure,
```
IMO we should keep the uom types as long as possible to avoid confusion about which unit is used.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+struct AirDataModuleRuntime {
+    /// If non-Duration::ZERO, the remaining time the runtime needs to initialize itself. Otherwise
+    /// the self-check has been completed.
+    remaining_startup: Duration,
+
+    initialized: bool,
+
+    pressure_filter: LowPassFilter<f64>,
```
Gurgel100, line 149:
```suggestion
    pressure_filter: LowPassFilter<Pressure>,
```
We should use uom types whereever possible.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
```diff
@@ -0,0 +1,260 @@
...
+        if self.remaining_startup > Duration::ZERO {
+            return;
+        }
+
+        if !self.initialized {
+            self.initialize(pressure_transucer_value, installation_position);
+        }
+
+        self.pressure_filter
+            .update(context.delta(), pressure_transucer_value);
```
Gurgel100, line 206: This should be equivalent.
```suggestion
        if !self.initialized {
            self.initialize(pressure_transucer_value, installation_position);
        } else {
            self.pressure_filter
                .update(context.delta(), pressure_transucer_value);
        }
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/mod.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub mode_select_m2: bool,
+    pub adr_off_command: bool,
+}
+
+#[derive(Default)]
+pub struct AdrAnalogInputs {
+    pub aoa_excitation_voltage_v: f64,
+    pub aoa_resolver_angle_deg: f64,
+    pub tat_value_deg_c: f64,
+}
```
Gurgel100, line 350: IMO we should use uom types here as well.

fbw-common/src/wasm/systems/systems_wasm/src/navigation.rs
```diff
@@ -0,0 +1,68 @@
...
+            Variable::named("BOARDING_STARTED_BY_USR"),
+            Variable::named("BOARDING_RATE"),
+        ],
+        |values| {
+            let boarding_rate = BoardingRate::from(values[1]);
+            let boarding_started_by_user = to_bool(values[0]);
+
+            match boarding_rate {
+                BoardingRate::Instant | BoardingRate::Fast => boarding_started_by_user,
+                _ => false,
```
Gurgel100:
```suggestion
                BoardingRate::Real => false,
```

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,605 @@
...
+pub(crate) struct A320AdirsElectricalHarness {
+    on_bat_light_id: VariableIdentifier,
+    on_bat_light: bool,
+
+    mech_horn_id: VariableIdentifier,
+    relay_delay_on_close_mech_call: ConfirmationNode,
+
+    pub adiru_1_electrical_harness: A320AdiruElectricalHarness,
+    pub adiru_2_electrical_harness: A320AdiruElectricalHarness,
+    pub adiru_3_electrical_harness: A320AdiruElectricalHarness,
```
Gurgel100:
```suggestion
    adiru_1_electrical_harness: A320AdiruElectricalHarness,
    adiru_2_electrical_harness: A320AdiruElectricalHarness,
    adiru_3_electrical_harness: A320AdiruElectricalHarness,
```

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,605 @@
...
+use uom::si::{f64::*, pressure::hectopascal};
+
+pub struct A320AirDataInertialReferenceSystem {
+    pub adiru_1: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,
+    pub adiru_2: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,
+    pub adiru_3: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,
+
+    pub sensor_complex_1: AirDataModuleAirDataSensorsComplex,
+    pub sensor_complex_2: AirDataModuleAirDataSensorsComplex,
+    pub sensor_complex_3: AirDataModuleAirDataSensorsComplex,
```
Gurgel100:
```suggestion
    adiru_1: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,
    adiru_2: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,
    adiru_3: AirDataInertialReferenceUnit<AdmAirDataReferenceRuntime>,

    sensor_complex_1: AirDataModuleAirDataSensorsComplex,
    sensor_complex_2: AirDataModuleAirDataSensorsComplex,
    sensor_complex_3: AirDataModuleAirDataSensorsComplex,
```

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,605 @@
...
+    ir_fault_warn: bool,
+    ir_align_discrete: bool,
+
+    // Simulator inputs
+    ir_fast_align_mode_id: VariableIdentifier,
+    ir_instant_align_id: VariableIdentifier,
+    ir_excess_motion_inhibit_id: VariableIdentifier,
+}
+impl A320AdiruElectricalHarness {
+    const RELAY_OPENING_TIME_DELAY: Duration = Duration::from_secs(300);
```
Gurgel100:
```suggestion
    const RELAY_OPENING_TIME_DELAY: Duration = Duration::from_mins(5);
```

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,605 @@
...
+        self.primary_powered = buses.is_powered(self.primary_powersupply);
+        self.backup_powered = buses.is_powered(self.backup_powersupply);
+        self.backup_powersupply_relay_switching_powered_1 = self
+            .backup_powersupply_relay_switching_1
+            .is_some_and(|bus| buses.is_powered(bus));
+        self.backup_powersupply_relay_switching_powered_2 = self
+            .backup_powersupply_relay_switching_2
+            .is_some_and(|bus| buses.is_powered(bus));
+
+        self.aoa_excitation_powered = buses.is_powered(self.aoa_excitation_power_source);
```
Gurgel100, line 647: If by chance you know how much power should be drawn it would be really cool to also have that so we can at some point get rid of the fake power drawing.

fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,668 @@
...
+use uom::si::pressure::hectopascal;
+
+pub struct A380AirDataInertialReferenceSystem {
+    pub adiru_1: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,
+    pub adiru_2: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,
+    pub adiru_3: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,
+
+    pub sensor_complex_1: IntegratedAirDataSensorsComplex,
+    pub sensor_complex_2: IntegratedAirDataSensorsComplex,
+    pub sensor_complex_3: IntegratedAirDataSensorsComplex,
```
Gurgel100:
```suggestion
    adiru_1: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,
    adiru_2: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,
    adiru_3: AirDataInertialReferenceUnit<IntegratedAirDataReferenceRuntime>,

    sensor_complex_1: IntegratedAirDataSensorsComplex,
    sensor_complex_2: IntegratedAirDataSensorsComplex,
    sensor_complex_3: IntegratedAirDataSensorsComplex,
```

fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,668 @@
...
+        self.adiru_3.accept(visitor);
+
+        visitor.visit(self);
+    }
+}
+
+pub(crate) struct A380AdirsElectricalHarness {
+    pub adiru_1_electrical_harness: A380AdiruElectricalHarness,
+    pub adiru_2_electrical_harness: A380AdiruElectricalHarness,
+    pub adiru_3_electrical_harness: A380AdiruElectricalHarness,
```
Gurgel100:
```suggestion
    adiru_1_electrical_harness: A380AdiruElectricalHarness,
    adiru_2_electrical_harness: A380AdiruElectricalHarness,
    adiru_3_electrical_harness: A380AdiruElectricalHarness,
```

fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,668 @@
...
+            _ => panic!("Invalid ADIRU Index {num}"),
+        }
+    }
+
+    fn switching_relay_logic(&self, input_1: bool, input_2: bool) -> (bool, bool) {
+        let output_1 = if self.probes_1_3_switching_relay_closed {
+            false
+        } else {
+            input_1
+        };
```
Gurgel100, line 342: This should be equivalent. You can decide if you want to take it or not :).

```suggestion
        let output_1 = !self.probes_1_3_switching_relay_closed && input_1;
```

fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,668 @@
...
+            buses.is_powered(ElectricalBusType::AlternatingCurrentEssentialShed);
+        // 491XP
+        self.ac_ess_powered = buses.is_powered(ElectricalBusType::AlternatingCurrentEssential);
+        // 103PP
+        self.dc_1_powered = buses.is_powered(ElectricalBusType::DirectCurrent(1));
+        // 204PP
+        self.dc_2_powered = buses.is_powered(ElectricalBusType::DirectCurrent(2));
+        // 407PP and 415PP
+        self.dc_ess_powered = buses.is_powered(ElectricalBusType::DirectCurrentEssential);
+    }
```
Gurgel100, line 443: Same as for the A32NX: would be cool to have the actual consumption as well.

fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs
```diff
@@ -0,0 +1,668 @@
...
+    ir_fault_warn: bool,
+    ir_align_discrete: bool,
+
+    // Simulator inputs
+    ir_fast_align_mode_id: VariableIdentifier,
+    ir_instant_align_id: VariableIdentifier,
+    ir_excess_motion_inhibit_id: VariableIdentifier,
+}
+impl A380AdiruElectricalHarness {
+    const RELAY_OPENING_TIME_DELAY: Duration = Duration::from_secs(300);
```
Gurgel100:
```suggestion
    const RELAY_OPENING_TIME_DELAY: Duration = Duration::from_mins(5);
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+        random_from_range,
+    },
+    simulation::{InitContext, SimulationElement, SimulationElementVisitor, UpdateContext},
+};
+
+#[derive(Default)]
+pub struct MultifunctionProbeBusOutputs {
+    pub total_pressure_hpa: Arinc429Word<f64>,
+    pub total_air_temperature_deg_c: Arinc429Word<f64>,
+    pub angle_of_attack_deg: Arinc429Word<f64>,
```
Gurgel100, line 25: Use uom types.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+struct MultifunctionProbeRuntime {
+    /// If non-Duration::ZERO, the remaining time the runtime needs to initialize itself. Otherwise
+    /// the self-check has been completed.
+    remaining_startup: Duration,
+
+    initialized: bool,
+
+    pressure_filter: LowPassFilter<f64>,
```
Gurgel100, line 164:
```suggestion
    pressure_filter: LowPassFilter<Pressure>,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+
+    tat_degree_c: f64,
+    aoa_deg: f64,
```
Gurgel100, line 167: Use uom types.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+
+    fn is_initialized(&self) -> bool {
+        self.initialized
+    }
+}
+
+#[derive(Default)]
+pub struct IntegratedStaticProbeBusOutputs {
+    pub left_static_pressure_hpa: Arinc429Word<f64>,
+    pub right_static_pressure_hpa: Arinc429Word<f64>,
```
Gurgel100, line 250: UOM types!

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+struct IntegratedStaticProbeRuntime {
+    /// If non-Duration::ZERO, the remaining time the runtime needs to initialize itself. Otherwise
+    /// the self-check has been completed.
+    remaining_startup: Duration,
+
+    initialized: bool,
+
+    pressure_filter: LowPassFilter<f64>,
```
Gurgel100, line 405:
```suggestion
    pressure_filter: LowPassFilter<Pressure>,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    fn is_initialized(&self) -> bool {
+        self.initialized
+    }
+}
+
+#[derive(Default)]
+pub struct SideslipAngleProbeBusOutputs {
+    pub sideslip_angle_deg: Arinc429Word<f64>,
```
Gurgel100, line 497:
```suggestion
    pub sideslip_angle_deg: Arinc429Word<Angle>,
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
Gurgel100, line 1: Many probes have kind of similar if not the same logic. Maybe there is a way to extract this into a common `Probe` type?

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    fn read(&mut self, reader: &mut SimulatorReader) {
+        // If it pitot blockage failure is active, no longer update the measured pressure, as it should retain the current value
+        if !self.pitot_blockage.is_active() {
+            let dynamic_pressure = reader.read(&self.dynamic_pressure_id);
+            self.dynamic_pressure = Pressure::new::<inch_of_mercury>(dynamic_pressure);
+
+            let static_pressure = reader.read(&self.static_pressure_id);
+            self.static_pressure = Pressure::new::<inch_of_mercury>(static_pressure);
```
Gurgel100, line 69: I think I know why you read the values here but just so you know these values are also available from the `UpdateContext`.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn accept<T: SimulationElementVisitor>(&mut self, visitor: &mut T) {
+        self.static_blockage.accept(visitor);
+
+        visitor.visit(self);
+    }
+
+    fn read(&mut self, reader: &mut SimulatorReader) {
+        if !self.static_blockage.is_active() {
+            let static_pressure = reader.read(&self.static_pressure_id);
+            self.static_pressure = Pressure::new::<inch_of_mercury>(static_pressure);
```
Gurgel100, line 109: I think I know why you read the value here but just so you know this value is also available from the `UpdateContext`.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+}
+impl TemperatureProbe for TotalAirTemperatureProbe {
+    fn get_temperature(&self) -> ThermodynamicTemperature {
+        self.total_air_temperature
+    }
+}
+impl SimulationElement for TotalAirTemperatureProbe {
+    fn read(&mut self, reader: &mut SimulatorReader) {
+        self.total_air_temperature = reader.read(&self.total_air_temperature_id);
```
Gurgel100, line 235: I think I know why you read the value here but just so you know this value is also available from the `UpdateContext`.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            OutputDataType::Adr => write!(f, "ADR"),
+            OutputDataType::Ir => write!(f, "IR"),
+        }
+    }
+}
+
+fn output_data_id(data_type: OutputDataType, number: usize, name: &str) -> String {
+    format!("ADIRS_{}_{}_{}", data_type, number, name)
```
Gurgel100: You can do this as well:
```suggestion
    format!("ADIRS_{data_type}_{number}_{name}")
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+        Self { lower, upper }
+    }
+}
+
+pub struct AirDataInertialReferenceUnit<AdrRuntime> {
+    adr_failure: Failure,
+    ir_failure: Failure,
+
+    // TODO replace with pin prog
+    #[allow(unused)]
```
Gurgel100:
```suggestion
    #[expect(unused)]
```
This way there is a warning when this will be used in the future.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+
+    adr_discrete_output_data: AdrDiscreteOutputs,
+    adr_bus_output_data: AirDataReferenceBusOutputs,
+
+    ir_discrete_output_data: IrDiscreteOutputs,
+    ir_bus_output_data: InertialReferenceBusOutputs,
+
+    ir_measurement_inputs: IrSimulatorData,
+
+    // Nonvalatile Programming
```
Gurgel100:
```suggestion
    // Nonvolatile Programming
```
Small typo.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+    adr_runtime: Option<AdrRuntime>,
+    ir_runtime: Option<InertialReferenceRuntime>,
+
+    adr_discrete_output_data: AdrDiscreteOutputs,
+    adr_bus_output_data: AirDataReferenceBusOutputs,
+
+    ir_discrete_output_data: IrDiscreteOutputs,
+    ir_bus_output_data: InertialReferenceBusOutputs,
+
+    ir_measurement_inputs: IrSimulatorData,
```
Gurgel100, line 139: I think this is not what you want. Now you have 3 instances of `IrSimulatorData` which is not what you wanted if I understand the idea correctly. Instead you'll have to pass a reference to a struct of this type to the `update` function.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+    const WIND_DIRECTION: &'static str = "WIND_DIRECTION";
+    const WIND_DIRECTION_BNR: &'static str = "WIND_DIRECTION_BNR";
+    const WIND_SPEED: &'static str = "WIND_SPEED";
+    const WIND_SPEED_BNR: &'static str = "WIND_SPEED_BNR";
+    const LATITUDE: &'static str = "LATITUDE";
+    const LONGITUDE: &'static str = "LONGITUDE";
+    const MAINT_WORD: &'static str = "MAINT_WORD";
+
+    pub(super) const ADR_AVERAGE_STARTUP_TIME_MILLIS: Duration = Duration::from_millis(3_000);
+    pub(super) const IR_AVERAGE_STARTUP_TIME_MILLIS: Duration = Duration::from_millis(3_000);
```
Gurgel100:
```suggestion
    pub(super) const ADR_AVERAGE_STARTUP_TIME: Duration = Duration::from_secs(3);
    pub(super) const IR_AVERAGE_STARTUP_TIME: Duration = Duration::from_secs(3);
```
We don't need `_MILLIS` in the name since the type is `Duration` where the consumer can decide which time fraction they prefer.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
```diff
@@ -0,0 +1,1194 @@
...
+    pub(super) const MINIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(200);
+    pub(super) const MAXIMUM_POWER_HOLDOVER: Duration = Duration::from_millis(300);
+
+    pub(super) const IR_FAULT_FLASH_DURATION: Duration = Duration::from_millis(50);
+
+    pub(super) const POWER_DOWN_DURATION: u64 = 15_000;
```
Gurgel100:
```suggestion
    pub(super) const POWER_DOWN_DURATION: Duration= Duration::from_secs(15);
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adm_adr_runtime.rs
```diff
@@ -0,0 +1,712 @@
...
+            self.mach = 0.;
+        }
+    }
+
+    fn update_temperatures(&mut self, analog_inputs: &AdrAnalogInputs) {
+        self.total_air_temperature_deg_c = analog_inputs.tat_value_deg_c;
+        self.total_air_temperature_valid = true;
+
+        self.static_air_temperature_valid = self.total_air_temperature_valid && self.mach_valid;
+        let tat_kelvin = self.total_air_temperature_deg_c - Self::ABSOLUTE_ZERO_DEG_C;
```
Gurgel100, line 390: When using uom you could get kelvin directly.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adm_adr_runtime.rs
```diff
@@ -0,0 +1,712 @@
...
+        discrete_outputs.low_speed_warning_3 =
+            self.low_speed_warning_3_hysteresis.get_output() && !self.output_inhibited;
+        discrete_outputs.low_speed_warning_4 =
+            !self.low_speed_warning_4_hysteresis.get_output() && !self.output_inhibited;
+    }
+
+    fn compute_discrete_word_1(&self, discrete_word_1: &mut Arinc429Word<u32>) {
+        let mut discrete_word = AdrDiscrete1Flags::default();
+
+        // FIXME implement icing detector heat
```
Gurgel100:
```suggestion
        // TODO: implement icing detector heat
```
I would say "TODO" fits better since "FIXME" sounds like something is broken.
(Also for the other instances)

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/integrated_adr_runtime.rs
```diff
@@ -0,0 +1,728 @@
...
+        }
+    }
+}
+impl IntegratedAirDataReferenceRuntime {
+    pub(super) const MINIMUM_TAS: f64 = 60.;
+    pub(super) const MINIMUM_CAS: f64 = 30.;
+    pub(super) const MINIMUM_MACH: f64 = 0.1;
+    pub(super) const MINIMUM_CAS_FOR_AOA: f64 = 60.;
+    pub(super) const MINIMUM_VALID_ALTITUDE: f64 = -2000.;
+    pub(super) const MAXIMUM_VALID_ALTITUDE: f64 = 50000.;
```
Gurgel100:
```suggestion
    const MINIMUM_TAS: f64 = 60.;
    const MINIMUM_CAS: f64 = 30.;
    const MINIMUM_MACH: f64 = 0.1;
    const MINIMUM_CAS_FOR_AOA: f64 = 60.;
    const MINIMUM_VALID_ALTITUDE: f64 = -2000.;
    const MAXIMUM_VALID_ALTITUDE: f64 = 50000.;
```
Doesn't need to be public.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/integrated_adr_runtime.rs
```diff
@@ -0,0 +1,728 @@
...
+    const STANDARD_SPEED_OF_SOUND_KNOT: f64 = 661.4746;
+    const STANDARD_SEA_LEVEL_PRESSURE_HPA: f64 = 1013.25;
+    const STANDARD_SEA_LEVEL_TEMPERATURE_KELVIN: f64 = 288.15;
+    const ABSOLUTE_ZERO_DEG_C: f64 = -273.15;
+
+    pub fn update(
+        &mut self,
+        context: &UpdateContext,
+        discrete_inputs: &AdrDiscreteInputs,
+        #[allow(unused)] analog_inputs: &AdrAnalogInputs, // TODO this is used for the OAT probe analog inputs, which are not implemented
```
Gurgel100:
```suggestion
        #[expect(unused)] analog_inputs: &AdrAnalogInputs, // TODO this is used for the OAT probe analog inputs, which are not implemented
```

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
```diff
@@ -0,0 +1,998 @@
...
+        discrete_inputs: &IrDiscreteInputs,
+        hardware_fault: bool,
+    ) {
+        self.update_alignment_excess_motion_monitor(context, discrete_inputs);
+        self.update_alignment_in_air_fault();
+
+        if let Some(timer) = self.one_hertz_clock.checked_sub(context.delta()) {
+            self.one_hertz_clock = timer;
+        } else {
+            self.one_hertz_clock = Duration::from_secs(2);
```
Gurgel100, line 462: Why is it called `one_hertz_clock` when it is set to 2 seconds?

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
```diff
@@ -0,0 +1,998 @@
...
+    //        self.ir_fault_flash_duration = Some(Self::IR_FAULT_FLASH_DURATION);
+    //    } else if let Some(flash_duration) = self.ir_fault_flash_duration {
+    //        let remaining = subtract_delta_from_duration(context, flash_duration);
+    //        self.ir_fault_flash_duration = if remaining > Duration::from_secs(0) {
+    //            Some(remaining)
+    //        } else {
+    //            None
+    //        };
+    //    }
+    //}
```
Gurgel100: Is this code for later? Otherwise, I suggest to remove it.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
```diff
@@ -0,0 +1,998 @@
...
+        bus.drift_angle.set(
+            if ground_speed_above_minimum_threshold {
+                let diff = self.measurement_inputs.track - self.measurement_inputs.heading;
+                if diff > Angle::new::<degree>(180.) {
+                    diff - Angle::new::<degree>(360.)
+                } else if diff < Angle::new::<degree>(-180.) {
+                    diff + Angle::new::<degree>(360.)
+                } else {
+                    diff
+                }
```
Gurgel100: Didn't you create a helper function for this? `normalised_180`?

Gurgel100 (commented):
Hui that's quite a PR!
Sorry, it took some time to go through everything :).

We probably need to work on a framework for simulating computers to not repeat the logic for the runtime resets.

fbw-common/src/wasm/systems/systems/src/auto_flight/mod.rs
tracernz: You need to be able to control the units, as there are both hPa and in.Hg bus outputs, and it's best that the names reflect explicitly which they are. For things that are not ARINC bus words, I agree, using uom makes sense.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs
lukecologne: The clock is used for 1 Hertz flashing of the IR FAULT legend, so the legend is off 1s and on another 1s. So this clock wraps around at 2s to be able to distinguish between the two phases

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
lukecologne: I do know the maximum allowable consumption for ADIRUs, but not the amount this specific ADIRU type uses. Not sure if we think this is enough for a first pass.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/integrated_probes.rs
lukecologne: I think most of duplication in the probe computer lies in the computer part, which would be included in a more generic computer implementation you mentioned. The rest is mostly either filtering or digital output management, which is mostly unique to the probes

fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/adiru.rs
lukecologne: This is intended for two reasons, first of all this data is not really something that is fed in as an input, instead it's generated internally from internal sensors (the accelerometers and gyrometers) which are not currently simulated. I wanted to keep the inputs as realistic as possible. Secondly, having these independent instances would allow to generate errors/drift/etc. in the data without having to fully implement the inertial integration.

lukecologne:
Thank you for the review (I know it's gotten rather large... ). I have adressed many of the comments, although for the ones regarding UOM, I think we need to discuss this further. For one thing, I find that it's usage is severly limited by the lack of fractional powers of units, which basically all of the computations in the ADR (compressible flow) need. This is one of the reasons that I have chosen to limit the use of UOM. Outside of this usability constraint, while it makes sense to use in physics computations which are not part of a computer system, in other cases it is different IMO. While it can be argued that use of it makes sense in internal computer computations or maybe also analog inputs which are then digitized in the computer, for bus interfaces this is no longer the case. The values are transmitted as fixed point numbers with a clearly defined unit, and when we decide to switch to a more realistic A429 (or also AFDX/A664) implementation, using UOM types will not be possible.

fbw-a32nx/src/systems/shared/src/EwdMessages.ts
```diff
@@ -227,6 +241,43 @@ const EwdMessages = new Map<string, { group?: string; text: string }>([
...
+  ['340002004', { text: '\x1b<5m -BARO REF.........CHECK' }],
+  ['340005001', { group: 'NAV$5', text: ' IR 1 FAULT' }],
+  ['340005002', { text: '\x1b<5m -ATT HDG SWTG....CAPT 3' }],
+  ['340005003', { text: '\x1b<5m -ATC/XPDR.........SYS 2' }],
+  ['340006001', { group: 'NAV$5', text: ' IR 2 FAULT' }],
+  ['340006002', { text: '\x1b<5m -ATT HDG SWTG.....F/O 3' }],
+  ['340006003', { text: '\x1b<5m -ATC/XPDR.........SYS 1' }],
+  ['340004001', { group: 'NAV$5', text: ' ADR 3 FAULT' }],
+  ['340004002', { text: '\x1b<5m -ADR 3 P/B..........OFF' }],
+  ['340004003', { text: '\x1b<5m -AIR DATA SWTG.....NORM' }],
```
FozzieHi: `-AIR DATA SWTG.....NORM` should be above `ADR 3 P/B..........OFF`

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4282,6 +4332,121 @@ export class PseudoFWC {
...
+    const adr1and2and3Faulty = adr1Faulty && adr2Faulty && adr3Faulty;
+
+    this.adr1and2and3FaultActive.set(
+      !(this.sdac00200Word.bitValue(18) && this.sdac00210Word.bitValue(20) && this.sdac00201Word.bitValue(20)) &&
+        adr1and2and3Faulty &&
+        !this.flightPhase110.get(),
+    );
+
+    this.adr1and2FaultActive.set(
+      !(this.flightPhase110.get() || this.sdac00210Word.bitValue(20) || this.sdac00200Word.bitValue(18)) &&
```
FozzieHi:
```suggestion
      !(this.flightPhase110.get() || (this.sdac00210Word.bitValue(20) && this.sdac00200Word.bitValue(18))) &&
```

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -6606,6 +6771,209 @@ export class PseudoFWC {
...
       side: 'LEFT',
     },
+    3400010: {
+      // ADR 1 FAULT
+      flightPhaseInhib: [1, 4, 8, 10],
+      simVarIsActive: this.adr1FaultActive,
+      whichCodeToReturn: () => [
+        0,
+        !this.adr3Used.get() ? 1 : null,
+        this.adr1DiscreteWord1.get().isNormalOperation() ? 2 : null,
```
FozzieHi: Prefer `!isInvalid()` over is NO?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -6606,6 +6771,209 @@ export class PseudoFWC {
...
+      sysPage: -1,
+      side: 'LEFT',
+    },
+    3400040: {
+      // ADR 3 FAULT
+      flightPhaseInhib: [1, 4, 5, 7, 8, 10],
+      simVarIsActive: this.adr3FaultActive,
+      whichCodeToReturn: () => [
+        0,
+        this.adr3DiscreteWord1.get().isNormalOperation() ? 1 : null,
```
FozzieHi: Remember to swap these around when you swap the text around

FozzieHi (commented):
Looks good, just a few comments on the FWC work

fbw-a32nx/src/systems/shared/src/EwdMessages.ts
lukecologne: Kept this way to follow currently used FWC standard

FozzieHi: We target F13 though, so regardless of UAMM standards `AIR DATA SWTG` should be above

lukecologne: Fixed

fbw-a32nx/src/wasm/systems/a320_systems/src/navigation/adirs.rs
lukecologne: I've changed the into_iter implementation and it now works. No more panics

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
tracernz: I don't like it for bus outputs, because there are a mix of pressure units on bus outputs, and passing uom through implicitly chooses one of them, leading to subtle bugs if we don't notice it's the wrong unit.

tracernz:
In the interests of moving this along, as it has much future work depending on it, I ran a review using Codex 5.6 Sol (High).

- **[P1] Alignment checks read the excess-motion flag.** [`[is_fully_aligned()](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs#L173-L176)`](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-a380x/src/wasm/systems/a380_systems/src/navigation/adirs.rs#L173-L176) calls `get_bit(13)`, which tests packed bit 12 (`EXCESS_MOTION_ERROR`). `NAV_MODE` is packed at bit 2. Consequently, aligned IRs normally report unaligned, preventing A380 autobrake arming and affecting GPWC/ND alignment logic. The A320 implementation has the same defect.

- **[P2] ADR input B is validated using input A.** In [`[ir_runtime.rs](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs#L414-L419)`](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs#L414-L419), both `adr_a_valid` and `adr_b_valid` inspect `adr_a`. Selecting a healthy B after A fails therefore raises a false ADR fault; selecting failed B while A is healthy suppresses it.

- **[P2] Excessive motion is direction-dependent.** [`[Vector2::max()](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs#L496-L499)`](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-common/src/wasm/systems/systems/src/navigation/adirs/hw_block3_adiru/ir_runtime.rs#L496-L499) returns the largest signed component. Reverse pushback or motion along a negative axis can exceed the alignment limit without triggering the fault. Use an absolute maximum (`amax()`) or the intended vector magnitude.

- **[P2] A32NX FWS reads the wrong packed ADR fault bit.** [`[PseudoFWC.ts](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts#L4388-L4394)`](https://github.com/flybywiresim/aircraft/blob/7c120b0a1abf9efa37c4737e428feb89405911d4/fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts#L4388-L4394) reads bit 13, while `ADR_STATUS_FAIL` is packed at bit 2—bit 3 through the one-based `bitValueOr()` API. Valid discrete words reporting sensor or excitation faults can therefore be missed.

fbw-common/src/wasm/systems/systems/src/navigation/adirs/air_data_sensors/air_data_module.rs
Gurgel100: I guess this makes sense.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1435,6 +1449,41 @@ export class PseudoFWC {
...
+  private readonly ir2and3FaultActive = Subject.create(false);
+
+  private readonly adr3FaultFlipFlop = new NXLogicMemoryNode(false);
+  private readonly adr3UsedLeft = Subject.create(false);
+  private readonly adr3UsedRight = Subject.create(false);
+  private readonly adr3Used = MappedSubject.create(
+    ([adr3UsedLeft, adr3UsedRight]) => adr3UsedLeft || adr3UsedRight,
+    this.adr3UsedLeft,
+    this.adr3UsedRight,
+  );
```
BravoMike99: If nothing is subscribing to these, we can probably skip the subjects? (It's an overall problem in the fwc, subjects being used many times when it's not needed)

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4359,6 +4409,124 @@ export class PseudoFWC {
...
+        this.dmcLeftDiscreteWord272.get().bitValueOr(12, false),
+    );
+
+    this.ir3Used.set(
+      (this.dmcLeftDiscreteWord272.get().bitValueOr(11, false) &&
+        this.dmcLeftDiscreteWord272.get().bitValueOr(12, false)) ||
+        (this.dmcRightDiscreteWord272.get().bitValueOr(11, false) &&
+          this.dmcRightDiscreteWord272.get().bitValueOr(12, false)),
+    );
+
```
BravoMike99: Could probably store some of these in helper variables just to reduce the .get bloat and unnecessary calls. (same comment regarding lower down)

BravoMike99 (commented):
Some minor comments, also, if the IR/ADR words changed, shouldn't we update the simvars.md?

PR #10461, opened 2026-01-02
fix(fms): use localizer station declination for true courses on final approach legs

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue where the final approach path of a localizer-like approach shown on the ND would have a kink in it because we used the magnetic variation of the airport, rather than the station declination of the localizer facility.

Also uses a type I transition between two CF legs. Using a type II occasionally caused a transition that was too large for courses that were very close.

## Screenshots (if necessary)

Before:
image
image

After:
image

## Additional context

## Testing instructions

If you have Navigraph navdata, install it for this test. The issues seemed to occur more often with it installed.

Load a few different approaches, particularly to Scandinavian airports (e.g ESSA, ESGG), make sure the path drawn on the ND looks the way you expect it to and doesn't show any of the kinks that the screenshots above do.


tracernz:
Please add test instructions.

calarandir:

Discord Username  : shadowca_
Object of testing : #10461
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 05/01/2026

Testing Process: 
Changing destination of Flight on INIT page and selecting various ILS/LOC approaches. I am not using Navigraph.
Unable to reproduce the effect at ESGG. Tried a lot of very north and very south airports but only had the problem at PANC:
PANC AMOTT4 ENA ILS07L
PANC ELLAM5 ILS15

Testing above approach configs with the PR build. Both final approach paths look good on the ND.
I would be happier with a second example tested but I cannot find any. Maybe someone else can.

Testing Results:
Passed

Media:
Before:
PANC AMOTT4 ENA ILS07L
image

PANC ELLAM5 ILS15
image


After:
PANC AMOTT4 ENA ILS07L
image


PANC ELLAM5 ILS15
image

PR #10470, opened 2026-01-06
feat(A380X/mfd): improved compute lrc

Lucas-IQ21:
Fixes #[issue_no]

## Summary of Changes

Improved LRC compute FMS perf cuize page.

## Screenshots (if necessary)

image

## References

https://www.fzt.haw-hamburg.de/pers/Scholz/arbeiten/HAW-TextCaers.pdf -> 2.1.4 Long-Range Cruise Speed
https://ansperformance.eu/library/airbus-fuel-economy.pdf

FCOM : 
image

image

## Additional context

Previously, the LCR speed was simply specified as a fixed value, but now it is calculated based on ISA temp, the weight, and the altitude of the aircraft.

Discord username (if different from GitHub): Lucas-IQ

## Testing instructions

1. Fill FMS as usual.
2. Before the cruise phase, the LRC speed in Mach does not appear.
3. Reach cruise phase. 
4. The LCR speed is displayed and varies depending on the cruising altitude, the weight of the aircraft, and the ISA temperature.
5. The LRC cannot be less than 0.75 or greater than 0.88.

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
@@ -4,12 +4,18 @@ import { TopTabNavigator, TopTabNavigatorPage } from 'instruments/src/MsfsAvioni
 
 import { ArraySubject, ClockEvents, FSComponent, MappedSubject, Subject, VNode } from '@microsoft/msfs-sdk';
 
-import { Button } from 'instruments/src/MsfsAvionicsCommon/UiWidgets/Button';
-import { RadioButtonGroup } from 'instruments/src/MsfsAvionicsCommon/UiWidgets/RadioButtonGroup';
 import { AbstractMfdPageProps } from 'instruments/src/MFD/MFD';
 import { Footer } from 'instruments/src/MFD/pages/common/Footer';
+import { Button } from 'instruments/src/MsfsAvionicsCommon/UiWidgets/Button';
+import { RadioButtonGroup } from 'instruments/src/MsfsAvionicsCommon/UiWidgets/RadioButtonGroup';
 
-import './MfdFmsPerf.scss';
+import { ApproachType } from '@flybywiresim/fbw-sdk';
+import { FlapConf } from '@fmgc/guidance/vnav/common';
+import { VerticalCheckpointReason } from '@fmgc/guidance/vnav/profile/NavGeometryProfile';
+import { FmgcFlightPhase } from '@shared/flightphase';
+import { A380SpeedsUtils } from '@shared/OperatingSpeeds';
+import { maxCertifiedAlt, Mmo, Vmo } from '@shared/PerformanceConstants';
+import { CostIndexMode, FmgcData, TakeoffDerated, TakeoffPowerSetting } from 'instruments/src/MFD/FMC/fmgc';
 import {
   AltitudeFormat,
   AltitudeOrFlightLevelFormat,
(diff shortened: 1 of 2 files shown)


fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
```diff
@@ -989,6 +990,37 @@ export class MfdFmsPerf extends FmsPage<MfdFmsPerfProps> {
             this.crzTablePredLine2.set(null);
           }
 
+          // Compute LRC
+          const pd = this.loadedFlightPlan?.performanceData;
+          const crzFL = pd?.cruiseFlightLevel?.get();
+          const grossWeightKg = SimVar.GetSimVarValue('TOTAL WEIGHT', 'pounds') * 0.453592;
+          const isaDev =
+            SimVar.GetSimVarValue('AMBIENT TEMPERATURE', 'celsius') -
```
Benjozork, line 998: Values from the ADIRS should be used, not magic sim values.

fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
```diff
@@ -989,6 +990,37 @@ export class MfdFmsPerf extends FmsPage<MfdFmsPerfProps> {
             this.crzTablePredLine2.set(null);
           }
 
+          // Compute LRC
+          const pd = this.loadedFlightPlan?.performanceData;
+          const crzFL = pd?.cruiseFlightLevel?.get();
+          const grossWeightKg = SimVar.GetSimVarValue('TOTAL WEIGHT', 'pounds') * 0.453592;
```
Benjozork, line 996: I think we may have a system that computes gross weight. Also, avoid conversions using magic numbers. @tracernz what's the conversion utilities we should use here?

Benjozork (changes requested):

fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
tracernz: The FMS indeed calculates GW itself already and that should be used. 

For future: when reading a simvar you can specify the units you want and the sim will do the conversion for you. If you need to do manual conversions in instruments with msfs-sdk you can use UnitType to do the conversion.

fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsPerf.tsx
```diff
@@ -989,6 +990,37 @@ export class MfdFmsPerf extends FmsPage<MfdFmsPerfProps> {
...
+          const isaDev =
+            SimVar.GetSimVarValue('AMBIENT TEMPERATURE', 'celsius') -
+            SimVar.GetSimVarValue('STANDARD ATM TEMPERATURE', 'celsius');
+          if (this.activeFlightPhase.get() >= FmgcFlightPhase.Cruise && crzFL > 0) {
+            const lrc = new Lrc({
```
tracernz, line 1001: We shouldn’t need to allocate new objects all the time for things like this. We want to avoid allocations as much as reasonably possible.

BravoMike99:
Could you perhaps also set the calculated value as the managed speed target, when the active cost index mode is LRC?
`getManagedCruiseSpeed` & `getManagedCruiseSpeedMach`  in fmgc.ts

donstim:
More information is needed on exactly what you are using for lift, drag, and fuel flow parameters as a function of weight, altitude, temperature, and speed.

PR #10476, opened 2026-01-13
fix(a380/fws): Suppress BTV, ROW, ROP, ARPT NAV faults as long as engines are not running

flogross89:
## Summary of Changes

Due to hydraulics not being supplied and engines not running, the FWS will report BTV, ROW/ROP and ARPT NAV FAULTs when setting up the aircraft. These faults and INOP SYS should however not show, as they are expected during that phase. 

## Additional context

## Testing instructions

none

### DIFF
--- a/fbw-a380x/src/systems/systems-host/PseudoPRIM/BrakeToVacate.ts
+++ b/fbw-a380x/src/systems/systems-host/PseudoPRIM/BrakeToVacate.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2023-2025 FlyByWire Simulations
 // SPDX-License-Identifier: GPL-3.0
 
-import { ConsumerSubject, EventBus, Instrument, MappedSubject, SimVarValueType } from '@microsoft/msfs-sdk';
+import { ConsumerSubject, EventBus, Instrument, MappedSubject } from '@microsoft/msfs-sdk';
 import {
   Arinc429LocalVarConsumerSubject,
   Arinc429RegisterSubject,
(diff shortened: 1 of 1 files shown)


fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
```diff
@@ -1530,54 +1532,62 @@ export class FwsCore {
...
+    this.eng1Or2AndEng3Or4RunningAndPhase,
   );
 
   public readonly oansFailedSimvar = RegisteredSimVar.createBoolean('L:A32NX_OANS_FAILED');
   public readonly oansFailed = Subject.create(false);
   public readonly oansPposLostSimVar = RegisteredSimVar.createBoolean('L:A32NX_ARPT_NAV_POS_LOST');
   public readonly oansPposLost = Subject.create(false);
   public readonly arptNavLost = MappedSubject.create(
-    ([oansFailed, pposLost]) => oansFailed || pposLost,
+    ([oansFailed, pposLost, engRunning]) => engRunning && (oansFailed || pposLost),
```
BravoMike99: The inop system could be directly hooked to the OANS failed. The fault itself should check for oansFailed && powered.

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
```diff
@@ -1530,54 +1532,62 @@ export class FwsCore {
...
 
   public readonly ldgPerfAffected = MappedSubject.create(
-    ([w1, w2]) =>
-      w1.bitValueOr(21, false) ||
-      w2.bitValueOr(21, false) ||
-      w1.bitValueOr(23, false) ||
-      w2.bitValueOr(23, false) ||
-      w1.bitValueOr(25, false) ||
-      w2.bitValueOr(25, false),
+    ([w1, w2, engRunning]) =>
```
BravoMike99: I think we can just inhibit it in phase 1 and 10 or if allEnginesFailure is active, and the same for ldg dist.

BravoMike99 (changes requested):

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
flogross89: good point

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
flogross89: Couldn't find any references for the OANS power supply but good idea

BravoMike99: Looks like it is AC4 while the CMV is DC1 (failure of CMV will render OANS inop) so probably inhibited in those two cases..

PR #10477, opened 2026-01-14
fix(a32nx/fwc): spurious low diff pr warnings

tracernz:
Fixes #[issue_no]

## Summary of Changes

Improve the logic for the CAB PR LO DIFF PR warning so it does not trigger spuriously for brief periods during descent when near the limit, and does not suggest reducing V/S if the cabin altitude is too high.

## Additional context

## Testing instructions

I don't think can reasonably be tested that well, other than descending from a high cruise to a low altitude with speedbrakes out and ensuring the warning does get triggered eventually.


fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1305,6 +1306,12 @@ export class PseudoFWC {
 
   private readonly voiceVhf3 = Subject.create(0);
 
+  private readonly excessCabAlt = MappedSubject.create(
```
FozzieHi, line 1317: I know this is a bit of a simplified representation of `excessCabAlt`, but do we want to add the 1s confirm node to this as well?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -4592,7 +4600,7 @@ export class PseudoFWC {
       // LO DIFF PR
       flightPhaseInhib: [2, 3, 4, 5, 7, 8, 9, 10],
       simVarIsActive: this.lowDiffPress,
-      whichCodeToReturn: () => [0, 1, 2],
+      whichCodeToReturn: () => [0, 1, this.excessCabAlt.get() ? null : 2],
```
FozzieHi, line 4618: Do we also want to add this to other codes? For example `213123304` should be checking `excessCabAlt`, but is currently checking `excessPressure` directly. This would also tie in with the confirm node I mentioned above.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2332,7 +2339,12 @@ export class PseudoFWC {
       this.excessResidualPrConfirm.write(this.enginesOffAndOnGroundSignal.read() && residualPressureSignal, deltaTime),
     );
 
-    this.lowDiffPress.set(activeCpc.bitValueOr(15, false));
+    this.lowDiffPress.set(
+      this.lowDiffPressConfirm.write(
+        this.cpc1DiscreteWord.bitValue(15) || this.cpc2DiscreteWord.bitValue(15),
```
BravoMike99, line 2359: Perhaps add a bitValueOf in case the cpcs are setting the bits regardless of failure state?

tracernz: The CPCs should be fixed if they are doing that. Will check.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
tracernz: I would like to keep the scope to just the warning with the bug that is being addressed here. It’s very easy for the scope to become massive with this kind of work.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
tracernz: You were correct, the CPCU logic was not right. Fixed it so bits are not sent when failed. cc @mjuhe

BravoMike99 (approved):

BravoMike99:
Dev team QA report

Object of testing : #10477  
Aircraft          : A32NX 
Simulator         : 2020 
Tier of Testing   : 1
Date              : 25/01/2026 

Testing Process: 
Slew to FL340, selected expedite descent, full speed-brake and waited till the LO DIFF pressure message appears.
Checked that LOW DIFF PRESS ECAM appeared.
Checked that with high cabin altitude the A/C V/S reduce line does not appear.

Testing Results: 
Passed

PR #10478, opened 2026-01-14
ci: direct cdn upload

FoxtrotSierra6829:
## Summary of Changes

- removes stale Bunny CDN script
- updates upload to Cloudflare R2 CDN to use rclone instead of our worker. See https://github.com/flybywiresim/installer/pull/538
- Also fixes purging. See https://github.com/flybywiresim/installer/pull/520 and https://github.com/flybywiresim/installer/pull/523 (hotfix)

## Additional context

foxtrotsierra

## Testing instructions

Not applicable

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanService.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanService.ts
@@ -210,7 +210,7 @@ export class FlightPlanService<P extends FlightPlanPerformanceData = FlightPlanP
     }
 
     if (temporaryPlan.alternateFlightPlan.pendingAirways) {
-      await temporaryPlan.finaliseAirwayEntry();
+      await temporaryPlan.alternateFlightPlan.finaliseAirwayEntry();
     }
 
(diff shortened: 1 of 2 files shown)


FozzieHi:
I'll take a proper look later today, but I guess the same comments I made on `cf-cdn.sh` in #10479 will apply here depending on what one gets merged first.

FoxtrotSierra6829:
> I'll take a proper look later today, but I guess the same comments I made on `cf-cdn.sh` in #10479 will apply here depending on what one gets merged first.

This one gets merged first. This PR is the reason why the other has a `Do not merge` tag

FozzieHi:
> This one gets merged first. This PR is the reason why the other has a `Do not merge` tag

I should've probably commented on this one first then, my bad! Everything else LGTM, might be worth setting the `permissions` on these as well (even if it's just top-level), but it's running against a trusted target so is less vital than the other one.

FoxtrotSierra6829:
> might be worth setting the `permissions` on these as well (even if it's just top-level), but it's running against a trusted target so is less vital than the other one.

I'll do separate PR's on all relevant repos and ask for the default organisation settings to be changed to be more restrictive

PR #10479, opened 2026-01-14
ci: upload QA builds for installer

FoxtrotSierra6829:
Requires #10478
## Summary of Changes

- [x] uploads PRs to Cloudflare CDN
- [x] updates QA config when PR is opened or updated
- [x] updates QA config when PR is closed
- [x] removes CDN files when PR is closed

## Screenshots (if necessary)

grafik

## References

https://github.com/FoxtrotSierra6829/fbw-aircraft-test/pull/7
https://github.com/FoxtrotSierra6829/fbw-aircraft-test/pull/8
https://github.com/FoxtrotSierra6829/fbw-aircraft-test/pull/9
https://github.com/FoxtrotSierra6829/fbw-aircraft-test/pull/10

## Additional context

https://github.com/flybywiresim/installer/pull/537

foxtrotsierra

## Testing instructions

1. Download the FlyByWire installer with QA functionality https://github.com/flybywiresim/installer/releases/tag/v3.6.0
2. Go into Settings.
3. Hold CTRL+ALT and click on the text Settings (in the navigation bar)
4. Go into Developer Settings
5. In QA Configuration URLs click add URL and add `https://flybywirecdn.com/installer/qa-config/pull-requests.json`
6. Click Reload Installer
7. Download/update/change your A32NX and A380X versions to those from the QA builds selection (PR's 7-10) and confirm they work without issues.

_NOTE: Clicking the PR title in the description will lead to a different PR, this is not an issue but working as intended. The test PRs are from another repo._
 

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/FMC/FlightManagementComputer.ts
+++ b/fbw-a380x/src/systems/instruments/src/MFD/FMC/FlightManagementComputer.ts
@@ -505,11 +505,9 @@ export class FlightManagementComputer implements FmcInterface {
           LOWEST_FUEL_ESTIMATE_KGS,
         );
       } else {
-        return Units.poundToKilogram(
-          Math.max(
-            destPred.estimatedFuelOnBoard - (this.fmgc.data.minimumFuelAtDestination.get() ?? 0),
-            A380AircraftConfig.vnavConfig.LOWEST_FUEL_ESTIMATE,
-          ),
+        return Math.max(
+          Units.poundToKilogram(destPred.estimatedFuelOnBoard) - (this.fmgc.data.minimumFuelAtDestination.get() ?? 0),
+          LOWEST_FUEL_ESTIMATE_KGS,
         );
       }
     }
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnFixInfo.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/F-PLN/MfdFmsFplnFixInfo.tsx
@@ -3,7 +3,6 @@ import { ObservableFlightPlan } from '@fmgc/flightplanning/plans/ObservableFligh
 import { FlightPlanIndex } from '@fmgc/flightplanning/FlightPlanManager';
 import {
   ComponentProps,
-  DateTimeFormatter,
   DisplayComponent,
   FSComponent,
   NumberFormatter,
@@ -27,6 +26,7 @@ import { FlightPlanFooter } from '../../common/FlightPlanFooter';
 
 import './MfdFmsFplnFixInfo.scss';
 import { Fix } from '@flybywiresim/fbw-sdk';
+import { hhmmFormatter } from '../../../shared/utils';
 
 export class MfdFmsFplnFixInfo extends FmsPage {
   private readonly flightPlanManager = new ObservableFlightPlanManager(
@@ -251,8 +251,6 @@ class FixInfoPredictionRow extends DisplayComponent<FixInfoPredictionRowProps> {
 
   private readonly ete = Subject.create(NaN);
 
-  private readonly eteFormatter = DateTimeFormatter.create('{HH}:{mm}', { nanString: '--:--' });
-
   private readonly eteText = Subject.create('');
 
   private readonly distance = Subject.create(NaN);
@@ -276,7 +274,7 @@ class FixInfoPredictionRow extends DisplayComponent<FixInfoPredictionRowProps> {
   onAfterRender(node: VNode) {
     super.onAfterRender(node);
 
-    this.subscriptions.push(this.ete.pipe(this.eteText, this.eteFormatter));
+    this.subscriptions.push(this.ete.pipe(this.eteText, hhmmFormatter));
     this.subscriptions.push(this.distance.pipe(this.distanceText, this.distanceFormatter));
     this.subscriptions.push(this.distance.pipe(this.distanceUnitVisible, (distance) => Number.isFinite(distance)));
     this.subscriptions.push(this.altitude.pipe(this.altitudeText, this.altitudeFormatter));
--- a/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsFuelLoad.tsx
+++ b/fbw-a380x/src/systems/instruments/src/MFD/pages/FMS/MfdFmsFuelLoad.tsx
@@ -29,7 +29,7 @@ import { FmsPage } from 'instruments/src/MFD/pages/common/FmsPage';
 import { MfdSimvars } from 'instruments/src/MFD/shared/MFDSimvarPublisher';
 import { FmgcFlightPhase } from '@shared/flightphase';
 import { AirlineModifiableInformation } from '@shared/AirlineModifiableInformation';
-import { getEtaFromUtcOrPresent } from '../../shared/utils';
+import { getEtaFromUtcOrPresent, hhmmFormatter } from '../../shared/utils';
 import { DropdownMenu } from 'instruments/src/MsfsAvionicsCommon/UiWidgets/DropdownMenu';
 import { CostIndexMode } from '../../FMC/fmgc';
 
@@ -87,7 +87,7 @@ export class MfdFmsFuelLoad extends FmsPage<MfdFmsFuelLoadProps> {
   private readonly extraFuelWeightText = this.extraFuelWeight.map((it) => (it ? (it / 1000).toFixed(1) : '---.-'));
 
   private readonly extraFuelTime = Subject.create<number | null>(null);
-  private readonly extraFuelTimeText = this.extraFuelTime.map((it) => new TimeHHMMFormat().format(it ?? 0));
+  private readonly extraFuelTimeText = this.extraFuelTime.map((it) => hhmmFormatter(it ?? NaN));
 
   private readonly blockLineRef = FSComponent.createRef<HTMLDivElement>();
 
--- a/fbw-a380x/src/systems/instruments/src/MFD/shared/utils.ts
+++ b/fbw-a380x/src/systems/instruments/src/MFD/shared/utils.ts
@@ -1,7 +1,8 @@
-// Copyright (c) 2023-2025 FlyByWire Simulations
+// Copyright (c) 2023-2026 FlyByWire Simulations
 // SPDX-License-Identifier: GPL-3.0
 
 import { Approach, ApproachType } from '@flybywiresim/fbw-sdk';
+import { DateTimeFormatter } from '@microsoft/msfs-sdk';
 
 export function getEtaFromUtcOrPresent(seconds: number | null | undefined, fromPresent: boolean) {
   if (seconds === null || seconds === undefined) {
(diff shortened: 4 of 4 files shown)


.github/workflows/pr-build-8k.yml
```diff
@@ -48,28 +48,30 @@ jobs:
...
-          echo SENTRY_DSN=\"${{ secrets.SENTRY_DSN }}\" >> fbw-a380x/.env
+          echo CLIENT_ID=\"\" >> fbw-a380x/.env
+          echo CLIENT_SECRET=\"\" >> fbw-a380x/.env
+          echo CHARTFOX_SECRET=\"\" >> fbw-a380x/.env
+          echo SENTRY_DSN=\"\" >> fbw-a380x/.env
           echo AIRCRAFT_PROJECT_PREFIX=${{ env.AIRCRAFT_PROJECT_PREFIX }} >> fbw-a380x/.env
           echo AIRCRAFT_VARIANT=${{ env.AIRCRAFT_VARIANT }} >> fbw-a380x/.env
           echo BUILD_DIR_NAME=${{ env.BUILD_DIR_NAME }} >> .env
-          echo LOCALAZY_READ_KEY=${{ secrets.LOCALAZY_READ_KEY }} >> .env
+          echo LOCALAZY_READ_KEY=\"\" >> .env
```
tracernz: Why do we want to remove these ones? That will prevent us from testing the Navigraph integration?

FoxtrotSierra6829: > # Additional context
> Secrets are now only accessed when handling code of the trusted targeted branch. Previously, secrets could be accessed on PRs from upstream branches but were inaccessible from PRs from forks.

This is running untrusted code which can potentially leak secrets. Previously the secrets just weren't available on fork PRs but now since the context is the target branch but except for the workflow file we are running PR code until later, these secrets would be available to be dumped.

.github/workflows/pr-build.yml
```diff
@@ -92,19 +100,63 @@ jobs:
...
+            [cloudflare-r2]
+            type = s3
+            provider = Cloudflare
+            env_auth = false
+            access_key_id = ${{ secrets.CLOUDFLARE_R2_ACCESS_KEY_ID }}
+            secret_access_key = ${{ secrets.CLOUDFLARE_R2_SECRET_ACCESS_KEY }}
+            endpoint = ${{ secrets.CLOUDFLARE_R2_ENDPOINT }}
+            region = auto
+            EOF
+      - name: Upload to CloudFlare CDN (Installer)
```
FozzieHi: Are you missing a `if: github.event.pull_request.auto_merge == false` condition here (And above for Checking out, and configuring Rclone?), otherwise there wouldn't be anything to upload as `/tmp/build-modules-a32nx` wouldn't exist?

scripts/cf-cdn.sh
```diff
@@ -1,39 +1,87 @@
 #!/bin/bash
```
FozzieHi, line 3: I know this isn't a new script, but worth adding a `set -eo pipefail` or equivilent so that the run fails if rclone fails (or something else)?

.github/workflows/pr-build.yml
FoxtrotSierra6829: Whoopsie, yes, didn't copy this over

scripts/cf-cdn.sh
FoxtrotSierra6829: Shouldn't rclone already exit non-zero on fail? The purging explicitly exits 1 when not successful

FozzieHi: rclone will return a non-zero exit code on failure yeah, the issue is that the script won't exit when it receives that non-zero exit code without `set -e` being set (set +e is the default) https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#index-set

FoxtrotSierra6829: Ahh right, one more reason I try to avoid bash scripts when possible😅

FozzieHi: Same here! I'm sure it's a useful default for just running a few commands, but not really for entire scripts

.github/workflows/pr-build-8k.yml
tracernz: The navigraph "secrets" are in plain text in the packages we distribute to users. The actions will still not run without dev team approval on outside PRs, right?

FoxtrotSierra6829: Yes. If you think it's safe enough to allow access to them, I can put them back, sure

tracernz: If you could add those 2 Navigraph ones it would be helpful for testing. The others are irrelevant (chartfox, sentry) and actually secret (localazy).

FoxtrotSierra6829: Actually no, pull_request_target workflows run regardless of the settings for workflow approval
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository

tracernz: Alright, we should probably keep your changes then. 👍

FoxtrotSierra6829: We could otherwise create a manual approval process similar to https://trstringer.com/github-actions-manual-approval , GitHub: https://github.com/trstringer/manual-approval but within the PR

tracernz: Let's see if it's an issue, and if it is tackle that in a different PR. This one is very valuable!

FoxtrotSierra6829:
QA not required, QA was performed with https://github.com/flybywiresim/installer/pull/537 and test PRs using these changes

tracernz:
Once straks sorts the workflows, this PR should actually create the first real QA build in the JSON?

FoxtrotSierra6829:
No this PR will not run the workflows at all because they have to run from the target branch (master) but do not exist there yet. Only runs triggered after the merge will upload it. But the fake QA builds were uploaded using the same workflows.

FoxtrotSierra6829:
On request the workflows were split so that the build continues to be triggered as `pull_request` instead of `pull_request_target`, which reduces possible security concerns regarding exploitability through outside contributors even after previous mitigation. The upload itself is triggered by `workflow_run` instead and runs in the context of the default branch after completion of the `pull_request` workflows. The fragmenter modules are passed between the workflows using artifacts.

.github/workflows/pr-closed.yml
```diff
@@ -0,0 +1,69 @@
...
+            EOF
+      - name: Remove QA config entries
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          ADDON_KEYS: a32nx-msfs2020,a32nx-msfs2024,a380x-msfs2020,a380x-msfs2024
+          CLOUDFLARE_CDN_ZONE_ID: ${{ secrets.CLOUDFLARE_CDN_ZONE_ID }}
+          CLOUDFLARE_PURGE_TOKEN: ${{ secrets.CLOUDFLARE_PURGE_TOKEN }}
+          CDN_BUCKET_DESTINATION: installer/qa-config
+        run: |
+          node ./scripts/remove_qa_config.js
```
pdellaert, line 58: Is there a risk where someone creates a PR with a change to these files (because they are taken from the PR), with something trying to steal secrets, then immediately closes it, that this could leak things?

Wonder if workflow splitting here could also work here? Trigger this workflow (that barely does anything), which then is monitored and then triggers a delete and qa-config change?

.github/workflows/pr-updated.yml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+          ADDON_KEYS: a32nx-msfs2020,a32nx-msfs2024,a380x-msfs2020,a380x-msfs2024
+          CLOUDFLARE_CDN_ZONE_ID: ${{ secrets.CLOUDFLARE_CDN_ZONE_ID }}
+          CLOUDFLARE_PURGE_TOKEN: ${{ secrets.CLOUDFLARE_PURGE_TOKEN }}
+          CDN_BUCKET_DESTINATION: installer/qa-config/
+        run: |
+          node ./scripts/update_qa_config.js
```
pdellaert, line 45: Same here with workflow splitting. Seems like it could rely on the pr-upload and pr-upload-8k workflows here as well, just like the build ones?

.github/workflows/pr-upload-8k.yml
```diff
@@ -0,0 +1,113 @@
...
+            region = auto
+            EOF
+      - name: Upload A380X to CloudFlare CDN (Installer)
+        if: steps.artifacts.outputs.found == 'true'
+        env:
+          CLOUDFLARE_CDN_ZONE_ID: ${{ secrets.CLOUDFLARE_CDN_ZONE_ID }}
+          CLOUDFLARE_PURGE_TOKEN: ${{ secrets.CLOUDFLARE_PURGE_TOKEN }}
+          CDN_BUCKET_DESTINATION: addons/a380x/pr-${{ steps.pr.outputs.number }}-8k
+        run: |
+          ./scripts/cf-cdn.sh /tmp/artifacts/build-modules-a380x-8k $CDN_BUCKET_DESTINATION
```
pdellaert, line 99: Given this still executes code from the PR in a workflow with access to secrets, I still suggest to use a label in the pr-build.yaml & pr-build-8k.yml dependencies, that must be added. 

Or at a minimum, force the setting that all non-team member fork pull request workflow runs must be approved, for all contributors (even if they are second, third, ... time).

pdellaert (commented):
Think it's a great move, but believe the closed and updated ones could also move to a split model. Bit annoying, but more secure, as these still run code from the PR (scripts), so while the workflow itself runs from master, the scripts inside run from the PR, which can be maliciously manipulated.

.github/workflows/pr-upload-8k.yml
FoxtrotSierra6829: It does not execute code from the PR but from the default branch.

.github/workflows/pr-closed.yml
FoxtrotSierra6829: It does not execute code from the PR but from the base branch. I'll change it to the default branch so it's even safer. (In 99% of the cases they're the same branch)

.github/workflows/pr-updated.yml
FoxtrotSierra6829: It does not execute code from the PR but from the base branch. I'll change it to the default branch so it's even safer. (In 99% of the cases they're the same branch)

FoxtrotSierra6829:
> Think it's a great move, but believe the closed and updated ones could also move to a split model. Bit annoying, but more secure, as these still run code from the PR (scripts), so while the workflow itself runs from master, the scripts inside run from the PR, which can be maliciously manipulated.

This is not the case. The scripts are not executed from PR code. Since these only handle metadata of the PR, PR code is not necessary to be pulled.

pdellaert (approved):

ghostxyz990:

Discord Username  : adb_1872
Object of testing : #10479 
Aircraft          : A380X/A32NX 
Simulator         : 2020 
Tier of Testing   : 1
Date              : 05/02/2026 

Testing Process: 
Download the FlyByWire installer with QA functionality https://github.com/flybywiresim/installer/releases/tag/v3.6.0
Go into Settings.
Hold CTRL+ALT and click on the text Settings (in the navigation bar)
Go into Developer Settings
In QA Configuration URLs click add URL and add https://flybywirecdn.com/installer/qa-config/pull-requests.json
Click Reload Installer
Download/update/change your A32NX and A380X versions to those from the QA builds selection (PR's 7-10) and confirm they work without issues.

Testing Results: 
Passed  

Negatives:


Conclusions: 


Media:

PR #10487, opened 2026-01-17
feat(a32nx/fms): Estimated Takeoff Time

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Implements manual estimated takeoff time entry on the RTA page & associated effects on predictions.
## Screenshots (if necessary)

image
image

## Additional context

## Testing instructions

Program FMS so that you have predictions then:
1. Verify time predictions headers appear as "UTC" on the F-PLN, PERF, STEP ALTS & FUEL PRED page (accessible after engine start) and that they are displayed correctly in relation to the absolute time/expected flight time.
2. Perform a vertical revision on any waypoint and access the RTA page, check that ETT can be inserted with 3, 4 5 or 6 digits. Anything above 23 hours or 59 minutes or 59 seconds or beyond 20hours of current time should be rejected (ENTRY OUT OF RANGE).
Once a valid ETT is inserted, verify it appears as magenta and all previous time predictions now show "UTC" as the header and are in relation to the ETT, rather than absolute time. On the F-PLN page, origin should have the ETT time displayed as magenta.
3. Wait till the ETT expires. Check that "CLK TIME IS TAKE OFF TIME" message appears and that time predictions are continuously updated based on the current time. On the RTA page the ETT should no longer be visible.
4. Insert an ETT in a past hour (following day).  Change the sim time and date and verify the message appears once again once the time is expired. Do keep in mind that anything ahead of 20 hours of present time will be rejected. E.g. If time is 11:17Z, anything between 07:17 to 11:17 won't be accepted.
5. Manually create a secondary flight-plan and insert an ETT. Wait till it expires and then activate or swap secondary. Check that the scratch pad message immediately appears assuming there wasn't a previously expired ETT.
6. In flight check that the time predictions headers appear as "TIME" and display an arrival time. ETT insertion should only be possible at this point on secondary flight plans which were created manually (not via copy active or swap).


fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
```diff
@@ -603,6 +610,8 @@ export class A320FlightPlanPerformanceData implements FlightPlanPerformanceData
     );
     other.pipe('pilotFinalHoldingFuel', this.pilotFinalHoldingFuel, other.pilotFinalHoldingFuel);
     other.pipe('pilotFinalHoldingTime', this.pilotFinalHoldingTime, other.pilotFinalHoldingTime);
+    other.pipe('estimatedTakeoffTime', this.estimatedTakeoffTime, other.estimatedTakeoffTime);
```
BlueberryKing, line 613: You should only pipe these properties if an ETT entry should automatically be copied from the active to an existing secondary flight plan created as a copy from the active plan. Are we sure this is the case?

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -5443,7 +5449,20 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
     const phase = this.getFlightPhase();
 
     if (phase === FmgcFlightPhase.Preflight || phase === FmgcFlightPhase.Done) {
+      const ett = this.flightPlanService.active.performanceData.estimatedTakeoffTime.get();
+      if (ett !== null) {
+        if (this.checkEttExpired()) {
+          this.flightPlanService.active.performanceData.estimatedTakeoffTimeExpired.set(true);
+        } else {
+          this.ettInterval = setInterval(() => {
```
BlueberryKing: Is it possible that we already have an `ettInterval` that we should clear before creating a new one?

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -5627,6 +5646,158 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
...
+
+    // If its not active or copy of active allow setting in any phase
+    if (!activeOrCopyOfActive || this.getFlightPhase() === FmgcFlightPhase.Preflight) {
+      // If ETT has expired we don't show it on the RTA page, as such, ignore clearing if thats the case
+      if (!flightplan.performanceData.estimatedTakeoffTimeExpired.get() && text == Keypad.clrValue) {
+        this.clearEttAndInterval(forPlan);
+        return;
+      }
+
+      if (/^\d{3,6}$/.exec(text)) {
```
BlueberryKing: You could use `/(\d{1,2})(\d{2})(\d{0,2})/` as a regex to automatically match the hours, minutes, seconds. That way, you don't have to use `.split`

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -5627,6 +5646,158 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
...
+        let minutes: number;
+        let hours: number;
+        let seconds: number = 0;
+        if (text.length <= 4) {
+          const is24H = text.length === 4;
+          hours = Number(text.slice(0, is24H ? 2 : 1));
```
BlueberryKing: We should use `parseInt` to parse strings

fbw-a32nx/src/systems/instruments/src/MCDU/legacy_pages/A320_Neo_CDU_VerticalRevisionPage.ts
```diff
@@ -196,12 +199,16 @@ export class CDUVerticalRevisionPage {
       }
     }
 
+    // Disable RTA if EET insertion not possible as nothing else is simulated for the time being
```
BlueberryKing:
```suggestion
    // Disable RTA if ETT insertion not possible as nothing else is simulated for the time being
```

fbw-a32nx/src/systems/instruments/src/MCDU/legacy_pages/CDU_RTA_Page.ts
```diff
@@ -0,0 +1,87 @@
...
+    index: number,
+    verticalWaypoint: VerticalWaypointPrediction,
+    forplan: FlightPlanIndex,
+  ): void {
+    mcdu.clearDisplay();
+    mcdu.page.Current = mcdu.page.RTAPage;
+    const isPreFlight = mcdu.flightPhaseManager.phase === FmgcFlightPhase.Preflight;
+    const flightplan = mcdu.flightPlanService.get(forplan);
+    const ett = flightplan.performanceData.estimatedTakeoffTime.get();
+    const ettExists = ett != null && !flightplan.performanceData.estimatedTakeoffTimeExpired.get();
```
BlueberryKing:
```suggestion
    const ettExists = ett !== null && !flightplan.performanceData.estimatedTakeoffTimeExpired.get();
```

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
BravoMike99: Looks like it carries over as per the latest references. Others should be addressed now

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
```diff
@@ -1231,6 +1240,16 @@ export class A320FlightPlanPerformanceData implements FlightPlanPerformanceData
    */
   readonly approachFlapsThreeSelected = Subject.create(false);
 
+  /**
+   * Estimated takeoff time in seconds since 1/1/AD
```
tracernz: I would suggest instead to use JS timestamps for all actual times. That is milliseconds since the UNIX epoch. You can then use JS Date functions, and any date/time functions from msfs-sdk directly, and also the ClockEvents topics in msfs-sdk.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -322,6 +321,14 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
     SimVarValueType.Enum,
   );
 
+  /** Seconds since midnight */
+  private readonly zuluTime = RegisteredSimVar.create<number>('E:ZULU TIME', SimVarValueType.Seconds);
+
+  /** Seconds since 1/1/1AD */
+  private readonly absoluteTime = RegisteredSimVar.create<number>('E:ABSOLUTE TIME', SimVarValueType.Seconds);
```
tracernz: Use ClockEvents instead rather than having a bunch of different things in the instrument reading the same simvars. They also do the conversion to JS timestamp format for you.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -5629,6 +5650,150 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
...
+            this.setScratchpadMessage(NXSystemMessages.entryOutOfRange);
+          } else {
+            this.flightPlanService.setPerformanceData(
+              'estimatedTakeoffTime',
+              Math.round(this.absoluteTime.get() + ettSecondsFromNow),
+              forPlan,
+            );
+            this.flightPlanService.setPerformanceData('estimatedTakeoffTimeExpired', false, forPlan);
+            if (activeOrCopyOfActive && this.ettInterval === null) {
+              this.ettInterval = setInterval(() => {
```
tracernz: Better to do these kinds of things in the normal instrument update loop rather than setInterval/setTimeout that run outside the VCockpit scheduler.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/FmsFormatters.ts
```diff
@@ -42,4 +46,21 @@ export class FmsFormatters {
     const m = parseInt(hhmm.substring(2, 4));
     return h * 60 + m;
   }
+
+  public static secondsTohhmmss(seconds: number): string {
+    const h = Math.floor(seconds / 3600);
+    const m = Math.floor((seconds - h * 3600) / 60);
+    const s = Math.floor(seconds - h * 3600 - m * 60);
+    return h.toFixed(0).padStart(2, '0') + ':' + m.toFixed(0).padStart(2, '0') + ':' + s.toFixed(0).padStart(2, '0');
+  }
```
tracernz: Use [DateTimeFormatter](https://microsoft.github.io/msfs-avionics-mirror/docs/api/framework/classes/DateTimeFormatter) in the msfs-sdk instead.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/FmsFormatters.ts
```diff
@@ -42,4 +46,21 @@ export class FmsFormatters {
...
+
+  /**
+   * Returns the number of seconds since midnight UTC for a given date.
+   * @param date the date to extract the seconds from
+   * @returns number of seconds since midnight UTC
+   */
+  public static dateToSecondsSinceMidnightUTC(date: Date): number {
+    const midNight = new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate(), 0, 0, 0);
```
tracernz: Use a cache object rather than allocating new dates all the time. If you use JS timestamps like suggested you could set your cache object to the timestamp, then set `setUTCHours(0, 0, 0, 0) on it`. 

Sidenote: I always prefer to pass timestamps (ms since epoch) around in APIs rather than date objects because you can test them for equality (important for subscribables) and easily put them into a date object if you need it. This is the problem with Date objects, and objects in general... they are compared by reference rather than value:
```js
const foo = new Date();
const bar = new Date(foo.getTime())
console.log(foo); // Date Fri Jan 30 2026 23:02:55 GMT+1300 (New Zealand Daylight Time)
console.log(bar); // Date Fri Jan 30 2026 23:02:55 GMT+1300 (New Zealand Daylight Time)
console.log(foo === bar); // false
console.log(foo.getTime() === bar.getTime()); // true
```

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
tracernz: These plans never get destroyed right? If they do these pipes will leak and keep the related objects alive unless something explicitly destroys them.

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
BravoMike99: Initially I was using epoch but then ended up swapping to absolute time due to having to convert from the msfs time simvars. Clock events and what not will make it much simpler 👍

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
BlueberryKing: We should destroy the performance data when the flight plan is destroyed

tracernz: You specifically need to destroy subscriptions to external subscribables as they won't be destroyed when the perf plan is, and will holds refs in their subscription list.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/FmsFormatters.ts
BravoMike99: This was leftover code which I forgot to delete 😅 . It's now gone

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
BravoMike99: All changed to use clock events and epoch now 👍

BlueberryKing (approved):

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts
BravoMike99: It looks like the pipe function is storing the subs which then get destroyed whenever the PD destroy method is called. That seems to be when the flight-plan is deleted. https://github.com/flybywiresim/aircraft/blob/fb834c87a19aecd6100fdb1bebe497fe6afd6905/fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/performance/FlightPlanPerformanceData.ts#L630
https://github.com/flybywiresim/aircraft/blob/fb834c87a19aecd6100fdb1bebe497fe6afd6905/fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/FlightPlan.ts#L72

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -321,17 +321,25 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
     SimVarValueType.Enum,
   );
 
-  /** Seconds since midnight */
-  private readonly zuluTime = RegisteredSimVar.create<number>('E:ZULU TIME', SimVarValueType.Seconds);
+  /** Unix epoch in milliseconds. Paused when no ETT present in active flightplan */
+  private readonly simTime = ConsumerSubject.create(
+    this.bus.getSubscriber<ClockEvents>().on('simTime').onlyAfter(1000),
+    0,
+  );
```
tracernz: I'd be vary wary of doing this in the main FMS class, as people are likely to use it and expect it to update every Update() from the publisher. Better to have an explicit ETT subscription that has is throttled and gets paused and resumed.

e.g. 

```ts
private readonly sub = this.bus.getSubscriber<ClockEvents>();

/** Simulation time in milliseconds since the UNIX epoch (JS timestamp). Hint: this clock is affected by sim rate. */
private readonly simTime = ConsumerSubject.create(this.sub.on('simTime'), 0);

private readonly ettUpdateSub = this.sub.on('realTime').onlyAfter(1000).handle(this.updateEtt.bind(this), true);
```

BravoMike99: Makes sense thanks, changed now to have a separate subscription.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy_pages/A320_Neo_CDU_FuelPredPage.ts
```diff
@@ -200,19 +198,16 @@ export class CDUFuelPredPage {
       }
 
       if (Number.isFinite(predictions.tripTime)) {
-        destTimeCell = isFlying
-          ? FmsFormatters.secondsToUTC(utcTime + FmsFormatters.minuteToSeconds(predictions.tripTime))
-          : FmsFormatters.minutesTohhmm(predictions.tripTime);
+        destTimeCell = mcdu.getTimePrediction(predictions.tripTime, FlightPlanIndex.Active);
```
FozzieHi: `getTimePrediction()` takes seconds as its first argument, yet isn't `predictions.tripTime` in minutes, which is why the code used to be `FmsFormatters.minuteToSeconds(predictions.tripTime)`?

fbw-a32nx/src/systems/instruments/src/MCDU/legacy_pages/A320_Neo_CDU_FuelPredPage.ts
```diff
@@ -200,19 +198,16 @@ export class CDUFuelPredPage {
...
 
       if (alternate) {
         if (Number.isFinite(predictions.alternateTime) && Number.isFinite(predictions.tripTime)) {
-          altTimeCell = isFlying
-            ? FmsFormatters.secondsToUTC(
-                utcTime + FmsFormatters.minuteToSeconds(predictions.tripTime + predictions.alternateTime),
-              )
-            : FmsFormatters.minutesTohhmm(predictions.tripTime + predictions.alternateTime);
+          altTimeCell = mcdu.getTimePrediction(
+            predictions.tripTime + predictions.alternateTime,
```
FozzieHi: Same as above in regards to `predictions.tripTime + predictions.alternateTime` being in minutes.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy_pages/A320_Neo_CDU_FuelPredPage.ts
BravoMike99: Well spotted, unit mess 😭

FozzieHi: I know we've had a few recently so I was on the lookout! I hate units 🙂

tracernz: Easiest to just do all time related things in milliseconds, then they are compatible with both JS Date class and msfs-sdk date and time topics and functions.

BravoMike99: I think that's something we could leave for the future as that would mean changing the vnav predictions which would affect the  380x too. Maybe when ETT gets brought there it would be a good opportunity to do so as I wanted to keep the score of this PR only to the 32nx.

fbw-a32nx/src/systems/instruments/src/MCDU/McduFsInstrument.tsx
```diff
@@ -35,6 +36,8 @@ export class McduFsInstrument implements FsInstrument {
 
   private readonly legacyFms = new A320_Neo_CDU_MainDisplay(this.bus);
 
+  private readonly clockPublisher = new ClockPublisher(this.bus);
```
tracernz: Clock a few lines above already instantiates this class.

BravoMike99: oops, now gone

calarandir:

Discord Username  : shadowca_
Object of testing : #10487
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 19/02/2026

Testing Process: 
Programmed FMC for a flight from SCEL gate 20 departing 35L
Starting engines and setting ETT 0810 with 3,4,5,6 digits and also out of range times
Checking predictions on F-PLN, PERF, FUEL PRED the later two do not show a valid prediction but the old one.
When reaching takeoff time the message is shown and RTA page does not show ETT anymore
Predictions on PERF page are updated but not related to ETT
Aborted test because it does not pass.

Testing Results:
Not Passed

Negatives:       
The predicted times are not related to ETT. I expected the calculation would take the ETT into account.
Also at some places like ETT entry the time is shown in a 12 hours format, at other places like T/D calculation it is shown in 24 hours format

Media:
image
image
image
image

24 hour fofmat vs 12 hour format:
image

BravoMike99:
> 
> Discord Username : shadowca_ Object of testing : #10487 Aircraft : A32NX Simulator : 2020 Tier of Testing : 1 Date : 19/02/2026
> 
> Testing Process: Programmed FMC for a flight from SCEL gate 20 departing 35L Starting engines and setting ETT 0810 with 3,4,5,6 digits and also out of range times Checking predictions on F-PLN, PERF, FUEL PRED the later two do not show a valid prediction but the old one. When reaching takeoff time the message is shown and RTA page does not show ETT anymore Predictions on PERF page are updated but not related to ETT Aborted test because it does not pass.
> 
> Testing Results: Not Passed
> 
> Negatives: The predicted times are not related to ETT. I expected the calculation would take the ETT into account. Also at some places like ETT entry the time is shown in a 12 hours format, at other places like T/D calculation it is shown in 24 hours format
> 
> Media: image image image image
> 
> 24 hour fofmat vs 12 hour format: image

Thanks for testing. There was indeed an issue with some times showing in 12H format rather than 24H. Should now be fixed

calarandir:

Discord Username  : shadowca_
Object of testing : #10487
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 20/02/2026

Testing Process: 
Programmed FMC for a flight from SCEL gate 20 departing 35L
Starting engines and setting ETT 1018 with 3,4,5,6 digits and also out of range times -> Works as expected
Checking predictions on F-PLN, PERF, FUEL PRED -> Working as expected (see negatives for TODOs)
When reaching takeoff time the message is shown and RTA page does not show ETT anymore.
Predictions on PERF, F_PLN and FUEL PRED page are updated

At 1040z I try to enter a time in the past:

1039, 959, 930 all result in "ENTRY OUT OF RANGE"
520 works fine for some reason
In fact it seems that all times, that are less than 4 hours in the past (So 640 and greater) are out of range. At 1043 I can enter up to 643.

Aborted test because it does not pass

Testing Results:
Not Passed

Negatives:
- Entering times in the past (next day) does not work as expected. -> Test failed

Other Observations
- The FUEL PRED Page does not use the VNAV calculations yet. Therefore time (and fuel) predictions are different than on F_PLN and PERF.
- User documentation on ETT and time prediction shown for example on CRZ page is not sufficient.
- ETA (and EFOB) are shown on CRZ and DES but not on CLB

BravoMike99:
> 
> Discord Username : shadowca_ Object of testing : #10487 Aircraft : A32NX Simulator : 2020 Tier of Testing : 1 Date : 20/02/2026
> 
> Testing Process: Programmed FMC for a flight from SCEL gate 20 departing 35L Starting engines and setting ETT 1018 with 3,4,5,6 digits and also out of range times -> Works as expected Checking predictions on F-PLN, PERF, FUEL PRED -> Working as expected (see negatives for TODOs) When reaching takeoff time the message is shown and RTA page does not show ETT anymore. Predictions on PERF, F_PLN and FUEL PRED page are updated
> 
> At 1040z I try to enter a time in the past:
> 
> 1039, 959, 930 all result in "ENTRY OUT OF RANGE" 520 works fine for some reason In fact it seems that all times, that are less than 4 hours in the past (So 640 and greater) are out of range. At 1043 I can enter up to 643.
> 
> Aborted test because it does not pass
> 
> Testing Results: Not Passed
> 
> Negatives:
> 
>     * Entering times in the past (next day) does not work as expected. -> Test failed
> 
> 
> Other Observations
> 
>     * The FUEL PRED Page does not use the VNAV calculations yet. Therefore time (and fuel) predictions are different than on F_PLN and PERF.
> 
>     * User documentation on ETT and time prediction shown for example on CRZ page is not sufficient.
> 
>     * ETA (and EFOB) are shown on CRZ and DES but not on CLB

The issue you experienced is per design as there's a 20 hours ahead of present time limitation on the real counterpart too. I updated the testing instructions to clarify that on the past hour bit.
The other observations are also the same on master. 
Regarding, documentation ill make up an issue on the docs PR.

calarandir:

Discord Username  : shadowca_
Object of testing : #10487
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 20/02/2026

Testing Process: 
Completing previous test after clarification of the 20 hour rule with same build and setup.

Creating a manual flightplan and entering an ETT (current time +1) and wait 2 minutes
Swapping flightplans creates the expected message instantly.

Testing Results:
Passed


Negatives:       
Bug: After deleting and starting over with the secondary flightplan on FUEL PRED page there is no fuel planning but BLOCK CONFIRM instantly which does nothing on LSK3

Also see negatives in previous test runs

PR #10499, opened 2026-01-24
chore: improve sim var util api

tracernz:
## Summary of Changes

Improve the `RegisteredSimVar` API:
- It now supports strings (`SimVarValueType.String`).
- ~~Unified booleans into `create`. You must use `SimVarValueType.Bool` for booleans.~~ Backed this one out to stay compatible with the msfs-sdk v2 implementation. It is objectively worse, but I guess less maint for us in future.
- A type specifier is no longer required unless you want a specific type for a number var (e.g. enum), and never required for string or bool.

The create signatures are as follows, disallowing bools in the main create to avoid bugs:
```ts
  public static create(name: string, unit: SimVarValueType.String, source?: string): RegisteredSimVar<string>;
  public static create<T extends number, U extends SimVarValueType | string = SimVarValueType | string>(
    name: string,
    unit: U,
    source?: string,
  ): U extends SimVarValueType.Bool | 'bool' | 'boolean' ? never : RegisteredSimVar<T>

  public static createBoolean(name: string, source = ''): RegisteredSimVar<boolean>
  ```

## Additional context

## Testing instructions

Smoke test a plane. No changes expected.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -70,6 +70,8 @@
 1. [A32NX/FWS] Fixed stall warning not working - @tracernz (Mike)
 1. [A380X/MFD] Add initial implementation of D-ATIS page in ATCCOM - @heclak (Heclak)
 1. [A32NX/FWS] Fixed takeoff memo not working in flight phase 9 (after landing) - @tracernz (Mike)
+1. [A380X/FUEL SYSTEM] Fixed inner tank transfer not starting in certain circumstances - @donstim (donbikes)
+1. [A32NX/FADEC] Fixed the PFD showing MAN FLX instead of MAN MCT post A/THR engagement after takeoff - @FozzieHi (Fozzie)
 
 ## 0.14.0
 
--- a/.github/workflows/ingamepanels-checklist-fix.yml
+++ b/.github/workflows/ingamepanels-checklist-fix.yml
@@ -17,6 +17,8 @@ jobs:
     if: github.repository_owner == 'flybywiresim'
     name: 'Build InGamePanels Checklist Fix'
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     env:
       FBW_PRODUCTION_BUILD: 1
       MASTER_ZIP_NAME: ingamepanels-checklist-fix-master.zip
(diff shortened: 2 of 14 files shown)


FozzieHi (commented):
I think you've missed some instances of `createBoolean`, such as https://github.com/flybywiresim/aircraft/blob/bdc6ce92ad9a48528d0de8bf4f26413a00d921da/fbw-a380x/src/systems/systems-host/PseudoPRIM/BrakeToVacate.ts#L68

Guessing you did a find and replace, then rebased onto master which has had some recent additions?

flogross89 (approved):

tracernz:
> I think you've missed some instances of `createBoolean`, such as
> 
> https://github.com/flybywiresim/aircraft/blob/bdc6ce92ad9a48528d0de8bf4f26413a00d921da/fbw-a380x/src/systems/systems-host/PseudoPRIM/BrakeToVacate.ts#L68
> 
> Guessing you did a find and replace, then rebased onto master which has had some recent additions?

Yeah, that's likely. Why does it build okay though... 😕

FozzieHi:
> Why does it build okay though... 😕

Looks like `FBW_TYPECHECK=1` is only set on `fbw-a32nx` for PR builds, and not `fbw-a380x`. Not sure if this is intentional or not, but it could result in some issues with people merging this into their branch just before they merge into master, with already existing `createBoolean` calls for example, and not realising.

https://github.com/flybywiresim/aircraft/blob/9d0528770fa35638fff13541a6f9bb57df631aca/.github/workflows/pr-build.yml#L70-L72
https://github.com/flybywiresim/aircraft/blob/9d0528770fa35638fff13541a6f9bb57df631aca/.github/workflows/pr-build.yml#L155-L157

It is present in `pr-build-8k.yml`, but that's skipped without the `pr-build-8k` label
https://github.com/flybywiresim/aircraft/blob/9d0528770fa35638fff13541a6f9bb57df631aca/.github/workflows/pr-build-8k.yml#L59-L61

tracernz:
> echo FBW_TYPECHECK=1 >> fbw-a380x/.env

Good find. We must have lost it when splitting the 8k build off.

BravoMike99:
Dev team QA report

Object of testing : #10499
Aircraft          :  A380X
Simulator         : 2020 
Tier of Testing   : 1
Date              : 06/02/2026

Testing Process: 
Loaded up the 380x and checked that systems which use registered simvars work correctly (ois page, fws & rmp)

Testing Results: 
Passed

PR #10507, opened 2026-01-25
feat(a32nx/fwc): Add more ELEC master cautions

FozzieHi:
## Summary of Changes
This PR adds more ELEC-based master cautions to the A32NX. Currently `IDG 1(2) DISCONNECTED` and `GEN 1(2) OFF`. I may add `GEN 1(2) FAULT` in a separate PR, but it's a bit more complicated.

## Screenshots (if necessary)

Screenshot 2026-01-25 120625
Screenshot 2026-01-25 140736

## Additional context

## Testing instructions

Probably easiest to spawn on a runway, or otherwise be in flight phase 2 (after 1st engine started), but the test will involve a takeoff.
1. Turn on the APU, wait for it to become available, and check it's been at least 60s since starting the engines before moving onto the next step.
2. Disconnect IDG 1, check that the `IDG 1 DISCONNECTED` master caution triggers after approximately a second.
3. Press the `CLR` button on the ECP, then press the `T.O. CONFIG` button on the ECP, check that it says `NORMAL` on the T.O. memo.
4. Turn off GEN 2, check that `GEN 2 OFF` master caution triggers after approximately five seconds.
5. Press the `CLR` button on the ECP, then press the `T.O CONFIG` button on the ECP, check that both `IDG 1 DISCONNECTED` and `GEN 2 OFF` re-appear.
6. Press the `CLR` button on the ECP, apply T/O power, check that that both `IDG 1 DISCONNECTED` and `GEN 2 OFF` re-trigger.


fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1723,6 +1772,12 @@ export class PseudoFWC {
     this.bat1PbOff.set(!SimVar.GetSimVarValue('L:A32NX_OVHD_ELEC_BAT_1_PB_IS_AUTO', 'bool'));
     this.bat2PbOff.set(!SimVar.GetSimVarValue('L:A32NX_OVHD_ELEC_BAT_2_PB_IS_AUTO', 'bool'));
 
+    this.gen1PbOff.set(!SimVar.GetSimVarValue('A:GENERAL ENG MASTER ALTERNATOR:1', 'bool'));
+    this.gen2PbOff.set(!SimVar.GetSimVarValue('A:GENERAL ENG MASTER ALTERNATOR:2', 'bool'));
+
+    this.idg1Disconnected.set(!SimVar.GetSimVarValue('L:A32NX_ELEC_ENG_GEN_1_IDG_IS_CONNECTED', 'bool'));
+    this.idg2Disconnected.set(!SimVar.GetSimVarValue('L:A32NX_ELEC_ENG_GEN_2_IDG_IS_CONNECTED', 'bool'));
```
BravoMike99: We could use some registered simvars for these.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2989,6 +3044,82 @@ export class PseudoFWC {
       this.iceNotDetTimer2.write(iceNotDetected1 && !(icePercentage >= 0.1 || (tat < 10 && inCloud === 1)), deltaTime),
     );
 
+    /* ELECTRICAL */
+    this.bat1Off.set(this.bat1PbOff.get() && (this.flightPhase6For60Seconds.read() || this.fwcFlightPhase.get() === 2));
+    this.bat2Off.set(this.bat2PbOff.get() && (this.flightPhase6For60Seconds.read() || this.fwcFlightPhase.get() === 2));
+
+    const engine1NotRunning = this.engine1State.get() === 0;
```
BravoMike99: Might be better to create an enum. It could be an util somewhere (there's one in the 380 fwscore file). 
Also, even when the engine is shutdown it has the value of 3 which only resets to 0 once the EGT is not cooling down anymore so you may want to check for that.
Or you could invert the condition and check for engine running instead.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2989,6 +3044,82 @@ export class PseudoFWC {
...
+    const gen1OffWarningConfirmNode2 = this.gen1OffWarningConfirmNode2.write(!engine1NotRunning && !phase6, deltaTime);
+    const gen2OffWarningConfirmNode2 = this.gen2OffWarningConfirmNode2.write(!engine2NotRunning && !phase6, deltaTime);
+
+    const gen1OffPart1 = gen1PbOffFor5Seconds && !this.idg1Disconnected.get(); // TODO: && !(Gen 1 fault memory || this.idg1Disconnected.get()) when implemented
+    const gen2OffPart1 = gen2PbOffFor5Seconds && !this.idg2Disconnected.get(); // TODO: && !(Gen 2 fault memory || this.idg2Disconnected.get()) when implemented
+
+    // Avoids one-cycle delay in exchange for looking awful
+    const gen1NotOperating =
+      (gen1OffPart1 &&
+        (gen1OffWarningConfirmNode2 || phase6For60Seconds || (phase6 && this.gen12NotOperating.get()))) ||
```
BravoMike99: maybe you could store the || phase6For60Seconds || phase6 && gen112..

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2989,6 +3044,82 @@ export class PseudoFWC {
...
+      idg1DisconnectWarnNext; // TODO: Gen 1 fault memory || when implemented
+    const gen2NotOperating =
+      (gen2OffPart1 &&
+        (gen2OffWarningConfirmNode2 || phase6For60Seconds || (phase6 && this.gen12NotOperating.get()))) ||
+      idg2DisconnectWarnNext; // TODO: Gen 2 fault memory || when implemented
+    const gen12NotOperatingNext =
+      (gen1NotOperating || !this.engine1Generator.get()) && (gen2NotOperating || !this.engine2Generator.get()); // TODO: Remove engine1Generator / engine2Generator checks when fault memory above implemented
+
+    const gen1OffPart2 = gen1OffWarningConfirmNode2 || phase6For60Seconds || (phase6 && gen12NotOperatingNext);
```
BravoMike99: and the same here and some other conditions down the line.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1723,6 +1772,12 @@ export class PseudoFWC {
     this.bat1PbOff.set(!SimVar.GetSimVarValue('L:A32NX_OVHD_ELEC_BAT_1_PB_IS_AUTO', 'bool'));
     this.bat2PbOff.set(!SimVar.GetSimVarValue('L:A32NX_OVHD_ELEC_BAT_2_PB_IS_AUTO', 'bool'));
 
+    this.gen1PbOff.set(!SimVar.GetSimVarValue('A:GENERAL ENG MASTER ALTERNATOR:1', 'bool'));
```
BravoMike99: Is there no `OVHD_ELEC_ENG_GEN_X_PB_IS_ON` variable?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -2989,6 +3044,82 @@ export class PseudoFWC {
...
+    // Avoids one-cycle delay in exchange for looking awful
+    const gen1NotOperating =
+      (gen1OffPart1 &&
+        (gen1OffWarningConfirmNode2 || phase6For60Seconds || (phase6 && this.gen12NotOperating.get()))) ||
+      idg1DisconnectWarnNext; // TODO: Gen 1 fault memory || when implemented
+    const gen2NotOperating =
+      (gen2OffPart1 &&
+        (gen2OffWarningConfirmNode2 || phase6For60Seconds || (phase6 && this.gen12NotOperating.get()))) ||
+      idg2DisconnectWarnNext; // TODO: Gen 2 fault memory || when implemented
+    const gen12NotOperatingNext =
```
BravoMike99: Do we need the next in the name? This one and the idgs confused me a bit 😅

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
tracernz: Another option is an acquisition layer that turns whatever data we get from our "FADEC" into discrete word bits, so it could be swapped out later for the actual words.

lukecologne: Depending on which discrete word and bit, it may already be implemented. I started to implement FADEC bus outputs during the FMGC implementation

FozzieHi: The proper Engine 1 Not Running here is actually an output from a separate data sheet (with inputs from the ECU and EIU). I have just used engine1State for now like other existing cautions are. I do want to properly map them at some point though (like I am doing for the L/G states from the LGCIU in #10521). To avoid going down another FWC rabbit hole in this PR though, shall we leave it as it is whilst taking into account the value of `3` which @BravoMike99 mentioned, either with an enum or without?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Could do, tbh I ran out of possible variable names at a point and didn't want to add even more. I don't think it's too bad in terms of readbility here, but I don't mind moving it out if you really want.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Technically no (the worst kind in this case?). An aspect is set as a copy of `GENERAL ENG MASTER ALTERNATOR:X` in https://github.com/flybywiresim/aircraft/blob/8be4f48843ed7f024378898ff078f3f08d689cb7/fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs#L435

However, as an aspect it's not exported to the instruments. Everything uses `GENERAL ENG MASTER ALTERNATOR:X`, including the ELEC SD page, which is where I copied it from initially. I guess we can changed it to a `named`, but I'm not familiar with the existing good practice of doing this within the Rust code.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Done

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: I tried to avoid shadowing the `this.gen12NotOperating` field, and `next` felt like the best description of it (it's going to be set next). Happy to take suggestions though.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Added the enum with the names directly from the FADEC for now.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: I've rewrote this with your other comment and added a helper func to reduce copy and paste issues.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Done, I've added a helper func.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1396,6 +1451,13 @@ export class PseudoFWC {
 
     this.toConfigMemoNormal.sub((normal) => SimVar.SetSimVarValue('L:A32NX_TO_CONFIG_NORMAL', 'bool', normal));
     this.fwcFlightPhase.sub(() => this.flightPhaseEndedPulseNode.write(true, 0));
+    this.fwcFlightPhase.sub((flightPhase) => {
```
BravoMike99: You should be able to reuse the subscription above. You can also start the wheelRpm as paused if you wanted.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -3027,6 +3109,99 @@ export class PseudoFWC {
...
+    const gen1NotOperatingBase =
+      !this.engine1Generator.get() || // TODO: Replace engine1Generator with fault memory
+      this.idg1DisconnectWarn.get();
+    const gen2NotOperatingBase =
+      !this.engine2Generator.get() || // TODO: Replace engine2Generator with fault memory
+      this.idg2DisconnectWarn.get();
+
+    const gen1OffConfirmOrPhase6 = this.gen1OffWarningConfirmNode2.read() || phase6For60Seconds;
+    const gen2OffConfirmOrPhase6 = this.gen2OffWarningConfirmNode2.read() || phase6For60Seconds;
+    const phase6WithOtherGenOff = (
```
BravoMike99, line 3147: Perhaps declare it outside the loop just so it isn't allocated on every update.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -3027,6 +3109,99 @@ export class PseudoFWC {
...
+    this.gen12NotOperatingPhase3Pulse.write(this.fwcFlightPhase.get() === 3 && this.gen12NotOperating.get(), deltaTime);
+
+    const gen1OffWarningPart1 = gen1OffPart1 && gen1OffPart2;
+    const gen1OffWarningPart2 = this.gen12NotOperatingPhase3Pulse.read();
+    const gen1OffWarningPart3 =
+      this.gen12NotOperating.get() && this.toConfigHalfSecondTriggeredNode.read() && this.flightPhase29.get();
+    const gen2OffWarningPart1 = gen2OffPart1 && gen2OffPart2;
+    const gen2OffWarningPart2 = this.gen12NotOperatingPhase3Pulse.read();
+    const gen2OffWarningPart3 =
+      this.gen12NotOperating.get() && this.toConfigHalfSecondTriggeredNode.read() && this.flightPhase29.get();
```
BravoMike99: These could be equal for now right?
```suggestion
    const gen2OffWarningPart2 = gen1OffWarningPart2;
    const gen2OffWarningPart3 = gen1OffWarningPart3
```

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -3027,6 +3109,99 @@ export class PseudoFWC {
...
+
+    this.gen1OffWarning.set(gen1OffWarningPart1 && !gen1OffWarningPart2 && !gen1OffWarningPart3);
+    this.gen2OffWarning.set(gen2OffWarningPart1 && !gen2OffWarningPart2 && !gen2OffWarningPart3);
+
+    const idg1DisconnectedWarningPart2 =
+      (this.flightPhase29.get() && this.toConfigHalfSecondTriggeredNode.read() && this.gen12NotOperating.get()) ||
+      this.gen12NotOperatingPhase3Pulse.read();
+    const idg2DisconnectedWarningPart2 =
+      (this.flightPhase29.get() && this.toConfigHalfSecondTriggeredNode.read() && this.gen12NotOperating.get()) ||
+      this.gen12NotOperatingPhase3Pulse.read();
```
BravoMike99: And the same for these? Look the same as the gen1OffParts

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -3027,6 +3109,99 @@ export class PseudoFWC {
...
+    const gen1PbOff = this.sdac05001Word.bitValue(19);
+    const gen2PbOff = this.sdac05010Word.bitValue(19);
+
+    const gen1PbOffFor5Seconds = this.gen1PbOffConfirmNode.write(gen1PbOff, deltaTime);
+    const gen2PbOffFor5Seconds = this.gen2PbOffConfirmNode.write(gen2PbOff, deltaTime);
+
+    const phase6 = this.fwcFlightPhase.get() === 6;
+    const phase6For60Seconds = this.flightPhase6For60Seconds.read();
+
+    this.gen1OffWarningConfirmNode2.write(!engine1NotRunning && !phase6, deltaTime);
```
BravoMike99: Maybe we could name these engine1NotRunningAndNotPhase6?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: So the reason I didn't is because I set `initialNotify` to true on this one to effectively start `wheelRpm` paused (which you can't ordinarily do with a `ConsumerSubject`). I don't think we'd want an initial notification on the `flightPhaseEndedPulseNode` though, so is it best to keep them separate?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: I think the allocation is trivial performance wise, and given how big the loop is I'd rather keep it in for readability purposes. Then again you know a lot more about the performance of the sim than I do, so if you think it's going to cause a big enough impact then I'm happy to move it out.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Yeah, done

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Done

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Done

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
BravoMike99: Would the initial notify make a big difference on that pulse node? That would only happen on the initial load in. If not, you could join them.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
BravoMike99: It's probably meaningless yeah but the less allocations the better.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
tracernz: The pulse node should not even be written in a sub IMO. Writing logic nodes from anything other than onUpdate is dangerous. While it works here, it sets a bad example.

I'm also not clear on why wheel1Rpm should be paused. That's also dangerous if it's not well documented on that var for anyone else that might try to use it. Also be aware when you pause it, wheel1SpeedAbove70kts will stay with it's previous state until you resume it again.

FozzieHi: Good points, I did check that nothing else in the FWC used it outside of those phases, but it is probably just an over optimisation and might lead to some unexpected situations in the future, so I've removed it

BravoMike99 (approved):

BravoMike99:
Dev team QA report

Object of testing : #10507 
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 28/02/2026

Testing Process: 
Spawned on runway and turned on APU.
Disconnected IDG 1 and ecam warning & SC appeared.
Cleared the message, T.O CONFIG normal upon takeoff config press.
Manually turned off generator 2, after aprox 5s the caution appeared. Cleared the caution shortly after.
Performed a T.O CONFIG test and this time both cautions appeared.
Cleared both cautions and applied takeoff thrust. Both cautions reappeared.

Testing Results: 
Passed

PR #10525, opened 2026-01-31
fix(fms): fms crash after enabling alternate

tracernz:
Fixes #[issue_no]

## Summary of Changes

Some legs were left behind in the alternate plan after enabling alternate as deleteAlternateFlightPlan did not properly clear it. This was partially because the enroute legs were not deleted, and partially because of the hack to use the primary plan destination as the origin airport for the alternate. The hack provides only a very tiny perf improvement while increasing complexity quite a lot (especially that getter is quite a footgun, and it would get even more complex to properly erase the alternate plan...), and caused other observable side-effects, so I elected to remove it as the easiest fix. I also re-ordered the alternate delete operations to be a little safer, and I actually saw exceptions from the airports being unset first while some things tried to use them to work with the legs still in the plan. I also fixed the type safety issues/bugs in some of the files while doing that which helped me find some of the issues with ALTN. I also added some code to clear the procedure arrays when the approach (for vias) or related airport is removed.

## Additional context

## Testing instructions

Create a flight plan and alternate flight plan. Ensure you are able to select the correct procedures at both the origin and destination in the primary and primary alternate flight plans. Activate the alternate and ensure the LNAV still works.


Benjozork (approved):

BravoMike99:
Dev team QA report

Object of testing : #10525 
Aircraft          : A32NX
Simulator         : 2020
Tier of Testing   : 1
Date              : 11/02/2026

Testing Process: 
Inserted LPBJ as departure and arrival & LPPT as alternate.
Checked that procedures could be selected in origin and destination in both primary and alternate flightplans.
Took off and enabled alternate and verified LNAV works. Performed a direct TO afterwards and it kept working with no issues.

Testing Results: 
Passed

PR #10537, opened 2026-02-04
fix(atc): interval leak when getting multiple handovers

Saschl:
Fixes #[issue_no]

## Summary of Changes

In rare cases (when getting multiple handover messages and having previous unanswered messages), the intervals which check for pending messages are not cleared properly, causing spam of LOGON/LOGOFF messages due to the leaking interval which is never cleared.

The issue was appearing with SayIntention CPDLC, and I added the case that caused the issue in the PR itself, as the condition is a bit complicated (at least for me :D)

## Additional context

      Cancel any existing handover interval before starting a new one.
     
      If a second HANDOVER arrives while the first interval is still polling
      (waiting for open messages to clear), the first interval becomes orphaned — it keeps
      running but its handle is lost when we overwrite this.handoverInterval.
     
      Once all messages are answered, BOTH intervals fire their setTimeout callbacks,
      causing repeated LOGOFF/LOGON every second. As the first interval is orphaned, it keeps running forever
     
      Example scenario:
        17:48:52  EDWW→SAS688  msg 32: "EXPECT MONA5A…" (requires R response)
        17:49:18  EDWW→SAS688  HANDOVER @EDGG  ← starts interval #1, does not clear as msg 32 is still open
        17:52:37  EDWW→SAS688  HANDOVER @EDWW  ← overwrites handle, interval #1 orphaned
        17:56:09  EDWW→SAS688  msg 4: "CONTACT BREMEN RADAR ON 123.225" (requires response)
        17:57:25  SAS688→EDWW  ROGER to msg 32 -> cleared
        18:00:07  SAS688→EDWW  WILCO to msg 4 -> cleared
        18:00:24  LOGOFF/LOGON spam starts (orphaned interval #1 fires infinitely)
     

## Testing instructions

1. Difficult to test as the preconditions are hard to reproduce. Make sure handovers work as expected.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -70,6 +70,8 @@
 1. [A32NX/FWS] Fixed stall warning not working - @tracernz (Mike)
 1. [A380X/MFD] Add initial implementation of D-ATIS page in ATCCOM - @heclak (Heclak)
 1. [A32NX/FWS] Fixed takeoff memo not working in flight phase 9 (after landing) - @tracernz (Mike)
+1. [A380X/FUEL SYSTEM] Fixed inner tank transfer not starting in certain circumstances - @donstim (donbikes)
+1. [A32NX/FADEC] Fixed the PFD showing MAN FLX instead of MAN MCT post A/THR engagement after takeoff - @FozzieHi (Fozzie)
 
 ## 0.14.0
 
--- a/.github/workflows/ingamepanels-checklist-fix.yml
+++ b/.github/workflows/ingamepanels-checklist-fix.yml
@@ -17,6 +17,8 @@ jobs:
     if: github.repository_owner == 'flybywiresim'
     name: 'Build InGamePanels Checklist Fix'
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     env:
       FBW_PRODUCTION_BUILD: 1
       MASTER_ZIP_NAME: ingamepanels-checklist-fix-master.zip
--- a/.github/workflows/localization.yml
+++ b/.github/workflows/localization.yml
@@ -9,6 +9,9 @@ jobs:
   localization-update:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'flybywiresim'
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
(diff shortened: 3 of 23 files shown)


fbw-common/src/systems/datalink/atc/src/ATC.ts
tracernz: Missing a null check?

tracernz: This does not get cleared if something happens in between, and it will still run and mutate state even though it's from an older handoverInterval.

Saschl:
> Doing this stuff with setInterval/setTimeout gives you lots of scope to shoot yourself in both feet 😁 . It might be simpler to just have a dumb loop checking if simtime reached a threshold to do each thing.

I don't disagree, this bug is proof enough :D

fbw-common/src/systems/datalink/atc/src/ATC.ts
```diff
@@ -514,15 +516,28 @@ export class Atc {
       return AtsuStatusCodes.SystemBusy;
     }
 
+    // Cancel any existing handover interval and timeout before starting a new one.
+    if (this.handoverInterval !== null) {
+      clearInterval(this.handoverInterval);
+      this.handoverInterval = null;
+    }
+
+    if (this.handoverTimeout != null) {
```
tracernz:
```suggestion
    if (this.handoverTimeout !== null) {
```
Dunno if falsy values like 0 are legal timer IDs, but just in case.

Saschl: Grr, that wasn't intended (too much Java lately :P). Will fix

Saschl:
Dev Team QA report (Testing own PR as the change is relatively small with low impact)

Discord Username : saschl
Object of testing : https://github.com/flybywiresim/aircraft/pull/10537
Aircraft : A32NX
Simulator : 2024
Tier of Testing : 1
Date : 09/02/2026

Testing Process:
Performed flight EDDM - EGLL with CPDLC. Handovers are registered and LOGOFF/LOGON messages are being sent. No spam could be observed

Testing Results:
Passed

PR #10472, opened 2026-01-09
feat(a380x): fuel penalty

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Adds fuel Penalty function
Updates DATA/STATUS page layout to batch 7.
Removes TMPY and EO page headers showing in pages where they shouldn't show.
Changes minimum fuel value computed by predictions to -35 tons instead of 0.
Increases font size of top tab navigator titles on the MFD.

## Screenshots (if necessary)

Before:
imagem

After:
Penalty

## Additional context

Discord username (if different from GitHub): bruno_pt99

## Testing instructions

Play around with the fuel penalty input field in the DATA/STATUS  MFD page and check that it accepts numbers up to 3 digits with a decimal (NNN.N). An optional + leading sign can also be used(+NNN.N). Anything else should give FORMAT ERROR (0.01, 9.02, 100.00, AAA.A etc).
Insert a valid fuel penalty (> 0 < 999.9). Verify PENALTY white on the page title section only appears in the following pages:
DATA/STATUS, FUEL & LOAD, F-PLN, HOLD
Verify that the fuel penalty has affected the fuel predictions (F-PLN/FUEL & LOAD page) close to the inserted value.
Delete a fuel penalty or insert a value of 0, the label should disappear and all go back to normal.
Do a quick pattern with fuel penalty inserted, check that it is deleted once the fms changes to the done phase.

TMPY:
Create a TMPY flightplan and check that the "TMPY" label only appears on the top of the MFD on the following pages:
DIR TO, FIX INFO, Revision pages (Airways, HOLD, STEP ALTS/ALT CSTR/SPD CSTR, DEPARTURE, ARRIVAL). PERF, INIT, FUEL & LOAD & F-PLN.
Check that the black bars located on top of the MFD next to where TMPY label appears, do not appear on a page outside the FMS (surveilance page for example)

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10542, opened 2026-02-07
fix(fms): fix estimated fuel at destination being -35t

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue where VNAV used a destination EFOB of -35t for its calculations after https://github.com/flybywiresim/aircraft/pull/10472. This would result in a ridiculuously low weight estimation and a steep descent path. I think the effect would only be seen if the descent path was recomputed during the descent, i.e if the flight plan route or PERF APPR values were changed.

## Screenshots (if necessary)

An example where the VNAV path was still at ~10'000 ft while only about 20 NM out.
image

## Additional context

## Testing instructions

Enter a normal route, make sure vertical predictions look ok. Take off, start your descent normally. Then force a recomputation of the descent path by changing the destination QNH or temperature on the PERF APPR page. Make sure the yoyo is still in a resonable place (based on a simple 3° calculation for example). Make sure the DECEL point is sensible.

### DIFF
--- a/fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
+++ b/fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
@@ -3604,8 +3604,8 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
   }
 
   /**
-   *
-   * @returns {number} Returns estimated fuel on board when arriving at the destination (in tonnes)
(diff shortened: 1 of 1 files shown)


tracernz:
Dev Team QA report

Discord Username  : tracernz
Object of testing : #10542
Aircraft          : A32NX 
Simulator         : 2024
Tier of Testing   : 1
Date              : 08/02/2026 

Testing Process: 
Performed part of a flight NZCH/NZWN. Edited some approach profile params during the descent and inspected the relevant code in the debugger. All looks good.

Testing Results: 
Passed

PR #10550, opened 2026-02-09
feat(a32nx/fwc): Add ENG REV SET master caution

FozzieHi:
## Summary of Changes

A small PR (for a change) which adds the `ENG REV SET` master caution, which triggers when the throttle TLA is moved into reverse whilst in flight.

To do this I have had to remove a clamp in `ThrottleAxisMapping.cpp`, which limited the throttle TLA whilst in flight to a minimum of 0. This was added 5 years ago, and may have been to prevent actual reverser deployment back then. Now we have other checks to prevent reverser deployment in flight, such as below which checks that the gear is compressed from the LGCIU:
https://github.com/flybywiresim/aircraft/blob/c4b9e42ce20cabf328b5d8452e9a61f5bcdff641/fbw-a32nx/src/wasm/systems/a320_systems/src/hydraulic/mod.rs#L5910-L5911

## Screenshots (if necessary)

Screenshot 2026-02-08 225848

## Additional context

## Testing instructions

1. Takeoff, then try and deploy the reversers in flight, check that the `ENG REV SET` master caution triggers.

**Bonus points:**
1. Set-up for a normal landing, deploy the reversers around 30ft from the ground.
2. Check that the master caution again triggers, and disappears when the main gear contacts the ground.

### DIFF
--- a/fbw-common/src/systems/datalink/atc/src/ATC.ts
+++ b/fbw-common/src/systems/datalink/atc/src/ATC.ts
@@ -51,6 +51,8 @@ export class Atc {
 
   private handoverInterval: number = null;
 
+  private handoverTimeout: number = null;
+
   private handoverOngoing = false;
 
   private currentAtc = '';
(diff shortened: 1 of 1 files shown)


fbw-common/src/wasm/fbw_common/src/ThrottleAxisMapping.cpp
```diff
@@ -253,11 +253,6 @@ void ThrottleAxisMapping::setCurrentValue(double value) {
     newTLA = thrustLeverAngleMapping.get(value);
   }
 
-  // ensure not in reverse when in flight
-  if (inFlight) {
```
BravoMike99, line 257: Does this also affect the 380x? If so, maybe the caution should be added there too if it doesn't take too much work.

FozzieHi: I assume it does given it's in common, and the A380 also checks for the compressed data from the LGCIU:
https://github.com/flybywiresim/aircraft/blob/c4b9e42ce20cabf328b5d8452e9a61f5bcdff641/fbw-a380x/src/wasm/systems/a380_systems/src/reverser/mod.rs#L99-L101

I'm happy to add it to the A380 as well, assuming it's the same implementation. That said, we could have removed this in its own PR without adding any cautions and it would have no impact on actual reverser deployment, so I wouldn't say it's needed.

BravoMike99: Just making sure, good to go then. Can be left for the future

Jonny23787:

Discord Username  : jonny_23
Object of testing : #10550 
Aircraft          : A32NX
Simulator         : 2024
Tier of Testing   : 1
Date              : 11/02/2026

Testing Process: 

1. Performed a take off, set thrust levers to reverse, confirmed ENG REV SET procedure appeared.
2. Setup for landing, at 30ft set thrust levers to reverse, ENG REV SET appeared and disappeared after MLG touchdown.

Testing Results: 
Passed

tracernz:
We should probably have checked that the reversers don't actually deploy in the air as part of the tests. I expect they won't, but just in case since we removed some code relating to that.

Jonny23787:
I don't think they did, at least on the EWD

FozzieHi:
I should have added that to the testing notes, but I did check on both the A32NX and A380X that they didn't.

PR #10553, opened 2026-02-11
ci: limit concurrency of qa config workflows

FoxtrotSierra6829:
## Summary of Changes

If timing is unfortunate, right now, there can be cases where two workflows try to modify the QA configuration at the same time. Whichever workflow writes later wins, while the other one's changes will get overridden. This is because modifying the file takes a short time and changes done by one workflow will not be taken into account if the other workflow has already copied the file from the server. To mitigate this, anti-concurrency measures are necessary.
Unfortunately, GitHub cannot handle workflows in a linear manner. Even using the built-in concurrency feature using a custom group, there can ever only be one active workflow and one waiting workflow. All other triggered workflows will simply cancel the already waiting one. 
So this is a solution where a lockfile gets created to protect against other workflows trying to modify the same file. When a workflow is done modifying it, it will release the file again.

## Additional context

### DIFF
--- a/.github/workflows/pr-updated.yml
+++ b/.github/workflows/pr-updated.yml
@@ -32,6 +32,7 @@ jobs:
             region = auto
             EOF
       - name: Lock QA config
+        if: github.event.pull_request.auto_merge == false
         run: ./scripts/acquire_lock.sh installer/qa-config/.lock
       - name: Update QA config entries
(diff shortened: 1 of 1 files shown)


FoxtrotSierra6829:
Hey @Gurgel100 please read the PR description😄

PR #10560, opened 2026-02-19
fix(gsx): Changes to support new Loader Behavior, Handlers and Service Cancellation

Fragtality:
Fixes: N/A - preventive maintenance

## Summary of Changes

- New Loader Behavior (released, live): They stay connected now after (Un-)Loading has finished.
**=>** Removed Code to close Cargo Doors (A380) after reaching 100%.
- Custom Aircraft Handler Scripts (released, experimental feature): Python Files running inside GSX which can override certain Service Interactions. Like setting power Availability beyond our Control.
**=>** Added custom Handler Files overriding all available Handler Methods.
- Service Cancellation (released, live): GSX Version 3.8.0 allows to cancel (stop) a Service earlier which comes with a new GSX Service State (7/completing).
**=>** Added new State to Enums in TS & Rust and checked Comparisons to Service States to be still semantically correct.

Along the Way I've made some further Optimizations:
- Harmonized XML Conditions for Chocks (A32NX_GND_EQP_IS_VISIBLE) on A320 & A380
  - Included Engines 3 and 4 for A380
- (TS/extra-host) Generalized GsxSync Base Class further 
  - Added new Loader Attached Variables
  - Rewrote GSX Power Sync
  - Changed Power Sync to use (GSX) GPU Connection Variable instead of Service State
- (TS/extra-host) Changed MSFS Power Sync
  - Syncing can be disabled via event
  - Added event to set ext avail state
  - Ensured that User Control via FlyPad is always given regardless of which Sync is active
- (Rust/systems) Changed Payload Synch to finish on 1/available and 7/completing (besides 6/completed)
- (Rust/systems) Boarding Complete sound is also triggered on 7/completing

## Additional context

## Testing instructions

WIP

Non-GSX:
1) Check if behavior is still the same (GSX Sync has to be disabled)

GSX:
1) Check if Power is synced as expected
  - Except Session-Start: Power is only avail when Jetway or GSX GPU is connected
  - No forceful Disconnection
  - Power Avail is removed during GSX Pushback
2) Check if Avail State can always be overridden by the User in FlyPad
3) Check if canceling a Service (Refuel, Board, Deboard) works without Problems (Fuel/Payload Sync should finish as normal, just earlier)
4) Check if Cargo-Door Sync is working without Issues


Fragtality:
Discussion Point: Usage of the A32NX_HYD_NW_STRG_DISC_ECAM_MEMO Variable in the XML Condition.
I get that it was added to remove the Chocks automatically when using the EFB Pushback. BUT: It is not an universal Indicator for "Pushback is active" since the Tug can already be connected (and Pin inserted) when Boarding is still ongoing. Which is possible with GSX. And GSX' Bypass Pin State is also synchronized to correctly trigger that Memo.
So in the End, the Chocks are already removed when the Tug is attached while Boarding. Without checking if the Aircraft is secured through the Parking Brake.

fbw-common/src/systems/shared/src/GPUManagement.ts
```diff
@@ -63,11 +65,17 @@ export class GPUManagement implements Instrument {
...
-      this.gpuHookedUp.sub((v) => this.setEXTpower(v));
+      this.sub.on('set_ext_power').handle(this.onExtPowerEvent.bind(this));
+      this.gpuHookedUp.sub((v) => {
+        if (this.msfsSyncEnabled.get() === true) this.setEXTpower(v);
+      });
       this.groundVelocity.sub((v) => {
         // disable ext power when aircraft starts moving
         if (v > 0.3 && this.anyGPUAvail()) {
-          this.toggleGPU();
+          this.setEXTpower(false);
```
Jonny23787, line 75: Should the `p` be capitalised?

Fragtality: ❔

Jonny23787: I just thought it should be for camel case and consistency

Fragtality: I didn't name it ... so maybe? 😅

.github/CHANGELOG.md
```diff
@@ -79,6 +79,8 @@
 1. [FMS] Show true bearings correctly in various FMS pages on the MCDU/MFD - @tracernz (Mike)
 1. [A380X/FWS] Improve various ABN SENSED procedure accuracy - @Jonny23787 (Jonathan)
 1. [A32NX/FMS] Add pilot entry ETT (Estimated Takeoff Time) functionality  - @BravoMike99 (bruno_pt99)
+1. [GSX] Changes to support new Features: Loader Behavior, Python Handlers and Service Cancellation - @Fragtality (Fragtality)
+1. [GSX] Reworked Ground Power Synch with Jetway/GSX GPU - @Fragtality (Fragtality)
```
tracernz, line 83: Note: wrong release.

PR #10564, opened 2026-02-20
feat(a32nx/fwc): Update TCAS and XPDR master cautions

FozzieHi:
## Summary of Changes

Updated `NAV TCAS STBY` master caution:
1. Now only triggers when the switch is in STBY and not for faults. The `TCAS STBY` amber memo still tracks faults.
2. Added a three second confirm timer.

Added `ATC/XPDR STBY` master caution which triggers when both XPDRs are not on.

## Screenshots (if necessary)

Screenshot 2026-02-21 182218

## Additional context

## Testing instructions

Feel free to spawn on a runway if it makes testing easier.
1. Whilst on the ground, set both the XPDR and TCAS switches to `STBY`.
2. Check that `TCAS STBY` appears in green as a right memo.
3. Take-off, wait until you are out of T.O. Inhibit
4. Check that both `NAV TCAS STBY` and `NAV ATC/XPDR STBY` master cautions appear, and that the `TCAS STBY` right memo has turned amber.
5. Move the TCAS switch from `STBY` to `TA/RA`, check that the `NAV TCAS STBY` master caution disappears, but the memo stays in amber.
6. Move the XPDR switch from `STBY` to `AUTO`, check that no amber now displays on the ECAM.


fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1587,7 +1618,25 @@ export class PseudoFWC {
...
 
+  private readonly xpdr1Status = RegisteredSimVar.create('A:TRANSPONDER STATE:1', SimVarValueType.Number);
+  private readonly xpdr2Status = RegisteredSimVar.create('A:TRANSPONDER STATE:2', SimVarValueType.Number);
+
   private acquireSdac(): void {
+    this.sdac00100Word.set(0);
+    this.sdac00100Word.setSsm(Arinc429SignStatusMatrix.NormalOperation);
+    this.sdac00100Word.setBitValue(
+      24,
+      this.xpdr1Status.get() === TransponderState.Off ||
```
BravoMike99: You could store the value just to avoid fetching the simvar 3x x2 in the worst case scenario

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -3115,6 +3166,23 @@ export class PseudoFWC {
...
 
+    // TODO: Check XPDR Fault below
+    this.tcasStby.set(this.tcasSensitivity.get() === 1);
+    this.tcasStbyConfirmNode.write(this.tcasStby.get(), deltaTime);
+    this.tcasControlPanelStbyConfirmNode.write(
+      this.tcasControlPanelPosition.get() === 0 && !this.tcasFault.get(),
+      deltaTime,
+    );
+
+    this.tcasStbyWarning.set(this.fwcFlightPhase.get() === 6 && this.tcasControlPanelStbyConfirmNode.read());
```
BravoMike99, line 3479: I think there's a flightPhase6 further down the loop which we could potentially use by moving it up. That way its all common.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: Fair point, although as it's a RegisteredSimVar it's not actually fetching the simvar every loop right?

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: I think we need to decide what we do for these. `fwcFlightPhase.get()` is currently called 53 times in the loop and I don't think there's much of an issue doing it here, but we also set `const flightPhase = this.fwcFlightPhase.get();` right at the top so I'm happy to change it to `flightPhase === 6` rather than having one var per phase?

BravoMike99: Indeed ,its a bit of a mess atm. I think it would be nice to have them all in one place so it can just be reused across different cases, although, that should be its own cleanup PR.

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
FozzieHi: You're correct (I double checked with tracer). I have moved it to a single var.

BravoMike99 (approved):

BravoMike99:
Dev team QA report

Object of testing : #10564 
Aircraft          : A32NX
Simulator         : 2020/
Tier of Testing   : 1
Date              : 21/03/2026

Testing Process: 
Checked that TCAS STBY memo appears on the ground with a green color, when the TCAS is in standby mode (switch or due to xpdr in standby)
After takeoff, with both TCAS and XPDR in stby the two caution appears and the TCAS STBY memo becomes amber.
Once the TCAS switch is out of STBY, the TCAS STBY caution appears but the memo remains visible.
With the XPDR switch out of STBY, the ATC/XPDR STBY caution disappears and so does the TCAS STBY memo.

Testing Results: 
Passed

fbw-a32nx/src/systems/systems-host/systems/FWC/PseudoFWC.ts
```diff
@@ -1683,6 +1710,8 @@ export class PseudoFWC {
     this.ecpEmergencyCancelLevel = warningButtons.bitValue(17) || this.ecpEmergencyCancelButtonHardwired.get();
   }
 
+  private readonly xpdr1StatusVar = RegisteredSimVar.create('A:TRANSPONDER STATE:1', SimVarValueType.Number);
+  private readonly xpdr2StatusVar = RegisteredSimVar.create('A:TRANSPONDER STATE:2', SimVarValueType.Number);
```
tracernz, line 1714: @FozzieHi FYI this doesn't actually do anything. MSFS only has one transponder. (Just noticed this when solving a merge conflict with an old branch and reading this code.)

Issue #10382, opened 2025-10-27
Seat Belts AUTO condition


becas22:
### Simulator

2024

### Aircraft Version

Development (4k)

### Build info

```json
{
    "built": "2025-10-20T20:56:19+00:00",
    "ref": "refs/heads/master",
    "sha": "b2fffc4edc0078875b198d2a5a4ce61cb8456834",
    "actor": "flogross89",
    "event_name": "manual",
    "pretty_release_name": "master:b2fffc4e",
    "version": "a380x-v0.14.0-dev.b2fffc4"
}
```

### Describe the bug

With SEAT BELTS sign AUTO, the SEAT BELTS memo is showing with Slats retracted and L/G up

### Expected behavior

With slats retracted and L/G up, seat belts in auto should not display the MEMO

### Steps to reproduce

1. Takeoff
2. Retract slats and L/G with SEAT BELS sw in AUTO
3. Check for SEAT BELS memo

### Discord Username (optional)

bernardor96

donstim:
I reproduced this issue. Not only does the SEAT BELTS memo remain displayed after flap/slat/gear retraction, it will also show if the overhead switch position is toggled to OFF and back to AUTO when engines are on with flaps/slats and gear up. The LVAR value for the overhead switch position (XMLVAR_SWITCH_OVHD_INTLT_SEATBELT_Position) is different for the 3 possible switch positions (off, on and auto), but the result of on and auto, which have LVAR values of 0 and 1, respectively (off is 2) are the same.

FlyByTim:
Unable to reproduce in the latest build. Are you still seeing this issue?

donstim:
Yes, exact same behavior still.
Auto:

![Image](https://github.com/user-attachments/assets/41e8a1be-2cca-4e99-a50e-ab2c6322a3e5)
On:

![Image](https://github.com/user-attachments/assets/2658dec1-a071-401f-aa89-0c1a32811336)
Off:

![Image](https://github.com/user-attachments/assets/0e48d950-ffd4-48f5-819a-74c1d59c3e1a)

BravoMike99:
Fixed by #10570

PR #10570, opened 2026-02-26
fix(a380x): Fixed auto seat belts

FlyByTim:
Fixes (#10382)

## Summary of Changes
Fixes the auto seat belt logic by adding a small margin for slat percentage not returning fully to 0.0. It tends to stop around 0.048-0.075 on my end for some reason.

## References

FCOM also mentions the No Smoking (and presumably the Mobile Devices too) signs will act the same, so that should be added at some point.

Though, I couldn't find anything about the engines necessarily having to be on.

image

## Additional context

## Testing instructions
1. Load A380X - I chose to load it midair for quick testing
2. Power up aircraft and turn engines on
3. Put seat belt switch into auto position
4. On ground, the seat belt memo should always show
5. In air, check the seat belt memo disappears when gear is up AND slats are retracted
6. Do quick individual tests by putting gear down and once locked, seat belt memo should appear.
7. Raise gear. Seat belt memo should go away instantly. Select flaps 1. Seat belt memo should appear.
8. Select flaps 0. Seat belt memo should disappear.
9. Bonus sanity test, put seat belt switch on, check memo is on. Put switch off, check memo is off. Put into auto and repeat 6-8.


Maximilian-Reuter:
Wouldnt it be better to check that the slats return to 0 percent again ? Or see why they dont at least ? Could also potentially handle the margin in the location that creates the LVAR. I am just not super comfortable with adding margins because of unexplained behavior :D

FlyByTim:
> Wouldnt it be better to check that the slats return to 0 percent again ? Or see why they dont at least ? Could also potentially handle the margin in the location that creates the LVAR. I am just not super comfortable with adding margins because of unexplained behavior :D

Sure, I don't know anything about Rust so I'll have to pass it on to someone else then

FlyByTim:
@Maximilian-Reuter Looking at the code, it looks like there's already a tolerance built in for the flaps, so maybe this was put in place due to the float always being a little different when extending/retracting steps of flaps?

In fbw-a380x\src\wasm\systems\a380_systems\src\hydraulic\sfcc\mod.rs line 35-42.

lukecologne:
Looking at the documentation for the A320, the CIDS, which controls the seat belt and PED/smoking signs, only gets one discrete input from each SFCC that indicates that the slats are at or above 17°, which is just below flaps 1. So we should probably use that as the condition, adapted for the A380 of course. I assume the condition will be similar there

FlyByTim:
So really we should be setting the LEFT(RIGHT)_SLATS_POSITION_PERCENT to >= the equivalent of 19 degrees, since CONF 1 on the A380 is 20 degrees.

I don't have any specific docs regarding CIDS just the FCOM.

fbw-a380x/src/base/flybywire-aircraft-a380-842/SimObjects/AirPlanes/FlyByWire_A380_842/model/A380_COCKPIT.xml
```diff
@@ -1912,8 +1912,8 @@
                         (L:A32NX_ENGINE_N1:3) 10 &gt; or
                         (L:A32NX_ENGINE_N1:4) 10 &gt; or
 
-                        (L:A32NX_LEFT_SLATS_POSITION_PERCENT, percent) 0 &gt;
-                        (L:A32NX_RIGHT_SLATS_POSITION_PERCENT, percent) 0 &gt; or
+                        (L:A32NX_LEFT_SLATS_ANGLE, degree) 19 &gt;
```
BravoMike99: Shouldn't we use greater or equal to 20? This way it can  trigger slightly before it reaches conf 1?

FlyByTim: Oops forgot to change the sign. But as far as triggering before flap 1, that's intentional using the logic Luke posted (triggers at 17, tho flap 1 is 18).

BravoMike99 (approved):

stratavelo92:

Discord Username  : spottersim 
Object of testing : #10570  
Aircraft          : A380X 
Simulator         : 2024 
Tier of Testing   : 1
Date              : 16/03/2026 

Testing Process: 
Tested as requested 

Testing Results: 
Passed

Negatives:       

4. Initially shows, then disappears (seen twice on GND in TO config)
7. Memo appears only once flaps are at commanded position...?

Conclusions: 


Media:

FlyByTim:
4. I see what you are seeing I think. The TO memo takes some parameters from the memo on the right side (such as spoilers armed, seat belts on), but seems to do it at times when TO memo doesn't even appear on the left, therefore making it look as if the memos just disappeared from the right. I was able to manipulate it to the point that I was all ready to go and the only thing I saw on the right was "NO MOBILE" and nothing else. Logically speaking though, the aim of this PR works as intended. The TO CONFIG thing should be it's own PR.

5. Memo should only appear when slats are at 19 degrees or greater as we discussed in above comments. Works as coded.

6. Off should always keep seat belt sign off regardless of configuration so that's working as intended.

FlyByTim:
Doing some more test, I start from the gate cold and dark as if I was going to do a whole new flight. Kept seat belts in auto for push. Start 1, and the memo appeared at 10% N1 as is coded. Then I started 2, and the seat belt memo got replaced with TCAS STBY. I armed the spoilers to check if that would appear, but it didn't. So I think my hunch that some of the info is being taken for the TO Memo is correct, but as to why it would disappear like this before the memo even appears, I don't know. Flipping seat belts to ON won't fix it because the seat belt logic is within the TO Memo logic at this point. Once TO memo does appear, logic still works correctly. And once in the air, logic works as well.

stratavelo92:
> 4. I see what you are seeing I think. The TO memo takes some parameters from the memo on the right side (such as spoilers armed, seat belts on), but seems to do it at times when TO memo doesn't even appear on the left, therefore making it look as if the memos just disappeared from the right. I was able to manipulate it to the point that I was all ready to go and the only thing I saw on the right was "NO MOBILE" and nothing else. Logically speaking though, the aim of this PR works as intended. The TO CONFIG thing should be it's own PR.
> 5. Memo should only appear when slats are at 19 degrees or greater as we discussed in above comments. Works as coded.
> 6. Off should always keep seat belt sign off regardless of configuration so that's working as intended.

Thanks for the input, I think I got the instructions mixed up. In this case, I have changed the report to passed

FlyByTim:
Did a full pattern from cold and dark with the new commit leaving seat belts on auto the entire time. TO Memo appears and disappears along with SEAT BELTS and GRD SPLRs ARMED showing up appropriately. Tested LDG Memo logic and transition into flight phase 11 for good measure and all looks good and similar to real world counter parts.

stratavelo92:

Discord Username  : spottersim 
Object of testing : #10570  
Aircraft          : A380X 
Simulator         : 2024 
Tier of Testing   : 1
Date              : 17/03/2026 

Testing Process: 
 
Re-tested with focus on SEAT BELTS memo message on ground

Testing Results: 
Passed  

Negatives:       


Conclusions: 

Works great!

Media:

PR #10312, opened 2025-08-21
fix(a32nx/flight model): A320 flight model update

donstim:
This update is intended, when combined with #10554 , to provide consistent takeoffs (no auto-rotation before VR or difficulty in rotating) when trim is set anywhere within the green trim band, regardless of weight or cg (as long as they are within the certified weight-cg envelope).

## Summary of Changes
This update removes the trim elasticity table and elevator trim effectiveness scalars that were used to prevent auto-rotation on takeoff and show an accurate trim control setting in flight. They have been replaced by code in #10554 that expands the capability to allow takeoffs with trim set anywhere in the green trim band without the airplane rotating before VR or being unable to rotate.

## Additional context

Discord username (if different from GitHub): donbikes

## Testing instructions
Perform takeoffs and landings at various weights and cg conditions. Confirm that no auto-rotation occurs before VR, there is no difficulty in rotating the airplane for takeoff with the trim set anywhere within the green trim band, and that no regressions in flight characteristics are present. Use takeoff speeds and flex thrust settings from the takeoff performance calculator in the EFB.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10585, opened 2026-03-12
fix(a32nx/flight model): A32NX Flight Model update for MSFS 2024

donstim:
Fixes #[issue_no]

## Summary of Changes
This PR is equivalent to #10312 but for MSFS 2024 only. It also, when combined with https://github.com/flybywiresim/aircraft/pull/10554 , is intended to provide consistent takeoffs (no auto-rotation before VR or difficulty in rotating) when trim is set anywhere within the green trim band, regardless of weight or cg (as long as they are within the certified weight-cg envelope). (Still need to verify whether the THS tables in that PR work for this one in MSFS 2024.) It also addresses flight model differences between MSFS 2020 and MSFS 2024.
## Additional context

## Testing instructions
Perform takeoffs and landings at various weights and cg conditions. Confirm that no auto-rotation occurs before VR, there is no difficulty in rotating the airplane for takeoff with the trim set anywhere within the green trim band, and that no regressions in flight characteristics are present. Use takeoff speeds and flex thrust settings from the takeoff performance calculator in the EFB.


donstim:
Not going to proceed with THS fudging

PR #10596, opened 2026-03-17
fix(ecam): GW not displaying until engine start

### LINKED ISSUE #10582: GW not initializing until engines are on
### Simulator Version

2020

### Aircraft Version

Development

### Build info

```json
{
    "built": "2026-03-01T10:03:38+00:00",
    "ref": "refs/heads/master",
    "sha": "95123dbe96c7a4f921319fdc5fb08d391c379539",
    "actor": "BlueberryKing",
    "event_name": "manual",
    "pretty_release_name": "master:95123dbe",
    "version": "a32nx-v0.15.0-dev.95123db"
}
```

### Describe the bug

With this most recent Dev version update, the plane will not display the AC's grows weight on the ECAM until at lease one engine is starting up and at least at 3% N1. I have tried turning off the Gens on both engines and the GW still pops up at 3% N1. This also prevents auto pasting weights into the "Fuel Pred" page on the MCDU by simply pushing the "ZFW/ZFWCG" button twice. This bug just showed up in this current Dev version and was not present in the most recent version.

### Expected behavior

In the previous version the "GW" would display live weight as soon as the APU, was available or Ground power was selected.

### Steps to reproduce

On initial startup with the AC running on ground power, GW is not displayed on the bottom right of the lower ECAM display.

On engine startup (doesn't matter which engine is started, both individually) once the first engine reached 3% N1, the GW pops up on the ECAM lower display, Then a message shows on the MCDU in orange saying "INITALIZE WEIGHT/CG"

### References (optional)

_No response_

_No response_

alex_n0ej

Ditoo29:
Fixes #10582

## Summary of Changes
- Removed engine running condition from GW display logic so GW now displays on the lower ECAM as soon as electrical power is available, instead of waiting for engine start
- Fixed GW value text style when engines are off (was incorrectly using `Text unit` instead of `Text value`)

## References
GW should display whenever AC power is available, consistent with real A320 behaviour.

## Additional context
Bug was introduced and reported in #10582. Previously GW would only appear at ~3% N1 during engine start.

Discord username (if different from GitHub): dito29

## Testing instructions
1. Load A32NX cold & dark at a gate
2. Connect External Power (EXT PWR on overhead ELEC panel)
3. Verify GW displays on the bottom right of the lower ECAM immediately
4. Verify GW value is correct and displayed in the correct style
5. Start an engine and verify GW continues to display correctly during and after engine start

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1863,3 +1863,4 @@
 1. [EFB] Restructured APIs and made Navigraph Auth a reusable component - @MicahBCode (Mischa Binder)
 1. [ECAM] Added F units to CRZ and COND pages for Cabin temps. Currently tied to kg/lbs option in EFB -Patrick Macken  (@PatM on Discord)
 1. [A32NX/FCU] Fixed bug where the FCU selected altitude would immediately switch to 1000s on selecting the 1000ft inteval knob - @elliot747 (Elliot)
+1. [ECAM] Fixed GW not displaying on lower ECAM until engine start. GW now displays correctly when electrical power is available - @Ditoo29 (dito29 on discord)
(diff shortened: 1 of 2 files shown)


tshomas:
Hi, can I check your reference with the GW being available before engine start?

That's not what I see in real life and also FCOM DSC-31-20 PERMANENT DATA says that it's only filled after the first engine start?

Also how does the aircraft know the GW of itself if you've just powered it on from cold and dark?

Edit: just seen the linked issue, unfortunately I don't think that user is recalling correctly

tshomas:
Screenshot_20260317-160645

Ditoo29:
I took a closer look and researched a bit more and that is correct. The current behaviour is as it's supposed to be.

PR #10598, opened 2026-03-18
fix(sound): boarding complete and welcome onboard announcements playing simultaneously

### LINKED ISSUE #10549: Anouncement Boarding completed and Welcome to close together
### Simulator Version

2020

### Aircraft Version

Development

### Build info

```json
Latest buildings
```

### Describe the bug

After adjustments to the sound system the anouncements of boarding completed and welcome are right after each other.

### Expected behavior

Some time in between those anouncements.

### Steps to reproduce

Start boarding
Wait till boarding complete

### References (optional)

_No response_

_No response_

_No response_

Ditoo29:
Fixes #10549

## Summary of Changes
Split the `welcomeonboard` and `boardingcompleted` announcements onto separate local variables (`A32NX_SOUND_BOARDING_COMPLETE` and `A32NX_SOUND_WELCOME_ONBOARD`) so they no longer trigger simultaneously. Added a delay timer in `BoardingSounds` that starts when boarding completes and fires the welcome announcement after 10 seconds, ensuring a natural gap between the two announcements. Works for both standard boarding and GSX boarding.

## Screenshots (if necessary)
N/A - audio only change

## References
Issue #10549 - reported that after sound system adjustments, boarding complete and welcome announcements play back to back with no gap.

## Additional context
The fix is contained entirely in `BoardingSounds`, `PayloadManager` and the corresponding XML sound definition. 

Discord username (if different from GitHub): dito29

## Testing instructions
1. Load the A32NX
2. Set boarding rate to Instant, Fast or Real
3. Start boarding and wait for it to complete
4. Verify `boardingcompleted` announcement plays
5. Verify `welcomeonboard` announcement plays ~10 seconds later
6. Start a second boarding session without deboarding and verify welcome plays again
7. Repeat with GSX enabled and verify both announcements behave correctly


fbw-a32nx/src/base/flybywire-aircraft-a320-neo/SimObjects/AirPlanes/FlyByWire_A320_NEO/sound/sound.xml
```diff
@@ -2183,7 +2183,7 @@
             <WwiseRTPC LocalVar="A32NX_SOUND_ANNOUNCEMENTS_ENABLED" Units="number" Index="0" RTPCName="LOCALVAR_A32NX_ANNOUNCEMENT_TOGGLE" />
         </Sound>
 
-        <Sound WwiseEvent="welcomeonboard" WwiseData="true" NodeName="SOUND_FWD_GALLEY" ConeHeading="180" CancelConeHeadingWhenInside="false" Continuous="false" LocalVar="A32NX_SOUND_BOARDING_COMPLETE">
+        <Sound WwiseEvent="welcomeonboard" WwiseData="true" NodeName="SOUND_FWD_GALLEY" ConeHeading="180" CancelConeHeadingWhenInside="false" Continuous="false" LocalVar="A32NX_SOUND_WELCOME_ONBOARD">
```
Benjozork, line 2186: Do we not need a similar change in the A380 sound.xml?

heclak: > Do we not need a similar change in the A380 sound.xml?

I think the A380 doesn't have announcements implemented yet. Although I'm curious about this bug. It was created when the new wwise soundpacks were updated. I'm not sure if this is a bug in the soundpacks or if it was a bug in the soundpacks previously but now "fixed".

Ditoo29: I found no pointers to this sound in the A380 just like @heclak said. But let me know if you need anything else to be added :D

Ditoo29:
Hey @Benjozork just bumping this in case it got forgotten. Thanks :D

Ditoo29:
@BravoMike99 sorry to ping you here, not sure what other developers do code reviews. Hopefully this isn't forgotten.

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
Maximilian-Reuter, line 749: in my opinion, there needs to be another entry here for `pax_welcome` so it is also properly stopped when the other ones are.

Maximilian-Reuter: Correct me if I am wrong here, but the call to `self.boarding_sounds.update_welcome(delta_time);` is in every branch of this function so why not call it once at the end or something ?

Maximilian-Reuter (changes requested):
In general I feel like the logic of playing the welcome aboard should be all contained in `BoardingSounds` as it doesnt depend on any external factors but elapsed time from the boarding completed sound. 

Not sure personally about the best way to deal with that as currently `BoardingSounds` doesnt have a concept of delta time. @Gurgel100 do you have ideas ?

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
Ditoo29: I just had it like that because the GSX branch returned. I simply added an else now and am calling the update_welcome once in the end :D

Ditoo29: The `pax_ambience `isn't set to false here as well. As for how I have the logic for this welcome sound to play not sure if setting this to false here would break anything. Will dig deeper.

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -709,13 +719,27 @@ impl BoardingSounds {
...
 
+    pub fn update_welcome(&mut self, delta: Duration) {
+        if self.welcome_timer > Duration::ZERO && !self.pax_welcome {
+            self.welcome_timer += delta;
+            if self.welcome_timer >= Self::BOARDING_COMPLETE_WELCOME_DELAY {
+                self.pax_welcome = true;
+                self.welcome_timer = Duration::ZERO;
+            }
+        }
+    }
```
Gurgel100: What about using `DelayedTrueLogicGate` which essentially implements exactly this logic?

Maximilian-Reuter: See this is why I wanted a Rust expert to look at this! What are the best ways to get an `UpdateContext` to `BoardingSounds` ?

Gurgel100: Since this function is ultimately called in the `update` function one could pass the `UpdateContext` there.

Ditoo29: Hello guys, I recently changed some parts of my desktop and couldn't yet reestablish the environment for developing and testing once again. Is anyway available, by any chance, to pick up where I left and carry on with the necessary changes before merging? You can work directly on my fork if needed, feel free to ask anything! @Gurgel100 @Maximilian-Reuter

.github/CHANGELOG.md
```diff
@@ -1913,3 +1913,4 @@
 1. [EFB] Restructured APIs and made Navigraph Auth a reusable component - @MicahBCode (Mischa Binder)
 1. [ECAM] Added F units to CRZ and COND pages for Cabin temps. Currently tied to kg/lbs option in EFB -Patrick Macken  (@PatM on Discord)
 1. [A32NX/FCU] Fixed bug where the FCU selected altitude would immediately switch to 1000s on selecting the 1000ft inteval knob - @elliot747 (Elliot)
+1. [Sounds] Fixed boarding complete and welcome onboard announcements playing simultaneously by adding a delay between them - @Ditoo29 (dito29 on Discord)
```

rust-toolchain.toml
```diff
@@ -1,3 +1,4 @@
 [toolchain]
 channel = "1.93.0"
 targets = [ "wasm32-wasip1" ]
+components = ["rust-analyzer"]
```
Saschl: To fix the VS code error

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -720,7 +724,8 @@ impl BoardingSounds {
 
     pub fn play_sound_pax_complete(&mut self, playing: bool) {
         if playing && !self.pax_complete {
-            self.welcome_timer = Duration::from_millis(1);
+            self.welcome_pending = true;
+            self.welcome_delay = DelayedTrueLogicGate::new(Self::BOARDING_COMPLETE_WELCOME_DELAY);
```
Gurgel100: I don't think this is needed tbh. If `welcome_pending` was reset to `false` in the previous frame I don't think that in this frame it will be set to `true` again and this would reset the `DelayedTrueLogicGate` as well.

Saschl: Good point. I think this would have even caused an issue with the GSX integration as the gate would have been recreated every frame.

Gurgel100 (approved):

PR #10611, opened 2026-03-28
feat(a380x/fws): Add non normal config overspeed warning

### LINKED ISSUE #10597: A380X/FWS: Non normal config overspeed warning
### Aircraft Version

Development

### Description

Overspeed warning should support configurations other than the normal flap configurations.

### References (optional)

Image

_No response_

bruno_pt99

matze-tech:
Fixes #10597 

## Summary of Changes

Added the overspeed warnings for non normal configurations and changed the overspeed warning for the normal configurations from the position of the flaps lever to the actual position of the flaps/slats.
## References

image

## Additional context

Discord username (if different from GitHub): matze2346

## Testing instructions

1. Make sure the overspeed warnings for the normal configurations are still working as expected
2. Check that all the non normal overspeed warnings are working as expected


fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
```diff
@@ -3742,11 +3762,107 @@ export class FwsCore {
...
+    // Non-normal flap configuration overspeed monitoring based on actual surface limits
+    const sfcc1Aw = this.slatFlapsSystem1ActualPositionWord;
+    const sfcc2Aw = this.slatFlapsSystem2ActualPositionWord;
+
+    const flaps0 = sfcc1Aw.bitValueOr(19, false) || sfcc2Aw.bitValueOr(19, false);
+    const flapsGt1 = sfcc1Aw.bitValueOr(21, false) || sfcc2Aw.bitValueOr(21, false);
+    const flapsGt2 = sfcc1Aw.bitValueOr(22, false) || sfcc2Aw.bitValueOr(22, false);
+    const flapsGt3 = sfcc1Aw.bitValueOr(23, false) || sfcc2Aw.bitValueOr(23, false);
+
+    const slatsRetracted = sfcc1Aw.bitValueOr(12, true) && sfcc2Aw.bitValueOr(12, true);
```
Eagle941: Should it not be `||` instead of `&&`?

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
```diff
@@ -3742,11 +3762,107 @@ export class FwsCore {
...
+      { isExceeded: isAbnConfig && isF2_3 && isSGt1 && isOverspeed(196), speed: 196, subj: this.overspeedVfeF2_3_SGt1 },
+      { isExceeded: isAbnConfig && isFGt3 && isS0 && isOverspeed(186), speed: 186, subj: this.overspeedVfeFGt3_S0 },
+      { isExceeded: isAbnConfig && isFGt3 && isS0_1 && isOverspeed(184), speed: 184, subj: this.overspeedVfeFGt3_S0_1 },
+      { isExceeded: isAbnConfig && isFGt3 && isSGt1 && isOverspeed(182), speed: 182, subj: this.overspeedVfeFGt3_SGt1 },
+    ];
+
+    let activeLimitSubj: Subject<boolean> | null = null;
+    let lowestSpeed = Infinity;
+
+    for (const limit of vfeLimits) {
```
Eagle941: The code needs two iterations on `vfeLimits`, but I don't understand which configuration matches multiple VFE limits because there is a one-to-one correspondance between flap/slat position and VFE.

Eagle941 (changes requested):

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
matze-tech: Because the default fallback value is true, we must use `&&` to ensure the fallback works. In the other examples, the fallback value is false thats why we use `||`.

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
matze-tech: A disagreement between SFCC 1 and 2 can cause multiple configuration states to be active simultaneously.

Eagle941: In that case, I would select the working SFCC (1 or 2 in order). If both are faulty, then VFE is null.

matze-tech: Sorry, but I don't understand how that should work.

Eagle941: I mean the following
```typescript
const sfcc1Aw = this.slatFlapsSystem1ActualPositionWord;
const sfcc2Aw = this.slatFlapsSystem2ActualPositionWord;

const sfccAw = sfcc1Aw.isInvalid() ? sfcc2Aw : sfcc1Aw;
```
Then you can use `sfccAw` for the rest of the code and simplify the for loop.

lukecologne:
Regarding the two review comments:
The way that it works on the 320 and likely here too, is that it is indeed a logical `OR` as in the first review comment, but guarded with checks if the actual position word is invalid or NCD. In case of a disagree, the lower VFe applies automatically, which makes sense. For the VMo and MMo, there is simply no check for the flaps position, as if this triggers, we are overspeeding no matter the flaps config, so this also makes sense.

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
```diff
@@ -3742,11 +3762,107 @@ export class FwsCore {
...
+    else if (lever === 3 && isFlaps2Physical) isNormalConfig = true;
+    else if ((lever === 4 || lever === 5) && isFlaps3Physical) isNormalConfig = true;
+    else if (lever === 6 && isFlapsFullPhysical) isNormalConfig = true;
+
+    const isAbnConfig = !isNormalConfig;
+
+    // Collect all limits to ensure we only display the MOST RESTRICTIVE (lowest) VFE warning at any time
+    const vfeLimits = [
+      {
+        isExceeded: isNormalConfig && isFlaps1Physical && isOverspeed(VfeF1),
```
lukecologne: I want to note that `isOverspeed` seems to be choosing an ADR depending on the FWS side, in reality the overspeed warning is checking all ADRs and emitting the warning if any ADR is above the threshold, but this would probably need a larger rewrite of the overspeed logic, so may be out of scope for this.

matze-tech:
@lukecologne So we can keep the logic as it is?

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
matze-tech: I think it's a better idea to fix it in a future PR.

fbw-a380x/src/systems/systems-host/CpiomC/FlightWarningSystem/FwsCore.ts
matze-tech: should be fixed now.

matze-tech:
@Eagle941 Ready for a new review.

Eagle941 (approved):

.github/CHANGELOG.md
```diff
@@ -125,6 +125,7 @@
 1. [A32NX/FE] Add pitch and low energy warnings, improve VCTrend computation - @lukecologne (luke)
 1. [A32NX/FWS] Add `ELEC APU GEN FAULT` master caution and APU GEN fault light - @FozzieHi (fozzie)
 1. [A32NX/FWS] Fix master warning / caution lights sometimes not extinguishing properly - @FozzieHi (fozzie)
+1. [A380X/FWS] Add non normal config overspeed warning - @matze-tech (matze2346)
```
tracernz: Note: wrong release, and please add at the end of the release in general.

BravoMike99:
Dev team QA report.

Object of testing : #10611 
Aircraft          : A380X
Simulator         : 2024
Tier of Testing   : 1
Date              : 28/06/2026

Testing Process: 
Tested multiple over speed configuration warnings and checked the threshold was correct on the EWD.

Testing Results: 
Not Passed

Negatives:       
While retracting the flaps, its possible that the over-speed limit is "bust" while there is no red band displayed on the PFD. Perhaps we should only consider the abnormal case if the flap/slat jammed signal is true? Perhaps @lukecologne has some ideas on how to approach this.
Under some 

Media:
image

lukecologne:
The VFe on the PFD is computed using the actual numerical flaps/slats angle with a continuous lookup table, so it corresponds to the actual VFe of the current config. I'm not really sure anymore about the exact implementation in this PR, but AFAICT it uses the flap/slat actual position words, which only gives rough position ranges with discretized configs. The real FWC would probably also use a lookup table.

PR #10634, opened 2026-04-11
fix(380X/fuel): only try to equalize feedtanks when each is below transfer limit

Maximilian-Reuter:
Fixes an issue discovered during testing in #10307 

## Summary of Changes

The equalization triggers for the fuel feed tanks did not consider whether the tanks in question were allowed to get fuel. So in a rare edge case where one tank was below the limit and one above it would overfill the one above. This should now be impossible as both tanks need to be below the limit to activate the equalization triggers.

EDIT: After additional issues came to light in testing it now also makes sure to properly track the transferstate of the feed tanks during equalization triggers. Also the thresholds have been tightened and EQ triggers cannot trigger if above the stop trigger reset threshold.

## Additional context

Discord username (if different from GitHub): \_chaoz_

## Testing instructions

Arguably it will be close to impossible to reproduce the conditions for this issue so the best that can be done is to make sure nothing broke. For that reason I propose doing a few full flights with similar fuel loads as #10307 suggests.


fbw-a380x/src/systems/systems-host/CpiomF/LegacyFuel.ts
```diff
@@ -256,6 +256,8 @@ export class LegacyFuel implements Instrument {
       }
       if (
         (Math.abs(this.feed1TankQty.get() - this.feed4TankQty.get()) < 2 &&
+          this.feed1TankQty.get() < 6765 &&
```
donstim: Other than for a failure or non-normal operations case, I'm having trouble seeing how this could ever be invoked. It would take fulfillment of a different equalization trigger involving feed tank 2 or 3 to cause a transfer to one of these tanks to be active, then it would need for the tank receiving the transfer to be just below the transfer activation threshold and the other to be just above. The only time feed 1 or 4 could be above 6765 is when the initial fuel load is above that value, and only until it reaches that value the first time. Afterwards, the transfer will stop at that value, so it will never be above. With normal fuel loading, feeds 2 and 3 would never be loaded with less fuel than 1 and 4. With mids above 8T, feeds 2 and 3 don't go below 6857, so cannot cause an equalization trigger with 1 or 4 to occur below 6765. With mids below 8T, I don't see how feed 2 or 3 could get to 6436, then start back up again and have an equalization trigger with 1 or 4 while one of those is still above 6765.

fbw-a380x/src/systems/systems-host/CpiomF/LegacyFuel.ts
```diff
@@ -264,6 +266,8 @@ export class LegacyFuel implements Instrument {
       }
       if (
         (Math.abs(this.feed2TankQty.get() - this.feed3TankQty.get()) < 2 &&
+          ((!this.triggerActive(13) && this.feed2TankQty.get() < 7186 && this.feed3TankQty.get() < 7186) ||
```
donstim: This is another one that I don't see how it could ever be invoked in normal operations. This case would cover having feed 2 or 3 rising due to activating an equalization trigger with 1 or 4, and then having the rising tank be just below 7186 and the falling tank being just above 7186.  But 1 or 4 cannot be rising unless it hit an activation trigger. But if it did that, the transfer would stop at 6765. For one of feed 2 or 3 to be above 7186, the other one couldn't be less than 6765, which is what it would have to be to activate an equalization trigger.

donstim (approved):
I think this covers the very rare corner case that was found through QA. Although I left some comments regarding a couple of the conditionals that I don't think would ever be invoked in normal operations, I'm fine with leaving them in. They shouldn't cause any harm and may be helpful for some type of non-normal or failure case. Thanks for doing this.

donstim:
Note that this should not be tested in MSFS 2024 unless the flight model.cfg file from either PR #10307 or #10634 is used. Those files fix a fuel system issue for MSFS 2024.

Maximilian-Reuter:
This was tested extensively on discord so I feel confident in merging this

PR #10639, opened 2026-04-15
fix(a380X/flight model): Drag Update for MSFS 2024 Flight Model

donstim:
Fixes #[issue_no]

## Summary of Changes
1. Adjusts the drag level to match IRL fuel burn data similarly to the MSFS 2020 version. One difference is that a more aft CG was used for some conservatism and to better match the target CG for the A380X in flight. (Unlike the real airplanes, most airliners in MSFS (except for our A32NX,) have higher drag as CG moves aft.)
2. Revises the VNAV drag parameters to match this drag level.

(Note:  This PR also contains the changes made in #10307. That PR is considered complete and just waiting for splitting the MSFS 2020 and 2024 versions of the A380X.) **As with PR #10307 this PR can only be used with MSFS 2024.**
## Additional context
In MSFS, the drag polar appears to be a function of gross weight, altitude, and CG. The drag parameters used for the VNAV path determination represent a mix of data using 3 different weights and 4 different altitudes.

Discord username (if different from GitHub): donbikes

## Testing instructions
1. Compare fuel burn to SimBrief fuel burn on full flights. Try to note fuel burned at each waypoint compared to the plan (considering actual winds vs what the plan was based on).
2. Confirm that VNAV descents are reasonable, that is, the airplane maintains the vertical path without the need for significant adjustments.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10641, opened 2026-04-17
fix(a380x/fuel system): Additional fuel system fixes for MSFS 2024

donstim:
This fixes an issue with the MSFS 2024 version of the A380X where the fuel transfers from the inner and middle tanks to the feed tanks will not be balanced if the transfer is not going to all 4 feed tanks at the same time. In other words, when only 2 or 3 of the 4 feed tanks are receiving a transfer, the transfer rates are not equal.

This PR also contains the fixes in PRs #10307  and #10639.

**As with PRs #10307 and #10639, this PR can only be used in MSFS 2024. It is only compatible with MSFS 2024 SU5 beta 1.7.21.0 and higher.**

## Summary of Changes
This updates the MSFS fuel system used for the A380X to version 5 and updates the fuel pump pressures and fuel line fuel flow per psi values to obtain the correct transfer flow rates and to balance the flows to each feed tank for each transfer condition.

Version 4 of the MSFS fuel system is used for the MSFS 2020 version of the A380X. This version resulted in imbalanced transfer flows when used in MSFS 2024. The transfers could be balanced when the transfers are simultaneously going to all 4 feed tanks using different fuel flow per psi values. This fix is what is in PR #10307.  Later, it was noticed that the transfers were no longer balanced if the transfers were not simultaneously going to all 4 feed tanks. This was not possible to fix while using version 4 of the MSFS fuel system.

It was found that the transfers could be balanced in all situations if the MSFS fuel system was upgraded to version 5. With the fuel flow per psi values used to balance the transfers, it was found that the fuel tank pump pressures needed to be changed to maintain the correct transfer flow rates. With SU5 beta 1.7.21.0, MSFS fuel system 7 can be used, and that is what is used in this PR. It enables removing the [FUEL} section of the flight_model.cfg file and placing the FUEL TYPE parameter that was only available in the {FUEL} section to the {FUEL_SYSTEM] section.

## Additional context

## Testing instructions
Perform a flight with a fuel load of 160 to 180 T. The flight should be long enough to transfer all fuel loaded into the inner, middle, trim, and outer fuel tanks. Make sure you have the FDR option turned on in the EFB and upload your FDR file here.


Maximilian-Reuter (approved):
LGTM

PR #10652, opened 2026-04-25
feat(a380x): wind page -2020

### LINKED ISSUE #9729: Wind Import Function For the A380
### Aircraft Version

Development

### Description

Wind import function for the A380, which is quite crucial for long flights, as wind can change both ETA and the associated fuel predictions quite a lot. Currently, both ETA and fuel pred change constantly and dramatically, one minute EFOB would be 13.4 and the next it would be 7.9. 

### References (optional)

I believe wind import is already implemented in the A32NX. 

_No response_

_No response_

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Adds wind page to the A380X.

## Screenshots (if necessary)

CLB Page
image
TMPY Restriction:
image
Wind Request:
WindReq2

## Additional context

## Testing instructions

A380X:
Check that you can access the wind page and create, modify or delete winds on the CLB, CRZ, DES pages alongside the alternate wind.
If a temporary flight plan exists, no wind modification should be possible on the active flightplan wind page.
Check that climb winds can no longer be edited once at or beyond the climb phase.
Check that cruise or descent winds can no longer be edited once at or beyond the descent phase.
Check that you can insert or reject a wind uplink and that the loaded winds match what the flight plan has.
Check that you can access the wind page via the F-PLN page through the revisions menu, or by selecting "T.WIND/EFOB" and clicking the wind cell.
Check that winds are displayed correctly on the F-PLN page if "T.WIND/EFOB" is selected (there is a text overlap bug atm but that is out of scope).
If you perform more than one flight, check that the HISTORY winds page reports the previously stored winds and that you can insert those into the nto the climb winds during preflight. For the wind entry at the cruise flight level a "CRZ FL" should appear next to the associated line.
Perform a normal flight and check for any other possible regressions in regards to wind, fms predictions or fms page behaviour.

A32NX:
Due to the nature of the changes in the wind structure a similar test to the A380X is required. No new functionality is expected, just a general regression test:
Check that wind uplink works and the displayed winds match what the flight plan has.
Check that you can edit/delete/add winds on the wind pages alongside the alternate wind.
Check that F-PLN B page displays winds in accordance to what was loaded in the flight-plan. 
Check that inserting an approach wind on the PERF page works as usual. This wind should also be copied automatically onto the descent winds.
If you perform more than one flight, check that the HISTORY winds page shows the previously stored winds and that you can insert those into the climb winds during pref-light.


fbw-a32nx/src/systems/fmgc/src/flightplanning/data/wind.ts
```diff
@@ -28,30 +35,100 @@ export enum FlightPlanWindEntryFlags {
...
+  return result;
+}
+
+export function extractTheta(vector: WindVector) {
+  return Math.atan2(vector.magnitude!, vector.direction!);
+}
+
+export function getVectorMagnitude(vector: WindVector) {
+  return vector.direction !== undefined && vector.magnitude !== undefined
+    ? Math.hypot(vector.direction, vector.magnitude)
```
BlueberryKing: How does `Math.hypot` work here?

fbw-a32nx/src/systems/fmgc/src/flightplanning/data/wind.ts
```diff
@@ -28,30 +35,100 @@ export enum FlightPlanWindEntryFlags {
...
+  //
+  if (theta === undefined) {
+    return {
+      direction: undefined,
+      magnitude: magnitude,
+    };
+  } else {
+    return {
+      direction: magnitude !== undefined ? magnitude * Math.cos(theta) : direction,
+      magnitude: magnitude !== undefined ? magnitude * Math.sin(theta) : undefined,
```
BlueberryKing: Why do we need `sin` and `cos` here? Don't we already have the direction and magnitude?

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -2871,53 +2891,68 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
...
-              vector: windEntry.vector,
-              type: windPropagationType,
-              sourceLegIndex: i,
-            });
+      const entries = this.draftCruiseWindExists ? this.draftCruiseWindEntries!.get(i) : element.cruiseWindEntries;
+      if (entries !== undefined) {
+        for (const windEntry of entries) {
+          let windPropagationType: PropagationType;
+          // Do not propagate entries will null FLs.
+          if (i === atIndex || windEntry.altitude === undefined) {
```
BlueberryKing, line 2902: Hmm, won't this mean that wind entries with undefined altitudes will always be added to the result array?

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -3067,6 +3164,71 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
...
+      vector: entry.vector !== undefined ? cloneWindVector(entry.vector) : undefined,
+    };
+  }
+
+  protected static isPartlyFilledWindVector(vector: WindVector) {
+    return (
+      (vector.direction !== undefined && vector.magnitude === undefined) ||
+      (vector.direction === undefined && vector.magnitude !== undefined)
+    );
+  }
```
BlueberryKing: Maybe these two functions should be in `wind.ts`?

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -99,13 +108,23 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
 
   private static readonly WindCache: PropagatedWindEntry[] = [];
 
+  protected readonly draftCruiseWindEntries: Map<number, WindEntry[]> | undefined;
+
+  protected draftCruiseWindExists = false;
```
BlueberryKing: Do we need this property? Seems like it's always just equal to `draftCruiseWindEntries !== undefined`.

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -99,13 +108,23 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
 
   private static readonly WindCache: PropagatedWindEntry[] = [];
 
+  protected readonly draftCruiseWindEntries: Map<number, WindEntry[]> | undefined;
```
BlueberryKing: I am not opposed to this, but why do we store these on the flight plan, rather than the leg, like the non-draft cruise wind entries?

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/FlightPlan.ts
```diff
@@ -63,9 +93,42 @@ export class FlightPlan<P extends FlightPlanPerformanceData = FlightPlanPerforma
...
+   * The draft descent wind entries used to store pilot edits prior to these being applied to the plan. If undefined, draft winds are disabled.
+   */
+  private readonly draftDescentWindEntries: FlightPlanWindEntry[] | undefined;
+  /**
+   * The draft cruise wind entry used to store pilot edits prior to these being applied to the plan. If undefined, draft winds are disabled.
+   */
+  private readonly alternateDraftWind: WindVector | undefined;
+  private draftClimbWindExists = false;
+  private draftDescentWindExists = false;
+  private alternateDraftWindExists = false;
```
BlueberryKing: Same here, can we replace them with a null check of the actual entry-arrays?

fbw-a32nx/src/systems/fmgc/src/flightplanning/uplink/WindUplinkUtilts.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const legPrediction = legPredictions?.get(i);
+    return legPrediction !== undefined
+      ? legPrediction.profilePhase === ProfilePhase.Cruise
+      : leg.segment.class === SegmentClass.Enroute;
+  }) as FlightPlanLeg[];
+
+  let cruiseWinds: CruiseWindRequest | undefined = undefined;
+  if (shouldRequestCruiseWinds && cruiseLegs.length > 0) {
+    const propagatedWinds = flightPlanService.propagateWindsAt(0, [], planIndex);
```
BlueberryKing: Can we do this? Then we don't need to pass in the flight plan service into this function.
```suggestion
    const propagatedWinds = plan.propagateWindsAt(0, []);
```

fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanManager.ts
```diff
@@ -69,6 +69,10 @@ export class FlightPlanManager<P extends FlightPlanPerformanceData> {
     private readonly performanceDataInit: P,
     private readonly syncClientID: number,
     private readonly master: boolean,
+    private readonly flightplanDraftWindsEnabled = false,
```
BlueberryKing: Maybe we should just pass in the config here. I think the flight plan manager deserves to know about its own config. 😄

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/wind/WindUtils.ts
```diff
@@ -14,41 +19,99 @@ export class WindUtils {
...
+  /**
+   * Sets the polar components of a wind vector.
+   * @param r - the new length (magnitude).
+   * @param theta - the new polar angle theta, in radians.
+   * @param vector - the wind vector to change.
+   * @returns the wind vector after it has been changed.
+   */
+  public static setPolar(r: number, theta: number, vector: WindVector) {
+    vector.direction = r * Math.cos(theta);
+    vector.magnitude = r * Math.sin(theta);
```
BlueberryKing: I don't think this is correct. You should not need to use `cos` and `sin` here. The mangitude will be equal to `r` directly. You might only have to convert the direction `theta` from radians to degrees if you conventionally store it in degrees

fbw-a32nx/src/systems/fmgc/src/flightplanning/data/wind.ts
```diff
@@ -1,16 +1,23 @@
...
 }
 
 export interface FlightPlanWindEntry extends WindEntry {
   flags: number;
 }
 
-export type WindVector = Float64Array;
+export type WindVector = {
+  magnitude: number | undefined;
+  direction: number | undefined;
```
BlueberryKing, line 20: Maybe we should write a doc comment to specify whether `direction` is in degrees or radians.

fbw-a32nx/src/systems/fmgc/src/guidance/vnav/wind/WindUtils.ts
```diff
@@ -14,41 +19,99 @@ export class WindUtils {
...
+  }
+
+  /**
+   * Gets the dot product of two wind vectors.
+   * @param v1 The first wind vector.
+   * @param v2 The second wind vector.
+   * @returns The dot product of the vectors.
+   */
+  public static dot(first: WindVector, second: WindVector): number {
+    return (first.direction ?? 0) * (second.direction ?? 0) + (first.magnitude ?? 0) * (second.magnitude ?? 0);
```
BlueberryKing: This doesn't work unfortunately. You'll want to make sure both vectors are complete, then you can use
```
return first.magnitude * second.magnitude * Math.cos((first.direction - second.direction) * MathUtils.DEGREES_TO_RADIANS)
```
assuming the direction of a vector is stored in degrees.

fbw-a380x/src/systems/instruments/src/MFD/ATCCOM/AtcDatalinkSystem.ts
```diff
@@ -46,24 +53,33 @@ export interface AtcErrorMessage {
...
   private requestAtsuStatusCodeCallbacks: ((code: AtsuStatusCodes, requestId: number) => boolean)[] = [];
 
   private atisAutoUpdates: string[] = [];
 
+  /** Map containing the wind request ids associated with the flightplan index which they belong to */
+  private readonly uplinkWindRequestFlightPlanMap: Map<number, number> = new Map();
+
   private atisReportsPrintActive: boolean = false;
 
+  private windsResponseCallbacks: ((
```
BlueberryKing: Is this still WIP?

fbw-a32nx/src/systems/fmgc/src/flightplanning/FlightPlanService.ts
```diff
@@ -984,29 +1060,79 @@ export class FlightPlanService<P extends FlightPlanPerformanceData = FlightPlanP
...
+    }
+    return Promise.resolve(false);
+  }
+
+  async historyWindInsertionAllowed(): Promise<boolean> {
+    const historyWindEntries = await this.getHistoryWindsEntries();
+    return Promise.resolve(
+      !this.hasTemporary &&
+        this.hasActive &&
+        this.flightPhase.get() === FmgcFlightPhase.Preflight &&
```
BlueberryKing: I am not sure the flight plan service should care about this. Checking whether it makes sense to insert should be handled by the caller, not the service. Checking certain conditions involving the flight plans makes sense, but introducing a dependency to the flight phase is not a good idea imo.

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -3836,13 +3827,17 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
...
+      : false;
 
-      this.flightPlanService.setDescentWindEntry(0, null, forPlan);
+    if (!hasDestination) {
+      this.setScratchpadMessage(NXSystemMessages.notAllowed);
+      return false;
+    }
 
+    if (s === Keypad.clrValue) {
+      this.bus.getPublisher<FmsWindEvents>().pub('delete_approach_wind', forPlan, false, false);
```
BlueberryKing: Why use the event bus for this and not just a function call?

fbw-a32nx/src/systems/instruments/src/MCDU/legacy/A32NX_FMCMainDisplay.ts
```diff
@@ -3856,20 +3851,17 @@ export abstract class FMCMainDisplay implements FmsDataInterface, FmsDisplayInte
...
-
-    const destinationMagVar = plan.destinationAirport
-      ? Facilities.getMagVar(plan.destinationAirport.location.lat, plan.destinationAirport.location.long)
-      : 0;
-
-    const theta = MagVar.magneticToTrue(dir, destinationMagVar) * MathUtils.DEGREES_TO_RADIANS;
-
-    const groundWind = { altitude: 0, vector: Vec2Math.setFromPolar(mag, theta, Vec2Math.create()), flags: 0 };
-    this.flightPlanService.setDescentWindEntry(0, groundWind, forPlan, false);
+    this.bus.getPublisher<FmsWindEvents>().pub(
```
BlueberryKing: Why use the event bus for this and not just a function call?

BlueberryKing (changes requested):

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
BravoMike99: Good idea, I didn't really think about modifying the leg tbh. The only issue I see is not being able to protect against arbitrary writes due to the loss of the FPM config flag and so on. It improves readability however, what do we want to go for here?

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/FlightPlan.ts
BravoMike99: I did it this way to make sure that nobody can modify the draft winds if its not enabled by the config as the undefined is forced upon flight-plan creation and the variables are readonly. As such, their existence is checked against the boolean flags. To use the null check we'd have to drop the readonly.

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
BravoMike99: Hmm good point, it would never match in `existingEntryIndex` before. There can only be one undefined altitude per leg (UI logic) so the new check should match it. If you prefer, we could also add a check for this in the edit/add new cruise wind entry just to be on the safe side.

fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -99,13 +108,21 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
 
   private static readonly WindCache: PropagatedWindEntry[] = [];
 
+  protected readonly draftCruiseWindEntries: Value<Map<number, WindEntry[]> | null> | undefined;
```
BlueberryKing, line 116: We should probably document the meaning of the null and undefined states here.

BravoMike99: Good idea, done

BlueberryKing (approved):

PR #10659, opened 2026-05-03
fix(a380x/fire): emit NCD on FDU discrete word when unpowered

bld481:
## Summary

The Fire Detection Unit's `update_discrete_word` emitted
`SignStatus::NormalOperation` unconditionally, with a `// TODO: Add
electrical supply for FDU, when not powered it should return NCD` at
fire_and_smoke_protection.rs:288. With the FDU's DC bus dead the discrete
word still claimed Normal Operation and downstream consumers (FWS, overhead)
had no way to know the FDU was offline.

## Change

Wire the FDU to a `DirectCurrentEssential` bus, and gate
`update_discrete_word` on `is_powered`. When unpowered, emit NCD with zero
data (matching the FDU init at line 184). Implementation pattern mirrors
`FireDetectionLoop` in the same file: `powered_by`, `is_powered`, and a
`receive_power` impl on `SimulationElement`.

## Test plan

- [ ] With DC ESS powered: verify FDU discrete word SSM=NormalOperation
- [ ] With DC ESS dead: verify FDU discrete word SSM=NoComputedData and data=0
- [ ] Verify FWS reacts appropriately to NCD on the FDU word

## Notes

**Working assumption:** `DirectCurrentEssential` matches loop A's bus
(the more-protected of the two detection loops). a380-simvars.md already
labels the FDU discrete word layout as "(assumed)" — no public reference
for the real A380 FDU bus. Happy to amend if the project has a different
intended bus (e.g., DC HOT, or a separate bus per redundant FDU).

### DIFF
--- a/fbw-a380x/src/wasm/systems/a380_systems/src/fire_and_smoke_protection.rs
+++ b/fbw-a380x/src/wasm/systems/a380_systems/src/fire_and_smoke_protection.rs
@@ -73,7 +73,10 @@ impl FireProtectionSystem {
 
     fn new(context: &mut InitContext) -> Self {
         Self {
-            fire_detection_unit: FireDetectionUnit::new(context),
+            fire_detection_unit: FireDetectionUnit::new(
+                context,
+                ElectricalBusType::DirectCurrentEssential,
+            ),
             fire_extinguishing_system: FireExtinguishingSystem::new(context),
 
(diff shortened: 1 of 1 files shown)


tracernz:
NCD would not be appropriate for unpowered. It should be no transmission at all, but our bus implementation isn't capable of that so we use FW instead of no transmission.

PR #10662, opened 2026-05-04
feat(a32nx): add TCAS bus, improve PFD TCAS display

lukecologne:
## Summary of Changes

* Implement the TCAS high speed output bus.
* Refactor the PFD V/S indicator to use MappedSubject pattern
* Improve the PFD TCAS RA indication graphics and logic, and switch to the new TCAS bus.
  * The V/S digital indicator is now colored red when inside the up or down red resolution advisory zones
  * The zone border is now properly angle so that the V/S needle aligns. The green fly to zone is now properly extended and flat during either up or down corrective resolution adivsories.
  * Multi-Aircraft RAs are also supported by the RA display, but our TCAS implementation does not support them currently.
  * The TCAS amber flag now also uses the TCAS bus.
* Also switch the FMGC from a TCAS bus shim to the new TCAS bus.

## Screenshots (if necessary)

Graphics comparison:
| Before | After|
|--------|--------|
| grafik| grafik|
| grafik| grafik| 

## References

Example of the PFD TCAS RA indications https://www.youtube.com/watch?v=nwL51Vs7sOw

## Additional context

## Testing instructions

* Trigger TCAS RAs, check that the RA indication is reasonable and corresponds to the aural command given. Also check that the AP/FD TCAS mode properly follows the indicated TCAS RA order.
* Check that the TCAS flag on the PFD appears when the TCAS is unable to provide RA information (Radio Altimeter, ADR off etc.), provided that the TCAS is not in standby.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -152,6 +152,7 @@
 1. [EFB] Add FAA and LIDO charts supplied by MSFS2024 - @tracernz (Mike)
 1. [EFB] Added a cold temperature correction calculator - @tracernz (Mike)
 1. [A32NX/FLIGHT MODEL] Updated A32NX flight model for MSFS 2024 - @donstim (donbikes)
+1. [A32NX/TCAS] Improve PFD Resolution Advisory graphics, introduce TCAS bus - @lukecologne (luke)
 
 ## 0.14.0
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -13,6 +13,7 @@
 ## 2024.2.0
 
 1. [ATSU] Add support for BeyondATC and SayIntentions AI as METAR/ATIS sources - @saschl
+1. [A32NX/TCAS] Improve PFD Resolution Advisory graphics, introduce TCAS bus - @lukecologne (luke)
 
 ## 2024.1.0
 
@@ -152,7 +153,6 @@
 1. [EFB] Add FAA and LIDO charts supplied by MSFS2024 - @tracernz (Mike)
 1. [EFB] Added a cold temperature correction calculator - @tracernz (Mike)
 1. [A32NX/FLIGHT MODEL] Updated A32NX flight model for MSFS 2024 - @donstim (donbikes)
-1. [A32NX/TCAS] Improve PFD Resolution Advisory graphics, introduce TCAS bus - @lukecologne (luke)
 
 ## 0.14.0
 
--- a/fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
@@ -78,13 +78,15 @@ export class VerticalSpeedIndicator extends DisplayComponent<VerticalSpeedIndica
 
   private readonly ra = ConsumerSubject.create(this.sub.on('chosenRa'), Arinc429Word.empty());
 
-  private verticalResolutionAdvisoryWord = Arinc429LocalVarConsumerSubject.create(
+  private readonly verticalResolutionAdvisoryWord = Arinc429LocalVarConsumerSubject.create(
     this.sub.on('a32nx_tcas_vertical_resolution_advisory_word'),
   );
 
-  private tcasModeWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_mode_word'));
+  private readonly tcasModeWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_mode_word'));
 
-  private tcasFaultSummaryWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_fault_summary_word'));
+  private readonly tcasFaultSummaryWord = Arinc429LocalVarConsumerSubject.create(
+    this.sub.on('a32nx_tcas_fault_summary_word'),
+  );
 
   private readonly vsFlagVisible = this.verticalSpeed.map((vs) => !vs.isNormalOperation());
 
--- a/fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
@@ -3,20 +3,58 @@
 // SPDX-License-Identifier: GPL-3.0
 
 import {
-  ClockEvents,
   ComponentProps,
+  ConsumerSubject,
   DisplayComponent,
   FSComponent,
+  MappedSubject,
   Subject,
   Subscribable,
   VNode,
 } from '@microsoft/msfs-sdk';
-import { ArincEventBus, Arinc429Word } from '@flybywiresim/fbw-sdk';
+import { ArincEventBus, Arinc429Word, Arinc429LocalVarConsumerSubject, Arinc429WordData } from '@flybywiresim/fbw-sdk';
 
 import { Arinc429Values } from './shared/ArincValueProvider';
-import { PFDSimvars } from './shared/PFDSimvarPublisher';
 import { LagFilter } from './PFDUtils';
 import { FlashOneHertz } from '../MsfsAvionicsCommon/FlashingElementUtils';
+import { A32NXTcasBusEvents } from '@shared/publishers/A32NXTcasBusPublisher';
+
+function bitsToInteger(word: Arinc429WordData, startBit: number, endBit: number): number {
+  const mask = ((1 << (endBit - startBit + 1)) - 1) << (startBit - 1);
+
+  return (word.value & mask) >> (startBit - 1);
+}
+
+function resolutionAdvisoryNumberToVs(resAdv: number, rateToMaintain: number, sign: number) {
+  if (resAdv == 1) {
+    return rateToMaintain;
+  } else if (resAdv == 2) {
+    return sign * 250;
+  } else if (resAdv == 3) {
+    return sign * 500;
+  } else if (resAdv == 4) {
+    return sign * 1000;
+  } else if (resAdv == 5) {
+    return sign * 2000;
+  } else {
+    return sign * 32767;
+  }
+}
+
+function yPosFromVs(vsValue: number): number {
+  const absVSpeed = Math.abs(vsValue);
+  const sign = Math.sign(vsValue);
+
+  if (absVSpeed < 1000) {
+    return (vsValue / 1000) * -27.22;
+  } else if (absVSpeed < 2000) {
+    return ((vsValue - sign * 1000) / 1000) * -10.1 - sign * 27.22;
+  } else if (absVSpeed < 6000) {
+    return ((vsValue - sign * 2000) / 4000) * -10.1 - sign * 37.32;
+  } else {
+    return sign * -47.37;
+  }
+}
 
 interface VerticalSpeedIndicatorProps {
   bus: ArincEventBus;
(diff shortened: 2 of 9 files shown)


fbw-a32nx/src/systems/shared/src/publishers/A32NXTcasBusPublisher.ts
```diff
@@ -0,0 +1,159 @@
+// @ts-strict-ignore
+// Copyright (c) 2025 FlyByWire Simulations
+// SPDX-License-Identifier: GPL-3.0
```
BravoMike99: again :P

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
```diff
@@ -208,6 +210,12 @@ export class TcasComputer {
 
   private downAdvisoryStatus: LocalSimVar<UpDownAdvisoryStatus>; // Down adivsory status (descend, don't climb, don't climb >500 etc.)
 
+  private tcasBusResolutionAdvisoryWord: Arinc429Register;
+
+  private tcasBusTcasModeWord: Arinc429Register;
+
+  private tcasBusTcasFaultSummaryWord: Arinc429Register;
```
BravoMike99: these could be readonly right?

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
```diff
@@ -1298,6 +1313,46 @@ export class TcasComputer {
     this.syncer.sendEvent('A32NX_TCAS_TRAFFIC', this.sendAirTraffic);
   }
 
+  private updateBus(): void {
+    this.tcasBusResolutionAdvisoryWord.setSsm(
+      !this.tcasPower ? Arinc429SignStatusMatrix.FailureWarning : Arinc429SignStatusMatrix.NormalOperation,
+    );
+
+    const uintRateToMaintain = Math.round(Math.abs(this.rateToMaintain.getVar()) / 100) & 0b00111111;
```
BravoMike99, line 1321: Do we want to continue updating the bits of the word even if its FW?

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
```diff
@@ -1298,6 +1313,46 @@ export class TcasComputer {
...
+      combinedControl = 6;
+    }
+
+    this.tcasBusResolutionAdvisoryWord.setValue(
+      (uintRateToMaintain << 10) |
+        (this.rateToMaintain.getVar() < 0 ? 1 << 16 : 0) |
+        (combinedControl << 17) |
+        (((this.raType.getVar() as number) & 0b00000111) << 20) |
+        (((this.upAdvisoryStatus.getVar() as number) & 0b00000111) << 23) |
+        (((this.downAdvisoryStatus.getVar() as number) & 0b00000111) << 26),
```
BravoMike99: Are the casts needed?

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
```diff
@@ -3,175 +3,211 @@
...
+
+  private readonly ra = ConsumerSubject.create(this.sub.on('chosenRa'), Arinc429Word.empty());
+
+  private verticalResolutionAdvisoryWord = Arinc429LocalVarConsumerSubject.create(
+    this.sub.on('a32nx_tcas_vertical_resolution_advisory_word'),
+  );
+
+  private tcasModeWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_mode_word'));
+
+  private tcasFaultSummaryWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_fault_summary_word'));
```
BravoMike99: These could also be readonly?

fbw-a32nx/src/systems/shared/src/publishers/A32NXTcasBusPublisher.ts
lukecologne: I should stop copy-pasting files and forgetting about this xD

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
lukecologne: Not really, but as the computer itself and thus failures are not really fully simulated, I think it's fine for now. This would be fixed if the TCAS computer is implemented as an actual computer

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
lukecologne: Technically yes, but they are initialized in the `init()` function, which prohibits using readonly here. As I don't want to break the pattern, I'll leave it as is.

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
lukecologne: Checked the resulting JS, and there is absolutely no difference and the compiler doesn't complain, so no, apparently not. Will remove

BravoMike99 (approved):

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
```diff
@@ -3,175 +3,213 @@
...
-    sub.on('simTime').handle((_r) => {
-      if (this.needsUpdate) {
-        if (this.tcasState.tcasState === 2) {
-          this.needleColour.set('White');
-        }
-        this.vspeedTcas.instance.update(this.tcasState);
+  private readonly sub = this.props.bus.getArincSubscriber<A32NXTcasBusEvents & Arinc429Values>();
+
+  private readonly verticalSpeed = ConsumerSubject.create(
+    this.sub.on('vs').withArinc429Precision(3),
```
tracernz: Do you really want this to only notify when it changes by 1000 (10^3)? These old school subscription filters don't really have any advantage over a .map where you have much more control over what you want..

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
```diff
@@ -3,175 +3,213 @@
...
+    this.sub.on('a32nx_tcas_vertical_resolution_advisory_word'),
+  );
+
+  private readonly tcasModeWord = Arinc429LocalVarConsumerSubject.create(this.sub.on('a32nx_tcas_mode_word'));
+
+  private readonly tcasFaultSummaryWord = Arinc429LocalVarConsumerSubject.create(
+    this.sub.on('a32nx_tcas_fault_summary_word'),
+  );
+
+  private readonly vsFlagVisible = this.verticalSpeed.map((vs) => !vs.isNormalOperation());
```
tracernz: Should probably be only .isFailureWarning().

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
```diff
@@ -330,200 +390,173 @@ class VSpeedText extends DisplayComponent<{
...
-          <VSpeedTcasZone
-            ref={this.redZoneElement}
-            zoneBoundLow={this.redZone}
-            zoneBoundHigh={this.redZoneHigh}
-            zoneClass="Fill Red"
-            isCorrective={this.isCorrective}
-            extended={Subject.create(false)}
+        <g
+          id="VerticalSpeedTCASGroup"
+          visibility={this.tcasIndicationsVisible.map((visible) => (visible ? 'visible' : 'hidden'))}
```
tracernz: In general 'inherit' is a better idea than 'visible', in case one of the parent elements is hidden. Not sure if that is possible in these components. It's safer to never use 'visible' unless you explicitly want it to be visible when a parent element is hidden.

fbw-a32nx/src/systems/tcas/src/components/TcasComputer.ts
tracernz: They should be typed optional like
```
 private tcasBusTcasFaultSummaryWord?: Arinc429Register;
```

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
lukecologne: I thought this was 10^-3? That would make sense, and the V/S worked as it should in testingm what would you suggest instead using map?

tracernz: 10^3 I'm afraid:
https://github.com/flybywiresim/aircraft/blob/f956b7d68ff9dfc5be4dc7460d95fd660c391385/fbw-common/src/systems/shared/src/ArincConsumer.ts#L61

It may be totally broken then if the VS still worked! 🤣

lukecologne: Well it does round(value *10^3)/10^3. That is three significant digits

tracernz: Looking at it more, it should probably be a bit different, so the filter runs all the time, as at the moment it will have a very inconsistent sampling interval unrelated to deltaTime if the VS word is not changing every frame.

```ts
private readonly verticalSpeed = ConsumerSubject.create(this.sub.on('vs'), Arinc429Word.empty());
private readonly filteredVerticalSpeed = Subject.create(0);

public override onAfterRender(node: VNode): void {
  super.onAfterRender(node);

  this.bus.getSubscriber<ClockEvents>().on('simTime').handle(() => {
    const filtered = this.lagFilter.step(this.verticalSpeed.get().value, this.props.instrument.deltaTime / 1000);
    this.filteredVerticalSpeed.set(MathUtils.round(filtered , 1)); // set the round value to what you like
  });
}
```

tracernz: > Well it does round(value *10^3)/10^3. That is three significant digits

Oh yeah, my brain is wired for the opposite pattern in MathUtils.round I guess!

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
lukecologne: Yeah, you're correct. Will fix

fbw-a32nx/src/systems/instruments/src/PFD/VerticalSpeedIndicator.tsx
lukecologne: > Looking at it more, it should probably be a bit different, so the filter runs all the time, as at the moment it will have a very inconsistent sampling interval unrelated to deltaTime if the VS word is not changing every frame.

Oh yeah, you're correct. I guess it's normally not very apparent, as the V/S is usually updated every frame

tracernz (approved):

tracernz (approved):

.github/CHANGELOG.md
```diff
@@ -147,6 +147,7 @@
 1. [EFB] Added a cold temperature correction calculator - @tracernz (Mike)
 1. [A380X/MFD] Fixed PERF APPR QNH entries in inHg being displayed as hPa - @Daboss57 (Daboss57)
 1. [EFB] Add FAA and LIDO charts supplied by MSFS2024 - @tracernz (Mike)
+1. [A32NX/TCAS] Improve PFD Resolution Advisory graphics, introduce TCAS bus - @lukecologne (luke)
```

lukecologne: Fixed

Saschl:

Discord Username  : saschl 
Object of testing : #10662
Aircraft          : A32NX 
Simulator         : 2024 
Tier of Testing   : 1
Date              : 11/06/2026

Testing Process: 
 
1. Attempted to cause some RAs by slewing to injected AI traffic
2. Watched indications and autopilot response
3. Checked if aural alerts match indication
4. Checked TCAS flag

Testing Results: 
Passed

Negatives:       


Conclusions: 

RAs seemed reasonable and the aural alerts matched, TCAS flag appears when ADR turned off.

Media:
Screenshot 2026-06-11 101026


Sorry for the messy video, was not that easy to trigger RAs
 

https://github.com/user-attachments/assets/36dae287-1a46-45dc-8581-5fbe9d51db40

PR #10670, opened 2026-05-09
fix: NXDataStore unset key handling on FS2024

tracernz:
Fixes #[issue_no]

## Summary of Changes

FS2024 GetStoredData returns null rather than '' for keys that don't exist. Fix up NXDataStore.getRaw so it coalesces null to '' and we get the same behaviour as FS2020 in all the consumers of the value (and it also makes it match the typing correctly).

This fixes the issue a few users have seen where the MCDU is in US units despite US units not being selected.

## Additional context

## Testing instructions

- Load the A32NX up in FS2024,
- Open the coherent debugger, select any instrument, and enter `DeleteStoredData('CONFIG_USING_METRIC')` in the console then press enter
- Restart the flight, enter any FROM/TO on INIT A, then go to INIT B and auto-fill the ZFW/ZFWCG. It should be 42.5 t rather than 93.7 klbs.


FozzieHi (approved):

BravoMike99:
Dev team QA report

Object of testing : #10670
Aircraft          : A32NX
Simulator         : 2024
Tier of Testing   : 1
Date              : 11/05/2025

Testing Process: 
Spawned in and cleared the datastore as per testing instructions and restarted flight.
Auto filled ZFW in INIT B,  verified it appeared in metric units.
Repeated the whole processes but for the FUEL PRED page, verified the same.

Testing Results: 
Passed

PR #10673, opened 2026-05-10
fix(fms): do not trigger lat disc if on hold without imm exit

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Fixes lateral discontinuity ahead activating on HM legs if IMM exit not selected.
## Additional context


fbw-a32nx/src/systems/fmgc/src/flightplanning/plans/BaseFlightPlan.ts
```diff
@@ -3070,8 +3070,13 @@ export abstract class BaseFlightPlan<P extends FlightPlanPerformanceData = Fligh
   getLastLegIndexBeforeDiscontinuity(): number | null {
     for (let i = this.activeLegIndex; i < this.allLegs.length; i++) {
       const nextLeg = this.maybeElementAt(i + 1);
-      // Handle case of end of flightplan or discontinuity outside of a manual leg.
-      if (isLeg(this.activeLeg) && !this.activeLeg.isVectors() && (!nextLeg || nextLeg.isDiscontinuity)) {
+      // Handle case of end of flightplan or discontinuity outside of a manual leg or hold without exit active.
+      if (
+        isLeg(this.activeLeg) &&
+        !this.activeLeg.isVectors() &&
+        (this.activeLeg.type !== LegType.HM || this.activeLeg.holdImmExit) &&
```
tracernz: I sort of think this is the wrong place to handle it, as this function no longer does what it's name implies. `shouldShowLatDiscontinuityAhead` should probably just inhibit the message if you're in an HM without IMM EXIT.

BravoMike99: Good idea, I just changed the whole active leg logic to that function instead so its all in one place.

tracernz (approved):

PR #10674, opened 2026-05-10
feat(fws): Add flight phase debug view on EWD

FozzieHi:
## Summary of Changes
Adds an EFB pin program option of "Avionics Test Mode" which displays the FWC flight phase number on the EWD when enabled. This pin program should also add a debug option for the FWC intermediate RA callouts when we add those.

Docs PR: flybywiresim/docs#1138

## Screenshots (if necessary)

Screenshot 2026-05-10 105924
image

Screenshot 2026-05-16 134615
image

## References

image
<br>
image
image

## Additional context

## Testing instructions

Perform for both the A32NX and A380X:
1. Spawn on a runway, enable the "Development Mode" pin program option in the EFB
2. Wait at least 30 seconds for the plane to enter flight phase 2, check that "2" displays on appropriate place on the EWD
3. Set takeoff power, check that "3" displays and "4" displays after 80kts, "5" ("6" for the A380) after lift off
4. Disable the "Development Mode" pin program option and check that the flight phase number disappears


fbw-a32nx/src/systems/instruments/src/EWD/FwcFlightPhaseIndicator.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  render(): VNode {
+    return (
+      <text class="Standard Center White" x={330} y={514} visibility={this.visibility}>
+        {this.displayedPhase}
+      </text>
+    );
+  }
+}
```
lukecologne, line 40: We should use the MappedSubject pattern, which is also simpler in this case. My (untested) suggestion:
```suggestion
export class FwcFlightPhaseIndicator extends DisplayComponent<FwcFlightPhaseIndicatorProps> {
  private readonly developmentMode = Subject.create(false);

  private readonly fwcFlightPhase = ConsumerSubject.create(
    this.props.bus.getSubscriber<EwdSimvars>().on('fwcFlightPhase'),
    0,
  );

  private cancelDevelopmentModeSub?: () => void;

  onAfterRender(node: VNode): void {
    super.onAfterRender(node);

    this.cancelDevelopmentModeSub = NXDataStore.getAndSubscribeLegacy(
      'CONFIG_A32NX_DEVELOPMENT_MODE',
      (_, developmentMode) => {
        this.developmentMode.set(developmentMode === '1');
      },
      '0',
    );
  }

  destroy(): void {
    this.cancelDevelopmentModeSub?.();
  }

  render(): VNode {
    return (
      <text
        class="Standard Center White"
        x={330}
        y={514}
        visibility={this.developmentMode.map((developmentModeEnabled) =>
          developmentModeEnabled ? 'inherit' : 'hidden',
        )}
      >
        {this.fwcFlightPhase}
      </text>
    );
  }
}
```

FozzieHi: So I did consider this, but I ended up pretty much copying and pasting `FOB.tsx`. I think it's simple enough to just do this, but I'm happy to change it to a MappedSubject if you think that's more readable.
https://github.com/flybywiresim/aircraft/blob/5ad053aaa42b6a89c8a0aad69bb91161b58c4cf0/fbw-a32nx/src/systems/instruments/src/EWD/FOB.tsx

BravoMike99:
Just an idea since this is related to the FWS. Perhaps it could be some variable which is only set by that specific code if the option is enabled and the EWD would show/hide it depending on SSM? (or change the pr title :d)  Also we should try to use the new datastore as it is a new setting I think.

tracernz:
How does this relate to the existing developer mode? The naming seems to invite confusion because it appears not connected.

fbw-a32nx/src/systems/instruments/src/EWD/FwcFlightPhaseIndicator.tsx
FozzieHi: Done, it makes much more sense now it's handled through the FWC

BravoMike99 (approved):
Just don't forget to change the testing instructions and the documentation with the new name later on.

.github/CHANGELOG.md
```diff
@@ -144,6 +144,7 @@
 1. [A32NX/FWS] Add `ELEC EMER CONFIG` master warning, `EMER GEN 1 LINE OFF` and `RAT FAULT` master cautions and `EMER GEN` memo - @FozzieHi (fozzie)
 1. [A32NX/FWS] Add `FAC 1(2)(1+2) FAULT` master cautions - @FozzieHi (fozzie)
 1. [A32NX/MCDU] Renamed "CHECK SPEED MODE" message to "SET MANAGED SPEED"- @BravoMike99 (bruno_pt99)
+1. [FWS] Add Avionics Test Mode pin program to display flight phase number on EWD - @FozzieHi (fozzie)
```

PR #10707, opened 2026-05-20
fix: prevent occasional freezing after tod pause

Saschl:
Fixes #[issue_no]

## Summary of Changes

Removes the input field focusing (which prevents inputs to prevent issues with additional bindings of the "P" key) as it sometimes does not give control back to the user after resuming from pause.

This can theoretically cause issues if someone has bound the "P" key to an action as this will not be intercepted anymore. But I consider this a valid tradeoff as the issue hard to reproduce and the cause unknown

## Additional context


Saschl:
QA not required IMO as the change is minimal and was tested by me in a flight.

tracernz (approved):

PR #10722, opened 2026-05-26
fix(a380x/fms): Auto-delete altitude constraints at or above CRZ FL

Popespice:
## Summary

Fixes #10619 — the A380X FMS does not display `CONSTRAINTS ABOVE CRZ FL : DELETED` or remove the relevant constraints when the cruise flight level is changed or a SID/STAR is inserted.

**Changes:**
- Added `deleteConstraintsAboveCruiseLevel()` to `FmcAircraftInterface` — iterates the active flight plan, clears any altitude constraint whose value is at or above the current CRZ FL, and queues the `CONSTRAINTS ABOVE CRZ FL : DELETED` scratchpad message if any were removed
- Added `cstrDelAboveCrzFl` message to `NXSystemMessages`
- Wired the trigger to all applicable events:
  - CRZ FL changed via FMS PERF page (`onUpdateCruiseLevel`)
  - CRZ FL changed via FCU alt knob (`onTrySetCruiseFlightLevel`)
  - Step climb/descent crossing detected (`onStepClimbDescent`)
  - SID selected or cleared (`MfdFmsFplnDep`)
  - SID enroute transition selected or cleared (`MfdFmsFplnDep`)
  - Destination runway changed (`MfdFmsFplnArr`)
  - STAR selected or cleared (`MfdFmsFplnArr`)
  - STAR enroute transition selected or cleared (`MfdFmsFplnArr`)
  - Cruise step inserted via VERT REV page (`MfdFmsFplnVertRev`)

**Constraint logic:**
- `AT` and `AT OR ABOVE` (alt1): deleted if `altitude1 >= CRZ FL`
- `AT OR ABOVE` alt2 descriptor ('C'): deleted if `altitude2 >= CRZ FL`
- `BETWEEN` (window): deleted if upper bound (`altitude1`) `>= CRZ FL`
- `AT OR BELOW` (ceiling): intentionally **excluded** — a ceiling above CRZ FL does not require the aircraft to fly above cruise; removing it would strip valid STAR descent restrictions

**Message queue fix (incidental):** `addMessageToQueue` previously only deduplicated against already-cleared entries, causing the same message to stack if triggered in rapid succession. The dedup now covers live entries too, so re-triggering moves the message to the top of the queue instead of duplicating it.

## Test plan

- [ ] Load A380X, plan a route with a STAR that has altitude constraints above a low CRZ FL (e.g. EDDM→EDDF, DEBH1B STAR, CRZ FL080)
- [ ] Verify scratchpad shows `CONSTRAINTS ABOVE CRZ FL : DELETED` on STAR insertion
- [ ] Verify the flagged constraints are removed from the F-PLN vertical profile
- [ ] Raise CRZ FL above the constraint and verify the message appears again / constraints cleared
- [ ] Verify `AT OR BELOW` constraints (STAR descent ceilings) are **not** deleted
- [ ] Verify no message appears when no constraints are above CRZ FL

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/CHANGELOG.md
```diff
@@ -10,8 +10,9 @@
 <!--  1. [Changed Area] Title of changes - @github username (Name)  -->
 <!-- The changed area should be prefixed with "A32NX/" or "A380X/" if only one of them is affected. -->
 
-## 2024.1.0
+## 0.15.0
 
+1. [A380X/FMS] Automatically delete altitude constraints at or above CRZ FL - @Popespice (Connor)
```
tracernz, line 16: Note: wrong release. and please add changelogs at the end of the release in general.

tracernz:
This kind of thing should be all systems code; not sprinkled through the UI.

Popespice:
Thanks for the feedback — reworked it so the rule lives entirely in systems code now. The MFD pages are back to identical with master.

What it looks like now:

- The deletion logic is a single method on `BaseFlightPlan` in fmgc. It runs from the triggers the FCOM actually lists (DSC-22-FMS-10-40-60): procedure changes (the runway/SID/STAR/VIA/approach setters, after the operation queue has flushed), pilot entry of a new CRZ FL, and step insertion.
- Deletions now bump the plan version and sync the touched segments to other instruments, instead of mutating legs directly with a "TODO sync" like before. Segment-level sync is used because leg indices on other instruments can be stale mid-revision, and the serialized segment is also what carries pilot-entered constraints.
- The plan publishes a new `flightPlan.constraintsDeletedAboveCruiseLevel` event, and the A380X FMC posts CONSTRAINTS ABOVE CRZ FL : DELETED from a subscription — same pattern as STEP DELETED. The message shows for the active and temporary plans only.
- Since fmgc is shared, the rule is gated behind a new `FpmConfig` flag (on for the A380, off for the A320), so A32NX behaviour is unchanged — the rule and message text are only sourced from the A380 FCOM.

While moving it I rechecked the FCOM and fixed a couple of things in the rule itself: AT OR BELOW constraints at/above the CRZ FL are deleted too (I had wrongly excluded them — a ceiling at or above cruise can never bind anyway), and the scan now only covers the active leg onwards, stopping before the missed approach and skipping altitude-terminated legs (CA/FA/VA), where altitude1 is a termination altitude rather than a constraint.

Added specs in `FlightPlan.spec.ts` covering the descriptor cases, the triggers, the alternate-plan guard, and the event emission.

PR #10739, opened 2026-06-01
refactor(a32nx/pfd): speedtape refactor

lukecologne:
## Summary of Changes

Refactor PFD speed tape using modern avionics-framework. This also facilitates changes to the speedtape for certain new features.

## Additional context

### DIFF
--- a/fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
+++ b/fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
@@ -160,7 +160,7 @@ export class AirspeedIndicator extends DisplayComponent<{
           {/* Outline */}
           <path id="SpeedTapeOutlineRight" class="NormalStroke White" d={this.outlinePath} />
           <VerticalTape
-            tapeValue={this.spdValue}
+            tapeValue={this.spdValue.map((value) => MathUtils.round(value, 1e-2))}
             lowerLimit={30}
             upperLimit={660}
             valueSpacing={ValueSpacing}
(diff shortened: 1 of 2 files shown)


fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -1294,56 +983,72 @@ class SpeedMargins extends DisplayComponent<{ bus: ArincEventBus }> {
...
+      } else if (!(mach.isFailureWarning() || mach.isNoComputedData()) && machHysteresis) {
+        return 3;
+      } else {
+        return 1;
+      }
+    },
+    this.mach,
+    this.machHysteresis,
+    this.onGround,
+  );
```
BravoMike99: Could be an enum instead of magic values?

fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -33,48 +34,20 @@ const ValueSpacing = 10;
...
-
-  private flightPhase = 0;
-
-  private v1Speed = 0;
+const getSpeedTapeOffset = (speed: number): number => (-speed * DistanceSpacing) / ValueSpacing;
 
-  onAfterRender(node: VNode): void {
-    super.onAfterRender(node);
+class V1BugElement extends DisplayComponent<{ bus: ArincEventBus }> {
+  private flightPhase = ConsumerSubject.create(this.props.bus.getSubscriber<PFDSimvars>().on('fwcFlightPhase'), 0);
```
BravoMike99: could be readonly amongst a few others.

fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -344,169 +265,97 @@ class FlapsSpeedPointBugs extends DisplayComponent<{ bus: ArincEventBus }> {
...
+  );
 
-  private leftMainGearCompressed = true;
+  private readonly rightMainGearCompressed = ConsumerSubject.create(
+    this.props.bus.getSubscriber<PFDSimvars>().on('rightMainGearCompressed'),
+    false,
+  );
 
-  private rightMainGearCompressed = true;
+  private readonly onGround = MappedSubject.create(
```
BravoMike99, line 288: As the onGround needs to be used across multiple components, perhaps it could be derived higher up in the chain and received as prop? Could possibily do the same for the normal  law.

fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -960,242 +712,179 @@ class VStallWarnBar extends DisplayComponent<{ bus: ArincEventBus }> {
...
   render(): VNode {
     return (
       <>
         <text
-          ref={this.lowerBoundRef}
           id="SelectedSpeedLowerText"
-          class="FontSmallest EndAlign Cyan"
+          class={this.isSpeedManaged
+            .map((managed) => (managed ? 'Magenta' : 'Cyan'))
+            .map((className) => `FontSmallest EndAlign ${className}`)}
```
BravoMike99: I think we could do something like this and on the others too
```suggestion
          class={{FontSmallest : true, EndAlign: true, magenta: this.isSpeedManaged, Cyan: this.isSpeedSelected}}
```

fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
```diff
@@ -960,242 +712,179 @@ class VStallWarnBar extends DisplayComponent<{ bus: ArincEventBus }> {
...
-    } else {
-      this.lowerBoundRef.instance.style.visibility = 'hidden';
-      this.upperBoundRef.instance.style.visibility = 'hidden';
-      this.speedTargetRef.instance.style.visibility = 'hidden';
-      this.spdSelFlagVisible.set(false);
-    }
-    return inRange;
+    sub.on('realTime').handle(() => {
+      this.airSpeedWord.get();
+    });
```
BravoMike99: Is this a left over?

BravoMike99 (commented):
Not 100% sure if you were finished but just some minor comments.

fbw-a32nx/src/systems/instruments/src/PFD/SpeedIndicator.tsx
tracernz: I would create a LgicuDataProvider (specific to the PFD) or something like that for this, rather than passing down a whole bunch of individual props. You can also encapsulate the common logic in the provider as well. This pattern would help in a lot of areas by reducing logic duplication and reducing the number of subjects and handlers that need to be called constantly.

lukecologne: Yep, I fully agree. Maybe just a "common logic" provider in general would be a good idea, not specifically for the LGCIU

PR #10743, opened 2026-06-06
feat(throttle): handle next/previous detent key binds

### LINKED ISSUE #9131: Add an option for instant detent snapping
### Aircraft Version

Development

### Description

To cater to users with XBOX controllers who do not have a throttle quadrant, it would be nice to include an option in the EFB that allows you to quickly snap into different detents in the aircraft. I recommend the following ones:

- TOGA
- FLEX/MCT
- CLB
- IDLE
- Reverse Thrust
  -    Idle reverse
  -    Full reverse
 
Right now, it is kind of difficult to blindly figure out where the detents are without looking at the throttle, making things difficult for critical operations such as landing or takeoff.

Of course, the user will still have the opportunity to switch back to the current system, should they need to add minimal thrust during taxi in order to get going.

### References (optional)

The Fenix A3xx has a similar system implemented, which is very reliable and convenient for users with a controller.

_No response_

ccute

tracernz:
Fixes #9131 

## Summary of Changes

THROTTLE_DETENT_PREV and THROTTLE_DETENT_NEXT key binds were added to MSFS, with default action of THROTTLE_CUT and THROTTLE_FULL. This PR creates an inverse mapping method for the LUT implementation used by the throttles (just a copy-paste and adjustment of the normal get), and then sets up key intercepts to make these events work properly, jumping to the next detent in the desired direction.

**Important**: The events don't seem to ever actually get sent by the sim at the moment though. Let's hope it will be fixed soon: https://devsupport.flightsimulator.com/t/cannot-intercept-throttle-detent-next-or-throttle-detent-prev-key-events/18029. Until then this is a do not merge, but I will keep it open/RTR so there are builds that Asobo can test if they wish.

## Additional context

## Testing instructions

Bind a key or button to each of these events:
image
Press the buttons and ensure the throttles jump to the next available detent. Try moving the throttle into intermediate positions and then repeat, and make sure the next detent is always selected.


.github/CHANGELOG.md
```diff
@@ -147,6 +147,7 @@
 1. [EFB] Added a cold temperature correction calculator - @tracernz (Mike)
 1. [A380X/MFD] Fixed PERF APPR QNH entries in inHg being displayed as hPa - @Daboss57 (Daboss57)
 1. [EFB] Add FAA and LIDO charts supplied by MSFS2024 - @tracernz (Mike)
+1. [THROTTLE] Added handling of previous and next detent throttle bindings - @tracernz (Mike)
```
tracernz, line 150: Note: wrong release.

tracernz:
Seems like this will not go ahead on the Asobo/MS side, so we'll have to create our own K events, and an input profile so they can be bound in the FS2024 control settings.

PR #10754, opened 2026-06-10
refactor(systems): move rust systems to 2024 sdk

Saschl:
Fixes #[issue_no]

This is still very much WIP and depending on some changes on msfs-rs and potentially dev-env, but I wanted to share the first results already, as some adaptions were needed.

## Summary of Changes

Moves the gauge to a msfs 2024 system and the respective APIs.

## Additional context

### DIFF
--- a/fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/payload.rs
+++ b/fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/payload.rs
@@ -3,117 +3,28 @@ use std::error::Error;
 use msfs::sim_connect;
 use msfs::{sim_connect::SimConnect, sim_connect::SIMCONNECT_OBJECT_ID_USER};
 
-use systems_wasm::aspects::{MsfsAspectBuilder, ObjectWrite, VariablesToObject};
+use systems_wasm::aspects::{ExecuteOn, MsfsAspectBuilder, ObjectWrite, VariablesToObject};
 use systems_wasm::{set_data_on_sim_object, Variable};
 
 pub(super) fn payload(builder: &mut MsfsAspectBuilder) -> Result<(), Box<dyn Error>> {
-    builder.copy(
-        Variable::aircraft("PAYLOAD STATION WEIGHT", "Pounds", 1),
-        Variable::aspect("PAYLOAD_STATION_1_REQ"),
-    );
-    builder.copy(
-        Variable::aircraft("PAYLOAD STATION WEIGHT", "Pounds", 2),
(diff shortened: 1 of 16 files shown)


fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/payload.rs
```diff
@@ -3,117 +3,28 @@ use std::error::Error;
...
+        builder.map_many(
+            ExecuteOn::PostTick,
+            vec![
+                Variable::aspect(&format!("PAYLOAD_STATION_{i}_REQ")),
+                //  Variable::named("BOARDING_STARTED_BY_USR"),
+                //  Variable::named("FSDT_GSX_BOARDING_STATE"),
+            ],
+            |values| values[0],
+            //|values| values[1] > 0. || values[2] >= 4. && values[2] < 6.,
+            Variable::aircraft("PAYLOAD STATION WEIGHT", "POUNDS", i),
```
Saschl, line 22: Aircraft variables can now be written, so we can use the new API directly and ditch simconnect here. I think it is even needed as I encountered issues when trying to load the plane. Probably due to timing issues in simconnect write vs, aircraft variable read, will check again

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -498,8 +498,8 @@ impl SimulationElement for Pax {
     }
     fn write(&self, writer: &mut SimulatorWriter) {
         writer.write(&self.pax_id, self.pax);
-        writer.write(&self.pax_target_id, self.pax_target);
-        writer.write(&self.payload_id, self.payload.get::<pound>());
+        //writer.write(&self.pax_target_id, self.pax_target);
```
Saschl, line 501: If we read and write the LVar all the time, the value set by the EFB may not be picked up. Essentially the whole payload functionality was not really working as the pax target kept getting reset to 0.

fbw-common/src/wasm/systems/systems/src/payload/mod.rs
```diff
@@ -1048,7 +1050,10 @@ impl BoardingInputs {
     }
 
     pub fn stop_boarding(&mut self) {
-        self.is_boarding = false;
+        if self.is_boarding {
+            self.is_boarding = false;
+            self.is_boarding_write_pending.set(true);
```
Saschl, line 1055: Same as before, we need some mechanism to only write when it is really needed, otherwise the boarding may never start as the value set by the EFB is not read quickly enough. I'm not sure if this approach is the best solution here, but it works

PR #10797, opened 2026-07-02
feat(a380x/prim): add FE part to PRIMs

lukecologne:
## Summary of Changes

This PR adds the flight envelope (FE) part to the PRIMs, which was previously temporarily located in legacy FACs, taken over and adapted from the A320. The FE part is moved from there into the PRIMs. The speedscale computation should be mostly identical compared to before.

The PFD has been adapted to use the PRIMs data, by choosing the speedscale data coming from the master PRIM. The speed trend arrow also has been adapted to use the PRIM data instead of local computations.

This also means that, if there is no master PRIM or the speedscale computation of the master PRIM has failed, the characteristic speeds will disappear and a SPD LIM flag will be shown.

## Additional context

## Testing instructions

This PR touches mainly
a) The PFD characteristic and protection speeds computation
b) The alpha floor computation

Check that the PFD characteristic and protections speeds are reasonable throughout the flight envelope (various altitudes, speeds and weights), and that the alpha floor protection activates in appropriate conditions.

A full flight is not required, however the speedscale should be tested throughout as much of the flight envelope as possible, as mentioned before.


tracernz (approved):

fbw-a380x/src/systems/shared/src/publishers/PrimFePublisher.ts
BravoMike99, line 1: Perhaps some of these publishers should go to systems/shared instead? We have quite a few in instruments but that shouldn't be there :(

BravoMike99:
Dev team QA report

Object of testing : #10797 
Aircraft          : A380X 
Simulator         : 2024
Tier of Testing   : 1
Date              : 08/07/2026

Testing Process: 
Did a few test patterns, and verified that alpha floor and toga lock work as they should.
Characteristic speed computation mostly agrees with the FMS at various weights and disappears from the PFD when all three prims are failed.

Testing Results: 
Passed

PR #10818, opened 2026-07-05
fix(oanc): prevent race conditions on load/unload

Saschl:
Also adds improved error handling to prevent "Please wait" staying forever

Fixes #[issue_no]

## Summary of Changes

Adds an opId to the oanc load/unload airport functionality to discard older requests in case mutliple are fired in quick succession. Might be a fix for overlayed airports being shown.

I could reproduce the issue by adding an additional delay inside the code, but not reliably with the dev version

## Additional context

## Testing instructions

A bit hard to test as the issue cannot be reproduced reliably. But you can make sure the OANS works as before and allows seleciton of multiple airports (but only one should be shown at once)

### DIFF
--- a/fbw-common/src/systems/instruments/src/OANC/Oanc.tsx
+++ b/fbw-common/src/systems/instruments/src/OANC/Oanc.tsx
@@ -25,6 +25,7 @@ import {
   AmdbFeatureTypeStrings,
   AmdbProjection,
   AmdbProperties,
+  AmdbResponse,
   Arinc429LocalVarConsumerSubject,
   EfisNdMode,
   EfisSide,
@@ -215,6 +216,8 @@ export class Oanc<T extends number> extends DisplayComponent<OancProps<T>> {
   private readonly labelManager = new OancLabelManager<T>(this);
 
   private readonly positionComputer = new OancPositionComputer<T>(this);
+
+  private opId = 0;
 
   public dataLoading = false;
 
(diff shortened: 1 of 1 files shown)


fbw-common/src/systems/instruments/src/OANC/Oanc.tsx
```diff
@@ -612,6 +615,8 @@ export class Oanc<T extends number> extends DisplayComponent<OancProps<T>> {
   }
 
   public unloadAirportMap(performanceModeUnload: boolean = false) {
+    this.opId++;
```
Saschl, line 618: Needs to be done here as unloads should also invalidate any in-flight requests

fbw-common/src/systems/instruments/src/OANC/Oanc.tsx
```diff
@@ -638,6 +643,8 @@ export class Oanc<T extends number> extends DisplayComponent<OancProps<T>> {
 
     this.unloadAirportMap(performanceModeUnload);
 
+    const opId = this.opId;
```
tracernz, line 646:
```suggestion
    const opId = ++this.opId;
```

Saschl: Don't think I can do the increment here, as I'd also like to cancel pending load commands if "unloadAirportMap" was called before. If that makes sense

PR #10831, opened 2026-07-12
refactor: log unknown enum values instead of panicking

Saschl:
Fixes #[issue_no]

## Summary of Changes

To avoid crashing the whole wasm gauge when an unknown value/discriminator is being read, we force the usage of try_into and a fallback value. The offending value is being logged once and can be seen on the efb troubleshooting page.

Also adds a custom hook for panics to print the error message. The default behaviour does not print the backtrace anymore, which makes troubleshooting harder as we get not much information in the logs besides the info that the gauge crashed.

## Screenshots (if necessary)

New log when the wasm panics

image

Troubleshooting logs in EFB:

Screenshot 2026-07-14 212017

## Additional context

## Testing instructions

1. Make sure the plane starts up fine (systems + engines)
2. Enable Developer mode in the Sim, open the behaviours menu from the "Tools" Section. Select the variables tab, uncheck "only show variables get/set by this behaviour" and search for A32NX_ENGINE_STATE:1 
3. Change the value to 42 and observe the EFB. The red status bar should appear and show an error log about Engine State like in the screenshot


fbw-a32nx/src/wasm/systems/a320_systems_wasm/src/lib.rs
```diff
@@ -38,6 +38,12 @@ use trimmable_horizontal_stabilizer::trimmable_horizontal_stabilizer;
 
 #[msfs::gauge(name=systems)]
 async fn systems(mut gauge: msfs::Gauge) -> Result<(), Box<dyn Error>> {
+    // The default panic output is lost when the WASM instance aborts, so print the
+    // panic message and location to the MSFS console before the trap happens.
+    std::panic::set_hook(Box::new(|panic_info| {
+        println!("A32NX_SYSTEMS PANIC: {panic_info}");
```
Saschl, line 44: I'd also like to print to stderr but the formatting in the console is off then as it prints multiple lines. Like this the offending line is printed in the same log line which makes search easier

beheh:
How about we change to the `TryFrom` trait here instead? That forces the consumer to deal with this (because then the conversion can `Err`), rather than hiding behind a panic in the enum.

fbw-common/src/wasm/systems/systems/src/shared/diagnostics.rs
```diff
@@ -0,0 +1,45 @@
+use std::cell::RefCell;
+
+use rustc_hash::FxHashSet;
+
+type OptionalFunction = Option<Box<dyn Fn(&str) + 'static>>;
+type MutableOptionalFunction = RefCell<OptionalFunction>;
+
+thread_local! {
```
Saschl: Even though our wasm is only single threaded we need this to be in a thread local as the compiler does not know about runtime

heclak:
@Saschl should this include a changelog saying it fixes the bug with the use of GSX's complete boarding?

tracernz:
Dev Team Report

Discord Username  : tracernz
Object of testing : #10831  
Aircraft          : A32NX 
Simulator         : 2024  
Tier of Testing   : 1
Date              : 18/07/2026 

Testing Process:
1. Changed the value to 42 and some other invalid values, notice the messages are logged and the gauge continues running. ✅ 
 

Testing Results: 
Passed  

Negatives:       


Conclusions: 


Media:

image

fbw-common/src/wasm/systems/systems/src/shared/diagnostics.rs
```diff
@@ -0,0 +1,45 @@
...
+) -> T {
+    let first_report =
+        LOGGED_UNEXPECTED_DISCRETES.with(|logged| logged.borrow_mut().insert((context, value)));
+    if first_report {
+        report_diagnostic(&format!(
+            "unexpected {context} discrete value {value}; falling back to {fallback:?}."
+        ));
+    }
+    fallback
+}
```
Gurgel100: May I suggest to rewrite this part a bit:
```suggestion
use rustc_hash::FxHashSet;
use std::{cell::RefCell, sync::OnceLock};

type DiagnosisReporter = fn(&str);

// as this sits in the sim agnostic module the msfs layer can register the commbus reporter here
static DIAGNOSTICS_REPORTER: OnceLock<DiagnosisReporter> = OnceLock::new();

pub fn set_diagnostics_reporter(reporter: DiagnosisReporter) {
    if DIAGNOSTICS_REPORTER.set(reporter).is_err() {
        println!("WARNING: diagnostics reporter already set, ignoring");
    }
}

fn report_diagnostic(message: &str) {
    match DIAGNOSTICS_REPORTER.get() {
        Some(reporter) => reporter(message),
        None => println!("{message}"),
    }
}

pub fn fallback_on_unexpected_discrete<T: std::fmt::Debug>(
    context: &'static str,
    value: u64,
    fallback: T,
) -> T {
    thread_local! {
        // log context and value pairs of unexpected values to avoid flooding the log
        static LOGGED_UNEXPECTED_DISCRETES: RefCell<FxHashSet<(&'static str, u64)>> =
            RefCell::new(FxHashSet::default());
    }

    let first_report =
        LOGGED_UNEXPECTED_DISCRETES.with(|logged| logged.borrow_mut().insert((context, value)));
    if first_report {
        report_diagnostic(&format!(
            "unexpected {context} discrete value {value}; falling back to {fallback:?}."
        ));
    }
    fallback
}
```

fbw-common/src/wasm/systems/systems/src/simulation/mod.rs
```diff
@@ -538,6 +538,23 @@ impl SimulationElementVisitor for SimulationToSimulatorVisitor<'_> {
...
+    where
+        Self: Sized + Read<Result<T, <T as TryFrom<f64>>::Error>>,
+        T: Copy + std::fmt::Debug + TryFrom<f64>,
+        <T as TryFrom<f64>>::Error: Copy + Into<u64>,
+    {
+        let value: Result<T, _> = self.read(identifier);
+        value.unwrap_or_else(|unexpected| {
+            fallback_on_unexpected_discrete(context, unexpected.into(), fallback)
+        })
+    }
```
Gurgel100, line 557: I'm not sure if this is the correct position to put this. I think it would fit better in the `Read` trait but I didn't get it to work either. So let's keep it here for now.

fbw-common/src/wasm/systems/systems/src/structural_flex/mod.rs
```diff
@@ -62,7 +64,7 @@ fn to_surface_vibration_coeff(surface: SurfaceTypeMsfs) -> f64 {
         SurfaceTypeMsfs::Tarmac => 1.2,
         SurfaceTypeMsfs::Asphalt => 1.,
         SurfaceTypeMsfs::Macadam => 1.5,
-        _ => 3.,
+        unexpected => fallback_on_unexpected_discrete("SurfaceTypeMsfs", unexpected as u64, 3.),
```
Gurgel100: This is not log worthy because this is actually not unexpected.
On the other hand we should actually log the unexpected value when reading the `SurfaceTypeMsfs` and convert `From` to `TryFrom` in `update_context.rs:82`.

fbw-common/src/wasm/systems/systems/src/macros/mod.rs
```diff
@@ -45,11 +45,11 @@ macro_rules! provide_potential {
...
 
-macro_rules! read_write_enum {
+macro_rules! try_read_write_enum {
     ($t: ty) => {
-        impl<T: Reader> Read<$t> for T {
-            fn convert(&mut self, value: f64) -> $t {
-                value.into()
+        impl<T: Reader> Read<Result<$t, <$t as TryFrom<f64>>::Error>> for T {
+            fn convert(&mut self, value: f64) -> Result<$t, <$t as TryFrom<f64>>::Error> {
+                <$t>::try_from(value)
```
Gurgel100: It's easier to read this way:
```suggestion
                value.try_into()
```

**Explaination why this works:**
Implementing `From` or `TryFrom` always implements the reciprocal `Into` or `TryInto` respectively.
_Note: The same is not true when implementing `Into` or `TryInto`._

Gurgel100 (commented):
Great improvements! I have some small comments though.

fbw-common/src/wasm/systems/systems/src/shared/diagnostics.rs
Saschl: That looks much better to read, thanks!

fbw-common/src/wasm/systems/systems/src/structural_flex/mod.rs
Saschl: Good catch, applied the same pattern as with the other enums

fbw-common/src/wasm/systems/systems/src/simulation/mod.rs
Saschl: Yeah, I also tried to get it to work on the Read trait but no luck

Saschl:
Note, did a quick retest after the changes and errors are logged as before and the gauge keeps running. Also took the chance to report panics to the efb as well:


image

fbw-common/src/wasm/systems/systems/src/simulation/update_context.rs
```diff
@@ -79,38 +79,42 @@ pub enum SurfaceTypeMsfs {
...
+            15 => Ok(SurfaceTypeMsfs::OilTreated),
+            16 => Ok(SurfaceTypeMsfs::SteelMats),
+            17 => Ok(SurfaceTypeMsfs::Bituminus),
+            18 => Ok(SurfaceTypeMsfs::Brick),
+            19 => Ok(SurfaceTypeMsfs::Macadam),
+            20 => Ok(SurfaceTypeMsfs::Planks),
+            21 => Ok(SurfaceTypeMsfs::Sand),
+            22 => Ok(SurfaceTypeMsfs::Shale),
+            23 => Ok(SurfaceTypeMsfs::Tarmac),
+            255 => Ok(SurfaceTypeMsfs::Macadam),
```
Gurgel100: Where does this come from?

Saschl: The value 255 appeared during testing and while not documented, I wouldn't want a troubleshooting log entry in the EFB when it appears. Probably a sim bug

Saschl: Checking the docs again, seems like we're missing value 24 as well. Added it for completeness.

Gurgel100: Maybe we should add a comment then for this "special" value?

Saschl: Yes definitely 👍 Done

tracernz: There's a slightly more complete enum here https://github.com/microsoft/msfs-avionics-mirror/blob/366be5056166c639a2189e09e5af7143174fd910/src/sdk/navigation/Facilities.ts#L1203

Perhaps SurfaceTypeMsfs::Unknown would make more sense for the 0xff value...

tracernz: 0xfe is also documented as unknown here https://www.fsdeveloper.com/wiki/index.php/BGL_File_Format#Runway 🤔

Gurgel100 (approved):

Gurgel100:
LGTM. Great work!

PR #10836, opened 2026-07-17
feat(a32nx): replace sentry client with custom implementation

tracernz:
Supercedes https://github.com/flybywiresim/aircraft/pull/8816

## Summary of Changes

The old sentry client we have no longer works with the sentry API, and the newer libraries from sentry do not support the WebKit we have in MSFS.

- Replace the sentry library with a custom implementation that captures console error/warn, unhandled exceptions and unhandled promise rejections, packages them up, and sends them to the Sentry API. This was mostly written by Codex.
- Remove the old senty-client.js that we used to used in the raw JS MCDU.
- Instantiate the sentry client in the new MCDU.
- Add the Sentry ID to the EFB troubleshooting page (shows "Disabled" and the settings path to enable it when error reporting is not enabled).
- Moved the telemetry consent popup to the EFB, as the old FS2020 popup system doesn't work anymore in FS2024. It uses a standard modal style, and appears over the EFB content when the user has never answered (using z-index). It appears even when the EFB is not on, as the pop-up would have as well.
- Added to the systems-host since it's starting to get bigger and it's important we know when things go wrong there.

Future steps:
- We should probably use `@sentry/esbuild-plugin` to send source maps for release and master builds to sentry so you can get source code in the sentry stack traces for ease-of-use.
- Setup an A380X project in sentry and add error reporting to the A380X MFD.

## Screenshots (if necessary)

image
image

image
image

image

## Additional context

## Testing instructions

Make sure that enabling error reporting in the EFB causes the sentry ID to show on the troubleshooting page, and that the MCDU functions normally with error reporting enabled.


Saschl (approved):

BravoMike99:
Dev team QA report

Object of testing : #10836 
Aircraft          : A32NX
Simulator         : 2024
Tier of Testing   : 1
Date              : 05/09/2026

Testing Process: 
Spawned on runway, verified the popup for error reporting showed on the EFB.
Upon accepting the popup, verified sentry ID is displayed on the troubleshooting page. If troubleshoot is disabled, no ID is displayed.
Programmed the FMS as usual and MCDU worked ok.

Testing Results: 
Passed


Conclusions: 
LGTM, would be nice to have it on the A380X too, also, why do we have the troubleshooting under AOC/ATC D: ?

PR #10850, opened 2026-07-20
fix(a380x): landing weight not present gross weight during approach phase

### LINKED ISSUE #10823: A380X: Landing weight is not gross weight during the approach phase
### Simulator

2024

### Aircraft Version

Development

### Build info

```json
master
```

### Describe the bug

As per title.

### Expected behavior

When the approach phase is activated, the calculated landing weight by the FMS, should be the present gross weight,
for the active flightplan, or secondary flightplans which are a copy of the active.
For a manually created secondary, it should assume the engines off/preflight logic.

### Steps to reproduce

1) Activate APPR phase and observe how "LW" doesn't change to current gross weight on the MFD PERF APPR page.

### References (optional)

_No response_

Also for 2020.

bruno_pt99

smchuk:
Fixes #10823

## Summary of Changes

`getLandingWeight()` in `FlightManagementComputer.ts` treated the Approach
phase the same as any other in-flight phase (`phase >= FmgcFlightPhase.Takeoff`),
so it kept computing `LW = GW - TRIP` after APPR was activated instead of
showing the present gross weight, per the expected behavior in the issue.

Changes:
- LW now snaps to present gross weight specifically on `FmgcFlightPhase.Approach`
  (not `>=`, since that range also includes GoAround, where the previous
  GW-minus-trip logic still applies)
- `forPlan` is now correctly passed through to `fmgc.getGrossWeightKg(forPlan)`,
  which was previously silently defaulting to the active flight plan
- Manually created secondary flight plans (not a copy of active) now use the
  static TOW − TRIP calculation regardless of live engine state, using the
  existing `isActiveOrCopiedFromActive()` check already used elsewhere in this
  file, consistent with the "assume engines off/preflight logic" behavior
  requested in the issue

Out of scope / follow-up needed: full trip-fuel prediction for secondary
flight plans is not implemented, `getTripFuel()` still returns `null` for any
non-active plan, since VNAV doesn't compute profiles for secondary plans yet.
This is a larger feature and will be tracked as a separate issue rather than
bundled here.

## Additional context

smchuco

## Testing instructions

1. Load the A380X, fly to cruise/descent phase.
2. Open MFD PERF APPR page, note the LW value.
3. Activate the APPR phase (either automatically at DECEL or via ACTIVATE
   APPR).
4. Confirm LW updates to match the current GW figure (cross-check against SD
   GW readout) rather than continuing to decrease with predicted trip fuel.
5. Trigger a go-around from APPR and confirm LW reverts to predicting
   GW − TRIP rather than staying frozen at the GW snapshot.
6. Repeat with a secondary flight plan that is a copy of active, and
   separately with a manually-created secondary, to confirm the
   engines-off/preflight-style calculation applies to the latter.

### DIFF
--- a/fbw-a380x/src/systems/instruments/src/MFD/FMC/FlightManagementComputer.ts
+++ b/fbw-a380x/src/systems/instruments/src/MFD/FMC/FlightManagementComputer.ts
@@ -498,28 +498,36 @@ export class FlightManagementComputer implements FmcInterface {
   }
 
   /** in kg */
-  public getLandingWeight(forPlan = FlightPlanIndex.Active): number | null {
-    const tow = this.getTakeoffWeight(forPlan);
-    const gw = this.fmgc.getGrossWeightKg();
-    const tf = this.getTripFuel(forPlan);
+ public getLandingWeight(forPlan = FlightPlanIndex.Active): number | null {
+  const plan = this.flightPlanInterface.get(forPlan);
+  const isActiveOrCopyOfActive = plan.isActiveOrCopiedFromActive();
+  const useLiveAircraftState = isActiveOrCopyOfActive && this.enginesWereStarted.get();
 
-    if (!this.enginesWereStarted.get()) {
-      // On ground, engines off
-      // LW = TOW - TRIP
-      return tow && tf ? tow - tf : null;
-    }
-    if (gw && tf && this.fmgc.getFlightPhase() >= FmgcFlightPhase.Takeoff) {
-      // In flight
-      // LW = GW - TRIP
-      return gw - tf;
-    }
-    if (gw) {
-      // Preflight, engines on
-      // LW = GW - TRIP - TAXI
-      return gw - (tf ?? 0) - (this.flightPlanInterface.get(forPlan).performanceData.taxiFuel.get() ?? 0) * 1000;
-    }
-    return null;
+  const tow = this.getTakeoffWeight(forPlan);
+  const tf = this.getTripFuel(forPlan);
+  const phase = this.fmgc.getFlightPhase();
+
+  if (!useLiveAircraftState) {
+    // LW = TOW - TRIP
+    return tow !== null && tf !== null ? tow - tf : null;
   }
+
+  const gw = this.fmgc.getGrossWeightKg(forPlan);
+
+  if (gw !== null && isActiveOrCopyOfActive && phase === FmgcFlightPhase.Approach) {
+    // APPR active: LW is the present gross weight
+    return gw;
+  }
+  if (gw !== null && tf !== null && phase >= FmgcFlightPhase.Takeoff) {
+    // LW = GW - TRIP
+    return gw - tf;
+  }
+  if (gw !== null) {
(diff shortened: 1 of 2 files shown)


fbw-a380x/src/systems/instruments/src/MFD/FMC/FlightManagementComputer.ts
```diff
@@ -498,29 +498,37 @@ export class FlightManagementComputer implements FmcInterface {
...
+  }
+  if (gw !== null && tf !== null && phase >= FmgcFlightPhase.Takeoff) {
+    // LW = GW - TRIP
+    return gw - tf;
+  }
+  if (gw !== null) {
+    // LW = GW - TRIP - TAXI
+    return gw - (tf ?? 0) - (plan.performanceData.taxiFuel.get() ?? 0) * 1000;
+  }
+  return null;
```
tracernz: I think the go around and done phase should also use the actual gross weight, as TOW - TRIP is no longer relevant after you've flown the approach already.

You can also simplify the logic by handling the gw === null case first, not using the active plan flight phase for other plans, and the approach/go around/done phases can be the default case at the end, like so:
```suggestion
  if (gw === null) {
    return null;
  }
  
  if (phase < FmgcFlightPhase.Takeoff || !isActiveOrCopyOfActive) {
    // Predicted LW = GW - TRIP - TAXI
    return gw - (tf ?? 0) - (plan.performanceData.taxiFuel.get() ?? 0) * 1000;
  }
  
  if (phase < FmgcFlightPhase.Approach) {
    // Predicted LW = GW - TRIP
    return gw - tf;
  }
  
  // APPR active: LW is the present gross weight
  return gw;
```

I wonder if we should be using live GW at all when isActiveOrCopyOfActive is false. Probably should use the init data for that plan (ZFW + FOB) and preflight phase. That should be addressed with this TODO rather than in this method however https://github.com/flybywiresim/aircraft/blob/6c3709ab9f87023051815a40e4db33cac3d8ce62/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.ts#L213

tracernz: The non-active plans issue would also need to be addressed here, not using the FQMS for non-active plans https://github.com/flybywiresim/aircraft/blob/6c3709ab9f87023051815a40e4db33cac3d8ce62/fbw-a380x/src/systems/instruments/src/MFD/FMC/fmgc.ts#L251

BravoMike99: Yep, always the init GW for not copied of active. ("engines off" state for those). The only question is what happens if a copy of active has a different ZFW from active but it seems they disable the ZFW input on the secondary in this case to prevent that from happening.

smchuk: Hiya! I’ve seen the review and applied the suggested change. Currently I am away from my computer, so I will have to address the rest when I am back. Thanks!

PR #10851, opened 2026-07-21
fix(a380x/flight model):  Fixes CONF 3 premature nosewheel liftoff

donstim:
Note: This PR is now coordinated with and dependent on #10861. It should not be merged before #10861.
As noted on Discord, a premature nose wheel liftoff without stick input is possible in CONF 3. The specific condition is shown below along with a video:
Screenshot 2026-08-25 155818
https://1drv.ms/v/c/85cbffb84cec323b/IQBF1gSR1G7pS7MfyX49hDr_Ac-mBqqamBiLX7BGnKOLJVw?e=F2knaM

## Summary of Changes

This PR changes:
1) The CONF 3 flap parameters to not induce as much nose up lift. 
2) The trim effectiveness is reduced at takeoff speeds. 
3) The drag in CONF 3 is increased because it was inadvertently made too low in the current MSFS 2024 flight model. 

## Screenshots (if necessary)
The issue of auto-rotation is affected by the waviness/bumpiness of the runways, particularly in MSFS 2024, and the accuracy of the takeoff speeds. The primary scenarios that were used to develop this PR were the one shown in the video above and a max weight (575T) takeoff at OMDB. The effect on the takeoff condition that this PR was started for is shown in the following video:
https://1drv.ms/v/c/85cbffb84cec323b/IQAoR-2FYKnvRpTmE1Cx2iT8AVWRF6g52YkBNU-SqSOZIEk?e=aRxfSa

For OMDB, the takeoff speeds varied greatly due to runway slope (which had twice the effect I would have thought it would have. I used the runway with the highest VR speed, which was 30L. The following video shows that with this PR incorporated, the nose wheel does lift before VR, but then comes down again by VR, lifts again just after VR, but then comes down again and does not lift for good until after V2: 
https://1drv.ms/v/c/85cbffb84cec323b/IQCY88OD6wxtSaKrx_82qO2lAVAhfbdHOioTu8EN2Uh3iEQ?e=G9kMmh

## Additional context
The downsides of these changes is that automatic pitch trim in CONF 3 (such as during an approach) was increased. Pitch trim was also increased on approach in CONF FULL, but to a lesser degree.

## Testing instructions
Test CONF 3 takeoffs at low and high weights on long runways to confirm that the nose wheel does not lift off before the normal rotation speed. The most critical conditions are typically high VR speed (so, a long runway) and aft cg loading (but using the normal loading procedure via the EFB, for example, don't use the cargo sliders to move the cg aft.


tracernz (approved):

donstim:
Superseded by #10924

PR #10858, opened 2026-07-24
refactor: move terronnd gauge to native 2024 wasm

Saschl:
Fixes #[issue_no]

## Additional context


fbw-common/src/wasm/terronnd/src/simconnect/clientdataarea.hpp
```diff
@@ -25,10 +27,10 @@ class ClientDataAreaBase : public base::Changeable {
   friend Connection;
 
  protected:
-  HANDLE* _connection;
+  HANDLE*       _connection;
   std::uint32_t _dataId;
   std::uint32_t _definitionId;
-  bool _alwaysChanges;
+  bool          _alwaysChanges;
```
tracernz, line 33: We should really remove the config that does this, and put it in the Fadec + Extras host only. I don't think anybody wants a lot of diff on unrelated lines every time we make a change.

PR #10859, opened 2026-07-25
feat(a380x/fms): update rnp ar nomenculature

BravoMike99:
Fixes #[issue_no]

## Summary of Changes

Updates the RNP AR approaches nomenclature to show RNP (AR) instead of the previous RNAV (RNP)

## Screenshots (if necessary)

RNPARR
RNP SID

## Additional context

## Testing instructions

A380X:
1) Select an airport with RNP AR approaches, check that the RNP AR approaches appear as RNP XXX(AR) on the approach list.
2) Verify that it also shows as RNP in the PERF arrival page and F-PLN page.
3) Verify that normal rnav approaches are shown as "RNV" on the PERF ARR page, arrivals page and F-PLN page.
4) Check that SID name and approach name on the ND are displayed correctly as before.

A32NX:
Check that the approach name on the ND is displayed correctly as before.


fbw-common/src/systems/shared/src/ApproachUtils.ts
```diff
@@ -94,14 +100,20 @@ export class ApproachUtils {
...
         return 'VOR';
+      case ApproachType.Gls:
+        return 'GLS';
       default:
         return '';
     }
   }
+
+  public static isRnpArApproach(approach: Approach): boolean {
+    return approach.authorisationRequired || approach.missedApproachAuthorisationRequired; // TODO for LPV it shouldn't be applicable.
```
tracernz: I don't really understand the TODO... RNP AR !== LPV.

fbw-common/src/systems/instruments/src/ND/ND.tsx
```diff
@@ -700,13 +713,22 @@ class TopMessages extends DisplayComponent<{
...
   private readonly trueTrackWord = Arinc429RegisterSubject.createEmpty();
 
-  private needApprMessageUpdate = true;
+  private readonly apprMessage0 = ConsumerSubject.create(this.sub.on('apprMessage0'), 0);
 
-  private apprMessage0: number;
+  private readonly apprMessage1 = ConsumerSubject.create(this.sub.on('apprMessage1'), 0);
 
-  private apprMessage1: number;
+  private readonly approachIsRnpAr = ConsumerSubject.create(this.sub.on('rnpArApproach'), false);
```
tracernz: Rather than this hack, the FMS should publish the full text for display. That means this will have to become per-aircraft code.

fbw-common/src/systems/shared/src/ApproachUtils.ts
tracernz: Discussed in DMs.. we can handle this now.

tracernz (approved):

PR #10862, opened 2026-07-27
fix(fms): fix vnav crash with far-away speed constraint

BlueberryKing:
Fixes #[issue_no]

## Summary of Changes

Fixes an issue where a speed constraint that was far away from the departure would cause a crash in the computation of predictions. This happened because VNAV tried to build a really long climb segment to the speed constraint, reaching altitudes above performance envelope. To fix it, we stop the segment when we reach the cruise altitude.

## References

Before:
image
image

## Additional context

## Testing instructions

Setup a flight plan, enter an SID, a cruise altitude and your weights such that predictions are computed. Then, enter a speed constraint on a waypoint downroute. Alternatively, you can enter a random waypoint after the SID and set a climb speed constraint on that waypoint, make sure predictions are computed as you would expect them to.


tracernz (approved):

Saschl:
Dev Team Tester report

Discord Username  : saschl
Object of testing : #10862
Aircraft          : A32NX 
Simulator         : 2024 
Tier of Testing   : 1
Date              : 06/08/2026 

Testing Process: 
 
1. Entered a SID and cruise ALT
2. Added a far away waypoint after the SID
3. Verified predictions are still working

Testing Results: 
Passed


Media:
image

PR #10887, opened 2026-08-10
chore: added SimVar section to PR template

Eagle941:
Fixes #[issue_no]

## Summary of Changes

As per title, added `SimVar` section to PR template to clearly outline any change to SimVars.

## Screenshots (if necessary)

No change

## Additional context

The goal is to improve the traceability of interface changes to reduce issues with third party software.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,6 +8,10 @@ Fixes #[issue_no]
 ## Summary of Changes
 <!-- Please provide a summary of changes for this pull request, ensuring all changes are explained. -->
 
+## SimVar Changes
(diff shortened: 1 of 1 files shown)


tracernz:
It should also cover local vars, key events, HTML events, etc. i.e. everything that would be considered part of the "cockpit API". Probably easier to just call it "Cockpit API Changes". Also, there are other documents than the two mentioned. If they are updated, then this section in the PR is redundant.

FoxtrotSierra6829 (approved):

Eagle941:
> It should also cover local vars, key events, HTML events, etc. i.e. everything that would be considered part of the "cockpit API". Probably easier to just call it "Cockpit API Changes". Also, there are other documents than the two mentioned. If they are updated, then this section in the PR is redundant.

I have improved the wording, what do you think?

PR #10900, opened 2026-08-19
fix(fms): use ADIRS altitude in FMS

BlueberryKing:
Fixes #[issue_no]

## Additional context

### DIFF
--- a/fbw-a32nx/src/systems/fmgc/src/navigation/Navigation.ts
+++ b/fbw-a32nx/src/systems/fmgc/src/navigation/Navigation.ts
@@ -13,6 +13,7 @@ import {
   NearbyFacilityMonitor,
   NearbyFacilityType,
   NearbyFacility,
+  RegisteredSimVar,
 } from '@flybywiresim/fbw-sdk';
 
 import { LandingSystemSelectionManager } from '@fmgc/navigation/LandingSystemSelectionManager';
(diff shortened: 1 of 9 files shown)


fbw-a32nx/src/systems/fmgc/src/navigation/Navigation.ts
```diff
@@ -164,6 +177,12 @@ export class Navigation implements NavigationProvider {
 
   private nearbyAirportMonitor: NearbyFacilityMonitor;
 
+  private readonly isFmgc1Priority = ConsumerSubject.create(this.sub.on('fmgc_1_priority'), true);
+
+  private readonly register = Arinc429Register.empty();
+  private readonly eisDiscreteWord2Left = RegisteredSimVar.create('L:A32NX_FCU_LEFT_EIS_DISCRETE_WORD_2', 'number');
+  private readonly eisDiscreteWord2Right = RegisteredSimVar.create('L:A32NX_FCU_RIGHT_EIS_DISCRETE_WORD_2', 'number');
```
BlueberryKing, line 184: I think this is not the right place to fetch these variables. They should come from the bus as well, but then they are different between the A320 and the A380X (as they should be), so we probably need another component somewhere that is aircraft specific?

tracernz: Yeah, we should probably factor things that are not from ARINC standardised systems (e.g. ADR or IR), into injectable providers. The FMGC should just define the interface the providers need to supply.

PR #10546, opened 2026-02-08
fix(A380x/fight model): Flight model update

donstim:
## Summary of Changes
This update is primarily intended to reduce the sensitivity in pitch control. Associated changes are to increase ground effect and update lift and drag parameters to retain the correct pitch attitudes, N1s, and fuel usage. The drag was adjusted to match data used by SimBrief for flight planning, but at a more aft cg that is consistent with normal A380 loading practices. Since, in MSFS a more forward loading leads to lower drag, this should be conservative for most normal loadings. That is, for more forward cg positions than 40.5% MAC, fuel burn should be less than predicted (though this conservatism can be entirely negated by adverse winds).

This update has been revised to be consistent with PR #10639 in terms of pitch sensitivity, ground effect, drag, and pitch trim values on approach. Although this update is intended for the MSFS 2020 version when we have split versions, it can also be used in MSFS 2024 until then. It will offer improved pitch response (lower sensitivity) than the current dev version. Like the current dev version, the pitch attitudes and N1s will be inaccurate, and the fuel system transfers may result in unbalanced fuel tanks.

## Additional context

Discord username (if different from GitHub): donbikes

## Testing instructions
This PR has changed significantly since the tests conducted in Feb/Mar. New QA testing is needed to
1. Confirm good pitch handling characteristics during takeoff and landing, including in crosswinds up to 40 knots.
2. Confirm good hand flying characteristics in pitch.
3. Confirm no auto-rotation before Vr on takeoff.
4. Confirm fuel usage is reasonably close to simbrief flight plan fuel predictions.
5. Confirm VNAV path descents are not adversely affected.

## How to download the PR for QA

Every new commit to this PR will cause new A32NX and A380X artifacts to be created, built, and uploaded.

1. Make sure you are signed in to GitHub
1. Click on the **Checks** tab on the PR
1. On the left side, find and click on the **PR Build** tab
1. Click on either **flybywire-aircraft-a320-neo**, **flybywire-aircraft-a380-842 (4K)** or **flybywire-aircraft-a380-842 (8K)** download link at the bottom of the page

PR #10904, opened 2026-08-22
fix(a380x/a32nx/flightmodel): Flight Model and Engine Model tidy up (and rework) -FS20-

GPUthesteve:
Fixes #[nil]

## Summary of Changes

per flight model
 (A388)
- Proper flaps and spoilers drag. It is given per radians, not total.
- Reduce excessive aileron effectiveness factor (from 1.8 down to 1.0, thus removing that line altogether)
(this is essentially revert, or perhaps, goes against this https://github.com/flybywiresim/aircraft/pull/10546)
- Reassign clean CLmax from 1.7 to 1.4 (then add 0.15 back to flaps to ensure proper flight characteristics)

 (A32NX)
- Reassign clean CLmax from 1.7 to 1.4, remove CLflaps from 1.8 to 1.7 to ensure proper ANU pitch during approach
- New ground effect 
 
per engine model (A32NX&A388)
- Simplify N1 vs N3 (or N2 in the sim) table. MSFS is very good at interpolating these tables while giving good results.
- Thrust/Ram drag model rework so to match new drag values. This will somewhat affect initial climb rate and accerelation at 15deg ANU after liftoff to maintain V2+10 
- (A388 exclusive) Tidy N3 vs Throttle drive curve (low and high mach). There's nasty spike around 1.45 IAP (3rd graph/line as highlighted in AAM) so I removed everything in between 1 and 5.41 and let MSFS interpolate those.
image

## Cockpit API Changes
nil

## References
SL N1 VS N3 table and thrust model simulated using X-Plane PlaneMaker.

## Additional context
A32NX now should perform on par or similiar to FSLabs Airbus and/or ProSim/Fenix Airbus

Discord username (if different from GitHub): _jjthesteve_

## Testing instructions
Simply go get a ride...'full flight' with this thing. 


BlueberryKing:
Hi, thanks for your contribution! Unfortunately, we cannot merge the PR in this stage. The flight- and engine models of our aircraft have undergone many iterations to get as close to reference values of the real life aircraft as possible. We've provided some comparisons here: https://flybywiresim.com/notams/a32nx-flight-model/.

For this reason, we want to be sure that any adjustment of these numbers are an improvement to the existing behavior of the airplane that can be backed up by comparison to real life numbers. This is not possible in this PR.

To be clear, both flight- and engine model tuning is very difficult work with the limited amount of reference data available and the large amount of time involved in testing changes across the entire flight envelope. I don't want to discourage you from pursuing future contributions, but I can only recommend to start small by identifying existing issues, either on your own, or looking at the open issues on GitHub. Ideally, you can also reach out on Discord to discuss your ideas with people who have worked on the flight- or engine model previously. Thanks for your understanding!

PR #10908, opened 2026-08-28
fix: harden terronnd wasm gauge

Saschl:
Fixes #[issue_no]

## Summary of Changes

As we encounter some wasm crashes in the terrain module recently (hard to reproduce) I checked the code to find some possible trigger points (Claude assisted).

- Adds some sanity checks before reserving memory (will definitely crash if expectedBytes is a bogus number that comes from Simbridge via simconnect)
- Discard potential garbage data (e.g. if more is sent than expected)
- Fix potential UB in vector handling

Sorry for the formatting change. I adapted the formatting file to use the same rules as the FBW gauges now.

## Additional context

## Testing instructions

- Launch Plane and Simbridge and ensure the terrain display still works and updates
- Verify the gauge does not crash (ND turns orange in this case)


fbw-common/src/wasm/terronnd/src/simconnect/clientdataarea.hpp
```diff
@@ -220,45 +224,14 @@ class ClientDataAreaBuffered : public ClientDataAreaBase {
...
-  }
 
   /**
    * @brief Reserves internal data to receive the data
    * @param expectedByteCount Number of expected bytes in streaming cases
    */
   void reserve(std::size_t expectedByteCount) {
     this->_expectedByteCount = expectedByteCount;
-    this->_content.reserve(expectedByteCount);
+    this->_content.resize(expectedByteCount);
```
Saschl, line 234: Not entirely sure if this was an issue, but as size never gets initialized with `reserve`, the later usage with memcpy and the underlying pointer is not really defined when size = 0.

Image

tracernz (approved):

BravoMike99:
Dev team QA report

Object of testing : #10908 
Aircraft          : A32NX
Simulator         : 2024
Tier of Testing   : 1
Date              : 05/09/2026

Testing Process: 
Turned on TERR on ND, switched ranges and modes on the ND and terrain kept working normally.

Testing Results: 
Passed

Issue #10938, opened 2026-09-16
[FMGS/LNAV] VA Leg (Heading to an Altitude) incorrect lateral behavior and sequencing


PedroBazaglia:
### Simulator Version

2024

### Aircraft Version

Stable

### Build info

```json
{
    "built": "2026-08-22T18:37:38+00:00",
    "ref": "refs/tags/v2024.1.0",
    "sha": "161051add8a6959e3325289dbebe097c0793056f",
    "actor": "Saschl",
    "event_name": "manual",
    "pretty_release_name": "stable/v2024.1.0",
    "version": "a32nx-v2024.1.0-rel.161051a"
}
```

### Describe the bug

Following up on a brief discussion in the Discord support channel, I am opening this issue to track the LNAV sequencing behavior regarding VA (Heading to an Altitude) legs.

Currently, the aircraft does not seem to handle the ARINC 424 VA Path/Terminator logic strictly according to the A320 FMGS standard, which affects SIDs and Go-Arounds that rely on this leg type.

### Expected behavior

Based on the A320 FMGS logic and ARINC 424 standards, when a VA leg becomes active:

Lateral Guidance (NAV): The AP/FD should command the aircraft to maintain a specific magnetic heading (not track). The FMA should display NAV (the pilot should not need to pull the HDG knob).

Sequencing (Terminator): The leg termination is solely based on Barometric Altitude. Once the aircraft crosses the pre-defined target altitude for that leg, the condition is satisfied. The FMGS should automatically sequence to the next waypoint and initiate the turn (if applicable) while remaining in managed NAV.

FCU Priority (The "Trap"): The FCU selected altitude must always take priority. If the VA leg target altitude is 3000 ft but the FCU is set to 2000 ft, the aircraft should capture 2000 ft (ALT* -> ALT). Because the 3000 ft trigger was never reached, the VA leg must not sequence, and the aircraft should fly the heading indefinitely in NAV mode until pilot intervention (e.g., dialing a higher FCU altitude or performing a Direct To).

### Steps to reproduce

1. Load the aircraft at SBSP (Congonhas Airport, São Paulo, Brazil) with an empty weight configuration.
2. In the MCDU, program a departure from Runway 17R using the UGTIX 2A (UGTI2A) SID.
3. In the PERF page, enter the takeoff data: V1 128, VR 130, V2 133, and Flaps 1.
4. Take off, engage the Autopilot, and ensure NAV mode is active on the FMA.
[...]

PR #10956, opened 2026-09-22
fix(fmgc): predict CA/FA altitude termination from the runway end

PedroBazaglia:
Related to #10938

## Summary of Changes
CA and FA legs predict where the aircraft will reach the leg's target altitude, in order to place the leg's termination point and trigger lateral sequencing. On departure, this prediction used the wrong origin and starting altitude, which could place the termination point too close to the aircraft (or even behind it) and cause unstable / oscillating lateral guidance immediately after takeoff.

**The Problem:**
In `CALeg.recomputeEstimatedTermination()`, the origin altitude was hardcoded to 0, and the leg's start point was the runway *fix* rather than the runway *end*. For a CA leg immediately after the runway, this means the remaining-altitude calculation ignores field elevation, and the distance-to-termination is measured from the runway threshold, not accounting for the length of the runway. On short/high-elevation airports (e.g. SBSP), this placed the termination point unstably close to the aircraft, causing severe lateral looping during the initial NAV-mode climb.

**The Fix:**
- `AltitudeTerminationPrediction.ts`: Added `getRunwayDepartureReference()` / `isRunwayFix()`, deriving the runway end (DER) and threshold elevation from the runway navdata object, obtained from the FA leg's own `fix`, or from the preceding runway `IF` leg's `fix` for a CA leg, and `alongTrackDistance()` for projecting a position onto a leg's course line.
- `CA.ts` / `FA.ts`: Legs that start at (or follow) a runway now originate their prediction at the runway end, using the runway elevation as the starting altitude, both on the ground and while airborne before passing the runway end.
- `AltitudeTerminationPrediction.ts`: Filters out instantaneous vertical-speed readings below 50% of the leg's default climb rate, treating them as noise (early rotation, momentary level-off) instead of using them to compute an unrealistically distant termination point.

## Cockpit API Changes
None.

## Screenshots (if necessary)
image

## References
Cross-checked the predicted termination point and turn timing against a real-world cockpit video of the SBRF SID BUTOB1A.

## Additional context
**Related Issues:**
This is related to #10938, which reports lateral looping on a CA-type (VA) leg on the SBSP/17R departure. Reproducing its exact steps (V1 128 / VR 130 / V2 133, flaps 1) no longer showed the looping behaviour with this change. However, this PR **does not** implement the heading-vs-track guidance mode or the FCU-altitude priority behaviour also requested in that issue, which is why it is only marked as related and not closing it.

**Known limitations / follow-ups:**
- `defaultClimbRate`/`defaultSpeed` are still fixed constants (2000 fpm / 175 kt for CA, 500 fpm / 175 kt for FA), not derived from V2 or the VNAV takeoff/climb profile. Planned as a follow-up PR using `FlightPlanPerformanceData.v2`.
- `INDICATED ALTITUDE` is compared directly against the leg's MSL target, which is only correct below the transition altitude (QNH).
- No acceleration-altitude modeling; a single climb rate is assumed all the way to the target altitude.

Discord username: pedro_baza

## Testing instructions
1. Load into a short or high-elevation airport (e.g., SBBR, SBSP runway 17R, or SBRF SID BUTOB1A).
2. Program a standard departure that includes an immediate CA or FA leg after takeoff.
3. Perform the takeoff roll and initial climb in NAV mode.
4. Verify that the termination point on the ND sits ahead of the aircraft correctly.
5. Confirm the automatic turn triggers at the expected time without any looping or oscillating lateral guidance behaviour.

2. Click on the **Checks** tab on the PR
3. On the left side, find and click on the **PR Build** tab
4. Click on either **flybywire-aircraft-a320-neo** or **flybywire-aircraft-a380-842** download link at the bottom of the page


tracernz:
Unfortunately this approach will only work for a subset of xA legs. We need LNAV and VNAV to compute predictions in lockstep to solve this for the general case (an xA leg anywhere in the SID). This requires some larger architectural work rather than a small PR.

PedroBazaglia:
That's a fair point, and matches a limitation I called out under
"Known limitations" — this only fixes the origin for xA legs that
directly follow the departure runway, not the general case of an xA
leg anywhere in the route.

To be clear about scope: this PR targets the specific bug reported
in #10938 (severe LNAV looping right after takeoff on short/high-
elevation airports), which is the case an xA leg most commonly
appears in practice (SID initial climb). It doesn't attempt to fix
xA legs elsewhere in a procedure.

Given that, would you prefer:
1. Merging this as a scoped fix for the departure case, with a
   tracking issue for the general LNAV/VNAV lockstep prediction, or
2. Holding off entirely until that larger architectural work is
   planned/started?

Happy to open that tracking issue either way, and to help with the
larger work if there's already a design direction for it.

